PR #46, opened 2026-07-23
fix(canvas): context links work for agents launched by hand in plain terminals

### LINKED ISSUE #45: Connecting two terminals running Claude Code manually doesn't link their context
## Problem

Drawing a connection between two Claude Code nodes links their context, so each agent can read the other's session when needed.

This only works when the nodes were created as Claude Code nodes (e.g. ⌘⇧C). If you instead create two regular terminals and start `claude` inside each one yourself, connecting the two nodes does nothing — the link doesn't take, and the agents can't read each other's context.

## Steps to reproduce

1. Create two regular terminal nodes (⌘T).
2. Run `claude` inside each one.
3. Drag a connection from one node to the other.

**Expected:** the two sessions get linked, the same as when the nodes are created as Claude Code nodes.

**Actual:** no link is created; neither agent can read the other's context.

DIodide:
Fixes #45

## Problem

Connecting two plain terminal nodes where the user started Claude Code by hand (⌘T, then typing `claude`) does nothing — the link doesn't take and no context is shared. It only works when the nodes were created as agent nodes (⌘⇧C).

Cause: the link classifier resolves agent identity from the node's `data.agentId`, which is only stamped at creation time for explicit agent nodes. A plain terminal never gets it, so `classifyLink` sees both endpoints as context-incapable and refuses the edge before any link plumbing runs — even though the managed hooks (armed in every local session via the PTY env) have already reported exactly which agent is running inside.

## Fix

Consult the hook-reported identity as a fallback:

- **`Canvas.tsx` — `agentIdOf`** falls back to `useAgentStatus.byId[id].agentId` (the hook-reported identity) when the node has no `data.agentId`. This is the core change: classification, discovery notes, and the link map all key off `agentIdOf`, so everything downstream flows once it passes.
- **`Canvas.tsx` — `linkSessionSig`** now includes `agentId`, so the link map re-pushes when a node's identity arrives *after* the edge was drawn (e.g. right after an app restart).
- **`agentStatus.ts`** persists `agentId` alongside `sessionId`/`session`/`unread`. A plain terminal's agent identity exists nowhere else, and tmux keeps the session (and the agent) alive across app restarts — the link must keep classifying too.
- **`noteLink.ts` — `buildBackgroundLinkMaps`** accepts the same fallback (optional param, existing call sites unaffected), so background projects' plain-terminal links survive the full link-map rewrite. New unit test covers it.

Terminals that have never run an agent still refuse the connection — the gate is unchanged, it just consults an identity source that already existed.

## Testing

- `npm run typecheck` clean; noteLink suite 22/22 (incl. new fallback test); full `vitest` run has no new failures (the two failing files — `pty-single-user` temp-dir race, `viewMode` Node localStorage quirk — fail identically on unmodified `main`).
- Verified live in an isolated instance (`NT_MULTI`): two plain terminals with hand-launched `claude` → drag connect → ⇌ context edge forms, both sides receive the "[nodeterm] You are now linked…" note, link files carry `agent`/`sessionId`/`transcriptPath`, and the get-linked-context skill in one node correctly reads the other's transcript. Control: two fresh terminals that never ran an agent → same drag, edge refused.

## Not in scope

The other `data.agentId` gates on plain terminals (status badge, chat panel, context meter, resume) — promoting those for hand-launched agents would be a natural follow-up if wanted.

### DIFF
--- a/src/renderer/canvas/Canvas.tsx
+++ b/src/renderer/canvas/Canvas.tsx
@@ -1713,13 +1713,18 @@ export function Canvas() {
     [onNodesChange, markDirty]
   )
 
-  // Resolve a node's agent id, with a tags fallback for not-yet-migrated legacy nodes.
+  // Resolve a node's agent id, with a tags fallback for not-yet-migrated legacy nodes and a
+  // hook-status fallback for plain terminals where the user launched an agent CLI by hand:
+  // every local session carries the hook env (pty-manager defaults agentId to 'claude'), so
+  // the managed hooks report who's actually running inside even when data.agentId was never
+  // set at node creation.
   const agentIdOf = useCallback((id: string): AgentId | undefined => {
     const n = nodesRef.current.find((x) => x.id === id)
     if (!n || n.type !== 'terminal') return undefined
     return (
       (n.data.agentId as AgentId | undefined) ??
-      (((n.data.tags as string[]) ?? []).includes('claude') ? 'claude' : undefined)
+      (((n.data.tags as string[]) ?? []).includes('claude') ? 'claude' : undefined) ??
+      useAgentStatus.getState().byId[id]?.agentId
     )
   }, [])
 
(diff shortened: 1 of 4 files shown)


DIodide:
image

tested manually

eneskirca:
Merging — one note for a possible follow-up: the persisted `agentId` never expires, so a terminal that ran `claude` once stays link-classifiable as a Claude node even after weeks of plain-shell use (linking it then points at the last session's transcript, which is still resumable — harmless, just worth knowing). Dropping the identity when the agent process exits would tighten that up.

PR #74, opened 2026-08-02
fix(ssh): prompt for passphrase-protected keys in SSH projects

### LINKED ISSUE #70: "Connect over SSH…" never prompts for a key passphrase and fails
**What happens**

Connect over SSH… loads my saved servers, but connecting to any host whose private key is
passphrase-protected fails with:

Could not establish the SSH connection: Permission denied (publickey).

No passphrase prompt ever appears, so there is no way to use an encrypted key with an SSH project.
The quick-connect terminal node is fine, since it runs ssh inside a real tty, it can prompt there.

**Why (from reading the code)**

SshProjectManager spawns the ControlMaster with no tty and no SSH_ASKPASS, so OpenSSH has
nowhere to ask. It tries the key once with an empty passphrase and gives up with the generic
publickey error above.

**I'm working on a fix and will PR it**

My current approach: an SSH_ASKPASS helper that prompts in the app (loopback HTTP relay into the
main process, token-authenticated), and then letting **ssh-agent** hold the unlocked key via
AddKeysToAgent so reconnects and master respawns don't re-prompt. This is the same direction
VS Code Remote-SSH takes, and it means the app itself never stores a passphrase.

Two questions before I open it:

1. Is there another approach you'd prefer for this repo?
2. Which branch should I target the PR to, main or feat/remote-sessions?

parsa222:
Fixes #70.

## What broke

"Connect over SSH…" fails for any host whose key has a passphrase:

```
Could not establish the SSH connection: Permission denied (publickey).
```

The project ControlMaster runs ssh with no tty and no `SSH_ASKPASS`, so there is nowhere to ask. ssh tries the key with an empty passphrase and gives up. The quick-connect terminal always worked because it has a real pty.

## The fix

1. **An SSH_ASKPASS relay.** A generated shell helper POSTs the prompt to a loopback server in main (random port, bearer token), which asks the renderer and answers ssh on stdout. `curl` is already a dependency of the agent hooks, so nothing new is needed.

2. **An app-private ssh-agent.** `AddKeysToAgent=yes` pointed at an agent nodeterm spawns and kills itself. One prompt per app run: reconnects are silent, quitting forgets the key. The user's own login agent is never touched.

The tradeoff is deliberate: `SSH_AUTH_SOCK` names one socket, so a key already sitting in the login agent now costs one prompt per run. In exchange, nothing nodeterm unlocks outlives nodeterm. Agent-only credentials (1Password, Secretive, smartcards) keep working via `IdentityAgent`, and the error hint says so when that is the problem. Quit also sends `ssh -O exit` per connection so a daemonized ControlPersist master cannot outlive the app.

## Measured

`test/ssh-docker/` ships a disposable sshd container with three users (unencrypted key, passphrase key, password-only). `matrix.sh` measures the raw OpenSSH behaviors and a containerized e2e drives the real server, generated script and ssh binary: the first connect prompts once, a second connect prompts zero times (even with askpass removed), the key lands only in the app's agent, and killing that agent makes the next master prompt again. The suite refuses to run outside a container and self-skips, so `npm test` and CI never need docker.

## Also fixed on the way

- Tty-less masters were burning failed password attempts on every reconnect, enough to trip fail2ban.
- `IdentitiesOnly=yes` alongside `-i`, so unrelated agent keys cannot exhaust `MaxAuthTries` first.
- The connect wait is a wall-clock deadline instead of an attempt count, so a passphrase prompt can take minutes without a false failure.
- Concurrent connects for one project coalesce instead of killing each other's masters.
- Cancelling the connect dialog mid-connect no longer leaks its throwaway browse master.
- A SIGTERM quit no longer stalls as a windowless process after finishing its teardown.
- The connection banner shows ssh's real error, SSH tabs get a chip and a `user@host` tooltip, and every connecting state gets a spinner. Two tiny NT_MULTI dev-sandbox chores ride as separate `chore(dev)` commits; happy to split anything out.

## Not in scope

Password-only servers (a tty-less master cannot answer one, and the helper declines non-passphrase prompts), `UseKeychain` (the passphrase should stay a second factor), Windows (the helper is `/bin/sh` + `curl`).

## Testing

`npm run typecheck`, `npm test` and `npm run build` are green (one pre-existing `pty-single-user` teardown flake, identical on a clean main checkout). Unit tests cover the askpass server, prompt classification, retry and cancel attribution, connect coalescing and cancellation, and the agent lifecycle. Docker e2e: 8/8. Also verified by hand against a live server.

## References

- [ssh(1)](https://www.man7.org/linux/man-pages/man1/ssh.1.html): with `SSH_ASKPASS_REQUIRE=force` the helper is used "regardless of whether DISPLAY is set".
- [ssh_config(5)](https://www.man7.org/linux/man-pages/man5/ssh_config.5.html): `IdentityAgent` "overrides the SSH_AUTH_SOCK environment variable"; [1Password's setup](https://www.1password.dev/ssh/get-started/) uses exactly that option.
- [VS Code Remote-SSH](https://code.visualstudio.com/docs/remote/troubleshooting) does the same dance: system ssh, its own prompt shim, retention in the agent ([#10697](https://github.com/microsoft/vscode-remote-release/issues/10697) tracks their IdentityAgent gap).
- Checked npm before hand-rolling: [`ssh-agent`](https://www.npmjs.com/package/ssh-agent) (dormant client), [`node-ssh-agent`](https://www.npmjs.com/package/node-ssh-agent) (not the real protocol), [`ssh-agent-js`](https://www.npmjs.com/package/ssh-agent-js) (keys in process memory), [`ssh2`](https://www.npmjs.com/package/ssh2) (replaces the whole OpenSSH stack).


parsa222:
Pushed the unix socket change from the #70 discussion, plus a hardening pass on it:

The askpass relay now lives on a 0600 unix socket in its own `~/.nodeterm/askpass/` dir (0700), and the helper reaches it with `curl --unix-socket`. Kernel permissions are the access control now; the token stays as a second factor. The passphrase never touches argv or env.

If the socket cannot bind at all, the relay just disables itself and connects fail the old way. It can never block app boot.

The relay is stopped at quit next to the app agent, and the helper's curl uses `-f` so an auth error fails loudly instead of reading as an empty passphrase.

eneskirca (commented):
Reviewed the full diff and ran everything myself. This is exceptional work — thank you.

**Independently verified on my side:**
- `npm run typecheck` clean, full suite 3133 green at your head (docker suite self-skips as designed).
- Your harness, for real: `./run.sh up && ./e2e.sh` → **8/8** against the containerized sshd, including the no-identity-file regression case and the "user's own agent stays empty" isolation proof. The harness scripts themselves are clean (loopback-only publish, throwaway keys, dedicated known_hosts).
- Read `ssh-askpass.ts` and `ssh-agent.ts` line by line. The OpenSSH behavioral notes (byte-identical retry prompt → pid-keyed retry detection; empty answer = abandon key → clean cancel with no latch; the `NumberOfPasswordPrompts=0` trap) all check out, and the failure-attribution work (pid-exact cancel, inflight coalescing across projects sharing a key, stale-attempt ownership checks in `connectOnce`) is more careful than most first-party code.

**On the deliberate difference (app-private agent, login agent never consulted):** accepted. Your reasoning holds — `SSH_AUTH_SOCK` names one socket so "there is no both"; nothing nodeterm unlocks outlives it; agent-only credential users (1Password/Secretive/smartcards) are covered by `IdentityAgent` semantics plus the error hint. The one-prompt-per-run cost for login-agent users is disclosed and fair. The `PasswordAuthentication=no` fail2ban fix is a genuinely valuable bonus — we've been burned by exactly that class of problem before.

**Two follow-ups, neither blocking:**
1. The renderer-side cancel test you mentioned — happy to take it in this PR or right after, your call.
2. Quit now `-O exit`s daemonized masters, which also retires the warm-orphan adoption path for unencrypted-key users (next launch pays a fresh ~1s connect). Consistent with "nothing outlives the app" and fine by me — just noting it's a deliberate behavior change so it doesn't read as an accident later.

CI is green (quality + CodeQL + dependency review). Marking approved once you flip it out of draft — or say the word and it goes in as is.

parsa222:
i'll add the test cases push the new updates soon

parsa222:
Renderer test is in (bc38b33): cancelling during the connecting step and unmounting mid connect both tear the browse master down, and moving the mark back after the await fails both cases. First component test in the repo, so jsdom rides in as a devDependency, scoped per file with a vitest environment pragma (every other suite stays on node) and the renderer include glob now accepts .tsx.

second point: the -O exit at quit is intentional, exactly as you read it. A relaunch paying a fresh connect (and one prompt, for an encrypted key) is the price of quit meaning the key is forgotten.... adopting a warm orphan would silently skip both.

ready to be merged after review.

eneskirca:
The renderer test made it in — `bc38b33` landed inside the merge window, it's on `main` (fbb6a0e) and ran green in both my local full-suite run and CI. Nothing left owed on this one. Thanks again for the quality of this whole contribution — looking forward to the next one.

PR #79, opened 2026-08-04
fix(canvas): Pasting an image into a terminal silently does nothing in the browser (Server Edition)

dattran1232003:
## Environment

- nodeterm 0.2.34, Server Edition (`npm run server:dev`) on macOS 15 (Darwin 24.6.0)
- Client: Chrome on Windows 11, page at `http://localhost:8443` via a VS Code Remote-SSH port
  forward. (Port Forwarding)

## Steps to reproduce

1. Start the Server Edition and open it in Chrome on another machine.
2. Copy an image to the clipboard **as image data, with no text flavour** — e.g. `Win+Shift+S` on
   Windows, or "Copy image" from another app.
3. Click into a terminal node so the caret blinks (focus lands on `.xterm-helper-textarea`).
4. Press Ctrl+Shift+V.

## Expected

Same as a drop: the image is uploaded to the machine running the terminal and its absolute path is
pasted into the terminal, so an agent CLI can read it.

## Actual

Nothing happens. The paste is a silent no-op.

chrome_aPE59YfUPt

## Root cause

`pastedFiles()` reads `clipboardData.files` and `clipboardData.items`. With an image-only
clipboard and focus on xterm's helper `<textarea>`, Chromium delivers **both of them empty**, so
`pastedFiles()` correctly returns `[]` and `onBodyPaste` early-returns.

This is consistent with Chromium populating `clipboardData` only with formats the paste *target*
can accept: a plain `<textarea>` holds text and nothing else, and an image-only clipboard has no
text flavour that survives that filter. It is presumably why applications that accept pasted
images focus a `contenteditable` rather than a textarea.

The async Clipboard API is not filtered by the focused element and does see the image.

# Suggested fix

When a paste carries **neither files nor text**, fall back to `navigator.clipboard.read()`, which
is not filtered by the focused element, and feed any image it returns into the existing
`insertFiles` path.

Restricting the fallback to the "carried nothing at all" case keeps it from touching ordinary text
pastes, and it needs no `preventDefault()` — in that branch there is no text for xterm to paste
anyway, so a clipboard that turns out to hold no image stays the no-op it already was. The read
requires a secure context and the `clipboard-read` permission; failing open on either leaves
current behaviour unchanged.

I have this implemented and verified end-to-end against the setup above (image pasted, uploaded to
`~/.nodeterm-server/uploads/<token>/pasted-<ts>.png`, path inserted, agent read the file), with
unit tests for the new helpers. Happy to open a PR.

### DIFF
--- a/src/renderer/components/kanban/ModalTerminal.tsx
+++ b/src/renderer/components/kanban/ModalTerminal.tsx
@@ -10,7 +10,7 @@ import { useSession } from '../../session/session'
 import { useSettings } from '../../state/settings'
 import { useTerminalSearch } from '../../terminal/useTerminalSearch'
 import { LocalTransport } from '../../terminal/local-transport'
-import { droppedPaths, pastedFiles } from '../../terminal/file-drop'
+import { clipboardImages, droppedPaths, pasteHasText, pastedFiles } from '../../terminal/file-drop'
 import { parseOsc52 } from '../../terminal/osc52'
 import {
   attachReplay,
@@ -345,10 +345,17 @@ export function ModalTerminal({ nodeId, spawn, searchOpen, onCloseSearch }: Moda
   // Capture phase, because xterm's own paste listener sits on the textarea below this wrapper.
   const onPaste = (e: React.ClipboardEvent) => {
     const files = pastedFiles(e.clipboardData)
-    if (!files.length) return
-    e.preventDefault()
-    e.stopPropagation()
-    void insertFiles(files, { raiseWindow: false })
+    if (files.length) {
+      e.preventDefault()
+      e.stopPropagation()
+      void insertFiles(files, { raiseWindow: false })
+      return
+    }
+    // Files nor text — the filtered image-only clipboard. Same fallback as the canvas node.
+    if (pasteHasText(e.clipboardData)) return
+    void clipboardImages().then((images) => {
+      if (images.length) void insertFiles(images, { raiseWindow: false })
+    })
   }
 
   return (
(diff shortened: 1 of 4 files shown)


eneskirca (commented):
Reviewed and verified locally — this is a clean fix for a real Server Edition gap.

**What I checked:**
- Root cause holds up: Chromium populates `clipboardData` only with formats the paste target accepts, and xterm's helper `<textarea>` accepts text — so an image-only clipboard arrives empty and the old handler correctly (but uselessly) no-ops.
- The fallback gating is exactly right: it fires only when the paste carried **neither files nor text**, so ordinary text pastes never touch the async API, and skipping `preventDefault()` in that branch is sound — there's nothing for xterm to paste anyway, so a no-image clipboard stays the no-op it already was.
- `clipboardImages()` fails open on every edge (API absent, insecure context, permission declined) with tests for each; `navigator.clipboard.read()` is called synchronously inside the paste handler, so transient activation is intact.
- Desktop is a strict non-regression: the same filtered-clipboard behavior exists in Electron, so the fallback either fixes image paste there too or fails open to today's no-op.
- Both paste surfaces covered (canvas node + kanban modal terminal), reusing the existing `insertFiles` upload path — no new plumbing.
- Typecheck clean, full suite 3271 green at your head, and all four checks pass after the sync.

One micro-nit, take it or leave it: the ModalTerminal comment reads "Files nor text" — "Neither files nor text" parses better.

Nice first contribution — the fail-open discipline and the test for the permission-denied path are exactly this repo's style.

PR #90, opened 2026-08-09
Add two-way GitHub Issues sync to Kanban

anonymort:
## Summary

- add opt-in GitHub issue cards to project Kanban boards
- support GitHub CLI or a write-only fine-grained token
- map exact issue labels to columns and sync moves, close, and reopen in both directions
- add All, GitHub, and Sessions source filters plus grouped local and GitHub label filters
- keep approval, credentials, cache, polling, and GitHub HTTP on the trusted project host
- support desktop, Server Edition, and project-scoped relay sessions
- keep empty Kanban columns aligned to the same content-driven height

## Safety and behaviour

GitHub remains canonical. Private issue data is held in a bounded mode 0600 host cache and can be read offline after local approval. Relay requests are restricted to the shared project and deltas are sent only to matching subscribers. Incomplete repositories and incomplete configuration are read only.

Cache deletion uses a repository barrier and drains active writers before deleting private data. Polling fails over between approved projects that share a repository. Subscription ownership remains isolated across Electron, relay, and Server Edition hosts.

## Validation

- `npm run typecheck`
- `npm test` – 4,046 passed and 8 skipped
- focused GitHub, relay, Settings, and Kanban tests – 150 passed
- `npm run build`
- `npm run server:build`
- desktop, 390 px mobile, Settings, source filtering, issue movement, theme switching, keyboard focus, modal focus, and equal empty-column height verified in Chrome
- Sol closing review – PASS


eneskirca (commented):
## Review: two-way GitHub Issues sync on the Kanban board

**Verdict: needs work** — not because the design is wrong, but because two of the findings below will burn a user's GitHub rate limit under ordinary use, and one lets a stray drag reopen a closed issue with no confirmation.

Whether nodeterm should carry GitHub Issues on the board at all is a product call for the maintainer, and I'm not arguing either side of it. This review is about whether the implementation is safe if the answer is yes.

Read the diff at `7357682` (69 files, +8435/−63), not the description. Gates (typecheck / build / 4046 tests) were already verified separately, so I spent the time on judgement instead.

### What's genuinely well built

- **The token never reaches a renderer.** `GitHubAuthStatus` carries `tokenPresent: boolean`, `storage`, and `login` — never the value. The `token` field lives only inside `ResolvedGitHubCredential` in core.
- **The seam is right.** All the logic sits in `src/core/github/*` behind `CorePlatform`; only the two shell-specific secret stores are outside it. Server Edition really gets the feature, not a stub — that's the CLAUDE.md rule honoured rather than paid lip service.
- **`client.ts` response validation is unusually careful**: origin-pinned `html_url` / `api_url` / `avatar_url`, `redirect: 'manual'`, bounded streaming body reads, PR filtering, bounded error-metadata reads.
- **The relay scope guard is correctly built**: it refuses *before* dispatch (`relay-host.ts:288`), covers `cast` as well as `req` (`:306`), and handles both argument shapes across all seven channels. `githubControl:` is separately blocked at `platform-electron.ts:90`, and the settings section refuses to render for `project.remote` (`GitHubIssuesSection.tsx:134`). That's the board-log discipline applied properly.
- Issue bodies render as plain text (`GitHubIssueSummaryModal.tsx`), not markdown — no injection surface.
- Incidental fix worth calling out: `lib/kanban.ts` `deleteColumn` now spreads `...k`. Before this PR it returned only `{columns, assignments}`, so **deleting a column silently dropped the board's `meta` and `labels`**. That's an unrelated pre-existing bug fixed in passing; it deserves a line in the PR description.

---

## Findings, by severity

### 1. HIGH — every GitHub read re-resolves credentials: 2 `gh` subprocesses + 1 unmetered `api.github.com/user` call, several times per poll

`service.ts:661`, `:696`, and `readWithEpoch` (`:812`) each call `contextForProject` → `host.ts:180` `resolver.resolve()` → `credentials.ts:88-96`: `gh auth status`, then `gh auth token`, then `validate(token)` → `GET /user`. There is no memoization anywhere on that path.

Counting one 60-second poll of a **single-page** repo: 4 credential resolutions → **8 `gh` process spawns + 4 `/user` requests + 1 real issues request**. A first full sync of a 100-page repo does roughly 200 resolutions → ~400 `gh` spawns and ~200 `/user` calls inside one refresh.

Failure scenario: a user opens a board for a repo with ~8k issues. The initial `subscribe` fans out ~200 `/user` requests in a tight loop. GitHub's secondary limiter trips. `validate()` catches the 403 and returns `null` (`credentials.ts:95`), `resolve()` returns `null`, `contextForProject` throws `not-authenticated` — and the board says "GitHub issues unavailable" while Settings says *"Sign in with GitHub CLI or save a valid token first"* to a user who is correctly signed in.

Compounding it: those `/user` calls do **not** go through `GitHubRequestCoordinator`, so they are neither rate-gated nor backed off, and the coordinator can't learn about limits they hit.

This is invisible to the new tests because every one of them stubs `contextForProject` as a free async function (`service.test.ts:139`, `:170`, `:234`, …). The epoch-revalidation design is sound; it's the *cost of evaluating the epoch* that wasn't budgeted. Memoizing the resolved credential per (provider, generation) with a short TTL, and routing `/user` through the coordinator, would fix both halves.

### 2. HIGH — `githubIssues:refresh` is unthrottled and reachable from a relay guest

`handlers.ts:36`; relay path `relay-host.ts:288` → dispatch. The only coalescing is `service.ts:220` (`if (state.refresh) return state.refresh`), which merges *concurrent* calls but not sequential ones. Nothing enforces a minimum interval, and `full: true` is caller-supplied.

Failure scenario: a guest on a shared project — or a renderer retry loop, or the Retry button being held down — calls `refresh(projectId, true)` repeatedly. Each completed call restarts a full reconciliation over every page. With finding 1 that's ~5 API calls per page per pass. A few minutes exhausts the host's 5000/hr primary limit and can secondary-limit the **account**, which degrades the user's `gh` CLI and github.com session too — not just nodeterm. The host's own "Refresh now" button is gated only by local `busy` state.

Needs a server-side minimum-interval gate, tighter for `full`.

### 3. MEDIUM-HIGH — write authority crosses the relay boundary with no separate consent, and the guard is inert on unscoped sessions

`relay-host.ts:223`: `githubIssuesOutOfScope` returns false whenever `opts.sharedProjectId` is falsy. That matches the board-log precedent (an unscoped session is a whole-desktop share, so the registry is the only gate) — consistent, but the consequence is larger here: on an unscoped session a guest reaches `githubIssues:move` / `create-labels` / `clear-cache` for **every approved project**, driving the host's token. Even on a scoped session, the guest inherits full close/reopen/relabel authority. The approval gate (`host.ts:145`) is per project *per machine*; it says nothing about who is driving.

The fair counter, stated plainly: a relay guest already has pty write on the shared project and could run `gh issue close` in a terminal. So this isn't privilege escalation so much as a quieter path to authority the guest already has — but it's one with no confirmation, no audit line, and a phone-sized UI. `docs/github-issues-kanban.md` discloses it ("Relay tabs send issue queries and movements to that host"), so it reads as deliberate. I'd want it to be deliberate on the maintainer's side too, since it's the one place where "the trusted host does the GitHub work" and "only the host decides what GitHub work happens" come apart.

To answer the question as asked: **credentials, cache, polling and GitHub HTTP do stay on the host — that claim holds.** What does not stay on the host is the *authority to trigger writes*.

### 4. MEDIUM-HIGH — a drag closes or reopens an issue with no confirmation, and "Ungrouped" is a reopen target

`KanbanView.tsx:255` (`dropOnColumn`) and `:275` (`dropAtCard`) call `moveGitHubState` directly — no dialog. `service.ts:333`: `state: toColumnId === completionColumnId ? 'closed' : 'open'`.

Failure scenario: closed issues all land in the completion column (`service.ts:112`), so that column is a pile of historical closed issues. The user drags a card and releases slightly left, onto Ungrouped. `toColumnId` is `null`, so the write is `state: 'open'` plus a label array with every mapped label stripped — a two-year-old closed issue is **reopened on GitHub, relabelled, and every subscriber is emailed**. No undo, no confirm.

This is below the bar the codebase sets elsewhere: worktree Merge always confirms, node Delete always confirms, and both are less outward-facing than reopening someone else's issue. Minimum fix: confirm any write that changes issue *state*. Arguably a drop on Ungrouped should be a no-op rather than a reopen.

### 5. MEDIUM — the column mapping and completion column are git-shared but are not covered by the approval binding

`shared/types.ts:352` puts `github` inside `ProjectKanban`, i.e. in `.nodeterm/project.json` — shipped to everyone who clones. The approval (`control-store.ts:120`) binds `localApprovalId + projectId + repository`. Changing the **repository** correctly invalidates approval. Changing `completionColumnId` or `columnMappings` does not.

Failure scenario: a teammate pushes a `project.json` moving `completionColumnId` from "Done" to "In Progress". You `git pull`. Nothing re-prompts. The next card you drag into In Progress closes the issue on GitHub. `config.revision` (`config.ts:96`) is in the epoch, so *in-flight* operations abort — but the new config is authoritative for the next operation, with no re-approval and no notice.

This is the same shape as the git-shared `defaultPermissionMode`, which the codebase handles by warning in the tab menu. Either fold the config revision into the approval, or surface a "the GitHub mapping changed" notice the way the permission-mode override is surfaced.

Smaller, related: `GitHubIssuesSection.tsx:50` defaults `completionColumnId` to the **last** column on enable. Default boards end in "Done" so it's usually right, but a board ending in "Icebox" or "Blocked" silently acquires a close-on-drop column. The Select does display it, so this is a footgun rather than a trap.

### 6. MEDIUM — ETag support is plumbed through types, client and cache but never actually used

`service.ts:657` declares `const etags: Record<string, string> = {}`, saves it into the snapshot at `:698` — and never writes a key into it. `listIssues` is never called with `etag` (`:666`), so the client's 304 branch (`client.ts`) is dead on the poll path, and `cache.ts` carefully validates a field that is permanently `{}`.

GitHub does not count 304 responses against the primary rate limit. As written, every 60-second poll spends real budget even when nothing has changed. That's pure loss, and all the plumbing to fix it is already in place — which suggests it was intended and dropped rather than decided against.

### 7. MEDIUM — `localApprovalId` is minted at load but only persisted on the next workspace save

`workspace-store.ts:114` mutates the in-memory index (`entry.localApprovalId ||= randomUUID()`); it is written to disk only through the `save` path at `:222`.

Failure scenario: a user approves GitHub for a project, uses the board read-only, and quits without any change that marks the workspace dirty. On relaunch a **new** id is minted, `isApproved` (`control-store.ts:120`) no longer matches, and the board reports "Approval is required before nodeterm reads the repository." The stale approval row stays in `github-issues-control.json` as unreachable garbage, accumulating one row per restart.

It fails *closed*, which is the right direction. But it's a confusing re-prompt loop, and reading alone I could not establish that a load is always followed by a save in practice.

### 8. LOW-MEDIUM — revoking a machine leaves the private issue cache on disk in the clear

`host.ts:154` clears the approval and cancels in-flight work but never touches the cache. `cache.ts` writes plaintext JSON (mode 0600) under `github-issues-cache/`, containing full titles and bodies of what may be a private repo. The dialog says *"Stop this computer from reading or changing issues for this project?"* — after which the app genuinely can't read them (`projectContextForCache` throws `not-approved`), but the file remains. The docs state this is intended; the dialog wording just promises more than it delivers. Offering "also clear cached data" on revoke would close the gap.

### 9. LOW — an uncapped rate-limit sleep, and the user-visible symptom of a rate limit is "you're signed out"

`request-coordinator.ts` `waitForRate` loops on `retryAt` with no ceiling; `retryAt` comes from `retry-after` / `x-ratelimit-reset`. Over TLS to `api.github.com` those values are trustworthy, so this isn't an attack surface — but a primary limit legitimately sets a reset up to an hour out, and the read that hit it then sits in `sleep` holding one of four read slots for that hour, with no "rate limited until HH:MM" anywhere in the UI. Meanwhile the 60s poll keeps firing and keeps re-resolving credentials (finding 1), each of which makes a `/user` call that is *also* limited and surfaces as `not-authenticated`. Net effect: hitting a rate limit reads as "you've been signed out."

### 10. LOW — renderer-gone cleanup covers relay peers and server clients, not a crashed desktop renderer

`src/main/index.ts` wires `dropGitHubRelayClient` into `onPeerGone`; `src/server/index.ts` wires `dropClient` into `onClientGone`. There's no equivalent for the desktop main window. Normal board close unsubscribes from the renderer effect, and a reload reuses the same `webContents.id` so the Set dedupes — but a renderer **crash** yields a new id and leaves the old `uiId` subscribed forever, keeping the 60s timer (`service.ts:731`) alive for a viewer that no longer exists.

---

## Polling cost — the direct answer

Better than I expected on the structural question, worse on the constant factor. `KanbanView` is conditionally mounted (`Canvas.tsx`, `{kanbanOpen && <KanbanView/>}`), so **polling only runs while a board is open** — closing the board tears the subscription down and stops the timer. One timer per `(userId, repository)`, not per project, so two projects on the same repo share one poller. That's the right shape.

What's wrong is per-tick cost: with findings 1 and 6, one tick of one board costs ~5 GitHub requests and ~8 `gh` process spawns where it should cost one conditional request that usually returns 304 for free. There's also no backoff when the window is unfocused or the machine is idle.

## Three surfaces

- **Desktop**: wired.
- **Server Edition**: wired, and correctly — `ServerGitHubSecretStore` has no `safeStorage`, declares `availability: 'restricted-file'`, and the UI surfaces that warning. One thing worth stating out loud: `registerServerGitHubControl` exposes token save/clear to any authenticated browser session, so Server Edition's single-user assumption is now also guarding a GitHub write token.
- **Mobile**: **not addressed anywhere.** I grepped the entire diff and both docs — no mention of *nodeterm mobile* / `nodeterm-ios`, and no explicit "not applicable" call. CLAUDE.md asks for that decision to be made consciously rather than left to rot. The iOS board is a read/move mirror, so the question is concrete: does an iOS board show GitHub cards at all, and if a move arrives from the phone, can it reach `githubIssues:move`? A one-paragraph answer in the PR would satisfy the rule.

## What I could not verify without a GitHub token and a real repo

Stating these plainly rather than implying I checked them:

1. **Whether `PATCH /issues/{n}` with an unknown label silently creates it.** My recollection is that GitHub does. If so, "Create missing labels" is a convenience rather than a prerequisite, and an ordinary move can create labels in the repo as a side effect — which widens the blast radius of finding 5.
2. **Real rate-limit behaviour** — whether ~200 `/user` calls in a burst actually trips the secondary limiter, and what `retry-after` values GitHub returns in practice.
3. **Whether the 403-vs-secondary-limit classification in `client.ts` matches reality.** It's inferred from headers plus a regex over the error message (`/secondary rate limit|abuse detection|secondary-rate-limits/`); a wording change on GitHub's side would silently reclassify a secondary limit as `insufficient-permission`.
4. **Whether `gh auth token` resolves in a packaged macOS app.** `credentials.ts:20` prepends `/opt/homebrew/bin:/usr/local/bin:…`, which suggests the author already hit the GUI-apps-don't-inherit-PATH problem, but I couldn't test a packaged build.
5. **Anything end-to-end**: approve → map → create labels → drag → confirm on github.com was never exercised. All 105 new tests run against fakes, including the `contextForProject` stub that hides finding 1.

## Suggested gate for merge

Blocking, in my read: 1, 2, and 4. Findings 5 and 7 are cheap to fix and each turns a silent wrong outcome into a visible one. 6 is a small change that pays for itself immediately. The rest are fine as follow-ups.

eneskirca (commented):
## Addendum — one new finding, and fix directions for three of the existing ones

Follow-up to my review above. One finding I missed on the first pass, then concrete directions for
the three findings where the fix shape matters more than the diagnosis.

### NEW — MEDIUM: `state_reason` is silently destroyed on every close

`shared/github-issues.ts:104` — `UpdateIssueInput` is `{ state?, labels? }`. There is no
`state_reason`. `service.ts:332` builds exactly that and PATCHes it.

Meanwhile `client.ts` *does* parse `state_reason` into `GitHubIssue.stateReason`
(`'completed' | 'not_planned' | 'reopened' | null`), the cache validates it, and it rides all the
way to the renderer — where it is **never rendered**. So today it is read-only dead data, and on the
write path it is data loss.

Concrete scenario: an issue was closed as **not planned** (a wontfix). Closed issues all land in the
completion column (`service.ts:112`), so it shows in Done. Someone drags it out to triage it —
reopened, fine. They drag it back to Done. `service.ts:333` sends `state: 'closed'` with no reason,
GitHub defaults to `completed`, and the "not planned" fact is gone from the issue and from its
timeline. Nobody chose that; it's a side effect of a round trip through the board.

Two ways out. Add `state_reason` to `UpdateIssueInput` and preserve the existing reason on a
re-close; or route **state** through `gh issue close` / `gh issue reopen`, which can express the
full vocabulary the REST PATCH cannot (`--reason completed`, `--reason "not planned"`,
`--duplicate-of N`) while title/body/labels keep using their own endpoints. Either beats dropping
it. Minimum viable fix: when closing an issue that is *already* closed, don't send `state` at all.

### Finding 1 + 2 (rate limit): the budget needs to be *pre-flight*, not post-hoc

`GitHubRequestCoordinator` is purely reactive — it only learns a limit exists after a 429/403 comes
back, which is why nothing stops the unmetered `/user` storm or an unthrottled `refresh` from
spending the budget in the first place. The shape that actually prevents it:

- a per-bucket budget (REST core vs GraphQL vs search — they have separate quotas), cached with a
  short TTL;
- a guard consulted **before** each call that can answer "blocked", so the caller returns a typed
  rate-limited error instead of making the request;
- a spend note decrementing after each call;
- a **single-flight** probe of the quota-exempt `/rate_limit` endpoint to learn the real reset time,
  rather than inferring it — compare `client.ts`'s doubling `secondaryBackoffMs`, which is a guess
  by construction, and a 403 burst that currently guesses once per response.

Related, since it belongs to the same "which quota am I spending?" question: this PR assumes
`api.github.com` everywhere (`client.ts` hard-pins `API_ORIGIN`), so GitHub Enterprise is out of
scope. That's a fine v1 call — it's just worth stating in the docs rather than leaving users to
discover it from a validation error.

### Finding 9: a rate limit currently reads as "you're signed out"

The fix is one shared classifier that maps an error to `rate_limited | network | server_error` and
returns **null** for auth / permission / 404 — because those are user-actionable and genuinely
different from "GitHub is unreachable". Shared between core and renderer, so the two surfaces can
never disagree about whether GitHub is up.

That distinction is exactly what this PR loses when `validate()` swallows a 403 into `null` and the
user is told to sign in. And the detection needs to be wider than what's here: `client.ts` currently
separates a secondary limit from `insufficient-permission` using only
`/secondary rate limit|abuse detection|secondary-rate-limits/`. Since that regex is the *only* thing
standing between the two outcomes, a wording change on GitHub's side silently reclassifies a
throttle as a permissions error. Match on the 429 status and the "rate limit" family too, not just
the secondary-limit phrasings.

### The secret store should distinguish "no token" from "cannot decrypt"

`ElectronGitHubSecretStore.readForHost()` catches a `safeStorage.decryptString` failure and returns
`null` — indistinguishable from "no token saved", so it too surfaces as "not authenticated".

The failure mode this hides is real on macOS: the user denies the OS keychain prompt after the app
is re-signed. The token is intact, the keychain ACL isn't, and telling the user "sign in again"
makes them destroy a working credential. A distinct decryption error is worth the few lines. Worth
a concurrency test too, since `atomicWrite` in `main/github-control.ts` has no write-race guard of
its own.

### One design note behind finding 4

The risk that a drag silently closes or reopens an issue is inherent to putting issues on a
drag-and-drop board — it isn't a bug in this implementation so much as the cost of the interaction.
Boards that mutate issues through explicit edits (a cell, a menu, a confirmed action) don't carry it
at all. Whether nodeterm should take that cost on in exchange for the drag is the maintainer's call
and I'm not arguing it either way; if the answer is yes, a confirmation on state-changing moves is
what makes it survivable.

### Revised summary

Blocking, unchanged: **1** (per-request credential resolution), **2** (unthrottled `refresh`),
**4** (unconfirmed state-changing drag). The new `state_reason` finding is not blocking on its own,
but it is cheap to fix and it is silent data loss — the kind that gets discovered months later by
someone wondering why their wontfix pile turned into "completed".

eneskirca:
## Blocking findings addressed in `0942e5f` — merging

I've pushed fixes for the four blocking findings rather than deferring them, since two of them
could spend a user's GitHub **account** quota (not just nodeterm's) and one could reopen a closed
issue from a mis-aimed drag. Each fix has a test that was watched failing first.

**1. Credential resolution is memoised (30s, per provider).** Every epoch re-check called
`resolve()`, which spawned `gh auth status` + `gh auth token` *and* spent a `GET /user` — requests
that bypass the request coordinator entirely, so nothing rate-limited or backed them off. Every
in-app credential change drops the memo through the existing `onCredentialBoundaryChange` hook
instead of waiting out the TTL. `integration.test.ts` proves both halves: the memo saves the second
`/user`, and `clearToken()` invalidates it immediately rather than keeping a revoked credential
alive.

**2. Caller-driven refresh is floored** — 30s incremental, 2 minutes full. Three details are
load-bearing and are commented as such: the floor is checked *before* `contextForProject` (or the
expensive half of every rejected call is still paid); the background poll goes *under* the floor,
since it is already paced by `POLL_MS` and routing it through broke its per-project fallback — a
throttled call returns without throwing, which that loop reads as "this project worked"; and a
refresh that **failed** does not hold the floor, so Retry stays responsive after a network blip.

**3. Close and reopen ask first.** The decision is the pure `githubMoveIntent`
(`renderer/lib/githubIssueMove.ts`), and all three move paths — drop, the card's Move selector, the
summary modal — now go through one funnel so none can skip what the others ask. Label-only moves
stay immediate. Dropping a card back where it already sits writes nothing at all.

**4. `state_reason` is preserved.** `state` is sent only when it actually changes, since GitHub
rewrites `state_reason` for any write carrying it — re-closing an already-closed issue was silently
turning a deliberate `not planned` into `completed`.

Docs updated for all four. Gates on the merged result (main merged in, 5 commits): typecheck clean,
build clean, **4081 tests pass**.

### Deliberately left as follow-ups

- **ETag conditional requests** (finding 6) — `etags` is still declared, persisted and never
  populated, so each poll spends real quota where a 304 would be free. Small change, but it belongs
  with a proper look at the poll path rather than bolted onto a review fix.
- **Approval does not cover the column mapping** (finding 5) — `completionColumnId` travels in
  git-shared `project.json`, so a teammate's push can change which column closes issues without
  re-prompting. The confirmation dialog now blunts the worst of it (you are told the issue is about
  to close), but binding the config revision into the approval is the real fix.
- **`localApprovalId` persistence** (finding 7), **cache retained after revoke** (finding 8),
  **crashed-renderer subscriber leak** (finding 10), and **`stateReason` round-trip preservation**
  across a reopen→re-close cycle — the last needs `gh issue close --reason`, since the REST PATCH
  cannot express the reason vocabulary.
- **Mobile surface** — still undecided in writing. Worth a line in the docs before the iOS board
  grows a GitHub column.
- The relay guest write-authority question (finding 3) is a product call, not a defect, and it is
  documented behaviour.

Thanks for a genuinely careful PR — the response validation, the epoch discipline and the relay
scope guard were all well above the bar; the gaps were in cost and in confirmation, not in
structure.

PR #98, opened 2026-08-09
feat(canvas): add linked agent inbox notifications

brendon-jewell:
## Summary

- add an opt-in `notify --node <id>` canvas-control verb for linked agent coordination
- deliver only a fixed Nodeterm-authored inbox prompt to context-linked agent targets
- rate-limit each source-target pair while retaining confirmations for arbitrary `write` and `close` commands

## Validation

- `npm test` (4,220 passed, 8 skipped)
- `npm run typecheck`
- `npm run build`

### DIFF
--- a/docs/superpowers/specs/2026-08-09-agent-notify-control-design.md
+++ b/docs/superpowers/specs/2026-08-09-agent-notify-control-design.md
@@ -1,0 +1,24 @@
+# Agent Notify Canvas Control
+
+## Goal
+
+Let a canvas-control-capable agent wake a context-linked agent after updating an external coordination channel, without approving arbitrary terminal input.
+
+## Command
+
+`nodeterm notify --node <id>` sends a fixed Nodeterm-authored prompt to the target. The command accepts no message text.
+
+## Trust boundary
+
+- Disabled by default and enabled in Settings > Notifications.
+- Source must pass the existing canvas-control authorization check.
+- Target must be a context-link-capable agent in the active project.
+- Source and target must have a persisted context link.
+- Each source-target pair is limited to one notification every 10 seconds.
+- `write` and `close` retain their confirmation dialogs.
+
+## Prompt
+
+The target receives: `[nodeterm] A linked agent updated shared coordination context. Check your configured inbox before continuing.`
+
+The app owns the entire prompt so the source cannot inject instructions through command arguments.
--- a/src/main/canvas-control-core.test.ts
+++ b/src/main/canvas-control-core.test.ts
@@ -28,6 +28,18 @@ describe('parseControlRequest', () => {
     })
   })
 
+  it('requires a target for notify and does not accept message text', () => {
+    expect(parseControlRequest('notify', {})).toEqual({ error: 'notify requires --node <id>' })
+    expect(parseControlRequest('notify', { node: 'n1' })).toEqual({
+      verb: 'notify',
+      args: { node: 'n1' }
+    })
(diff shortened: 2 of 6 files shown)


eneskirca:
Two notes from a batch review of the open PRs.

**The rate limit is bypassable by concurrent requests.** `Canvas.tsx:6400` writes the throttle timestamp *after* `await api.pty.sendText(...)` at `:6395`. Nothing serializes control requests — the hook server awaits its handler per HTTP request on a concurrent `node:http` server — so N requests fired together all read `lastSentAt` at `:6389` before any of them records one, and all N pass. Each delivered line is a submitted turn in the target. A failed send also records nothing, so an unreachable target can be retried without limit. Moving the `set` above the await fixes both.

**The link check rests on an asserted identity.** `sourceNodeId` reaches the handler from the request body, and the hook token is one value shared by every spawned session, so an agent can name a source it is not. Combined with `list` (which returns every node id on the project), an unlinked agent can iterate ids as claimed sources until one satisfies the link test. The payload here is a fixed string, so this is noise rather than takeover — but it is the same gate #112 and #113 now lean on much harder. I have written the shared analysis and the fix (bind the hook token per node, resolve the acting node from the token instead of the body) on #112 and #113; this PR benefits from the same change rather than needing its own.

Worth deciding alongside #113: that PR covers the same ground with arbitrary-text `write` plus per-pair grants that are pruned when a node is deleted — which also solves the never-pruned `agentNotifyAtRef` map here. The two overlap enough that taking them together, on top of the token fix, is probably cleaner than landing both as-is.

Verified by reading `Canvas.tsx:6366-6410`, `hook-server.ts:171/180/453`; not reproduced with a crafted POST.

eneskirca:
Thank you for this, and sorry it sat open.

I am not going to merge it, but not because there is anything wrong with it. The agent-messaging design I am implementing right now covers the same ground, and it folds in both of the ideas that make this PR work:

- the **`notify` verb** as the shape for an agent telling a linked node something, rather than overloading `write`
- the **10 second throttle**, which is the right instinct — an inbox that can be flooded by a loop is worse than no inbox

Both are going in with credit to you. I would rather do that than have you rebase a 111-line PR onto a design that is about to move underneath it.

If you want to stay involved in the messaging work, say so and I will loop you in on the design before it lands — you clearly thought about the failure modes here.

brendon-jewell:
@eneskirca, thank you for the reply! Glad to see the feature implemented, and I was happy to contribute. I have thoroughly enjoyed using nodeterm thus far and am excited for things to come. Keep up the great work!

eneskirca:
Closing this as superseded — with thanks: the idea was right, and it's now in the product.

The `notify` verb landed in **#212** ("the send, reply and notify verbs"), and the commit that added it explicitly folds in this PR's design: `notify --node <id>` delivers a **fixed, app-authored** inbox prompt (any caller-supplied body is discarded — there is a test pinning that a hostile body never reaches the pane), goes through the one rate-limited delivery pipeline with typed outcomes, and is documented in the agent-facing skill text. The trust model around it came with **#237** (per-project `agentMessaging` capability switch + machine-local consent), and later waves (#459, #492) hardened the delivery path itself.

So everything this PR set out to do — linked-agent nudges without free-text injection, rate-limited, confirmation-free — ships today, just through the shared messaging pipeline rather than a standalone verb. Thank you for pushing on it early; the fixed-body design in particular survived into the final shape.

If you spot a gap the current `notify`/`send` mechanics don't cover, please open a fresh issue against today's `main` — happy to look.

PR #105, opened 2026-08-10
fix(workspace): propagate transient index read errors; sideline unrecognized shapes

jasondwhyte:
> **Reworked after v0.3.0** — the original version of this PR predated your corrupt-index recovery (sideline + recovery note + boot sweep), which now covers most of what it did, and covers the crash-between-renames note case mine never did. What remains is the delta: two gaps in the recovery as shipped, one of which quietly reintroduces the data-loss path this PR was born to close.

## 1. A transient read failure still clobbers the index

`loadInner`'s read catch treats every failure as "no index". For ENOENT that's right. But an EACCES/EIO boot degrades to `EMPTY_WORKSPACE`, and the renderer's unconditional post-load save writes that empty index over the original — one unreadable moment becomes permanent loss of every inline project, which exists nowhere else on disk. The store's own SSH-mirror rule — "a failed read is never evidence of absence" — applies to its own index too: non-ENOENT read errors now propagate. Every existing caller already catches a rejected load, and the renderer's bare `.then()` simply never hydrates, so no boot save fires and the file survives to the next boot.

## 2. Valid JSON in an unrecognized shape bypasses the sideline

A hand-edited or future-version index that parses but matches no shape any version ever wrote fell through `migrateLegacy` to a silent empty workspace — the same overwrite-the-only-copy path, with no sideline and no recovery note. It now takes the same sideline + note road as unparsable JSON.

## Tests

EACCES propagation (original bytes provably never clobbered), unrecognized-shape sideline (backup intact, note fired), ENOENT staying a quiet fresh start, and a pin that overlapping saves never reuse a tmp name — sequential under the new saveChain by design; uniqueness still protects the chain-bypassing writers (second instance, SSH poll) and the crash window. 67/67 green with your existing suite, typecheck clean.

### DIFF
--- a/src/core/workspace-store.test.ts
+++ b/src/core/workspace-store.test.ts
@@ -26,6 +26,7 @@ beforeEach(async () => {
   initPlatform(fake)
 })
 afterEach(async () => {
+  vi.restoreAllMocks()
   resetPlatformForTests()
   await fs.rm(userData, { recursive: true, force: true })
   await fs.rm(projRoot, { recursive: true, force: true })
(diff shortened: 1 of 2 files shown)


jasondwhyte:
Reworked in place: rebased onto current main and reduced to the delta your v0.3.0 recovery doesn't cover — see the updated description. The original single-commit version this replaces is preserved in this branch's history (b9136cb) if you want the archaeology. Same deal as #106/#107: maintainer edits welcome, and if the fork's branch protection blocks you again, a mirror PR closing this one is perfectly fine by me.

eneskirca:
Thanks @jasondwhyte — reviewed this properly, and the diagnosis is right even though the patch can't land as-is.

**Why it can't merge:** the branch is cut ~250 commits behind `main`, and everything it claims has since shipped: `b64ce692` (unique tmp names + save serialization + empty-write guards), `c634f018` + `2a833027` (corrupt-index sideline **with** a user-facing recovery note carrying the backup filename), `789152ce` (cleanup-on-failure + stale-tmp sweep), and `66a53e33` — which is your own #107 idea, landed. `git apply --3way` onto current main conflicts in `src/core/workspace-store.ts`, main already pins the sideline / read-only / tmp-uniqueness / note behaviours in `workspace-store.test.ts:878-1050`, and this branch's concurrency test deadlocks against main's `saveChain` serialization (measured: times out at 5000 ms).

**But one part of your diagnosis is still an open data-loss path on main, and it's the part worth having.** A **non-ENOENT** read failure (EACCES / EIO / EMFILE) still degrades to an empty index, sidelines nothing, and the boot save writes over a populated `workspace.json`. Main's empty-write guard can't save it: its disk re-read fails with the same errno and the catch falls through to "an empty write is fresh". Measured against `origin/main` with a mocked errno: a populated index became `{"version":3,"activeProjectId":"","entries":[]}` and no `.corrupt-` file was produced.

**If you're up for a small rebased PR, this is the shape:**
1. Propagate a non-ENOENT read failure instead of degrading to empty.
2. A **write-side latch** — propagation alone is only half the fix. With your `loadInner` grafted onto main I measured that the rejected load leaves the file intact, but the *first user action* (creating a project on the Welcome screen the silent failure just put in front of them) writes a fresh index over it. Latch "this store never read the index" and refuse **any** index write, not just the zero-project one.
3. A user-visible note — reuse the existing `workspace:corrupt-recovered` channel rather than a silent rename. Silence is the failure mode this repo has already been burned by (a stray empty workspace.json once blanked a phone browser and looked exactly like a first run).

Do **not** carry over the `writeAtomic` hunk (main has it plus a stale-tmp sweep) or the concurrency test (it deadlocks now).

Closing this one as superseded — the useful remainder is small and clearly separable. Genuinely good catch on the class of bug; #107 landing under a different hand doesn't change that you found it.

PR #106, opened 2026-08-10
fix(pairing): serialize agent.json and authorized_keys mutations

jasondwhyte:
## Problem

Both pairing-service entry points do unserialized read-modify-writes over whole files, from independent event sources (`pairing:revoke-device` IPC; the pairing POST on its own connection). Two overlapping mutations each read the original file and write back a copy carrying only their own change:

- Two rapid revokes: the loser's stale read republishes the device the winner just revoked — key line and agent token both — so a revoked phone silently keeps SSH and host-agent access.
- A pairing completing mid-revoke can resurrect the revoked device's `agent.json` entry, bearer token included.

## Fix

All mutations run through one closure-level chain (the `SshStore` `writeChain` idiom). Revoke's two-file read-modify-write (`agent.json` + `authorized_keys`) is one unit; the pairing POST's append+persist is another. A failed mutation neither cancels nor surfaces on the ones queued behind it, while each caller still sees its own failure.

## Verification

The revoke-vs-revoke race test goes red against the unserialized code with the exact symptom (revoked key republished). The pairing-POST half is exercised over the real HTTP listener. The suite also guards its own fs mocking: every path is derived from `os.homedir()` at module load, so the tests refuse to run outside an `nt-pairing-*` temp home — a silently failed mock can never touch a real `~/.ssh`. Typecheck clean.

### DIFF
--- a/src/main/pairing-service.test.ts
+++ b/src/main/pairing-service.test.ts
@@ -1,0 +1,279 @@
+import { describe, it, expect, beforeEach, afterEach, afterAll, vi } from 'vitest'
+import { promises as fs, mkdirSync, readFileSync, rmSync, statSync, writeFileSync } from 'fs'
+import { randomBytes } from 'crypto'
+import { request as httpRequest } from 'http'
+import path from 'path'
+
+const TEMP_HOME_MARKER = 'nt-pairing-'
+
+// pairing-service computes AGENT_DIR / AGENT_JSON_PATH / AUTH_KEYS_PATH from `os.homedir()` at
+// MODULE LOAD time, so the redirect has to be in place before the import — hence the mock factory
+// (hoisted by vitest) minting the temp home itself. `networkInterfaces` is faked too so start()
+// finds a LAN IP on a CI box with none. Everything else about `os` stays real.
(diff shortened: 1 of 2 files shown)


eneskirca:
Merged as 6957cde via #116, with your commit `f87945a` carried verbatim and authorship preserved — thank you, this closed a real hole (a revoked phone keeping its SSH key and bearer token). I verified the race independently: your revoke-vs-revoke test goes red against the pre-PR code with the exact republish symptom.

I could not push the rebase to this branch — the fork rejects maintainer pushes despite `maintainer_can_modify: true` — and the branch was behind a protected main, hence the mirror PR.

One review fix rode along: `revokeDevice` dropped the `agent.json` entry before rewriting `authorized_keys`. On a partial failure that hides the device from the UI while its SSH key is still live, and `PhoneSection` has no `catch`, so the owner sees no error and has no Revoke button left. Reversed so the bigger capability is cut first. Your serialization is what made that reversal safe to reason about.

Noted as follow-ups, not addressed: the chain is per-process (a second instance or the host agent can still race a revoke), and revoking a device does not revoke its minted `relayDeviceToken`.

Closing in favour of #116.

PR #107, opened 2026-08-10
fix(stores): unique per-call tmp names for atomic writers

jasondwhyte:
> **Stacked on #106** — its commit is included here because the atomic-write test suite asserts post-serialization behavior. Merging #106 first shrinks this PR to its own two commits.

## Problem

Eight atomic writers used a fixed `${file}.tmp`. Wherever two writes can overlap, they interleave in one temp file: one writer's rename publishes the other's half-written bytes, or moves the tmp out from under it so the loser's rename fails. Affected: `settings-store`, `usage/provider-cookie`, the GitHub PAT stores (`main/github-control`, `server/github-control`), `github/cache`'s `writePrivate`, pairing's `agent.json` / `authorized_keys` writers, and `remote/approved-devices`.

## Fix

- Every site moves to `<file>.<pid>.<seq>.tmp` with remove-own-tmp-on-failure and the error rethrown (each caller keeps its existing error contract).
- Credential files get an orphan-tmp sweep on save and clear — the legacy fixed name plus foreign-pid leftovers — because a leaked tmp of a PAT store or `agent.json` is a leaked secret. Own-pid temps are never swept.
- A failed `approved-devices` save provably leaves the old pinned list intact, which `revocation.ts` documents relying on.
- `control-store` and `ssh-store` are deliberately untouched: both serialize every write through an internal queue. `control-store` now carries a comment stating why its fixed name is load-bearing-safe, so a future caller that skips the queue knows what it's breaking.

## Verification

16 new tests across five suites, each watched to fail against the unfixed code first (shared-tmp races red with the exact interleaving symptom; sweep tests red on leftover temps). Combined run with the serialization suite: 50/50 green, typecheck clean.

### DIFF
--- a/src/core/github/cache.test.ts
+++ b/src/core/github/cache.test.ts
@@ -1,8 +1,8 @@
-import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { promises as fs } from 'node:fs'
 import os from 'node:os'
 import path from 'node:path'
-import { GitHubIssueCache, type GitHubCompleteSnapshot } from './cache'
+import { GitHubIssueCache, type GitHubCacheDocument, type GitHubCompleteSnapshot } from './cache'
 import type { GitHubIssue } from '../../shared/github-issues'
 
 let userDataDir: string
@@ -35,6 +35,7 @@ beforeEach(async () => {
 })
 
 afterEach(async () => {
+  vi.restoreAllMocks()
   await fs.rm(userDataDir, { recursive: true, force: true })
 })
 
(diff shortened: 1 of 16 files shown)


eneskirca:
Merged as 66a53e3 via #117 — both your commits cherry-picked verbatim with authorship preserved. Thank you.

Same reason as #106: the fork rejects maintainer pushes despite `maintainer_can_modify: true`, so the stacked branch could not be rebased in place. #106 landed first (6957cde), which reduced this to its own two commits.

I checked the two things that actually mattered here rather than taking the description's word for them:
- The sweep cannot delete a live write — own-pid temps are skipped, so a concurrent same-process writer sitting between `writeFile` and `rename` is safe.
- Secret temps are created `0600` **in the open(2) call**, before any secret byte lands, so there is no window where a PAT is world-readable.

Your call to leave `control-store` and `ssh-store` on fixed names checks out: both genuinely serialize through an internal queue.

One fix rode along. `settings-store` was the single writer in this family passing no `mode`, so settings.json landed 0644 under a predictable temp name. CodeQL had that open as `js/insecure-temporary-file` since 2026-07-24 (alert #45 on main) — the alert blocked this PR not because your change introduced it, but because your change touched the line. Its wording ("in the os temp dir") is wrong, the path is under userDataDir; the mitigation it wanted was right anyway, so it is closed now.

Follow-ups noted, not addressed: a hard crash between write and rename still leaves a 0600 temp holding a live secret until that store is next saved (a boot-time sweep would close it), and the sweep can delete a live *foreign* process's in-flight temp on a shared data dir.

Closing in favour of #117.

PR #109, opened 2026-08-10
fix(core): the session reaper is idleness, not attachment

eneskirca:
> **Depends on #108.** The reaper's safety story is "a reaped session resumes on next open". That is only true when the node knows its session id, and #108 is what guarantees it has one. Merging this first would turn "idle session reaped" into "conversation lost" for nodes whose id was never learned — measured at 18 of 40 on one host.

## The bug

The reaper never fired on a canvas. Its eligibility rule required a session to be **detached**, borrowed from the WebGL budget's "never evict a visible holder".

The two words do not mean the same thing:

| | WebGL `visible` | tmux `attached` |
|---|---|---|
| means | in the viewport, this instant | a mounted node exists somewhere |
| true when the user is looking elsewhere | no | **yes** |
| true for a node untouched for a week | no | **yes** |

Measured on a multi-tenant host: **54 of 54** `nt-` sessions reported `attached=1`, so `planReap()` returned `[]` on every sweep it had ever run. That host then died of a global OOM with ~32 GB held by 90 claude sessions, three quarters idle for over a day — precisely the accumulation this code exists to prevent.

Over a 3-minute sample on the same host, 5 of 53 sessions advanced their activity and 48 did not. **Attachment separates nothing; activity separates cleanly.**

## The change

- `planReap` no longer consults attachment — eligible is `nt-*` **and** idle past grace
- the kill-time re-verify re-reads the same rule that made a session eligible, so a session that wakes mid-sweep is spared on the signal that actually means it woke up (it used to re-check attachment)
- `maxDetached` → `maxIdle`, counting the **eligible** population: fifty sessions in active use are not accumulation, and a cap firing there would take work away from people
- grace default **6 h → 24 h**, because grace is now the only guard left — all that stands between a session left open over lunch and a sweep during tight memory

`NODETERM_SESSION_MAX_DETACHED` is still read as a fallback: an operator who tuned it meant "do not let more than N of these pile up", and that intent outlives the rename.

Reaping stays safe for the same reason as before — to the node a reap is a reboot, and cold restore replays the scrollback and resumes the agent.

## Verification

- `npm run typecheck` clean
- 98 test files / 1329 tests in `src/core` pass
- `session-budget.test.ts`: 21 → 25 tests; the two that asserted the old contract were replaced with ones asserting the new

### DIFF
--- a/src/core/session-budget.test.ts
+++ b/src/core/session-budget.test.ts
@@ -13,13 +13,13 @@ const NOW = 1_753_000_000 // fixed epoch seconds for every test
 const cfg = (over: Partial<SessionBudgetConfig> = {}): SessionBudgetConfig => ({
   disabled: false,
   minAvailableMb: 2048,
-  maxDetached: 48,
+  maxIdle: 48,
   graceSec: 6 * 3600,
   batchMax: 8,
   ...over
 })
 
-/** A detached nt- session idle for `idleH` hours. */
+/** An nt- session whose last activity was `idleH` hours ago. */
 const idle = (name: string, idleH: number, attached = false): SessionInfo => ({
   name,
   attached,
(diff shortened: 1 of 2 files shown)


eneskirca:
Closing for now — deferred, not rejected on the merits. The branch stays up (`fix/session-reap-idle`) and this can be reopened as-is.

**Where it stands.** The diagnosis holds and is measured: attachment vetoing eviction made the reaper a structural no-op (54 of 54 sessions reported `attached=1`, `planReap()` returned `[]` on every sweep it ever ran), and activity separates cleanly where attachment separates nothing (5 of 53 sessions advanced activity over a 3-minute sample). Typecheck clean, 1329 `src/core` tests pass.

**Why it waits.** Turning the reaper on makes it actually kill sessions, and its safety story — "a reaped session resumes on next open" — holds only for nodes that know their session id. #108 guarantees that for nodes created from now on, but every node already on a canvas still learns its id from hooks, over the SSH reverse tunnel that is the fragile path here (measured down again while investigating). So today a reaped pre-#108 node could still come back blank.

That risk retires on its own as existing nodes cycle through being closed and reopened with minted ids. Worth revisiting then, together with the hook `source` taxonomy follow-up described in #108, which is what keeps a minted id current across `/clear` and `--fork-session`.

PR #111, opened 2026-08-10
feat(windows): add psmux support and NSIS packaging

PaiCthulhu:
## Problem

nodeterm already contains several Windows-aware paths (PowerShell, ConPTY via `node-pty`, path handling), but two missing pieces prevent the desktop app from preserving terminal sessions there:

1. executable discovery probes extensionless names, so commands such as `tmux.exe`, `ssh.exe`, `git.exe`, and `claude.exe` are not found through a normal Windows `PATH`;
2. Windows has no native tmux package, and the UI previously hid the missing-multiplexer banner on `win32`.

Without a detected multiplexer, terminals silently fall back to a plain shell and lose reattach/scrollback continuity when the UI restarts.

## Change

- make executable discovery `PATHEXT`-aware on Windows while leaving POSIX lookup unchanged;
- centralize the executable-access check so Windows App Execution Aliases such as `winget.exe` can be detected;
- resolve `tmux` first and then `psmux`, a Windows tmux-compatible multiplexer;
- expose the missing-multiplexer banner on Windows and offer:

  ```text
  winget install -e --id marlocarlo.psmux
  ```

- make the banner's copy name the multiplexer the host actually has. Un-hiding the banner on `win32` made its no-installer fallback reachable there for the first time, and that fallback advised `brew install tmux` — a macOS-only package manager naming a package with no Windows build. Title and body now come from a pure `bannerCopy(phase, platform, hasInstallCommand)`; on Windows without winget it names winget as what is missing instead of inventing an install path;
- skip the well-known absolute tmux locations on `win32`. They are POSIX paths rooted without a drive letter, and the walk is a bare `existsSync` with no `PATHEXT` expansion, so it can never match a real `tmux.exe` — but `existsSync` is true for directories, so a stray directory named `tmux` under one of those paths would be returned as "the multiplexer" ahead of the PATH, and every spawn after it would fail with an opaque `EACCES`/`EISDIR`. macOS and Linux keep the same list, in the same order, ahead of the PATH;
- add focused tests for executable discovery, tmux/psmux selection, and the banner copy;
- add an unsigned Windows x64 NSIS target (`npm run dist:win`);
- keep locally packaged builds off the production updater feed (see below).

The psmux integration is capability/name detection rather than a separate Windows terminal backend; macOS/Linux continue to prefer the reference tmux binary.

### On the updater change

This one is not Windows-specific, and it is included here because the Windows package is what surfaced it.

A locally packaged app is indistinguishable from a release at runtime — `app.isPackaged` is true for both — so any `dist`-produced build initialized electron-updater against the production feed, polled it for a version nobody published, and logged a 404 on `latest*.yml` at boot and every six hours after. `npm run dist` on macOS and `npm run dist:linux` produce the same unpublished build and hit the same 404, so the opt-out marker is set by all three `dist` scripts rather than only `dist:win`, and `--publish never` comes with it (electron-builder defaults to `onTagOrDraft`, which on a tagged checkout with a token in the environment would try to publish a local smoke build).

`release` does not set the marker, so a promoted build wires up and updates itself exactly as before. The CI release workflow packages Linux through `electron-builder` directly and never goes through `dist:linux`, so it is unaffected either way. A guard test asserts the scripts themselves, since a `dist` script that forgets the marker is a build-config mistake no unit test of the pure function can catch and the failure mode is silent.

Happy to split this into its own PR if you would rather keep this one strictly to the Windows port.

## Verification

Tested on native Windows with Node 22 and psmux 3.3.7. Re-run on the current branch after rebasing onto `main`:

- `npm run typecheck` passes;
- 66 focused Windows/updater/banner tests pass across 7 files;
- 48 remote unit tests pass;
- `npm run build` passes;
- `npm run dist:win` produces an unsigned x64 NSIS installer, and the packaged `package.json` carries the update-opt-out marker;
- the full suite shows no regression from this branch. It is not green on a Windows host — 115 pre-existing failures across 29 files, all POSIX-only assertions (`/bin/sh`, Unix domain sockets, `0600` permissions, rooted `/...` paths) — and the count is identical with and without these commits. CI runs on `ubuntu-latest`, where those files pass.

Verified earlier on the same branch content: `npm ci` completes including Electron rebuilds for `node-pty` and `smart-whisper`; a manual GUI smoke test; and an end-to-end ConPTY/psmux probe covering session creation, UI-client disconnect, reattach to the same shell, preserved scrollback, detached `send-keys`, and session cleanup.

## Known limitations / follow-ups

- SSH askpass still tries to create a Unix-style socket path on Windows and logs `EACCES`; that is separate from terminal persistence and should be handled in a focused follow-up.
- psmux does not currently expose tmux's `terminal-features` option or `#{bracket_paste_flag}`. Basic terminal persistence and command transport work; `bracketPasteRequested` already treats a failed query as false, so paste delivery degrades to the legacy two-step path rather than misbehaving, but those optional capabilities need separate compatibility handling.
- The Windows artifact is intentionally unsigned; this PR does not add release publishing, signing, or production auto-update configuration for it.

🤖 Generated with Claude Code and independently verified by Ignis.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,8 +23,8 @@
     "server:dev": "npm run build && npm run server:build && node out/server/main.cjs",
     "make-icon": "node scripts/make-icon.mjs",
     "make-social": "node scripts/make-social.mjs",
-    "dist": "npm run make-icon && electron-vite build && electron-builder --mac --arm64 -c.mac.identity=null -c.mac.notarize=false",
-    "dist:linux": "npm run make-icon && electron-vite build && electron-builder --linux",
+    "dist": "npm run make-icon && electron-vite build && electron-builder --mac --arm64 --publish never -c.mac.identity=null -c.mac.notarize=false -c.extraMetadata.nodeTermUpdates=disabled",
+    "dist:linux": "npm run make-icon && electron-vite build && electron-builder --linux --publish never -c.extraMetadata.nodeTermUpdates=disabled",
     "dist:win": "npm run make-icon && electron-vite build && electron-builder --win --x64 --publish never -c.extraMetadata.nodeTermUpdates=disabled",
     "release": "npm run make-icon && electron-vite build && sh -c 'electron-builder --mac --arm64 --x64 --publish never; s=$?; npm run rebuild; exit $s'",
     "postinstall": "electron-rebuild -f -w node-pty,smart-whisper"
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "make-social": "node scripts/make-social.mjs",
     "dist": "npm run make-icon && electron-vite build && electron-builder --mac --arm64 -c.mac.identity=null -c.mac.notarize=false",
     "dist:linux": "npm run make-icon && electron-vite build && electron-builder --linux",
+    "dist:win": "npm run make-icon && electron-vite build && electron-builder --win --x64 --publish never -c.extraMetadata.nodeTermUpdates=disabled",
     "release": "npm run make-icon && electron-vite build && sh -c 'electron-builder --mac --arm64 --x64 --publish never; s=$?; npm run rebuild; exit $s'",
     "postinstall": "electron-rebuild -f -w node-pty,smart-whisper"
   },
@@ -95,6 +96,17 @@
           ]
         }
       ]
+    },
+    "win": {
+      "target": [
+        {
+          "target": "nsis",
+          "arch": [
+            "x64"
+          ]
+        }
+      ],
+      "icon": "build/icon.png"
     },
     "publish": {
       "provider": "generic",
--- a/src/core/exec-path.test.ts
+++ b/src/core/exec-path.test.ts
@@ -1,0 +1,134 @@
+import { describe, it, expect, afterEach } from 'vitest'
+import fs from 'fs'
+import os from 'os'
+import path from 'path'
+import { execCandidates, findInPathString, isExecutable } from './exec-path'
+
+describe('isExecutable', () => {
+  it('accepts a real executable and rejects a path that is not there', () => {
+    const d = fs.mkdtempSync(path.join(os.tmpdir(), 'nt-exec-'))
+    try {
+      const p = path.join(d, os.platform() === 'win32' ? 'nt-fake.exe' : 'nt-fake')
+      fs.writeFileSync(p, '')
+      fs.chmodSync(p, 0o755)
+      expect(isExecutable(p)).toBe(true)
+      expect(isExecutable(path.join(d, 'nope'))).toBe(false)
+    } finally {
+      fs.rmSync(d, { recursive: true, force: true })
+    }
+  })
+
+  // THE REASON THIS HELPER EXISTS, and it cannot be expressed on any other platform.
+  //
+  // Windows ships its most important commands — winget above all — as APP EXECUTION ALIASES:
+  // zero-length reparse points under `…\AppData\Local\Microsoft\WindowsApps`. `fs.existsSync`
+  // answers FALSE for one (it resolves the link and gets EACCES), while the file is on PATH and
+  // runs perfectly. Any lookup written with `existsSync` therefore reports the single package
+  // manager Windows actually has as "not installed" — which silently emptied the tmux banner's
+  // install button. `fs.accessSync(X_OK)` sees them.
+  //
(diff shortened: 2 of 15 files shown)


eneskirca:
Thank you for this — it is unusually careful work. The commit messages explain *why* at a level most PRs here don't reach, the new functions take `platform` / `PATHEXT` / `exists` as parameters instead of reading `process` (which is why all of it is testable on Linux CI), and the `wellKnownTmux` finding — a stray `tmux` **directory** shadowing the PATH through `existsSync` — is a real bug you found in our code.

On Linux the branch is green as submitted: `npm run typecheck`, `npm run build`, and the full 4643-test suite all pass. No CI ran because of the fork-approval gate, not because anything failed.

Let me be straight about where this stands. nodeterm has no Windows CI, no Windows build target, and nobody who can verify a ConPTY or PSReadLine regression. Making Windows a supported platform is a commitment that hasn't been made, and I don't want you investing more here until it has.

Two things worth fixing regardless of that decision, because they are **not** dormant on macOS/Linux:

**1. The Ctrl-U → Ctrl-C change in `src/renderer/terminal/command-delivery.ts` (lines 42, 130) is unguarded.** Ctrl-U is a line edit; Ctrl-C is SIGINT to the pane's foreground process group. The retry fires on *any* echo-verify timeout, not only on a mangled line — zsh-autosuggestions or syntax-highlighting defeating `echoedIntact`, the 1500ms silence cap in `TerminalNode.tsx` landing mid-rc-file or mid-SSH-connect, `agent-restart.ts`'s "changed foreground command" branch which by construction fires on shells it could not identify. So the worst case goes from a stray `^U` to interrupting whatever is running. Your reasoning that this belongs to the shell rather than the OS is right; changing it *everywhere* is what mac/Linux users pay for. It needs to be platform-scoped. (The added test is also tautological — `expect(ABORT_LINE).toBe('\x03')`.)

**2. `TMUX_BIN_NAMES = ['tmux', 'psmux']` in `src/core/pty-manager.ts` is unguarded**, so mac/Linux also PATH-probe `psmux`. Harmless today, but `main` now ships a bundled tmux with fixed-paths → PATH → bundled precedence, and a naive rebase puts a PATH `psmux` ahead of the tmux we ship.

Also `src/core/pty-manager.ts:817` swaps the install-hint probe from `fs.existsSync` to an X_OK check, which tightens brew/apt/dnf detection on POSIX. Probably an improvement, but it is a mac/Linux behaviour change riding in a Windows PR.

**The updater opt-out and the `dist` / `dist:linux` changes are genuinely good and have nothing to do with Windows** — I confirmed `release` is untouched and the release workflow calls `electron-builder --linux` directly. Please do split that out; I will take it now.

On the rebase: it is not trivial. `findTmux` was rewritten on `main` for the bundled binary, so `src/core/pty-manager.ts` is a design conflict rather than a textual one, and `tmux-discovery.test.ts` would need rewriting against the new shape. `src/core/tmux-hint.ts` is adjacency-only and mechanical.

No security issues in the diff — no credential interpolation, no shell-string building, and `sessionName()` already sanitises the node id. One for later if Windows ever happens: PATHEXT resolving `claude.cmd` hits Node's `execFile` refusal for `.cmd`/`.bat`, and the obvious `shell: true` fix would put cmd.exe's parser in front of project paths, where `&`, `^`, `|` and `%VAR%` become live injection. That is the Windows quoting layer this port hasn't had to write yet, and it is the part I would want most carefully reviewed.

I will come back to you with a decision on Windows rather than leave this open indefinitely.

eneskirca:
Closing with thanks — and with context, because the landscape changed since August. When this was opened we had no Windows commitment; we since made one, but down a different architectural path: session continuity is now owned by a dedicated **session host** (#305 + hardening) rather than a multiplexer, packaging is **NSIS via electron-builder** groundwork (#302), atomic writes landed in #300, and the ConPTY native patch just merged (#526). So the psmux/Squirrel route is an architecture we deliberately didn't take, not one we overlooked.

That said, three pieces of this PR are still valuable and **still unlanded** — each would be welcome as a small standalone PR:
- **PATHEXT-aware executable discovery** (main still probes without extensions),
- the **wellKnownTmux directory-shadowing fix** (that bug is still present today — verified),
- the **X_OK install-hint probe**.

The updater opt-out shipped in spirit (nodeTermUpdates=disabled across dist targets). Thanks for being early on Windows before we were.

PR #115, opened 2026-08-10
fix(github-issues): hide token controls when the GitHub CLI is signed in

eneskirca:
## Question this answers

> If `gh` (GitHub CLI) is signed in, is a personal access token even needed?

**No.** The GitHub Issues auth model (`src/core/github/credentials.ts`) has three providers and defaults to **`auto`**, which prefers an authenticated `gh` (`gh auth token`) and only falls back to a stored PAT when `gh` is absent. So on any machine where `gh auth login` has been run, the token field is pure noise.

## Problem

The Authentication section showed the provider dropdown **and** the PAT field unconditionally — even with `gh` signed in — so a signed-in user was left wondering whether they had to paste a token.

## Fix — progressive disclosure

- **`gh` signed in** (the common case): a single line — `✓ Signed in via GitHub CLI as @login. No token needed.` — and the provider dropdown + token field move into a collapsed **"Advanced — use a personal access token instead"** disclosure. (The "No token needed" clause is dropped if the user has explicitly pinned the provider to token-only.)
- **`gh` not signed in**: the token field is surfaced prominently (it's the only way in) with a `gh auth login` hint; the provider dropdown stays under "Advanced".
- The token path is **preserved** for headless / Server Edition hosts without `gh` — nothing is removed, only relocated. Token DRY'd into one `tokenFieldRow` rendered in the correct slot per state.

## Testing

- `npm run typecheck` + `npm run build` clean.
- Note: the Playwright MCP was disconnected this session, so the visual before/after was **not** browser-verified — owed on Mac / a live browser. Both render paths were confirmed by code review (all controls remain reachable in each state).


eneskirca:
Reviewed (Opus): **Needs fixes** before merge. The relocation itself is clean (JSX moved verbatim, single `#github-personal-access-token` in either branch), and secret handling / Server Edition parity / auth precedence all check out — but the new layout keys off a flag that lies in two common states:

1. **High — unapproved project reads as "signed out".** `src/core/github/host.ts:100-108` masks the auth block (`ghAuthenticated: false`) whenever `approved` is false, and the component only calls `status(projectId)`. Approval comes *after* authentication in the visible flow, so during setup every signed-in user sees "GitHub CLI is not signed in. Run `gh auth login`…" with the token field front and center — the exact noise this PR removes, plus a wrong instruction. Fix: treat unapproved as **unknown** (neutral "Approve this machine to check authentication." branch), or read the unmasked status without a projectId for this line.
2. **Medium — copy derived from `ghAuthenticated` alone, but requests authenticate via `activeProvider` × `selectedProvider`.** (a) `selectedProvider='gh'` + gh signed out → hint says a pasted token will work; it's inert under that pinning, and the provider dropdown moved into Advanced. (b) gh signed in + `selectedProvider='token'` + no stored token → green "✓ Signed in via GitHub CLI." renders while `activeProvider` is null. Gate the ✓ on `activeProvider === 'gh'`; state the pinning inline when `selectedProvider !== 'auto'`.
3. **Medium — zero tests for the new branch** (both existing tests fixture `ghAuthenticated: false`). Pin: signed-in render keeps the token input reachable inside `<details>`; the two copy states above.
4. Low: settings search can land on a control hidden inside collapsed `<details>` (consider `open` when searching); a stored token becomes hard to discover/clear ("A saved token is kept as a fallback." when `tokenPresent`); no way to re-check after running `gh auth login` (a "Check again" button calling refreshStatus — the resolver bypasses the credential cache, so it would be accurate); `<details> summary` lacks a `:focus-visible` style.

eneskirca:
Review findings fixed — but this PR was already **merged** when the review landed, so the branch was reset to `origin/main` + one commit and the fixes are up as #165 (not merged).

- **[High] Unapproved project reading as "signed out"** — confirmed at `src/core/github/host.ts:100-108`. The section now reads the project-independent auth block (`status()` with no projectId) in exactly the masked case, so the approved path spawns nothing extra; a read that could not run is *unknown*, never "signed out".
- **[Medium] Copy vs `activeProvider` × `selectedProvider`** — confirmed both mismatches. The ✓ is gated on `activeProvider === "gh"`, the pinning is stated inline, and a token the pinning makes inert moves into Advanced instead of being offered as the way in.
- **[Medium] Tests** — 10 render tests now, up from 2.
- **[Low]** "Check again" (the resolver's `status()` uses `fromGitHubCli()` directly, so it really does bypass `CREDENTIAL_CACHE_MS` — your parenthetical checks out); "A saved token is kept as a fallback." when `tokenPresent`; Advanced opens while a settings search is active.
- **[Nit]** `:focus-visible` ring on the summary; `style={{ marginTop: 12 }}` → `mt-3`.

Nothing refuted. One thing found while fixing that the review did not list: the old copy also mis-described `activeProvider === "token"` — a working saved token rendered "GitHub CLI is not signed in… paste a personal access token below", which is the same "derived from one flag" defect. That state now has its own line.

PR #121, opened 2026-08-11
fix(sidebar): clicking project row switches active project

### LINKED ISSUE #120: Sidebar: clicking a project only collapses/expands sessions, doesn't switch to it
**Current behavior**

In the left sessions sidebar (`SessionsSidebar.tsx`), left-clicking a project header only toggles collapse/expand of its session list. Switching the active project is only available via right-click → context menu, which is not discoverable — most users expect a left click on a project row to open/switch to it.

**Expected behavior**

Clicking a project row should switch the active project (and expand its sessions), matching the mental model of a typical sidebar project tree. Collapsing/expanding without switching should still be possible via the chevron icon.

**Location**

`src/renderer/components/SessionsSidebar.tsx` — project header block, `onClick` around line 278, chevron span around line 292.

gor3a:
## Summary
- Left-click on a project row in the sessions sidebar now switches the active project (and expands it if collapsed), instead of only toggling collapse.
- The chevron icon keeps its own click handler (`stopPropagation`) so users can still collapse/expand a project without switching to it.

Fixes #120

## Test plan
- [x] `npx tsc --noEmit` passes
- [ ] Manual: click inactive project row → becomes active + expands
- [ ] Manual: click chevron on active project → collapses without losing active state
- [ ] Manual: right-click project header → context menu (switch/rename/etc) still works unchanged

### DIFF
--- a/src/renderer/components/SessionsSidebar.tsx
+++ b/src/renderer/components/SessionsSidebar.tsx
@@ -46,6 +46,7 @@ export function SessionsSidebar(props: SessionsSidebarProps): JSX.Element | null
   // Closed projects are hidden from the tab bar; hide them from the sidebar too.
   const projects = useMemo(() => allProjects.filter((p) => !p.closed), [allProjects])
   const activeProjectId = useProjects((s) => s.activeProjectId)
+  const setActiveProject = useProjects((s) => s.setActive)
   const statusById = useAgentStatus((s) => s.byId)
   const namingById = useSessionNaming((s) => s.byId)
   // This sidebar's core api (a stable context read — the branch lookups run on the session's git).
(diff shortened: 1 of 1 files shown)


eneskirca:
Thanks @gor3a — the fix is in, via #157. Your commit was carried over with `git cherry-pick -x` so authorship is preserved.

Two things we changed on top before merging:
- The switch now goes through Canvas's `switchProject` (an `onSwitchProject` prop, matching the sibling `onAddToProject`/`onFocusNode`/`onReorderProject` props) instead of calling `useProjects.setActive` directly. That path does `commitActiveToStore()` → `setActive()` → `writeDisk()`; without it, live React Flow edits to the outgoing project since the last debounced persist were discarded, and the new `activeProjectId` was never written to disk (the app reopened on the old project).
- The row-click/collapse interaction was made coherent with the documented `sidebarAutoCollapse` contract, extracted into a pure tested `projectHeadClickAction`, and the chevron became a real button with a proper hit area and `aria-expanded`.

Closing this one as superseded. Appreciate the contribution — the behaviour you were after is what shipped.

PR #122, opened 2026-08-11
feat(projects): add idle project state to skip PTY/SSH auto-resume

gor3a:
## Summary
- Projects can now be marked **idle**: they stay visible/selectable in the sidebar and tab bar, but PTY spawn / SSH ControlMaster connect / tmux reattach are suppressed until the user explicitly resumes them — frees memory without hiding the project like `closed` does.
- Toggle via sidebar right-click → "Mark idle" / "Resume" (dimmed row + pause badge while idle), or via a "Project idle" banner + Resume button shown over the canvas when the active project is idle.
- Persisted (`Project.idle`), machine-local like `closed` — never leaks into the shared `.nodeterm/project.json`.
- Passive semantics: marking an already-live project idle never kills its running sessions, only blocks future auto-resume.

## Notable fixes during review
- Resume didn't actually respawn terminals (React Flow re-renders existing node components on reload rather than remounting them, so `TerminalNode`'s spawn effect never re-ran) — fixed by bumping each resumed node's `respawnNonce`.
- Idle toggles weren't persisted to disk (missing `persist()`, unlike sibling project-mutation actions).
- Git remote routing could leak into an idle local project after switching away from an SSH project.

## Test plan
- [x] `node_modules/.bin/vitest run` — 4586/4613 passing; the 20 failing tests are in 5 pre-existing native-PTY/sandbox-environment test files (`pty-coattach`, `pty-require-remote`, `pty-spawn-diagnosis`, `pty-typing`, `server-e2e`), confirmed unrelated to this branch (none of its files are touched by this diff; failure count/category matches pre-branch baseline)
- [x] `npm run typecheck` — clean
- [x] `npm run build` (`electron-vite build`) — clean
- [ ] Manual: mark a project idle, switch to it — canvas freezes (no terminal output), banner shows, no new PTY/SSH connect fires
- [ ] Manual: click Resume — terminals actually spawn/reattach
- [ ] Manual: mark the *active* project idle while its sessions are live — sessions keep running (not killed)
- [ ] Manual: restart the app — idle flag survives, idle project's sessions do not auto-resume

### DIFF
--- a/src/core/workspace-files.test.ts
+++ b/src/core/workspace-files.test.ts
@@ -51,6 +51,16 @@ describe('projectToFile / fileToProject round-trip', () => {
     expect(back.nodes[0]).toMatchObject({ cwd: '/new/root/x', agentId: 'claude', accountId: 'acct1' })
     expect(back.unavailable).toBeUndefined()
   })
+
+  it('drops idle from the shared file, restores it from base like closed', () => {
+    const p = project({ idle: true, defaultAccountId: 'acct1' })
+    const f = projectToFile(p, 1, '2026-08-11T00:00:00.000Z')
+    expect((f as any).idle).toBeUndefined()
+    const back = fileToProject(f, { cwd: '/x', idle: true })
+    expect(back.idle).toBe(true)
+    const notIdle = fileToProject(f, { cwd: '/x' })
+    expect(notIdle.idle).toBeUndefined()
+  })
 })
 
 describe('sameProjectContent', () => {
@@ -135,6 +145,18 @@ describe('splitWorkspace', () => {
       () => 1, '2026-07-11T00:00:00.000Z')
     expect(remote.index.entries[0]).toMatchObject({ id: 'p1', ssh: { remoteCwd: '~/app' } })
     expect(remote.index.entries[0].cache).toBeUndefined() // no cache fabricated from the placeholder
+  })
+})
+
+describe('splitWorkspace', () => {
+  it('carries idle into the index header, like closed', () => {
+    const ws: Workspace = {
+      version: 2,
+      activeProjectId: 'p1',
+      projects: [project({ idle: true, cwd: '/a/b' })]
+    }
+    const { index } = splitWorkspace(ws, () => 1, '2026-08-11T00:00:00.000Z')
+    expect(index.entries[0].idle).toBe(true)
   })
 })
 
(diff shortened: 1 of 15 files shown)


eneskirca:
Thanks @gor3a — reviewed this carefully, and I'm going to close it. The craft is genuinely high (the `idle` flag rides the machine-local index rather than the git-shared `project.json`, `data.projectIdle` is correctly kept out of `flowToNodeStates`, old workspace files read tolerantly, and the logic sits in `src/core` so the Server Edition gets it for free). The reason is scope overlap, not quality.

**The memory argument no longer holds.** Since this branch was cut, `main` gained a five-phase RAM system that covers it from several directions:

- `terminal/offscreen-policy.ts` — a terminal offscreen for 10 minutes has its xterm + PTY client released in place (~15 MB/node back), reattaching on approach.
- `terminal/hibernation-policy.ts` — opt-in "Eco": an idle, offscreen agent CLI is exited with its own exit command and resumed with `--resume` when you look at it. That reclaims the *agent process*, which is the actual heavyweight (a field report had 95 idle sessions ≈ 34 GB).
- `core/session-budget.ts` + `core/pty-reap.ts` — idle detached tmux sessions and stranded client PTYs are reaped under memory pressure.
- `core/memory-pressure.ts` — a two-signal monitor that drives the above when the host is actually under pressure.

Against that, an idle *project* buys little on the memory axis specifically: only the **active** project's nodes are mounted at all, so a background project was already costing nothing in the renderer, and `idle` doesn't touch tmux or the agent processes — which is where the memory actually is.

**What it does buy that nothing else does** is real, and it's a different feature than the PR is pitched as: **suppressing SSH ControlMaster auto-connect and cold-restore agent `--resume` for a project you deliberately don't want waking up.** That is not a memory feature, it's an "open this project without it phoning home" feature — for a machine you're offline from, a host you don't want to hit, or an agent you don't want resuming a conversation. If you want to bring that back framed that way, I'd look at it seriously.

Two things that would need fixing in that version:
1. **The kanban card modal bypasses it entirely** — `ModalTerminal` calls `transport.create()` with no idle guard, so opening any card in an idle project spawns the PTY. This repo's rule is that the canvas and the board are two views of the same nodes and get wired in the same change.
2. **No per-node explanation** — `closed`/`ended` nodes render an overlay saying what happened; an idle node renders blank with only a canvas-level banner. Also `resumeProject` calls `requestReload()` where `reloadActiveProject()` is the one that preserves the camera (it happens to work today only because a synchronous prefix beats the effect — fragile).

Closing as overlapping in its current framing. The SSH/resume-suppression idea is welcome as its own PR, and thanks for the care in the implementation — the index-vs-project-file split in particular was the right instinct.

PR #123, opened 2026-08-11
feat(canvas): Cmd/Ctrl+1-9 jumps to the Nth project in sidebar order

gor3a:
## Summary
- Adds a Cmd/Ctrl+1-9 shortcut that switches directly to the Nth project, matching sidebar/tab-bar display order (1-indexed, no wraparound, silent no-op out of range or when already active)
- Fixes a pre-existing gap where Ctrl+digit combos leaked a raw control byte into the focused terminal's PTY on Linux/Windows before the app-level handler could intercept them (e.g. Ctrl+3 both sent `^[` to a focused vim and switched projects)
- Adds a "Jump to project" row to the shortcuts help panel (Cmd/Ctrl+/)

## Test plan
- [x] `projectIdAtIndex` unit tests (in-range, out-of-range, empty list)
- [x] `terminalKeyAction` unit tests for the new ctrl+digit swallow branch (ctrl/meta/alt/shift/digit-range/keyup combinations)
- [x] `npx tsc -p tsconfig.web.json --noEmit` clean
- [ ] Manual: Cmd/Ctrl+1-9 while a terminal is focused, on macOS and Linux/Windows
- [ ] Manual: AltGr+digit still types correctly on non-US keyboard layouts


eneskirca:
Thanks for this @gor3a — the feature is a genuinely nice addition, and two things in particular were done right: resolving through a pure `projectIdAtIndex` with its own tests, and spotting that xterm's own key handler runs *before* the window listener, so a `preventDefault()` in `Canvas.tsx` is too late to stop the control byte. That second one is subtle and easy to miss; the fix below builds directly on it rather than replacing it.

I've pushed one commit (`da04e905`) addressing two findings from review, plus a small structural change that makes both of them checkable.

**1. The chord is browser-reserved, so the feature is desktop-only.** Chrome, Firefox and Safari all own Ctrl+1-9 (Cmd+1-9 on macOS) for tab switching, and a page cannot `preventDefault()` them. In the Server Edition (`src/server` + the browser bridge) the switch can therefore never happen — while the swallow still stripped the terminal's control codes, for no benefit at all. The whole feature is now gated on `isBrowserRuntime()` (`src/renderer/bridge/runtime.ts`): in a browser tab we neither handle the key nor swallow it, and `ShortcutsPanel` drops the "Jump to project" row rather than advertising a shortcut that never fires.

**2. The swallow now only fires when the digit resolves.** It was unconditional across the digit row, which permanently removed Ctrl+2..Ctrl+8 from every terminal — `^@ ^[ ^\ ^] ^^ ^_`, and `^]` (jump to tag) and `^^` (alternate file) are daily-use vim keys — even when the chord could do nothing, e.g. Ctrl+5 with three projects open. The swallow now mirrors the handler's decision exactly: no target, no swallow.

**Structural change that ties the two together.** Both fixes are really "the app owns this key only when it can act on it", asked in three places (the Canvas handler, plus the two xterm handlers — `TerminalNode` and the kanban `ModalTerminal`, which also needed the swallow and was missing it). Rather than repeat the matcher, it now lives once in `src/renderer/lib/projectJump.ts`: a pure `projectJumpDigit` / `projectJumpTarget(e, ctx)` plus a `liveProjectJumpTarget` bound to the shell flag and the project store. `terminalKeyAction` takes the answer as a parameter instead of re-deriving it (defaulting to `false` = the byte-identical pre-feature behavior), so the swallow and the switch cannot drift apart.

**On reusing `@shared/shortcut`** (the third review note): I looked, and decided against it. `matchesShortcut` matches one named key off `e.key`, whereas this is a family of nine matched *positionally* off `e.code` — which is the right call and your original code already had it, since an AZERTY digit row prints letters unshifted. Bending the shared helper to cover both would blur the settings capture field's contract for no gain. Your matcher is kept, just moved.

**Behavior notes.** Your key-shape rules are preserved verbatim: AltGr (which reports as ctrl+alt) still types normally, shifted digits and `Digit0` are untouched, keyup does nothing. One deliberate difference: the chord now `preventDefault()`s even when the digit addresses the *already active* project, so the swallow and the handler agree on the exact same set of events (`switchProject` no-ops on the active id, as before).

**Tests.** New `projectJump` suites cover key shape, the closed-project filter, the out-of-range null, and the browser gate (in its own file, since `markBrowserRuntime` is a one-way boot switch); the `terminal-config` tests keep a swallow/pass pair. Your key-shape cases moved into the new suite along with the matcher. I checked both gates are mutation-sensitive — deleting the desktop gate or the resolve-before-swallow makes tests go red rather than passing quietly.

**Verification.** `npm run typecheck` clean; full `npm test` green (5104 passed) apart from 3 pre-existing `node-pty-patch` failures that are environmental in my checkout and unrelated to this change. The branch also rebases onto current `main` with no conflicts and stays green there — I pushed a plain fast-forward commit rather than a rebase, so your history is untouched.

Still worth a manual pass on a real machine before merge: Cmd+1-9 on macOS and Ctrl+1-9 on Linux with a terminal focused, and confirming Ctrl+] / Ctrl+^ still reach vim when the digit addresses nothing.

PR #124, opened 2026-08-11
feat(sidebar): show working/total badge on project rows with active sessions

gor3a:
## Summary
- Adds a working-count badge with a spinning glyph to project header rows in the sidebar whenever at least one session in that project is actively running an agent turn — hidden entirely when nothing is working. It reads like its `attention` / `unread` siblings: a glyph plus its own count, with the project's existing total still shown once by `.ss-group__count`.
- Reuses the existing `nt-spin` animation and introduces an `--agent-working` CSS custom property so the colour stops being duplicated between the per-session dot and the new badge. The token is defined in **both** the dark and the light palette (`#d97757` / `#b04a28`) — the badge paints a COUNT with it, so it owes the light palette's 4.3:1 text floor.
- Respects `prefers-reduced-motion`: the spinning ring/arc badges freeze in place.

## Also in this PR: the theme guard that should have caught the missing light token

`styles.theme.test.ts` sliced the light block with `indexOf(":root[data-theme='light']")`, which matches the **doc comment** inside the `:root` block ~90 lines earlier. Two suites had been passing vacuously as a result:

- *"the light theme overrides every themeable token"* compared against an **empty** dark-token list, so every token added since went unchecked.
- The light **contrast floors** resolved `LIGHT` to the tail of the *dark* block — they were measuring the dark palette against itself.

The slice is now anchored to the selector at the start of a line. With it fixed the suite flagged one further token, `--text`; that one is a genuine false positive (it is mixed purely from `--tint-rgb`, which the light block *does* override, so it flips by construction) and is now exempted by shape rather than silenced. `--agent-working` was added to the asserted hue list, and both fixes were mutation-checked: removing the light token, or reverting it to the dark clay, each fail the suite.

## Test plan
- [x] `projectSignalCounts` unit tests: working count, zero case, grouped-session traversal, attention-over-unread precedence, working-not-double-counted-as-unread — the original a–f matrix is kept, with the working assertions added alongside it
- [x] `npx vitest run src/renderer` — 153 files / 2085 tests passing
- [ ] Manual: badge shows/hides correctly as sessions start and finish
- [ ] Manual: spinner and count legible in **both** themes

### DIFF
--- a/src/renderer/components/SessionsSidebar.tsx
+++ b/src/renderer/components/SessionsSidebar.tsx
@@ -309,6 +309,12 @@ export function SessionsSidebar(props: SessionsSidebarProps): JSX.Element | null
                     {signals.unread}
                   </span>
                 )}
+                {signals.working > 0 && (
+                  <span className="ss-group__sig ss-group__sig--working" title="Sessions running right now">
+                    <span className="ss-group__sig-spin" />
+                    {signals.working}/{projectCount(g)}
+                  </span>
+                )}
                 <span className="ss-group__count">{projectCount(g)}</span>
                 <button
                   className="ss-group__add"
--- a/src/renderer/components/SessionsSidebar.tsx
+++ b/src/renderer/components/SessionsSidebar.tsx
@@ -312,7 +312,7 @@ export function SessionsSidebar(props: SessionsSidebarProps): JSX.Element | null
                 {signals.working > 0 && (
                   <span className="ss-group__sig ss-group__sig--working" title="Sessions running right now">
                     <span className="ss-group__sig-spin" />
-                    {signals.working}/{projectCount(g)}
+                    {signals.working}
                   </span>
                 )}
                 <span className="ss-group__count">{projectCount(g)}</span>
--- a/src/renderer/lib/sessionList.test.ts
+++ b/src/renderer/lib/sessionList.test.ts
@@ -161,6 +161,51 @@ describe('projectSignalCounts', () => {
     }))
   })
 
+  // Restored from before the working badge: the original a–f matrix. It pins the row-glyph
+  // PRECEDENCE (attention beats unread, a working session is not yet unread) across both
+  // ungrouped and grouped sessions, which the narrower fixtures below do not reach. The working
+  // count is asserted alongside it rather than replacing it.
+  it('counts attention and unread across ungrouped and grouped sessions', () => {
+    const proj: ProjectInput[] = [
+      {
+        id: 'p1',
+        name: 'P1',
+        color: '#123',
+        nodes: [
+          node('g1', { kind: 'group', title: 'G', color: '#abc' }),
+          node('a'), // waiting → attention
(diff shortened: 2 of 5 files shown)


eneskirca:
Thanks @gor3a — nice feature, and the `--agent-working` token was the right instinct: pulling the clay out of `.ss-dot--working` is what made the badge and the dot one decision instead of two. I've pushed a few maintainer commits to your branch (rebased onto current `main`) so this can go in.

**What I changed**

1. **Light theme for `--agent-working`.** The token only landed in `:root`, so on a light page the badge kept its dark-field clay and measured **2.95 / 2.72 / 2.73:1** against `--bg` / `--panel` / `--canvas-bg` — under this palette's own 4.3 floor, and it matters more than usual here because the badge paints a *count*, not just a dot. Added `#b04a28` to the light block: same hue (15°), darkened into the band `--danger` / `--success` already occupy, so "working" still reads as the same colour in both themes.

2. **Fixed the guard that let that through** — the more important half, and not your bug. `styles.theme.test.ts` sliced the light block with `indexOf(":root[data-theme='light']")`, which matches the *doc comment* inside the `:root` block ~90 lines earlier. So "the light theme overrides every themeable token" had been running against an **empty** token list, and the light contrast floors were measuring the **dark** palette against itself — vacuous for every token added since. Now anchored to the selector at the start of a line, and mutation-checked both ways (removing the light token, and reverting it to the dark clay, each fail).

   With the slice repaired the suite flagged one more token, `--text`. That one is a real false positive rather than a miss: it is mixed purely from `--tint-rgb`, which the light block *does* override, so it already flips by construction — the stylesheet says as much. Tokens of that shape are now exempt by shape, and `token()` falls back to the dark block for them, so a hue that genuinely went missing still resolves to its dark value and fails the contrast floor loudly. **No other light tokens were missing.**

3. **Restored the deleted tests.** The change removed `counts attention and unread across ungrouped and grouped sessions` (the a–f matrix) and `returns zeros for a quiet project`, replacing them with narrower fixtures. The a–f matrix is the only one that reaches node `e` — working **and** unread — through `buildSessionList`, which is exactly the precedence the new counter has to preserve. Both are back, with the working count asserted *alongside* rather than instead of them; your new cases are kept.

4. **Dropped the duplicated total.** The badge rendered `{working}/{projectCount(g)}` immediately before the existing `.ss-group__count`, so a project read `⟳ 3/8  8`. It now shows just the working count, matching how the `attention` and `unread` badges already sit next to that one shared total.

5. **PR body.** It claimed the badge is suppressed on idle-marked projects and referenced an `.is-idle` row — that depends on #122, which isn't landing, and there's no idle concept in the tree today. Body updated so the description matches the diff.

Gates on the rebased branch: `npm run typecheck` clean, `npx vitest run src/renderer` 153 files / 2085 tests passing. Leaving the merge call to @eneskirca — the two manual checks (badge show/hide as sessions start and finish, and legibility in **both** themes now) are still unticked.

PR #137, opened 2026-08-12
test(core): darwin-gated behavioural guard for the reaper's default memory reader

eneskirca:
## What

Adds the behavioural test the source-text guard in `session-budget.test.ts` could not be: proof that `createSessionReaper()` **without an injected `readMem`** never reaps on memory grounds on macOS, no matter how full memory is.

- 20 detached `nt-` sessions idle far past grace, no `readMem` injected.
- The watermark is env-pinned above any physically possible host (1 TB available), so **any** byte reading counts as pressure; only `hostMemReader`'s darwin `null` refuses. This encodes *"memory fullness must never reap on macOS"* without depending on the host's current load.
- A companion assertion pins `readMemInfo()` non-null on darwin so the main test can never pass vacuously (both readers null).
- `it.skipIf(process.platform !== 'darwin')`: on Linux the two readers are the same function and the real `/proc/meminfo` reading legitimately trips the impossible watermark — the test would fail there for the wrong reason. The existing source-text check remains as CI's stand-in; its comment now points here.

## Mutation verification (on a real 24 GB M-series Mac, macOS 26.5.2)

Reverting the default to `readMemInfo` (the pre-#136 state):

```
× without an injected readMem, sessions survive NO MATTER how full memory is
  AssertionError: expected 8 to be +0
```

A full batch of 8 reaped — the exact field bug ("my sessions keep disappearing"). With `hostMemReader()` restored: 31/31 pass.

## Also

`docs/session-memory.md` §10: item 6 re-verified independently (shipping `parseVmStat` vs a same-tick Activity Monitor capture: 18.87 vs 18.88 GB parts, 0.05% off; 4096-page mutation under-reports exactly 4.00×), item 5 partial (BSD `ps rss` units are kB, but RSS vs AM footprint gaps up to 8× measured).

Gate: `npm run typecheck` clean, `npx vitest run` 5021 passed / 8 pre-existing skips.

https://claude.ai/code/session_017TxMV9B96o8U7rHxob5zB1

### DIFF
--- a/docs/session-memory.md
+++ b/docs/session-memory.md
@@ -427,10 +427,23 @@ macOS (the ps path never runs on Linux)
  5. Open the panel on a Mac: `defaultProcessTableReader` returns null there, so the whole table
     comes from `ps -eo pid,ppid,rss`. Rows must populate, and the totals must be plausible — BSD ps
     reports rss in kB, but confirm one known process against Activity Monitor before trusting it.
+    PARTIAL 2026-08-12: the units ARE kB, but `ps rss` and AM's "Memory" column measure different
+    things — WindowServer read 92,768 kB (90.6 MB) of RSS while AM showed 763.5 MB of footprint on
+    the same tick, an 8× gap for a compressed/IOKit-heavy process. The panel will understate such
+    processes; "plausible" must mean plausible-as-RSS, not AM-equal. Panel populate still unchecked.
  6. ~~macOS: check `parseVmStat` against Activity Monitor.~~ **DONE 2026-08-12** — 19.1 GB vs
     AM's 19.00 GB of parts on a 24 GB machine (was 23.9/24.0 before the fix). Still open on
     macOS: give the memory-PRESSURE monitor a real signal (`kern.memorystatus_vm_pressure_level`)
     rather than a byte watermark — the same capture showed 82% used with AM's pressure graph GREEN.
+    RE-VERIFIED independently 2026-08-12 (second session, shipping function imported via esbuild
+    bundle, vm_stat captured on the same tick as an AM screenshot): ours 18.87 GB used vs AM's
+    parts 7.23 + 2.99 + 8.66 = 18.88 GB — 0.05% off; AM's HEADLINE "Memory Used" was 19.70 GB,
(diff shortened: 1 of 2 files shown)


eneskirca:
**D) Reaper caught live — PASS** (main built with one disclosed deviation: `TMUX_SOCKET` → isolated `nt-verify-sock`, so no real user session was ever at risk; reaper/memory logic byte-identical to main).

Setup: 60 detached `nt-bait-*` sessions, `NODETERM_SESSION_GRACE_HOURS=0.9` (→ `graceSec: 0` via the envInt quirk below), ~66 min observation, `os.freemem()` measured 0.10–0.23 GB throughout — the reading under which the pre-#134 reaper reaps 8 on every sweep.

Every `session-budget` line the run produced — 12 of 12 count-cap-driven (60 > 48), zero memory-driven:
```
sweep @05:02 → 8 lines:  nt-bait-1, 10, 11, 12, 13, 14, 15, 16   (60 → 52)
sweep @~05:23 → 4 lines: nt-bait-17, 18, 19, 2                    (52 → 48, converged AT the cap)
```
Heartbeat (raw): `05:02 alive=60 lines=0` → `05:03 alive=52 lines=8` → `05:24 alive=48 lines=12` → **`05:59 alive=48 lines=12`** — ~36 min of silent sweeps with 48 eligible grace-0 sessions on the socket. (The gap between sweeps 1 and 2 was ~20 min, not 10 — macOS App Nap coalescing the backgrounded instance's timer; disclosed, does not affect the verdict.)

Caveats recorded in §10: on a healthy host this discriminates the original `os.freemem()` field bug, not #136 alone (an honest byte reading sits above the watermark here and is also silent) — the darwin-gated test in this PR is what pins #136. At measurement time no shipped build contained either fix; v0.2.46/v0.2.47, cut later the same morning, are the first that do — updating off 0.2.43 is what closes the field bug on a user's machine.

Found along the way: `envInt` floors AFTER its `> 0` guard, so `NODETERM_SESSION_GRACE_HOURS=0.9` silently yields `graceSec: 0` (grace protection off). One-liner fix available on request.

PR #148, opened 2026-08-12
fix(core): a fractional NODETERM_SESSION_* value must not disarm the reaper

eneskirca:
Found while running the live reaper experiment with `NODETERM_SESSION_GRACE_HOURS=1`.

`envInt` used `Math.floor(n)` behind `n > 0`, so a fractional setting became **zero** — and a zero here is not a smaller setting, it is the removal of a safety:

| Input | Before | Effect |
|---|---|---|
| `GRACE_HOURS=0.5` | 0 s | a session was reapable the moment it detached |
| `MAX_DETACHED=0.5` | cap 0 | every detached session counted as over-cap — a full batch died every sweep |

**The asymmetry is what made it dangerous rather than merely wrong.** `abc`, `''`, `0` and `-3` all fell back to the safe default, while `0.5` — the most plausible thing an operator would type, meaning "half" — silently disarmed the guard. A hand-editable value must degrade to the safe default, never to something more destructive than it.

`envInt` now requires `>= 1`. Sub-hour grace is a legitimate wish, so it is served properly by `envHours`, which keeps the fraction: `0.5` is 1800 seconds rather than nothing.

Mutation-verified: restoring the old `> 0` floor fails with `expected +0 to be 48` and `expected +0 to be 1800`.


eneskirca:
Reviewed (Opus): **Approve** — the bug is real and destructive (`MAX_DETACHED=0.5` floored to 0 ⇒ `overCap = detached.length` ⇒ a full batch reaped every sweep with healthy memory; `GRACE_HOURS=0.5` ⇒ zero grace). Both legs verified mutation-covered; suite green on the branch. Merging.

Non-blocking nits for a future pass:
1. `envHours`: a sub-second fraction still rounds to zero grace (`Math.round(0.0001*3600) === 0`) — same failure class, less plausible input. Cheap belt: `Math.max(1, Math.round(...))`.
2. The "a zero here is the REMOVAL of a safety" comment/doc claim holds for `MAX_DETACHED`/`GRACE_HOURS` but not `REAP_BATCH`/`MIN_AVAILABLE_MB` (where zero *disarmed* the reaper — the fix makes those slightly more reap-prone than before, correctly). One clause naming which keys were dangerous would keep the doc from teaching a wrong generalization.
3. `MAX_DETACHED='1.5' → 1` rounds in the unsafe direction (tighter cap than asked); `Math.ceil` for the cap only would make "never more destructive than asked" hold literally.
4. Naming: `envHours` returns seconds.

PR #161, opened 2026-08-13
[superseded by #193] feat(canvas): Cmd/Ctrl+0 fits view

gor3a:
## Summary
- Bind existing `fitAll()` to Cmd/Ctrl+0. Digit0 was deliberately excluded from the Cmd/Ctrl+1-9 project-jump chord's regex, so no collision.
- Matched on `e.code` (physical key), same convention as the project-jump chord, so it fires on non-US layouts too.
- Added row to the shortcuts panel.

## Test plan
- [x] `npm run typecheck` passes
- [ ] Manual: press Cmd/Ctrl+0 on canvas, confirm fit view triggers same as dock/Controls button/⌘K "Fit view"

### DIFF
--- a/src/renderer/canvas/Canvas.tsx
+++ b/src/renderer/canvas/Canvas.tsx
@@ -4682,6 +4682,12 @@ export function Canvas() {
       } else if ((e.metaKey || e.ctrlKey) && e.key === '/') {
         e.preventDefault()
         setShortcutsOpen((v) => !v)
+      } else if ((e.metaKey || e.ctrlKey) && !e.altKey && !e.shiftKey && e.code === 'Digit0') {
+        // Fit view. Matched on `e.code` (physical key), same as the Cmd/Ctrl+1-9 project-jump
+        // chord, so it fires from a non-US keyboard layout too. Digit0 is deliberately excluded
+        // from that chord's regex, so the two can never collide.
+        e.preventDefault()
+        fitAll()
       } else if (projectJumpDigit(e) !== null) {
         // Cmd/Ctrl+1-9 jumps to the Nth project — but only when the app actually owns the key
         // (desktop shell, and the digit addresses an open project). `liveProjectJumpTarget`
@@ -4702,7 +4708,7 @@ export function Canvas() {
     }
     window.addEventListener('keydown', onKey)
     return () => window.removeEventListener('keydown', onKey)
-  }, [toggleSessionsPin, switchProject])
+  }, [toggleSessionsPin, switchProject, fitAll])
 
   // Apply the accent color as a CSS variable.
   useEffect(() => {
--- a/src/renderer/components/ShortcutsPanel.tsx
+++ b/src/renderer/components/ShortcutsPanel.tsx
@@ -45,7 +45,8 @@ function buildSections(dictationKeys: string[], dictationLabel: string): { title
         { keys: ['Left-drag'], label: 'Box-select (touch to select)' },
         { keys: ['Middle / Right-drag'], label: 'Pan the canvas' },
         { keys: ['Double-click'], label: 'Center & focus a node' },
-        { keys: ['⌘', 'wheel'], label: 'Zoom in / out' }
+        { keys: ['⌘', 'wheel'], label: 'Zoom in / out' },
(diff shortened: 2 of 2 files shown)


eneskirca:
Thanks for this — the implementation itself is sound: it reuses the existing `fitAll` callback (same one behind the dock button, Controls, ⌘K and the context menu), so it does **not** reintroduce the unmeasured-node camera jump that `goToNode` was fixed for in b9b4147. `preventDefault` is scoped correctly, there's no listener leak, and your comment about `Digit0` being excluded from `projectJump`'s digit handling is accurate (pinned by `projectJump.test.ts:35-37`).

Four things before it can land, then one question for the maintainer.

**1. Guard the Kanban board.** This repo has a canvas-only-shortcut discipline (see the `isKanbanOpen(...)` early-returns used by undo/redo `Canvas.tsx:2309`, dictation `:3511`, Delete `:3850`). Today ⌘0 fires with the board open — and it isn't a no-op: `fitView` → `onMove` → `markDirty` + a persisted viewport + a team-sync cast, all under an opaque overlay.

**2. Guard text focus.** It currently fires while typing in a terminal, Monaco, an input, or with the ⌘K palette / Settings open. The ⌘C branch at `Canvas.tsx:4967` is the pattern to copy (`tagName` check, optionally `closest('.monaco-editor, .xterm')`).

**3. Server Edition (browser) story.** ⌘/Ctrl+0 is the browser's reserved reset-zoom, so in a browser tab the user gets **both**: the canvas fits *and* their page zoom is wiped. The repo already ruled on this class for ⌘1-9 — `projectJump.ts` gates on `isBrowserRuntime()` and `ShortcutsPanel.tsx:31-33` drops the row rather than "advertising a shortcut that never fires". Either gate handler + panel row the same way, or state in the PR that the double effect is intended.

**4. macOS menu check.** The app never calls `Menu.setApplicationMenu`, so Electron's default View → "Actual Size" is live on `CommandOrControl+0` (verified in the bundled Electron 42.7.1). Renderer `preventDefault` in `before-input-event` order *should* win, but this repo has already needed main-process interception for exactly this class (`src/main/index.ts:471-483`, ⌘M/⌘W). One Mac smoke check; if it double-fires, add a `key === '0'` branch there.

**Question for the maintainer (@eneskirca), worth settling before the above work:** ⌘0 conventionally means *actual size / 100%*, not *fit* — that's what the Electron menu item bound to it does, and Figma/tldraw/Excalidraw all use Shift+1 for fit and reserve 0 for 100%. This app has `zoomTo100` (`Canvas.tsx:1077`, palette `zoom-100`) with no shortcut at all. So one option is ⌘0 → `zoomTo100` and Shift+1 → fit, which would also make findings 3 and 4 mostly evaporate.

Happy to take it either way — just don't want you doing the guard work twice if the binding changes.

eneskirca:
Landed as #193 (merged: `8f5837b0`), with your commit credited via `Co-authored-by`. Thank you — this turned out to be a more interesting change than it looked.

The maintainer chose the convention-preserving split: **⌘0 → zoom to 100%** (which `zoomTo100` already implemented but had no binding for) and **Shift+1 → fit view**, matching Figma/tldraw/Excalidraw.

Two things worth knowing, since they came out of reviewing your PR:

**⌘0 was dead on desktop as written.** The repo never calls `Menu.setApplicationMenu`, so Electron's default menu is live and its View ▸ Actual Size owns ⌘0 — and a menu accelerator is handled *before* the page. A renderer-only binding would have silently reset the window's page zoom and never touched the canvas. It's intercepted in `before-input-event` now and forwarded, following the same pattern the repo already uses for ⌘M and ⌘W. Not something you could have known without a Mac build in front of you.

**Guards.** Both chords now early-return when the Kanban board is open or focus is in a terminal/editor/input, and on key repeat. Without the board guard the shortcut fired under an opaque overlay and still triggered `onMove` → `markDirty` → a persisted viewport and a team-sync cast.

Your read on `fitAll` was right, for what it's worth: reusing the existing callback meant it never went near the unmeasured-node camera jump that `goToNode` was fixed for, and your note about `Digit0` being excluded from `projectJump` was accurate.

Closing this in favour of #193.

PR #169, opened 2026-08-13
fix(ssh): keep large saved-server lists usable

### LINKED ISSUE #141: SSH server picker: list overflows the dialog with many saved servers (no scroll)
## Problem

With dozens of saved servers (easy to reach via **Settings → SSH → Import from ~/.ssh/config**), the server picker in the "Connect over SSH" dialog runs off the bottom of the box — reported by a user who could only reach roughly the first ten hosts.

`.confirm` caps the dialog at `max-height: 85vh` and makes `.confirm__msg` the scrolling region (`flex: 1 1 auto; min-height: 0; overflow-y: auto`). The server list in `SshProjectDialog.tsx` is a **separate** `<div style={{ display: 'flex', flexDirection: 'column', gap: 6, ... }}>` with no `min-height: 0` and no `overflow`, so it is neither the scroll region nor scrollable itself: the rows past the cap are simply unreachable.

Repro: save/import ~20 SSH servers, then New project → Connect over SSH.

## Fix

- Give the server list its own scroll region: `flex: 1 1 auto; min-height: 0; overflow-y: auto` (mirroring `.confirm__msg`), or a `maxHeight` + `overflowY: auto` like the directory list further down the same file (which already does exactly this at `maxHeight: 240`).
- The Cancel / Manage servers buttons must stay visible — that is the whole point of scrolling the list rather than the dialog.

## Worth doing at the same time

A type-to-filter input above the list. At ~30 imported hosts, scrolling is a workaround; filtering by label/host is the real answer (same pattern as the SCM branch quick-pick).

Check the same list in **Settings → SSH** (`SshSection.tsx`) for the same overflow — it renders the full server list too.

Reported by a user over email.

anonymort:
## What changed

- Put the saved-server rows in a dedicated flexible scroll region while keeping the title, filter, Cancel, and Add server actions fixed in the dialog.
- Added a pinned type-to-filter input matching label, host, user, and `user@host` case-insensitively.
- Added a clear no-results state without changing the existing empty-list state or SSH connection flow.
- Left Settings unchanged because its main content surface already owns vertical scrolling.

## Root cause

The capped `.confirm` flex column contained an unconstrained server-list child. That child had neither `min-height: 0` nor vertical overflow, so a large imported host collection extended beyond the visible dialog and pushed its actions out of reach.

## Verification

- `npx vitest run src/renderer/components/SshProjectDialog.test.tsx` with 6 tests passing
- Mutation check confirmed the layout regression test fails when `overflowY: 'auto'` is removed
- `npm run typecheck`
- `npm test` with 5,200 tests passing and 8 skipped
- `npm run build`
- `git diff --check main...HEAD`

## Surfaces

- Desktop and Server Edition use the same renderer component and are covered by this change.
- Mobile is not applicable because the native companion does not render this dialog.

## Not verified

No live SSH connection or manual Electron screenshot was needed for this renderer-only list-layout change. Existing connect, cancel, and browse-master teardown tests remain green.

Closes #141

### DIFF
--- a/docs/superpowers/plans/2026-08-13-ssh-server-picker.md
+++ b/docs/superpowers/plans/2026-08-13-ssh-server-picker.md
@@ -1,0 +1,216 @@
+# SSH Server Picker Overflow and Filtering Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make every saved SSH server reachable in the Connect over SSH dialog and add fast case-insensitive filtering for large imported server collections.
+
+**Architecture:** Keep the change within `SshProjectDialog`. The pick step becomes a bounded flex column whose server-list child owns vertical scrolling, while local React state filters the existing Zustand server collection without changing persistence or connection behaviour. The Settings surface remains unchanged because its parent page already scrolls.
+
+**Tech Stack:** React 18, TypeScript, Zustand, Vitest, jsdom, inline layout styles consistent with the existing component.
+
+## Global Constraints
+
+- Implement against `eneskirca/nodeterm` main at commit `711f34b1656a370c673b02e2b63a4bde699953d4`.
+- Keep the change restricted to issue 141, its tests, and the approved design and plan.
+- Preserve existing SSH connection, cancellation, browse-master teardown, folder browsing, and Add server behaviour.
+- Only the server-row list scrolls. The title, explanatory text, filter, Cancel, and Add server actions remain outside the scroll region.
+- Match label, host, user, and `user@host` case-insensitively after trimming the query.
+- Do not modify Settings because `SettingsPage` already supplies page-level `overflow-y-auto`.
+- Use test-first development and watch every new regression test fail for the expected missing behaviour before changing production code.
+- Required final gates are the focused component test, `npm run typecheck`, full `npm test`, `npm run build`, and `git diff --check`.
+
+---
+
+### Task 1: Pin the overflow and filtering regressions, then implement the picker fix
+
+**Files:**
+- Modify: `src/renderer/components/SshProjectDialog.test.tsx`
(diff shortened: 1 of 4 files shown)


eneskirca:
Merged — thank you, this is a clean fix and the tests have real teeth. I broke `overflowY: 'auto'` and swapped `visibleServers.map` back to `servers.map`, and both mutations were caught.

Verified before merging: typecheck, the full suite (baseline failures only), and `npm run build` all pass; merges into current `main` with zero conflicts. No security surface — the filter is `String.includes` on a lower-cased needle, no regex is built from user text, and no host/user/port reaches a shell or a path.

Five small follow-ups, none blocking, if you want a quick second PR:

1. **`SshProjectDialog.tsx:79-80` — guard against a malformed saved entry.** `ssh-servers.json` is loaded with only an `Array.isArray` check and cast straight through (`src/main/ssh-store.ts:23`, `parsed as SshServer[]`), so a legacy or hand-edited row missing `user` would make `.toLowerCase()` throw and blank the dialog on the first keystroke — where the old code merely rendered `undefined@host`. Narrow, but it is a robustness regression:
   ```ts
   [saved.label, saved.host, saved.user, `${saved.user}@${saved.host}`]
     .some((value) => (value ?? '').toLowerCase().includes(query))
   ```
2. **Enter-to-connect.** Issue #141 asked for "the same pattern as the SCM branch quick-pick", and that one acts on Enter. `onKeyDown={(e) => { if (e.key === 'Enter' && visibleServers.length === 1) void connect(visibleServers[0]) }}`.
3. `spellCheck={false}` on the filter input — hostnames otherwise get spellcheck squiggles. The SCM filter already sets this.
4. `role="status"` on the "No saved servers match …" paragraph, so screen-reader users are told the list emptied.
5. The two message paragraphs at :231 and :235 still inherit `overflow-y: auto` from `.confirm__msg`; `overflow: 'visible'` would match the heading paragraphs at :199/:202.

For the changelog: a user with 50 saved servers now gets a scrollable list with Cancel and "Add server…" pinned below it and a focused filter box on top, instead of a list whose last ~40 hosts *and both action buttons* ran off the bottom of the dialog.

PR #168, opened 2026-08-13
feat(sidebar): group sessions by status

dkattan:
## Summary

Adds a **Group sessions by status** preference so the sessions sidebar can regroup by live agent status — sessions needing attention float to the top across all local-core projects, instead of being buried in a collapsed project.
image

- New setting `sidebarGrouping: 'project' | 'status'` (default `'project'`, preserves current behavior)
- Status mode flattens across projects into always-expanded sections: **Needs you → Idle → Done → Running** (working sinks to the bottom)
- Each row shows its project monogram + name so you still know where an attention session lives
- Two controls: a compact **Project | Status** segmented toggle in the sidebar header, and a Settings row under Behavior
- Secondary sort within a section: project order, then title — stable, no reshuffle as statuses change

### Design constraints
- **No read/unread concept in grouping** — keys off `AgentNodeStatus.state` only. nodeterm already has the right property: clicking a row clears the cosmetic `unread` badge but never touches `state`, so a row never jumps/re-sorts on view.
- **Done/idle are never hidden or auto-collapsed.** A broken scraper regex can misclassify a genuinely-waiting session as done/idle; those rows must stay visible so the safety property holds: best-effort-needs-attention floats up, everything else stays visible below, never behind a click.

### Out of scope
Remote/relay live status in the sidebar is a pre-existing gap (relay `onAgentStatus` is unsubscribed + bare node ids would collide), orthogonal to this preference. Remote sessions render as `idle` in **both** modes today, so this introduces no regression.

### Bug fix included
`buildStatusList` keys ownership off the persisted `p.nodes` map (live nodes layered on top for the active project) to dedup rows across the project-switch window. Clicking a cross-project row flips `activeProjectId` synchronously while React Flow's nodes still hold the previous project's; without the owner-map dedup the same session appeared twice, tagged under both projects. Project mode hid this behind collapse — it only surfaced in status mode.

## Test plan
- [x] `npm run typecheck` — clean (node + web)
- [x] `vitest run` — full suite 5086 passed, 8 skipped; includes a new `buildStatusList` suite (section order, cross-project flatten + project tags, within-section sort, filter, empty-section omission, idle fallback) and a regression test for the cross-project dedup window
- [x] `npm run dev` — status mode renders attention-first sections with project monograms; toggle back to project mode is unchanged; cross-project clicking no longer duplicates rows
- [ ] Reviewer: spot-check the status priority order and the always-expanded behavior against the design intent

🤖 Generated with [Claude Code](https://claude.com/claude-code)
[...]

PR #171, opened 2026-08-13
feat(custom-agents): env/args/baseAgent inheritance + ${env:VAR} expansion

dkattan:
## Motivation

Let's say you run your own inference server and want to use Claude Code as your harness while preserving vanilla Claude Code for work with frontier models and remote MCP (which gets disabled when ANTHROPIC_API_KEY/ANTHROPIC_AUTH_TOKEN is set). 

You would typically accomplish this with a wrapper script that sets these env vars at the process level rather than system/user level as that would modify the default behavior of Claude Code. 

Nodeterm Custom Agents cannot currently be set as default and do not have a way to inject the base harness extension as there is no way to specify what the base harness is.

## What

Custom agents can be set as default
image

Custom agents get env/args/baseAgent inheritance with `${env:VAR}` expansion.

image

- **Inheritance**: a custom agent can declare a `baseAgent` (e.g. `claude`) and inherit its spawn command, capabilities, and config — then override only what it needs.
- **Env + args**: per-agent `env` and `args` are resolved and injected into the spawn environment / command line.
- **\`${env:VAR}\` / \`${env:VAR:fallback}\` expansion**: values expand against your shell environment at spawn time. These win over values nodeterm sets, so they can override an account login — e.g. set `ANTHROPIC_AUTH_TOKEN` / `ANTHROPIC_BASE_URL` to point a Claude-wrapping agent at your own proxy. A custom `PATH` should reference `${env:PATH}`, or CLI resolution may break. Custom `PATH` is not applied to remote sessions.

New pure modules in `src/shared/agents/` (`custom-agent.ts`, `expansion.ts`, `launch.ts`, `shell-quote.ts`) with full vitest coverage (56 tests).

## Status

- `npm run typecheck` ✅
- new test suites ✅ (56/56)

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/core/agent-env-ipc.ts
+++ b/src/core/agent-env-ipc.ts
@@ -1,0 +1,26 @@
+// Main/server IPC handlers for the custom-agent preview: env-var expansion and command assembly
+// run here (against the real `process.env`) because the renderer has no `process.env` of its own.
+// The preview is therefore guaranteed to match what `pty-manager` will actually type into the shell.
+
+import { IPC } from '../shared/ipc'
+import { platform } from './platform'
+import { assembleLaunchCommand, type LaunchInputs } from '../shared/agents/launch'
+
+/** A string-only snapshot of `process.env` (undefined entries omitted), for `${env:VAR}` expansion
+ *  in the renderer's preview. */
+function envSnapshot(): Record<string, string> {
+  const out: Record<string, string> = {}
+  for (const [k, v] of Object.entries(process.env)) {
+    if (typeof v === 'string') out[k] = v
+  }
+  return out
+}
+
+/** Register the env-snapshot + preview-command handlers. Call once at startup, beside the other
+ *  `registerIpc()` calls (main: src/main/index.ts; server: src/server/index.ts). */
+export function registerAgentEnvIpc(): void {
+  platform().handle(IPC.envSnapshot, () => envSnapshot())
+  platform().handle(IPC.agentPreviewCommand, (inputs: LaunchInputs) =>
+    assembleLaunchCommand(inputs, envSnapshot())
+  )
+}
--- a/src/core/custom-agent-env.test.ts
+++ b/src/core/custom-agent-env.test.ts
@@ -1,0 +1,132 @@
+import { describe, expect, it } from 'vitest'
+import { applyCustomAgentEnv, customAgentEnvArgs } from './custom-agent-env'
+import type { CustomAgent } from '../shared/types'
+
+const PROC = { MY_TOKEN: 'sk-abc', PATH: '/usr/bin:/bin' }
+
+describe('applyCustomAgentEnv', () => {
+  it('returns the env unchanged (shallow copy) when there is no custom agent / no env', () => {
+    expect(applyCustomAgentEnv({ A: '1' }, undefined, PROC)).toEqual({
+      env: { A: '1' },
+      warnings: []
+    })
+    const noEnv: CustomAgent = { id: 'custom:x', label: 'X', launchCmd: 'x' }
+    expect(applyCustomAgentEnv({ A: '1' }, noEnv, PROC)).toEqual({
+      env: { A: '1' },
+      warnings: []
+    })
+  })
+
+  it('merges custom env LAST so it wins over the existing env', () => {
+    const custom: CustomAgent = {
+      id: 'custom:p',
+      label: 'Proxy',
+      launchCmd: 'claude',
+      env: { ANTHROPIC_AUTH_TOKEN: 'proxy-token', NEW: 'n' }
+    }
+    const r = applyCustomAgentEnv({ ANTHROPIC_AUTH_TOKEN: 'account-token', KEEP: 'k' }, custom, PROC)
+    expect(r.env).toEqual({
+      ANTHROPIC_AUTH_TOKEN: 'proxy-token',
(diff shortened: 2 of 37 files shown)


dkattan:
This branch has been rebased directly onto the homogenize-menus base and now lives in the replacement stacked PR: https://github.com/dkattan/nodeterm/pull/3 (base: https://github.com/dkattan/nodeterm/pull/2). GitHub cannot retarget an upstream-repository PR to a fork-only base branch, so this PR is superseded.

eneskirca:
Thanks for these — I reviewed all three together, since they're a stack.

**The topology is clean and I'd like to land it as-is.** #171, #172 and #168 share `7ac95425..cede0be3` byte-for-byte (same SHAs, not rebased copies), and they merge in the order **#171 → #172 → #168** with zero conflicts — I verified by actually merging them in sequence; typecheck passes on the final state. Once #171 lands, GitHub will show #172 collapse to ~224 lines / 16 files and #168 to ~400 lines / 12 files. So no restructure needed.

#171 also has to go first for a correctness reason, not just size: its unique commit `ed35acea` fixes a launch bug the trunk introduces — a custom agent with `baseAgent: 'claude'` and a stale `promptInjectionMode: 'flag-prompt'` emits `claude --prompt '…'`, which the current CLI rejects. #168 and #172 carry the broken trunk without that fix.

The `--`/prompt-splicing rules survived intact (`launch.ts:126` keeps flags before `--` and the prompt after, matching `docs/grok-agent.md` §6), and `addMenuSpec.tsx` is a genuinely good DRY refactor. Four things to fix first:

**1. `remoteTmuxPtyArgs` (`control-master.ts:500`) splices `extraEnv` unquoted into the remote shell command.** Every other token there goes through `posixQuote`. This was latent while `extraEnv` was nodeterm-controlled — but `customAgentEnvArgs` now feeds it **expanded OS env values**, which is the feature's headline use case. Proven:

```
ANTHROPIC_AUTH_TOKEN = "${env:MY_TOKEN}",  MY_TOKEN = "`id > /tmp/OWNED`"
→ tmux -L nodeterm-rmt new-session -A -e ANTHROPIC_AUTH_TOKEN=`id > /tmp/OWNED` -s 'nt-x' ...
```

The value isn't authored by the user, so a proxy token containing a backtick, `$()`, `;` or even a plain space executes on the SSH host or corrupts the tmux invocation. Fix is `extraEnv.map(posixQuote).join(' ')` plus the one expectation at `control-master.test.ts:247`. (I have a separate hardening PR in flight for the same line, so if it lands first you'll just need the rebase.)

**2. `env:snapshot` and `agent:preview-command` are registered via `platform().handle`,** which puts them in the peer-dispatchable table — a paired phone, a collab desktop, or any authenticated Server Edition browser client can call `env:snapshot` and get the host's entire `process.env`, API keys included. `agent:preview-command` is a second read of the same data (put `${env:ANTHROPIC_API_KEY}` in `args`, read it back out of `command`). The rule is spelled out at `platform-electron.ts:41-47`. These want raw `ipcMain` on desktop, and a decision on the server.

**3. `${env:…}` in `launchCmd`/`args` never expands at launch — only in the preview.** All four real call sites pass `{}` (`workspace.ts:366`, `TerminalNode.tsx:2779/2849/2957`); `api.agent.envSnapshot` is plumbed through preload and types but has zero renderer consumers. So the settings card shows `claude-wopr '--model' 'opus' 'fix the bug'` while the shell receives `'--model' 'fix the bug'` — no program, and `--model` swallows the prompt. `launch.ts`'s header claims the preview can't drift from the real launch; today it always does.

Please pick one: thread the snapshot through, or restrict `${env:}` to the `env` map (which already works correctly, main-side). **Important coupling:** `launchCmd` is expanded and then used unquoted (by design — it's a command line). Today the blanking accidentally masks that. Threading the real env in without settling the `launchCmd` quoting story converts a silent-blank bug into a live local injection, so those two changes need to land together.

**4. #168 only — the tip commit `5b71d8bb` reverts PR #183 and #184.** It deletes `char-size-quantize.ts` + its test and the `.xterm-rows { -webkit-font-smoothing: auto }` rule — the two commits immediately before your merge-base, both still at the Mac verification gate. Its own message says it was committed "to free the index for a rebase." Please drop it; it's also unrelated to the sidebar feature.

Non-blocking notes:
- `WebNode.tsx:105` and `SubagentNode.tsx:63` import `NODE_MIN_SIZES` but still hardcode the literals, so the single-source-of-truth claim isn't quite true for 2 of 11 kinds (values match, no behavior change).
- `ws-bridge.ts` never got the `agent:` namespace the stub comment claims it overrides, so the Server Edition preview card silently never renders.
- `CustomAgent.promptInjectionMode` went required → optional. `settings:load` is peer-reachable, so if the iOS client decodes it as a non-optional Swift `Codable`, an inheriting agent that omits it will fail to decode. Worth confirming before this ships.
- The quoting in `launch.ts` is real and my mutation test confirms it's protected — but there's **no adversarial-payload test**: nothing pushes `;`, `$()` or a backtick through `args`/`env` and asserts it stays one literal argument. Right now it survives on incidental golden-string equality, so a future golden update would silently remove the guard. One such test would be very welcome.

Also FYI: your fork PRs sit at `action_required` until a maintainer approves the workflow runs, which is why CI looked dead — I've approved the current ones.

PR #185, opened 2026-08-14
fix(usage): the quota bar fills by the used/remaining mode

Temikus:
## The bug

Settings → Usage lets a provider limit read as **"% used"** or **"% left"**, but the toggle only ever flipped the number. Both bars — the popover's `LimitRow` and the compact pill's minibar — were hardcoded to fill with what was left.

So at 92% used the UI showed **"92% used" beside a nearly-empty sliver**: a bar contradicting its own label. The severity red was already correct there, just too small to notice.

## The fix

One pure helper, `barFillPercent(usedPercent, mode)`, now backs both fills.

```
92% used, mode 'used'       ▓▓▓▓▓▓▓▓▓░   red
92% used, mode 'remaining'  ▓░░░░░░░░░   red
```

## Blast radius

`usagePercentMode` defaults to `remaining`, so nothing should move for anyone until the setting is switched, making the default rendering unchanged.

`ContextMeter`'s bars are a different feature (per-node context window, not subscription quota) and are deliberately untouched.

**Surfaces:** pure renderer, so Desktop and Server Edition both get it with no bridge work. The mobile companion renders its own usage UI and is unaffected — no protocol change here.

## Verified

- `npm run typecheck` — clean, both projects.
- `npx vitest run src/renderer` — **2026 tests pass**. New unit tests cover `barFillPercent`, including the invariant that the fill always agrees with the number `percentNumber` renders beside it, and that colour stays keyed to remaining quota.

### DIFF
--- a/src/renderer/components/UsageIndicator.tsx
+++ b/src/renderer/components/UsageIndicator.tsx
@@ -5,7 +5,14 @@ import { useSettings } from '../state/settings'
 import { useProjects } from '../state/projects'
 import { useSshConn } from '../state/sshConn'
 import { scopeFromKey, scopeUsage, usageScopeKey } from '../lib/usageScope'
-import { formatResetCountdown, formatTimeAgo, percentNumber, percentText, severityColor } from '../lib/usageFormat'
+import {
+  barFillPercent,
+  formatResetCountdown,
+  formatTimeAgo,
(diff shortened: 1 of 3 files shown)


Temikus:
Amazing project, btw. Kudos!

PR #189, opened 2026-08-14
Switch the default Claude account from the usage popover

### LINKED ISSUE #142: Switch the default Claude account from the usage pill
## Request

> "I love seeing the current usage limits in the bottom left corner, I'm running two Claude accounts and would be awesome to switch between them via this element."

The usage popover already lists **one row per Claude identity** (System first, then each managed account). A user running two accounts reads that list, sees which one has headroom, and then has to go somewhere else entirely to act on it. The switch belongs where the decision is made.

## What "switching" means here

There is no global "active account" — `data.accountId` is resolved **once at node creation** and is then immutable (see CLAUDE.md, Managed Claude accounts). What a click can honestly do:

- **Set `project.defaultAccountId`** for the active project, so the next agent node opens under that account. This is exactly what the TabBar caret menu already offers; the popover row would be a second, better-placed entrance to the same action.
- Show which account is currently the project default (a check/dot on the row).

What it must **not** do is imply that existing sessions move — they cannot; a running session is tied to its `CLAUDE_CONFIG_DIR`.

## Sketch

In the usage popover, each account row gets a "Use for new sessions" affordance (or the row itself becomes clickable) that writes `project.defaultAccountId`. The active default is marked. Rows stay scoped by `usageScopeKey` as they are today, so on an SSH project only that host's accounts are offered — the same rule the panel already follows.

## Surfaces

- Desktop: primary.
- Server Edition: the popover is pure renderer + settings/workspace writes, so it should follow for free — verify.
- Mobile: N/A for now (launch-account selection already exists there separately).

Requested by a user over email.

anonymort:
## Why

The usage popover already shows every Claude identity and its remaining capacity, but users had to leave that decision point and open the project menu before choosing which account new sessions should use.

## What changed

- Make local and SSH Claude identity rows selectable when an account choice exists.
- Mark the active project default and expose the selection through native button pressed state.
- Map System back to the existing undefined default and managed identities to their existing account IDs.
- Persist through the existing workspace-dirty path without changing any existing node or running session.
- Preserve pointer toggling and make keyboard focus flow from the usage pill through the account rows before Refresh.
- Keep non-Claude provider rows read-only and keep the local System-only presentation unchanged.

## Surfaces

- Desktop – supported through the shared renderer.
- Server Edition – supported through the same renderer, usage bridge, and workspace-save path.
- Mobile – not applicable because mobile launch-account selection is a separate existing surface.

## Verification

- `npx vitest run src/renderer/components/UsageIndicator.test.tsx` – 7 tests passed.
- `npm run typecheck` – passed.
- `npm run build` – passed.
- `npm test` – 413 files passed, 5,325 tests passed, 8 skipped.
- Mutation check – removing the project default write made all original account-selection tests fail.
- Independent task review and whole-branch re-review found no remaining findings.

## Not verified manually

- The desktop and Server Edition popovers were not exercised in a running packaged UI. Their shared renderer path is covered by component tests, typecheck, production build, and the full suite.

Closes #142


eneskirca:
Reviewed this in depth. The security questions I had all came back clean, and the tests are the strongest I have seen on a contributed PR here. A few behavioural issues to fix before it lands.

## Security: clean

I checked three things specifically, because managed Claude accounts touch `CLAUDE_CONFIG_DIR` and per-account directories.

- **No cross-account config leakage.** The rows the popover offers are host-scoped by the pre-existing `scopeUsage()`, so a local account id can never be offered on an SSH project or vice versa. The click writes only `project.defaultAccountId` and never touches any node's `data.accountId`. Crucially there is **no retroactive repointing**: live sessions read `data.accountId`, and `CLAUDE_CONFIG_DIR` is computed once at spawn in `pty-manager.ts`; the only consumer of `defaultAccountId` is `resolveNewNodeAccount()`, which runs at **node creation**. Changing the default cannot move a running session's config dir.
- **No new `project.json` sink.** The ids that can be written come from `settings.claudeAccounts`, which is app-generated and Settings-owned. Nothing from a repo-carried `project.json` reaches a new filesystem path or HMAC input.
- **No command-line surface at all.** Zero `exec` / `spawn` / `path.join` / `.replace(` / token interpolation in the whole diff — it is renderer + CSS + tests.

## Tests: real

I broke six things and all six were caught, including the one that normally survives here (persistence written but untested):

| Mutation | Result |
| --- | --- |
| Delete the `setProjectDefaultAccount(...)` call | 3 tests fail |
| Delete `markWorkspaceDirty()` only | 3 tests fail |
| `accountId={r.accountId ?? undefined}` → `undefined` on the SSH row | 1 fails |
| Revert the `event.detail === 0` pill branch | 1 fails |
| Turn an inner `<span className="usage-row__title">` back into a `<div>` | 1 fails |
| Drop the `scoped.accounts.length > 0` guard | 1 fails |

CI is green on the updated branch and it merges into current main with zero conflicts.

## Requested changes

1. **The whole popover body is now a mutating control on a hover-opened readout.** `.usage-popover` opens on `onMouseEnter`, and its entire content area is now full-width `<button>` rows. A user who hovers to read their numbers and clicks anywhere — including to dismiss — silently changes the project's default account for new sessions, with no confirmation and no undo. Please scope the affordance to an explicit target inside the row (a "Use for new sessions" control, matching the design doc's own wording), or show a visible confirmation of what changed.

2. **On SSH, the "no choice ⇒ read-only" rule is missing.** The local branch correctly falls back to read-only when there are no managed accounts, but `visibleRemote.map(...)` is unconditional. I checked: an SSH host with only its system `~/.claude` row — per the code's own comment, "the common case" — renders one clickable button titled *"Default for new sessions"*, and clicking it fires `markWorkspaceDirty()`, i.e. a workspace write for a selection that changed nothing. Gate the remote rows the same way you gate the local ones.

3. **A no-op re-selection still dirties the workspace.** `selectDefaultAccount` writes and marks dirty even when the id is unchanged. Early-return when it matches.

4. **The checkmark ignores the stale-id normalisation the rest of the app applies.** `selected={defaultAccountId === a.id}` compares the raw stored value. With `project.defaultAccountId = 'gone'` (an id not in `claudeAccounts` — reachable from a `project.json` that travelled in a cloned repo) **no row is marked at all**, while `resolveNewNodeAccount` and the launch menus resolve it to System. So the popover says "nothing is default" where the app behaves as System. Normalise first, as `Canvas.tsx` and `Dock.tsx` already do:

   ```ts
   const effectiveDefault = scoped.accounts.some((a) => a.id === defaultAccountId)
     ? defaultAccountId
     : undefined
   ```

   and please add a test — none of the current seven covers a stale id.

5. **No keyboard way to dismiss the popover.** The new `event.detail === 0` branch makes keyboard activation of the pill always `setOpen(true)`, so Enter/Space can open but never close, and there is no Escape handler.

6. *(Optional, a11y)* These rows are a mutually-exclusive set but use `aria-pressed`, so a screen reader announces each as an independent toggle. `role="radiogroup"` + `aria-checked` would describe what is actually happening.

Items 1 and 4 are the ones I would insist on. Everything else is small.

For the record, the user-visible change once this lands: clicking a Claude identity row in the usage popover makes that account the default for **new** agent nodes in the active project, replacing a trip to the TabBar caret menu — existing and running nodes are untouched.

anonymort:
@eneskirca I have pushed an update addressing the requested changes.

- Account selection now uses an explicit "Use for new sessions" control, so reading or dismissing the popover cannot change the project default.
- SSH System-only rows are read-only.
- Re-selecting the existing default does not dirty the workspace.
- Stale account ids now display System as the effective default.
- Keyboard activation toggles the popover, Escape dismisses it, and the selectors are an accessible radio group with Arrow-key navigation.

The focused UsageIndicator suite, typecheck, and production build pass. I also synced the branch with current main.

eneskirca:
Closing this with an apology owed, and I want the record to say clearly: this was **overtaken, not rejected**.

The timeline wasn't fair to you: I left a six-point review, you addressed every point within a day — including the mutation-style tests, which were genuinely well built — and the very next day my own parallel implementation (#253) merged and closed #142, and this PR never got the follow-up it had earned. That's on us, not on the work.

Today the feature lives in main via #253 and the later #429 rework, and this branch is ~765 commits behind with conflicts, so merging it no longer makes sense. But if you're still interested in contributing, I'd gladly review a fresh PR — and issues you open get real attention. Thanks for the quality work and the fast turnaround; sorry it ended up racing an internal change.

PR #203, opened 2026-08-15
feat(agents): read pane ownership from the kernel, not a name allowlist

eneskirca:
PR 2 of the agent-messaging plan: gate 1. Independent of the rest of the series — nothing else in
the plan has to land for this to be correct, and nothing here is wired into a delivery path yet
(PR 3 is the first consumer).

## What `isShellCommand` misses today

The current "is this an agent pane?" gate is `!isShellCommand(pane_current_command)`
(`src/shared/agents/pane.ts:7`), and `isShellCommand` is a **seven-name allowlist**:
`{zsh, bash, sh, fish, dash, ksh, tcsh}`. Negate it and every other foreground command reads as an
agent. So a plain terminal running **`nu`**, **`pwsh`** or **`xonsh`** — also `elvish`, `ssh`,
`python` — is admitted as an agent pane today. Text delivered there **executes**, because `sendText`
appends Enter (`pty-manager.ts`).

It is wrong in the other direction too, and measured on a real pane rather than reasoned about:

```
$ tmux -L probe display-message -p -t owner-probe '#{pane_pid}|#{pane_tty}|#{pane_current_command}'
2485382|/dev/pts/0|node          <- the pane is running `…/bin/claude --resume x`
```

`#{pane_current_command}` answers `node` for every npm-installed agent CLI, so the field cannot tell
claude from codex from any other node process either. And `agentId` — the persisted field the old
gate leaned on — is durable and outlives the CLI, so a terminal whose claude exited weeks ago still
claimed to be an agent node.

## The new predicate

Kernel truth. Every terminal has a **foreground process group** (`tcgetpgrp`) and only that group
can read the tty, so whatever is in it is what will receive the bytes we are about to write:

1. `PANE_OWNER_FMT` — one `display-message -p` for `#{pane_pid}|#{pane_tty}|#{pane_current_command}`.
2. one `ps` on that tty, and the foreground group taken from `stat`'s `+` flag.
3. `isAgentPane(owner, agentId)` matches the **basename of argv[0]** — after resolving an
   interpreter to the script it runs (`node …/bin/claude` → `claude`) — against that agent's known
   binaries. Never a substring of the command line: `vim /etc/claude.conf` and `grep -r claude .`
   are not claude panes.

It subsumes "target must be an agent node" as a side effect, and it is unspoofable by a stale field.
Three-valued on purpose: `agent` / `not-agent` / `unknown`, because "the probe failed" and "the
kernel said no" need different replies.

`isShellCommand` is **not** deleted. Its existing callers (the restart exit poll, the SSH reconnect
resync) ask a different question — "has the CLI let go of the pane?" — for which a shell allowlist is
the right shape. Its docblock now says so.

### What the prototype measured, and what it ruled out

The plan's sketch assumed Linux would read the foreground pgid from `ps -o tpgid=` and then list it
with `ps -g <pgid>`, with a separate `darwin` branch. Run against a real pane on this host
(Linux 6.8, procps-ng 4.0.4), that is **wrong in its second half**:

- `ps -g <pgid>` does not select a process group here at all — `-g` is session/effective-group-NAME
  — and the GNU long option `--pgid` does not exist (`error: unknown gnu long option`).
- selecting by **tty** works identically on Linux and BSD (`-t /dev/pts/0` and `-t pts/0` both
  accepted), and `stat`'s `+` flag means "in the foreground process group of its controlling
  terminal" on **both**. `tpgid` is the procps-only column — asking for it is what would have forced
  the platform branch.

So the committed shape is **one invocation valid on both platforms**,
`ps -ww -o pid=,pgid=,stat=,args= -t <tty>`, with no `process.platform` branch. `-ww` is
load-bearing on BSD, which truncates the argv to the terminal width by default and would cut off
exactly the script path that names the agent. macOS was not reachable to measure — **Mac
verification owed** — but it is unverified for a shape whose every piece is POSIX/BSD-documented,
rather than for a column BSD is documented not to have. The measurement is recorded in the module
docblock, including what it ruled out.

## The SSH leg

`paneOwner` mirrors `paneCommand`'s dispatch exactly: an `sshRemote` session is read on the REMOTE
tmux over the project's ControlMaster (`remotePaneOwnerArgs`), everything else on the local socket.
Both round-trips ride the same master.

`#{pane_tty}` is a value the **remote host printed back at us**, crossing into a line the remote
login shell parses — the class `remote-safety.ts` exists for. Two layers, both tested:

- `isSafeTty` refuses anything outside `[A-Za-z0-9/._-]` (and any `..`), so for a hostile tty
  **nothing is built at all** — the second round-trip never happens;
- what survives is `posixQuote`d at the splice, proven with `control-master.injection.test.ts`'s own
  shell parser (`unquotedMeta` must be empty).

**No credential is on either command line, and none may be added** — a tty path and a pid are the
whole payload (Global Constraint 6). Said so in a comment so nobody adds a `-H` later.

## Fail-open directions

The direction is the same everywhere: **a read that did not happen is never evidence that a pane is
an agent.** Every failure lands on `unknown`, and `unknown` is refusable/retryable, never `agent`.

| failure | `paneOwner` | `isAgentPane` | what the gate does |
| --- | --- | --- | --- |
| pane unreadable (`display-message` errors, empty output) | `null` | `unknown` | refuse, retryable |
| pane dead / "can't find session" | `null` | `unknown` | refuse, retryable |
| tmux not installed, or `ssh` not found for an SSH node | `null` | `unknown` | refuse, retryable |
| `ps` missing these flags on some platform | `null` | `unknown` | refuse — a refusal, never a guess |
| `ps` lists the tty but marks no foreground group | `null` | `unknown` | refuse, retryable |
| unsafe tty | `null`, and the `ps` is never built | `unknown` | refuse |
| slow read (wedged tmux, half-dead ControlMaster) | bounded by the caller: `probeWithin` answers `null` at `PANE_PROBE_TIMEOUT_MS` = **2s** | `unknown` | refuse, retryable — the deadline is a refusal, not a hang |
| a custom agent whose binary nothing can name | owner is fine | `unknown` | refuse, retryable (or pass an explicit binary list) |

`paneOwner` never returns a partial object: an owner carrying no argv would read to the predicate as
"the kernel answered", so that case is `null` at the source. And `paneOwner` deliberately has no
deadline of its own — the caller bounds it, exactly as the restart poll bounds `paneCommand`.

## Three surfaces

- **Desktop (Electron)** — both legs live: the local socket and SSH over the project's ControlMaster.
- **Server Edition** — everything here is `src/core` + `src/shared`, so it ships on **both shells**
  and `paneOwner` works there identically (the Server Edition drives the same `PtyManager`). It is
  reachable but has no caller there yet: the messaging *verbs* do not exist on the Server Edition
  (plan Task 5.3), so the gate ships and simply is not asked. This is invariant 4 — a hook-server
  change that lands on one shell only has bitten this repo three times.
- **Mobile (phone)** — N/A as a sender. As a **target** its pane is read by the host exactly like any
  other: a phone-spawned session is a tmux session on the host, with a tty and a foreground group,
  and nothing in this read cares who spawned it.

## Proof, and the mutation results

Proved against the **real thing**, not a mocked `display-message`: `pane-owner.test.ts` spawns a real
tmux server on a private socket in its own `TMUX_TMPDIR`, runs a real command in a real pane, calls
the real `ps`, and asserts the foreground group is the command — not the pane's own shell. A second
case asserts an idle pane reports its shell, so an idle pane can never read as an agent. The server
and its tmpdir are removed afterwards; `kill-server` alone leaves the socket file behind, which on a
shared host is litter.

**Mutation testing: 16 mutants, all red.** Reverted one behaviour at a time, confirmed the suite went
red, restored.

| # | mutation | result |
| --- | --- | --- |
| M1 | foreground `+` filter dropped — any row counts as owner | RED (8 failed) |
| M2 | pgid filter dropped — a backgrounded job counts as owner | RED (6) |
| M3 | `-ww` dropped — BSD would truncate the argv | RED (3) |
| M4 | `isSafeTty` always true — hostile tty reaches the remote shell | RED (9) |
| M5 | `paneOwnerFrom` returns a partial owner (empty argv) | **survived → fixed**, then RED (3) |
| M6 | `parsePaneOwner` accepts pid 0 / malformed | RED (2) |
| M7 | `isAgentPane` matches a substring of the command line | RED (2) |
| M8 | `isAgentPane` upgrades `unknown` to `agent` | RED (1) |
| M9 | interpreter resolution reaches past an option | **survived → fixed**, then RED (1) |
| M10 | unknown agent id falls back to allowing anything | RED (1) |
| M11 | SSH branch ignored — an SSH node read on the LOCAL socket | RED (2) |
| M12 | `paneOwner` throws instead of answering null | RED (3) |
| M13 | `probeWithin` never times out | RED (3) |
| M14 | `probeWithin` lets a rejection escape | RED (2) |
| M15 | `probeWithin` leaves the timer running | RED (1) |
| M16 | remote `ps` args unquoted | RED (1) |

The two survivors were real gaps and are fixed in `d7364eb0`:

- **M5** — deleting `paneOwnerFrom`'s empty-argv guard changed nothing, because the pgid guard above
  it already covered the same fact. Two guards for one decision means one is a line no test can
  reach; collapsed to a single one.
- **M9** — deleting the "an interpreter's next token must not be an option" rule changed nothing,
  because no test used an option that basenames to an agent name. `node -claude` is that case: strip
  the login-shell dash and an interpreter FLAG reads as the script it is not.

## Gates

- Per-task suites: `pane-owner.test.ts` 33, `pty-manager.pane-owner.test.ts` 11,
  `pane-owner-predicate.test.ts` 12, `pane-probe.test.ts` 7, plus 3 added to
  `control-master.injection.test.ts` and the untouched `pane.test.ts` 3 — **74 passing**.
- Full `npx vitest run` — **5854 passed, 12 skipped, 4 failed**: 3 `node-pty-patch` + 1
  `webgl-addon-pair`, both environmental on this host (they read `node_modules`), failing on
  `origin/main` too.

Read-only against everything on this host that this branch did not create: `tmux display -p` and
`/proc`/`ps` reads only. No keys sent, nothing killed, the shared `node-terminal` and `nodeterm-rmt`
sockets untouched (71 sessions before and after).


eneskirca:
Review fixes pushed as `db182ab5`. Both required items done, the recommended one too.

**REQUIRED 1 — agreed, and applied.** The reading of FreeBSD `bin/ps/keyword.c` `parsefmt()` is
correct: an item containing `=` is a column header, `tempstr = NULL`, always the last item — so
`-o pid=,pgid=,stat=,args=` is the keyword `pid` with the header `,pgid=,stat=,args=`, one column,
exit 0, silently useless. I have no counter-evidence and would not have caught it: my measurement
was procps-only and macOS-unreachable, so the PR's darwin claim was resting on a shape that happens
to be right on Apple for a reason (`#ifndef __APPLE__`) I had not established. The remote host is the
one that matters here and it is not macOS by default.

Now `['-ww', '-o', 'pid=', '-o', 'pgid=', '-o', 'stat=', '-o', 'args=', '-t', tty]`. Verified
byte-identical to the old form on this host (procps-ng 4.0.4) against a real pane running
`sleep 300 | cat` — `diff` clean, both exit 0. The `keyword.c` reasoning and the FreeBSD ps(1)
sentence that sanctions the multi-`-o` form are in the docblock, plus its own test asserting four
`-o` flags and no comma in any argument. The three argv assertions (`pane-owner`,
`pty-manager.pane-owner`, `control-master.injection`) moved with it.

**REQUIRED 2 — done.** `binariesFor(agentId, customAgents)` + `binaryFromLaunchCmd`, alongside the
predicate; `isAgentPane` now defaults to `binariesFor(expected)` instead of raw `AGENT_BINARIES`, so
an unrecognised non-custom id (`aider`) resolves too, mirroring `resolveAgent`'s fallback.

`launchCmd` is free text, so the walk skips options, env assignments (`env FOO=1 …`), package
runners and their verbs (`npx -y`, `pnpm dlx`, `bun run`) and interpreters, strips a trailing npm
version (`@acme/my-agent@1.2.3` → `my-agent`), and — the part that matters — **refuses** when the
first real token is in `NEVER_A_BINARY` (`bash -lc "…"`, `sh -c`, `ssh`, `sudo`, `tmux`, …). A wrong
name there would not fail to match, it would match every idle shell on the canvas: the
`isShellCommand` hole re-opened through the custom-agent door. Null still means `unknown`, but it is
now reachable only for a custom agent whose definition the caller did not pass, or one whose launch
command genuinely cannot be named.

**RECOMMENDED 3 — added.** The 128-char `isSafeTty` cap has its own case, which asserts the value is
charset-clean (so it passes every other check) and still refused.

**Carried into the docblocks, for the caller PR:**

- `pane-probe.ts` — a section headed "DO NOT RETRY `unknown` ON A FIXED SHORT TIMER WITHOUT A
  CIRCUIT BREAKER", with your measurement: the deadline abandons the wait at 2s, the `ssh` child
  lives to `runAsync`'s 15s reap, `childArgs` carries `ControlMaster=auto`, so against a dead master
  each retry is a real login — naming the 72k-logins/day incident so the next reader knows it is not
  hypothetical.
- `isAgentPane` — "WHAT THIS DOES NOT PROVE": the `sh -c "sleep 600 | claude"` case answers `agent`
  correctly while claude's stdin is the pipe, so the verdict means *in* the foreground group, not
  *reading* the tty. Plus the two terminal false negatives (script path with a space; differently
  named symlink), documented as refusals rather than fixed. Both have tests pinning the current
  answers, so a future change to either is deliberate.
- `paneOwner` — notes that `remotePaneOwnerArgs` splices the session id unquoted like every sibling
  builder and that the guarantee is `sessionName()`'s `[A-Za-z0-9_-]` sanitisation, living there and
  not in this call. Left as-is, as you said.

**Gates:** typecheck clean. Six new mutants for the new assertions, all red — single-`-o` restored
(5 failed), length cap dropped (1), `NEVER_A_BINARY` bypassed (2), `customAgents` ignored (2),
unknown `custom:` id guessed (2), runner/interpreter skip removed (4). Full suite:
**5869 passed, 12 skipped, 4 failed** — the same 3 `node-pty-patch` + 1 `webgl-addon-pair`
environmental ones, unchanged from `origin/main`.

PR #207, opened 2026-08-15
feat(agents): the agent-message delivery primitive

eneskirca:
PR 3 of the agent-messaging plan: the delivery primitive. Builds on #203 (kernel-truth pane
ownership) and #204 (`stateVerified` / `verifiedAt` / `clientRevision`, committed through one
`commitState`).

**This ships INERT.** Nothing an agent can reach calls any of it. There is no verb, no route, no
handler, no shim — those are PR 5, behind the per-project switch of PR 6. Every module here is
`src/core`, so it ships on both shells, and on the Server Edition the verbs will answer the named
`control-unsupported-on-this-edition` refusal (#206). Merging this changes no product behaviour: the
only reachable diffs are a board-log rotation, one new `BoardLogEvent` type, and an `idleInferred`
marker on the status mirror.

## The six gates, and which task implements each

| Gate | Where it is enforced | Task |
| --- | --- | --- |
| **G1** the target's pane is really owned by the expected agent | `decideDelivery` gate 1 over `isAgentPane(paneOwner(...))`, probed inside the lock and bounded by `probeWithin` | 3.1 + 3.3 |
| **G2** the target is KNOWN idle on a VERIFIED status | `idleRefusal` + `identityRefusal` — `undefined`, `working`/`waiting`/`blocked`, `restored`, an `idle_prompt`-inferred `done`, and any `done` whose evidence was `stateVerified: false` are all refused, each with its own reason | 3.1 + 3.6 |
| **G3** the pane did not change hands under us | the post-write re-verify → `deliveredToReplacedTarget`, never `delivered` | 3.3 |
| **G4** the message actually landed | `awaitReceipt` — an observed, **verified** advance, or a typed `stalled` | 3.4 |
| **G5** the unread bit is untouched | no `clearUnread`/`setActive` in `DeliveryDeps`, asserted by running a delivery through a Proxy that throws on any undeclared key | 3.3 |
| **G6** every delivery leaves a trace | `recordDelivery` → board log where there is one, bounded ring where there is not, `traced` on the wire either way | 3.5 |

Gate 0 (the per-project switch) and the rate limiter arrive in PRs 6 and 4; the decider already
carries their outcomes (`notPermitted`, `rateLimited`) so those PRs add a fact, not a branch.

## The typed outcome table

`AgentMessageOutcome` is a 16-member discriminated union, and `RETRYABLE` is a
`Record<Kind, boolean>` so a missing row is a **compile error**, not a runtime `undefined` that reads
as "do not retry". The caller is a language model: an outcome that is not retryable must say so, or
it retries a permanent refusal in a busy-loop.

| Outcome | Retry? | Means |
| --- | --- | --- |
| `delivered` | — | written, pane unchanged, and an advance observed (`signal: newTurn \| working`) |
| `deliveredToReplacedTarget` | no | the bytes went out; the pane was not the same afterwards. Never reported as success |
| `stalled` | no | written, nothing advanced within `RECEIPT_DEADLINE_MS` |
| `queued` / `expired` / `queueFull` | –/yes/yes | PR 7's deliver-on-idle queue |
| `rateLimited` | yes | PR 4, carries `retryAfterMs` |
| `targetBusy` | yes | `working` / `waiting` / `blocked` |
| `targetNotIdleUnknown` | yes | no status, between sessions, restored from disk, or an `idle_prompt`-inferred `done` — with the reason in prose |
| `targetHookScriptStale` | **no** | Finding F2 — the script predates per-node identity. Names the action per surface |
| `targetStatusUnverified` | **no** | no token file: the node cannot prove itself. Names the action |
| `targetStatusStale` | yes | token file present, current script, no verified event yet |
| `targetNotAgentPane` | no | the kernel said no, or could not be read (`observed: 'unknown'`) |
| `targetNotPasteAware` | no | herdr `:260` — never send a multi-line body unframed |
| `targetGone` | no | no live session, or the write itself failed |
| `notPermitted` | no | `switch-off` / `cross-project` / `self-send` / `unsupported-edition` |

**Order is cheapest-and-most-permanent first, and it is real rather than decorative.**
`decidePreProbe` runs the free gates *before* any pane is touched, so a caller that is not permitted
or is over budget is refused with **no tmux round-trip** (and, on an SSH project, no `ssh` one). The
ordering test walks `DECISION_ORDER` by clearing one fact at a time, asserted by running the decider.

**The three unverified refusals** are three populations needing three actions, and collapsing them is
worse than a bare refusal. The stale script is checked FIRST because it is the outer cause: the token
file is present in that case, so a token-file check alone answers "retry after its next turn" —
forever, because the script cannot read the file. Measured 2026-08-15: a phone-spawned session lands
in none of the three. It verifies, because of the persist-time token sweep, **not**
`ptyManager.create`, which it never touches.

## Why the envelope cannot be forged

A fixed literal prefix is forgeable by construction: A writes `--- NODETERM MESSAGE ---` into its own
body and B cannot tell A's forgery from the app's frame, because they are byte-identical.

The frame carries a **per-delivery nonce the sender never sees** (`FRAME_NONCE_BYTES = 9` → 12
base64url chars). A supplies a body; the app frames it with a value A was never given. Two separate
mechanisms, defeating two different attacks:

- **Header fields cannot start a line.** `from:` / `reply-to:` are prefixed, and `oneLine` strips
  every newline from the title and the ids — so a user-editable node title cannot produce a frame
  line whatever it contains. (`frameLineRe` is anchored, and the test asserts exactly two matching
  lines with a title that tries.)
- **A body CAN start a line** — legitimately, a body may quote another message's envelope verbatim.
  That quoted frame survives as data, exactly as it should, and fails to match because it carries a
  different delivery's nonce.

The body gets **one** sanitizing pass, against both delimiters: `sanitizePasteText` removes every ESC
byte, so no byte of the body can express paste structure at all. And the two rules stay separate, as
Global Constraint 7 requires — `sanitizePasteText('a\nb') === 'a\nb'` (delivery: a newline is
legitimate), `oneLine('a\nb') === 'a b'` (composition: it never is). Neither is a general-purpose
"sanitizer"; a merged one would have to guess which caller it had.

**Stated in the code, not assumed:** within one uid the nonce probably cannot be kept from A — a
sibling that can read the token file can likely read the tmux session env too. It raises the cost of
forgery. It is **not** a boundary, and nothing downstream may treat it as one.

## The receipt's mechanism

Snapshot nothing, write, then require an **observed advance** within 8 s or return `stalled`. rev. 2
had no receipt at all, so it could not distinguish "B read it" from "the body is sitting in B's
composer with no submit" — the failure herdr's CHANGELOG records twice.

- The signal is `newTurn`, emitted by **all five** supported agents on turn start (claude
  UserPromptSubmit, codex, gemini, opencode, grok — measured in the normalizers). A bare `working`
  transition is a documented fallback for a misconfigured install, and the outcome records **which**
  signal was used, so a stalling host is diagnosable from the trace.
- The receipt must itself be **verified**. An adversary who could fake a receipt could confirm its
  own deliveries — the same hole gate 2 exists to close, and PR 7's queue would consider such a
  message consumed.
- An event for another node never satisfies it.
- `RECEIPT_DEADLINE_MS = 8000`, asserted numerically against the real `CONTROL_CEILING_MS` (now
  exported for that purpose). The 130 s ceiling and the desktop's 120 s await exist so a **human
  confirmation dialog** can stay open; borrowing that headroom for an automated wait would make a
  stuck receipt look like a stuck dialog, which is the one thing the person debugging this must be
  able to tell apart.

## The three carried warnings — each in a docblock where the code lives

1. **Do NOT retry an `unknown` pane verdict on a fixed short timer without a circuit breaker.**
   (`agent-message.ts`, at the pre-flight probe.) `probeWithin` abandons the WAIT, not the WORK: it
   answers at 2 s while the `ssh` child lives to `runAsync`'s 15 s reap, so a 2 s retry loop stacks
   ~7 overlapping children per pane — and `childArgs` carries `-o ControlMaster=auto`, so against a
   dead master **each retry opens a real login**. A pane unreadable *because* its master died is the
   worst case. That is the 72k-logins/day shape this repo already lived through. This module probes
   at most twice per delivery and never loops; `targetNotAgentPane` is marked not-retryable partly
   for this reason.
2. **`isAgentPane` proves the agent is IN the foreground process group, not that it is the thing
   READING the tty.** (`agent-message-decide.ts`, on the `pane` fact.) `sh -c "sleep 600 | claude"`
   answers `agent` correctly, but claude's stdin is the pipe, so text written to the pane sits in the
   tty buffer until the shell reads it after the pipeline exits — the "rename splice = lost launch"
   shape. The **receipt** is the only thing in this feature that makes that case observable rather
   than silent, and it is why `stalled` exists.
3. **Two terminal false negatives, expected and not papered over** (same docblock): an agent script
   path containing a **space**, and a differently-named **symlink** to the agent binary. Both answer
   `not-agent` ⇒ a refusal, which is the safe direction. Do not "fix" either by matching a substring
   of the command line — `vim /etc/claude.conf` would become a claude pane.

## Proved against a real reader

`agent-message.realtty.test.ts` runs a real tmux server on its **own socket inside its own
`TMUX_TMPDIR`**, spawns real bash panes via `exec -a claude bash …` so the **kernel** really reports
the pane under the agent's name, calls `deliverAgentMessage` with a real `paneOwner` (#203's parser
over real `tmux` + `ps`) and a real `sendFramed` (`tmux send-keys -l --`), and reads back with
`capture-pane`. The server is killed in `afterAll` and **verified gone**.

- the whole multi-line envelope arrives as **one block**: every frame/header/body line is present and
  the shell's first error lands *after* the closing frame line;
- a body carrying `\x1b[201~` + `touch <marker>` does **not** execute — the marker is never created,
  and the printable tail (`hello[201~`) still arrives as content;
- with bracketed paste off the delivery is **refused** and `capture-pane` is byte-identical before
  and after;
- a pane a plain `bash` owns is refused by gate 1 against kernel truth, with nothing written.

Two **mutation controls run, not described**: the same body framed *unsanitized* escapes the frame
and creates the marker; the same envelope delivered *unframed* is submitted line-by-line (the shell's
first error lands *between* the frame lines) — herdr `:260`, demonstrated.

## Measurement worth flagging for review

On this host (tmux **3.4**, Ubuntu 24.04) `#{bracket_paste_flag}` **does not exist as a format** —
it is absent from the binary's format table and from the man page's FORMATS list, and expands to
empty exactly like a bogus name (`insert_flag`, `keypad_flag`, `wrap_flag` are all present). Since
`PtyManager.bracketPasteRequested` compares against `'1'`, on such a host it answers `false` for
every pane. Consequences: `sendText` always takes the legacy two-step path there, and **every
agent-message delivery would be refused with `targetNotPasteAware`** — the correct fail direction,
and the direction a test in this PR pins (whatever the running tmux answers, an answer that is not
`'1'` must refuse, never fall back to the unframed path). The app bundles tmux 3.7b, which was not
measurable from here. Flagging rather than fixing: it is not this PR's defect, and PR 5's rollout
should know about it.

## Deviations from the plan, stated

- **`clientRevision` is read off `MirrorEntry`**, not passed as a separate fact. #204 landed it on
  the entry, committed through `commitState` with the state — reading it anywhere else would
  reintroduce the drift that PR fixed.
- **The "two lines contain NODETERM" assertion is expressed against the anchored, nonce-bound frame
  pattern** instead of the substring. The substring form would force us to mangle a legitimate title
  containing the words; the property that matters is that no line *other than ours* matches the frame
  for **this** delivery's nonce, and that is what is asserted.
- **`targetLive` is a caller-supplied fact**, not derived from an unreadable pane: `paneOwner`
  answers null for a dead pane *and* for a missing tmux, a failed `ps`, a lapsed deadline. Calling
  all of those "the node is gone" would be a confident answer to a question we did not ask.
- **The union has 16 members** (the plan's prose says 14 in one place and 15 in another; the listing
  itself has 16).
- **`MirrorEntry.idleInferred`** is new here: gate 2 must refuse a `done` rescued from `idle_prompt`,
  and nothing recorded which kind of `done` an entry held. Written inside `commitState`, so it can
  never describe a state it did not arrive with.
- **Board-log rotation is local-only**: `RemoteLogExec` exposes `append`/`tail` and nothing that can
  stat or rename. Documented in the constant's docblock rather than left silently uneven.

## Gates

`npm run typecheck` clean. Full `npx vitest run`: **6049 passed**, 4 failed — the 3 `node-pty-patch`
and 1 `webgl-addon-pair` failures that are environmental on this host (no rebuilt native module).

**Mutation-checked: 26 mutations applied to production code, each run against the suite that should
catch it, then restored. 26/26 killed** — but only after the first pass found **one false green**:
the trace ring's eviction test was green with the eviction removed, because it read back with
`recentDeliveries()`'s default limit, which *is* the capacity — a **read cap masquerading as
eviction**, the same shape as the board-log 500 this very PR rotates around. Fixed by asking for more
than the capacity; the mutation now fails.

### DIFF
--- a/src/core/agent-status-mirror.test.ts
+++ b/src/core/agent-status-mirror.test.ts
@@ -1620,6 +1620,35 @@ describe('idle_prompt rescue (Esc that ran no Stop hook)', () => {
     expect(next.state).toBe('done')
     expect(next.updatedAt).toBe(1000)
   })
+
+  it('marks the RESCUED done as inferred, so a consumer can tell it from a turn end', () => {
+    // A node blocked on an approval is also "idle at its prompt". Messaging's gate 2 refuses this
+    // shape, and it can only refuse what the entry remembers.
+    const working = reduceEntry(undefined, ev({ state: 'working' }), 1000)
+    const rescued = reduceEntry(working, ev({ state: 'done', idle: true }), 2000)
+    expect(rescued).toMatchObject({ state: 'done', idleInferred: true })
+    const genuine = reduceEntry(undefined, ev({ state: 'done' }), 1000)
+    expect(genuine.idleInferred).toBeUndefined()
+  })
+
+  it('a later genuine turn-end CLEARS the marker — it describes the current state only', () => {
+    const working = reduceEntry(undefined, ev({ state: 'working' }), 1000)
+    const rescued = reduceEntry(working, ev({ state: 'done', idle: true }), 2000)
+    expect(rescued.idleInferred).toBe(true)
+    const nextTurn = reduceEntry(rescued, ev({ state: 'working', newTurn: true }), 3000)
+    expect(nextTurn.idleInferred).toBeUndefined()
+    const ended = reduceEntry(nextTurn, ev({ state: 'done' }), 4000)
+    expect(ended.idleInferred).toBeUndefined()
+  })
+
+  it('an idle event that commits NOTHING leaves the marker alone', () => {
+    // The rescue may only move a node that is still `working`; on a blocked node it returns early
+    // and must not stamp a state it did not commit.
+    const blocked = reduceEntry(undefined, ev({ state: 'blocked' }), 1000)
+    const next = reduceEntry(blocked, ev({ state: 'done', idle: true }), 2000)
+    expect(next.state).toBe('blocked')
+    expect(next.idleInferred).toBeUndefined()
+  })
 })
 
 
--- a/src/core/agent-status-mirror.ts
+++ b/src/core/agent-status-mirror.ts
@@ -95,6 +95,22 @@ export interface MirrorEntry {
    * safest. Cleared by the first live event.
    */
   restored?: true
+  /**
+   * The CURRENT `done` was inferred from the CLI going idle at its prompt (Claude's `idle_prompt`
+   * notification, `normalize.ts` `idle`), not from a turn-end hook.
+   *
+   * `normalize.ts` already calls that a RESCUE signal, and `reduceEntry` already honours it as one
+   * — it may only move a node that is still `working`. What was missing is that the resulting entry
+   * did not remember WHICH kind of `done` it holds, so a consumer downstream could not tell a turn
+   * that ended from a CLI that merely went quiet. Gate 2 of agent messaging is the consumer that
+   * must: a node blocked on an approval is also "idle at its prompt", and delivering into it is
+   * delivering into a modal. `Canvas.tsx` already discards the idle rescue for an `undefined` node
+   * for the same reason; messaging must not be the one consumer that trusts it.
+   *
+   * Written on the SAME edge as `state`/`stateVerified` (inside `commitState`), so it can never
+   * describe a state it did not arrive with.
+   */
+  idleInferred?: true
 }
 
 /** This host's Server-Edition install metadata (spec: server-update). Written by the installer
@@ -365,6 +381,11 @@ export function reduceEntry(
     next.stateVerified = proof
     if (proof) next.verifiedAt = now
     next.clientRevision = ev.clientRevision
+    // Which KIND of `done` this is, recorded on the same edge as the state itself. `idle` is only
+    // ever meaningful on a `done`; assigning (not merging) is the point — a later, genuine turn-end
+    // `done` must clear the marker, or a node would stay tainted for the rest of its session.
+    if (state === 'done' && ev.idle) next.idleInferred = true
+    else delete next.idleInferred
     // The entry's STATE now comes from this run, so it is no longer the one restored off disk.
     // Cleared HERE and only here: an event that commits no state — a context/usage event, a
     // held-off late `working`, the idle rescue — leaves a restored `done` exactly as restored as
--- a/src/core/agents/agent-message-decide.test.ts
+++ b/src/core/agents/agent-message-decide.test.ts
@@ -1,6 +1,8 @@
 import { describe, it, expect } from 'vitest'
 import {
   decideDelivery,
+  decidePreProbe,
+  FIRST_PAID_DECISION,
   type AgentMessageOutcome,
   type Proceed,
   DECISION_ORDER,
@@ -230,12 +232,12 @@ describe('the decision ORDER is load-bearing', () => {
       (p) => ({ ...p, notPermitted: undefined }),
       (p) => ({ ...p, retryAfterMs: undefined }),
       (p) => ({ ...p, targetLive: true }),
-      (p) => ({ ...p, pane: 'agent' }),
       (p) => ({ ...p, target: { ...p.target!, clientRevision: MANAGED_SCRIPT_REVISION } }),
       (p) => ({ ...p, tokenFilePresent: true }),
       (p) => ({ ...p, target: { ...p.target!, stateVerified: true, state: undefined } }),
       (p) => ({ ...p, target: { ...p.target!, state: 'working' } }),
       (p) => ({ ...p, target: { ...p.target!, state: 'done' } }),
+      (p) => ({ ...p, pane: 'agent' }),
       (p) => ({ ...p, pasteAware: true })
     ]
     const walked: string[] = []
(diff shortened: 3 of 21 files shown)


eneskirca:
## Review fixes pushed — `03483ea4`

All four, plus the three named siblings. Full suite: **6073 passed**, 4 failed (the 3 `node-pty-patch` + 1 `webgl-addon-pair` environmental ones). `npm run typecheck` clean. **37 mutations, 37/37 killed.**

### FIX 1 — the lost-receipt race
`watchForReceipt` opens the subscription **before** `sendFramed` and buffers the first qualifying advance; `wait()` returns it immediately if it already arrived. The **deadline still starts at `wait()`**, after the write and the post-write probe — deliberately, so the 8 s stays a budget for the target's response and does not mean something different on SSH than locally.

Two tests, both at the `deliverAgentMessage` level: a target that emits `newTurn` from inside `sendFramed` while the post-write probe is still in flight is now `delivered` (`M26`, reverting to subscribe-after-write, is red); and an advance that arrives during the **pre**-flight probe is *not* counted — that is the target's previous turn ending, not our receipt. Plus: the watch is cancelled on every exit, so a refused or failed write leaves no listener.

### FIX 2 — `samePane`
`PANE_OWNER_FMT` now carries `#{pane_id}`, appended (a three-field read still parses, so an older format string or a tmux that expands it to nothing degrades rather than breaks), and validated to `%\d+` — a *garbage* id that compared equal to another garbage id would read as "confirmed", which is the dangerous direction. `paneOwnerFrom` carries `pids` for every foreground row (`ps` already printed the column), and the new `agentPidIn` names the pid of the row that made the pane an agent pane. `samePane` requires all four: pane id, tty, pane pid, and the agent's own pid — with **every unknown treated as "no"**.

The exploit has its own test: same wrapper, same name, same tty, same pane pid, same pane id, only the agent pid moved ⇒ `deliveredToReplacedTarget`. And each field is isolated: `M28` (drop panePid), `M29` (drop paneId), `M30` (drop the agent pid), `M34` (stop requesting `pane_id`), `M36` (carry a garbage id), `M37` (`agentPidIn` answers for a non-agent pane) — **all red**. `M28` is the one that survived your run; it dies now.

The real-tty suite proves the plumbing end to end: those deliveries still reach `stalled`, which is only reachable if `samePane` held, i.e. real tmux answered `#{pane_id}` and real `ps` gave a pid column that lined up with argv — and the test now asserts both explicitly.

### FIX 3 — self-send
Two lines at the top of `decidePreProbe`, ahead of everything but an explicit `notPermitted`. Refused with no round-trip, whatever the target status says, and `M31` is red.

### FIX 4 — the two false claims, made true
- **Free gates.** `identityRefusal` and `idleRefusal` moved into `decidePreProbe`. `DECISION_ORDER` is now `notPermitted → rateLimited → targetGone → targetHookScriptStale → targetStatusUnverified → targetStatusStale → targetNotIdleUnknown → targetBusy → targetNotAgentPane → targetNotPasteAware`, with `FIRST_PAID_DECISION = 'targetNotAgentPane'` marking the boundary. Two tests assert it by **running** the decider: nine refusals decided with the pane fact withheld entirely, and a delivery-level test that a busy / unverified / stale / restored / never-posted target costs **zero** `paneOwner` calls. Accepted consequence, stated in the docblock: a target that is both busy and on a stranger's pane now reports `targetBusy` — retryable, so the caller comes back and learns the rest.
- **The trace headline.** Every refusal is traced, including the pre-write ones and the `targetGone` from a `sendFramed` that failed after a partial write. Amplification is bounded on both legs by design — the ring evicts, the board log rotates. Refusal *outcomes* deliberately gain no `traceId`: the trace is a record for the human, not a handle for the sender.
- **Siblings.** `read()` now spans the rotation boundary (previous generation consulted only when the current file has not satisfied the request, so the common case is still one read; cap re-applied to the join). `M33` red.

### One thing I removed rather than fixed
`paneOwnerFrom` briefly guarded "attach `pids` only if they line up with `argv`". That mutation **survived** — and correctly: both arrays are the same `parsePsRows` output filtered by the same pgid, so a mismatch is not expressible. It was a line no test could reach, which this very file already argues against ("a second guard for the second phrasing would be a line no test can reach"). Deleted, and replaced with an assertion on real `ps` output that each pid is paired with *its own* argv row and that the pane shell appears in neither. `M35` now mutates the pgid filter itself, and is red.

### Found while fixing — a literal-copy guard, the pattern you flagged
`control-master.injection.test.ts` asserted the SSH leg's format by comparing against a **hardcoded copy** of `PANE_OWNER_FMT`. It reddened the moment the local constant gained a field — the exact "guard that trains people to edit the guard" shape. It now compares against the imported constant (which is what "verbatim" means here) and additionally asserts the format contains shell metacharacters, so the sibling `unquotedMeta` check is provably non-vacuous.

### Carried into PR 5, in the code
- **F6** is in a docblock at the refusal site: `#{bracket_paste_flag}` first shipped in tmux **3.7**; Ubuntu 24.04 → 3.4, 22.04 → 3.2a, Debian 12 → 3.3a, Debian 13 → 3.5a, Ubuntu 26.04 → 3.6a; plus every SSH target (the remote's tmux decides) and the Server Edition; and the bundled 3.7b is `extraResources` under `"mac"` only with `bundledTmuxPath` deliberately last. It names what PR 5 owes: the three-way probe, a distinct outcome carrying `tmux -V`, and an explicit ship/no-ship decision for Linux and SSH. It also says why `targetNotPasteAware` cannot carry it — it says the pane did not ask, when the truth is we could not ask.
- **F7** is in the envelope docblock as the convention: **only the outermost frame is authentic; everything between the first opening line and the last closing line is data.** I did **not** add `body-lines: <n>` — the plan pins the envelope shape exactly, and the outermost-pair rule is decidable without the nonce, which is what makes it the right thing to ship. Say the word if you want the header anyway.

### Disagreement: none on the four
Every one reproduced. The only place I went a different way is the `pids` guard above, and that was your standard applied to my own code.

PR #261, opened 2026-08-16
feat(canvas): `sticky` verb — agents write into notes, notes render markdown

### LINKED ISSUE #144: Let agents write into a sticky note, so external sources (Linear, etc.) can live on the canvas
## Request

> "I would love to be able to pull external sources into this — even if it's just a sticky note. Not sure what would be the easiest, but maybe consider shipping this with an MCP and allow Claude to update/sync a sticky note or markdown element. I would use this to pull ticket updates from Linear into this."

So: a canvas element whose content is written by an agent (or a sync job) rather than typed by hand — a live Linear/Jira/GitHub view living next to the sessions working on it.

## Why this is close to free

The write path already exists. Agents control the canvas today through the **manage-nodeterm-canvas** shim: a POSIX `sh + curl` script that POSTs form-urlencoded to the hook server's `/control/<verb>` routes, env-gated on `NODETERM_CANVAS_CONTROL`, installed locally and on SSH hosts. It can already create nodes. What is missing is a verb to **write into an existing sticky**, and a sticky that renders markdown.

Rough shape:

- `nodeterm sticky --node <id> --text <markdown>` (replace) and/or `--append`, going through the same control routes and the same write-approval path as the other verbs.
- Sticky notes render markdown (the renderer already has `lib/markdown.ts` — `marked` + DOMPurify — used by the terminal and editor markdown views).
- A stable way for the agent to address the note: either the node id it was told, or a `--title`/`--tag` lookup, so a cron/loop session can keep updating "Linear: my tickets" without being handed the id every run.

With that, "sync Linear into a sticky" is a scheduled agent turn using the user's own Linear MCP — we do not ship or maintain a Linear integration, which is the right trade.

## Open questions

- Is an MCP server the right entrance in addition to the sh shim? The shim reaches every agent (claude/codex/gemini/opencode/grok) with zero per-agent work, and the MCP bridge was deliberately retired once for Context Link. Probably: keep the shim as the mechanism, and consider MCP only if non-nodeterm clients need it.
- Refresh cadence and staleness: a synced note should show when it was last written.
- Auth/scope stays entirely with the user's own MCP servers; nodeterm stores no third-party tokens.

## Surfaces

Desktop + Server Edition (the control routes and stickies exist on both). The kanban board should get it too — a synced note is exactly the kind of thing the card modal should show. Mobile: read-only at best, follow-up in the iOS repo.

Requested by a user over email.

eneskirca:
Closes #144 — "Let agents write into a sticky note, so external sources (Linear, etc.) can live on the canvas".

## What

A new canvas-control verb, reachable from every control-capable agent (claude/codex/gemini/opencode/grok) through the existing sh shim, locally and on SSH hosts:

```sh
nodeterm sticky --node <id|title> (--text "markdown" | --append "markdown") [--create]
```

- **Addressing**: `--node` matches a node id first, then a note's header title (case-insensitive). An ambiguous title is refused with the candidate ids; `--create` turns exactly the not-found case into a new note titled after the ref — never a typo'd id. This is what lets a cron/loop session keep updating `"Linear: my tickets"` without being handed an id every run.
- **Markdown**: sticky bodies now render as sanitized markdown (the existing `marked` + DOMPurify lib, lazy-loaded so it stays out of the boot bundle, `breaks: true` so pre-markdown plain notes keep their line structure) — on the canvas node **and** on the kanban card modal. Click flips back into the familiar textarea; blur/Esc renders again.
- **Staleness**: an agent write stamps the note — `↻ <agent title> · 5m ago` in the footer (persisted `textUpdatedAt`/`textUpdatedBy`). Any hand edit clears the stamp, so it only ever vouches for text an agent actually wrote.
- **Size**: body capped at 100 KB UTF-8, a spoken refusal well under canvas-mutations' 256 KB ceiling where an oversized note would silently stop syncing to team peers.

## Decisions worth reviewing

- **Not confirm-gated** (unlike `write`/`close`): nothing here reaches a PTY — the text lands in node data and is DOMPurify-sanitized on render — and a Linear-sync loop that raises a dialog every run is a sync loop the user turns off. The stamp is the accountability surface instead. `sticky` is therefore *not* added to `DESTRUCTIVE_VERBS`, and `control-destructive.test.ts` stays green.
- **Title = header title** (`data.title`, what `list`/`rename` use), not the kanban first-line label — the label changes with every rewrite of the body, which is exactly what a synced note does, so it can't be a stable address.
- **No MCP server**: as the issue leans, the shim reaches every agent with zero per-agent work; the user's own Linear/Jira MCP supplies the data on their side.

Server Edition / relay inherit the existing control-verb absence unchanged (`setControlHandler` is desktop-only).

## Mechanics

- `src/renderer/lib/stickyWrite.ts` — pure resolver (`resolveStickyRef`, tri-state like `resolveColumnRef`) + body composer (`applyStickyWrite`) with its own unit suite.
- `canvas-control-core.ts` — verb model, `parseControlRequest` rules (presence-not-truthiness: `--text=""` clears a note), shim bare-positional table, and both agent-facing doc templates (skill + codex/gemini/opencode block).
- `Canvas.tsx` — `case 'sticky'` next to `rename`; `--create` places the note below the source via `placeBelow` + `addAndConnect`; titles pass through `oneLine` at the door like `rename`'s.
- `NoteMarkdown.tsx` — shared rendered view (StickyNode + CardModal); anchor clicks are already routed to `shell.openExternal` by main's `will-navigate` guard and don't flip edit mode.
- CardModal's capture-phase Esc handler learned the note edit the same way it knows the title edit (Esc exits the edit, not the modal).

## Tests

`npm test`: 495 files, 6776 passed. New coverage: resolver/composer unit suite; `parseControlRequest` sticky rules; shim end-to-end through a real `/bin/sh` + hook server (bare positional, markdown body with backticks/`#`/newlines/`$` verbatim, bare `--create` arriving as an empty-valued arg); both skill texts document the verb.


eneskirca:
Self-review (multi-agent, 8 finder angles → 16 deduped candidates → 6 verifiers, shim findings verified by executing the extracted script under `/bin/sh`) surfaced 10 confirmed issues; all are fixed in 6b727b3f:

1. **Silent note wipe** — a body starting with `--` (e.g. a markdown `---` rule) arrived as empty `--text` + a junk flag and cleared the note with `ok`. `parseStickyArgs` now refuses unknown flags with a pointer at the `--text=<value>` form; a shim e2e pins the wire shape.
2. **Camera hijack** — `sticky` from a non-active project traveled the view there on every sync run. Now store-answered (like `send`/`reply`, same G5 reasoning): the write lands in the owning project's serialized nodes via `applyNodeMutation` + `writeDisk`.
3. **Duplicate-note lockout** — title matching normalizes both sides through `oneLine`, so a ref with an interior control char re-matches the note its own `--create` minted.
4. **`--create` is now `--create yes`** — every flag carries a value (the repo rule that keeps the stale pre-peek shim on unreconnected SSH hosts parsing identically), and a JSON caller's `{create: false}` no longer creates.
5. **Forged byline** — `sticky` joins `requiresVerified` (per-node token only; own one-sentence refusal `Sticky write refused.`). The stamp replaces the confirm dialog, so it must not be forgeable by a bearer-holder naming someone else's node id. New verb ⇒ no legacy population stranded.
6. **Lost concurrent append** — the write re-applies inside the `setNodes` updater against the freshest text; chained appends compose instead of the second overwriting the first.
7. **Frozen stamp** — `relativeTime` (day unit, future-skew clamp) + a minute tick; the card modal shows the stamp too.
8. **Copy-from-note impossible** — ending a drag-selection no longer flips into the textarea (selection guard, both surfaces); Enter opens the editor for keyboard users.
9. **Surface disagreement** — the modal renders with the note's own `.sticky-node__md` typography; the board card label strips leading `#` markers.
10. **Cap unit mismatch** — the 100k cap measures JSON-serialized length, the same unit as canvas-mutations' 256k ceiling it exists to stay under, so escape-heavy bodies can't slip past into the silent peer-desync zone.

Full suite: 6785 passed.

Issue #78, opened 2026-08-02
Roadmap: 4 bug fixes + 4 features I might work on + ideas


parsa222:
Hey! After #74 I want to keep contributing, I really like this project. Here is what I plan to
do. Filing the list first so you can tell me to skip or reshape anything before I start. I have
dug into the code for each one, so the bugs below come with the potential causes I found.

## Bugs

- [x] fix(window): closing the app in macOS fullscreen leaves a black screen. This looks like
  the known Electron behavior (electron/electron#20263): the close handler hides the window
  without leaving fullscreen first, so the empty Space seems to stay behind. Likely fix: leave
  fullscreen, wait for the transition, then hide; same guard on the quit path.
- [x] fix(terminal): canvas terminals do not match the user's real terminal colors (zsh). Three probable causes: the xterm theme is hardcoded to just background/foreground with
  no ANSI palette and no setting, the app's tmux conf unsets terminal-overrides so truecolor
  gets clamped to 256 colors, and COLORTERM is never set in the PTY env. Plan: verify with a
  truecolor probe, then truecolor passthrough + COLORTERM first, and a terminal theme setting
  next to the existing font settings after that.
- [x] fix(usage): usage percentages are inconsistent. From what I can tell, the usagePercentMode
  setting ("32% used" vs "68% left") only reaches the bottom-left pill; ContextMeter, SessionRow
  and the notch HUD print bare unlabeled percentages with their own duplicated color thresholds,
  and the pill's bar always fills as remaining even when the number says used. Plan: route
  everything through the existing percentText/percentNumber helpers and one shared threshold
  function.
- [x] fix(notch): the notch capsule covers menu bar status items on crowded menu bars. It looks
  like the capsule pads itself symmetrically (notchWidth + 2x content width), so every
  running agent icon widens it over the status-item side, and hovering it also swallows their
  clicks (the window flips off click-through). Possible fix: grow left-only or clamp the
  overhang; optionally read the real notch width natively instead of asking the user, or have a a separate editor for it .

## Features

- [x] feat(canvas): user-facing "spawn a team". The spawn-team control section already builds the
  whole thing for agents (grouped grid of up to 8 agent nodes, context-linked back to the
  conductor). This adds an entry point (Dock/command palette): type a task, get the
  role split, call the same code path, with optional worktree per member.
- [x] feat(canvas): focus mode. One terminal/agent node fills the window and everything else
[...]

PR #265, opened 2026-08-16
feat(settings): live debug log panel — redacted in-memory ring behind a toggle

eneskirca:
The #78 log-panel item, in the agreed shape: ring buffer in main, batched IPC only while the panel is open, filter chips + auto-scroll viewer, secrets redacted at the ring-buffer boundary, manual export only (no crash-report sending — telemetry consent is its own topic).

## Core (desktop AND Server Edition — registered via CorePlatform)
- `log-redact`: deny-by-name redaction seeded from `AUTH_ENV_STRIP` + every `NODETERM_*` token name + `GH_TOKEN`/`GROK_CLI_AUTH_HEADER` + Authorization/cookie headers + well-known token shapes (`sk-…`, `ghp_…`, `github_pat_…`). Deliberately NOT an entropy matcher: SHAs, node ids and paths are the debugging payload. Heavily unit-tested.
- `log-buffer`: 2000-entry ring; redaction happens INSIDE `push()` so nothing unredacted ever exists in the ring — pairing-core's "NEVER sent to the renderer" invariant enforced at the source, not the IPC boundary.
- `log-sink`: console wrap (forwards to the original, mirrors into the ring, parses the existing `[subsystem]` convention) + `uncaughtExceptionMonitor` — the monitor hook on purpose: a plain `uncaughtException` listener would change crash semantics, and main currently has NO handler at all.
- `log-handlers`: ref-counted subscribe → 200ms-debounced `logBatch` broadcasts, gated on the setting; snapshot/clear round-trips. Client dedupes by monotonic `seq`.

## Shell
- **main**: sink installed at module load (boot warnings land); renderer consoles mirrored via `console-message` (packaged apps swallow those too, and React error boundaries report through `console.error`); handlers registered next to `registerBoardLogHandlers`.
- **server**: same registrar — headless is where a swallowed console hurts most; the browser panel reads the server process's ring over the ws bridge (real implementation, not a stub).
- **UI**: Settings → Application → Debug (`debugLogPanel`, default off) + Open button; palette entry "Show debug log" while enabled; the viewer (level chips, text filter, stick-to-bottom autoscroll, Copy/Clear) is code-split via `lazyPanels`. Kanban: not node-scoped, so the board deliberately gets nothing.

Refs #78

### DIFF
--- a/src/core/log-buffer.test.ts
+++ b/src/core/log-buffer.test.ts
@@ -1,0 +1,61 @@
+import { describe, it, expect } from 'vitest'
+import { LogBuffer, LOG_BUFFER_CAP, LOG_LINE_MAX } from './log-buffer'
+
+const rec = (msg: string) => ({ level: 'info' as const, tag: 't', msg })
+
+describe('LogBuffer', () => {
+  it('assigns monotonic seqs and snapshots in order', () => {
+    const b = new LogBuffer()
+    b.push(rec('a'))
+    b.push(rec('b'))
+    const snap = b.snapshot()
+    expect(snap.map((r) => r.msg)).toEqual(['a', 'b'])
+    expect(snap[1].seq).toBeGreaterThan(snap[0].seq)
+  })
+
+  it('drops oldest past the cap — the ring never grows unbounded', () => {
+    const b = new LogBuffer()
+    for (let i = 0; i < LOG_BUFFER_CAP + 10; i++) b.push(rec(`m${i}`))
+    const snap = b.snapshot()
+    expect(snap.length).toBe(LOG_BUFFER_CAP)
+    expect(snap[0].msg).toBe('m10')
+    // seq keeps counting across drops — since() stays correct after eviction.
+    expect(b.lastSeq()).toBe(LOG_BUFFER_CAP + 10)
+  })
+
+  it('since(seq) returns only newer records', () => {
+    const b = new LogBuffer()
+    b.push(rec('a'))
+    const mark = b.lastSeq()
+    b.push(rec('b'))
+    b.push(rec('c'))
+    expect(b.since(mark).map((r) => r.msg)).toEqual(['b', 'c'])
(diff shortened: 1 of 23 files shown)


eneskirca:
Not a review of the panel itself (it doesn't touch the render path) — one note from the terminal-rendering side, because this panel is about to become where those failures get diagnosed.

The renderer emits a small set of deliberate field traces, and they are the fastest way to tell three otherwise identical "the terminal looks wrong" reports apart:

- `[nodeterm] healed a broken webgl swap (<context>)` — a renderer swap died midway and left a stray canvas; `<context>` says which strand fired (`acquire-failed`, `release`, `visible`, `context-restored`).
- `[nodeterm] webgl dispose threw mid-release` and `[nodeterm] healed a renderer-less webgl release` — the signature of an addon/core version skew; this is what the black-terminal bug looked like from the outside before it was root-caused.
- `[glyphgrid] …` warnings — per-node attach/geometry failures that leave a terminal on its own renderer.

If the ring captures `console.warn` from the renderer process, these come along for free and a user report can be triaged from the panel instead of asking for a devtools console. If it only captures the main process, they won't — and that would be worth closing, since the failures that produce them are renderer-side by definition. Either way: the redaction pass is what matters for these (they carry node ids and a context string, no paths or tokens), so nothing here should be sensitive.

eneskirca:
Both halves of the note are now true by construction: the ring **was** already capturing renderer consoles (the `console-message` mirror is registered for every webContents, main window included), and as of `a81be934` the mirror splits the renderer's own `[tag]` prefix exactly like the main-process sink — so `[nodeterm]` heal strands and `[glyphgrid]` attach warnings land under their own tag chips instead of a blanket `renderer`, and a report can be triaged from the panel's tag filter directly. Untagged lines keep the `renderer` fallback. Agreed on redaction: those traces carry node ids and context strands, nothing the deny-list needs to touch.

PR #266, opened 2026-08-16
feat(ui): polish pass — enter animations for every shared shell + quiet scrollbars

eneskirca:
The #78 polish item, in the reshaped scope agreed there: **no component-library adoption** — the hand-written macOS-native token CSS stays the identity. Two purely-CSS changes, zero TSX churn, all themed off the existing tokens:

- **Enter animations** (transform/opacity only, one-shot on mount): scrims fade 120ms; dialogs (`.confirm`/`.palette`/`.shortcuts`) pop with the kanban modal's spring curve; context/tab menus settle in 120ms; drawers slide from their anchored edge; tooltips fade. One rule per shared shell class covers every dialog and menu in the app, and `prefers-reduced-motion` turns all of it off (the mascot-pulse contract: motion is decoration). No exit animations on purpose — every shell hard-unmounts via `{open && portal}`, and threading a closing state through 15 dialogs isn't worth a 100ms tween; the kanban modal set this precedent.
- **Quiet scrollbars** via the STANDARD properties (`scrollbar-color` on `:root` — it inherits; `scrollbar-width: thin` on `*` — it doesn't), not `::-webkit-scrollbar`: macOS keeps its overlay behavior (no reserved gutter), the tint token themes both light and dark, the deliberately-hidden bars (tab strip, xterm viewports) keep working through the `scrollbar-width: none` they already declare, and the banner/menu hand-rolled webkit thumbs converge on the same thin themed look.

Theme guard (`styles.theme.test.ts`) green. Terminals keep streaming: nothing here animates layout, and everything is mount-time one-shot.

Refs #78

### DIFF
--- a/src/renderer/styles.css
+++ b/src/renderer/styles.css
@@ -10619,3 +10619,108 @@ body,
   text-overflow: ellipsis;
   white-space: nowrap;
 }
+
+/* ================================ Polish pass (issue #78) ==================================
+   Enter animations for the shared overlay shells + quiet scrollbars. Rules of the pass:
+   transform/opacity ONLY (nothing that forces layout — terminals must keep streaming at 60fps),
+   one-shot on mount (no exit mechanism: every shell hard-unmounts via `{open && portal}`, and
+   keeping portals alive for an exit tween would thread a closing state through 15 dialogs — the
+   kanban modal set this precedent and it reads fine), and the existing token palette only. */
+
+@keyframes nt-overlay-fade {
+  from {
+    opacity: 0;
+  }
+  to {
+    opacity: 1;
+  }
+}
+@keyframes nt-surface-pop {
+  from {
+    opacity: 0;
+    transform: scale(0.97) translateY(4px);
+  }
+  to {
+    opacity: 1;
+    transform: none;
+  }
+}
+@keyframes nt-menu-in {
+  from {
+    opacity: 0;
+    transform: translateY(-3px);
+  }
+  to {
+    opacity: 1;
+    transform: none;
+  }
+}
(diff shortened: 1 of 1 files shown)


eneskirca:
Checked this from the terminal-rendering side, since animating shells is one of the ways terminals historically got hurt (a transform on an ancestor makes the shared glyph layer's measured origins stale, and layout-triggering properties compete with a streaming pty for the main thread).

No objection — the pass stays clear of both hazards:

- the animated selectors are overlays, menus, drawers and tooltips only (`.palette-overlay`, `.drawer-overlay`, `.shortcuts`, `.palette`, `.ctx-menu`, `.tab-menu`, `.drawer`, `.tooltip`); nothing matches `.term-node`, `.xterm`, or the kanban card modal's terminal (`kanban-modal__term`), so no terminal is inside an animating box;
- transform/opacity only, as the header claims, so nothing forces layout while a pty streams.

One thing worth keeping in the header as a rule for the next pass: if an animation ever lands on a shell that *contains* a terminal, the constraint is stronger than "no layout" — an animating `transform` on an ancestor of a glyph-attached terminal desynchronizes the shared canvas (grid origins are measured from the DOM rect, the canvas draws from the camera), so such a shell has to be excluded from the shared renderer for the duration, the way `dragging` already is.

eneskirca:
The header now carries the rule verbatim (pushed to main as `3e12c2e2`, since this PR had already merged): never animate a shell that contains a terminal — an ancestor `transform` desynchronizes the shared canvas, so such a shell must leave the shared renderer for the duration, the way `dragging` already does. It sits at the top of the polish section where the next pass will read it.

PR #273, opened 2026-08-16
fix(build): keep every local package off the production update feed

PaiCthulhu:
Split out of #111 at your request — this is the part with nothing Windows-specific in it.

## Problem

A locally packaged app is indistinguishable from a release at runtime — `app.isPackaged` is true for both — so any `dist`-produced build initialized electron-updater against the production feed, polled it for a version nobody published, and logged a 404 on `latest*.yml` at boot and every six hours after.

## Change

- `shouldEnableUpdater(isPackaged, updateMode)` gates the packaged path on a `nodeTermUpdates` marker, which `packagedUpdateMode()` reads back from the packaged `package.json`;
- `dist` and `dist:linux` inject that marker through electron-builder's `extraMetadata`;
- `--publish never` comes with it: electron-builder defaults to `onTagOrDraft`, which on a tagged checkout with a token in the environment would try to publish a local smoke build;
- `release` does not set the marker, so a promoted build wires up and updates itself exactly as before. The CI release workflow packages Linux through `electron-builder` directly and never goes through `dist:linux`, so it is unaffected either way.

The trade-off, stated plainly: a `dist` package can no longer smoke-test the updater wiring itself — a manual check now answers "up to date" without going near the network. The old behaviour at least proved the wiring was live, at the cost of a recurring 404 in every local build's log. Verifying against the real feed is a release package's job.

The guard test asserts the **scripts themselves**, not just the pure function: a `dist` script that forgets the marker is a build-config mistake no unit test of `shouldEnableUpdater` can catch, and the failure mode is silent. Same shape as `src/core/no-electron.test.ts`.

## Verification

Run on Windows with Node 22 (CI on `ubuntu-latest` is the real check — nothing here is platform-dependent):

- `npm run typecheck` passes;
- `src/shared/update-platform.test.ts` — 10/10;
- `npm run test:unit` — 49/49 across 9 files;
- `npm run build` passes;
- the guard test was checked by **mutation**: dropping the marker from `dist:linux` fails it with `["dist:linux"]`, and restoring it passes. A build-config guard that can pass vacuously guards nothing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,8 +23,8 @@
     "server:dev": "npm run build && npm run server:build && node out/server/main.cjs",
     "make-icon": "node scripts/make-icon.mjs",
     "make-social": "node scripts/make-social.mjs",
-    "dist": "npm run make-icon && electron-vite build && electron-builder --mac --arm64 -c.mac.identity=null -c.mac.notarize=false",
-    "dist:linux": "npm run make-icon && electron-vite build && electron-builder --linux",
+    "dist": "npm run make-icon && electron-vite build && electron-builder --mac --arm64 --publish never -c.mac.identity=null -c.mac.notarize=false -c.extraMetadata.nodeTermUpdates=disabled",
+    "dist:linux": "npm run make-icon && electron-vite build && electron-builder --linux --publish never -c.extraMetadata.nodeTermUpdates=disabled",
     "release": "npm run make-icon && electron-vite build && sh -c 'electron-builder --mac --arm64 --x64 --publish never; s=$?; npm run rebuild; exit $s'",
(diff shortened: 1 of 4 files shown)


eneskirca (approved):
Reviewed — approving. The marker mechanism is sound (`extraMetadata` → packaged package.json → `shouldEnableUpdater` gate), `release` correctly carries no marker, and I verified the claim about the release workflow: release.yml packages Linux via `npx electron-builder --linux --publish never` directly, never through `dist:linux`, so promoted builds are unaffected. The script-level guard test (mutation-checked per your notes) is exactly the right shape for a silent build-config failure mode, and the fail-open `packagedUpdateMode()` read preserves release behavior if the marker read ever breaks. CI green (typecheck + tests + build). Thanks!

PR #274, opened 2026-08-16
feat(quit): confirm before quitting the app

gor3a:
## Summary
- Adds a native confirm dialog before the app actually quits, cross-platform (mac/win/linux)
- Gates both quit paths: the native title-bar close on Windows/Linux (closes the window directly, no `app.quit()` first) and `app.quit()` itself (Cmd/Ctrl+Q, menu Quit, `window-all-closed`)
- macOS's traffic-light close (hide, not quit) is untouched
- Auto-update's restart-to-install skips the dialog, since that's not a user exit needing re-confirmation
- CI: run the quality job (typecheck/test/build) on `windows-latest` too, not just `ubuntu-latest`

## Test plan
- [x] `npm run typecheck`
- [x] `npx vitest run src/main/main-window.test.ts`
- [ ] CI green on ubuntu-latest + windows-latest
- [ ] Manual: Cmd+Q on mac shows dialog; Cancel keeps app open; Quit exits
- [ ] Manual: title-bar × on Windows/Linux shows dialog before window closes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,11 @@ permissions:
 
 jobs:
   quality:
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v7
 
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -361,6 +361,39 @@ let activeRemote: { cwd: string; ref: GitRemoteRef } | null = null
 // True from the first before-quit on: lets window close-events through (see hide-on-close).
 let quitting = false
 
+// Confirm-before-quit gate. Set once the user has answered "Quit" in the dialog below, or when
+// a quit is app-initiated rather than user-initiated (auto-update restart) and should not be
+// interrupted by a prompt for a decision already made. `pending` dedupes concurrent triggers
+// (shortcut + menu + window-close firing in the same tick) into a single dialog.
+let quitConfirmed = false
+let skipQuitConfirmation = false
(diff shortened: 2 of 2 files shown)


gor3a:
windows-latest CI is red, but not from this PR's change: \`managed-script.test.ts\` and \`opencode.test.ts\` \`spawn('sh', ...)\` / shell out to \`curl\` directly with no platform guard, and Windows runners have neither on PATH by default — hence the ENOENT / 5s timeouts. \`main-window.test.ts\` (the logic this PR actually adds) passes on both ubuntu-latest and windows-latest.

Pre-existing gap in the test suite, surfaced now because this PR is the first to add windows-latest to the CI matrix. Leaving as a known follow-up rather than broadening this PR's scope — the fix is \`describe.skipIf(process.platform === 'win32')\` (or similar) on the sh/curl-spawning suites in those two files.

eneskirca:
Reviewed. The quit-gate logic itself reads correctly (both paths covered: title-bar × via the close handler, Cmd/Ctrl+Q via before-quit; `quitConfirmed`/`pending` dedupe; auto-update restart skips the dialog) — but two things block approval as-is:

1. **The windows-latest CI matrix doesn't belong in this PR.** The failures are pre-existing Windows incompatibilities in the test suite (chmod 0600 assertions, POSIX path separators, etc.), not your quit code — making the whole suite Windows-clean is what the Windows-support extraction phases (#300+) are doing. Please drop the ci.yml change here so the PR can go green on its own merits.

2. **An always-on confirm dialog with no opt-out is a default-behavior change** — some users quit-and-relaunch frequently and will find a modal on every ⌘Q hostile. A `confirmBeforeQuit` setting (Behavior section) would make this an easy approve; whether it defaults on or off is the maintainer's call.

Happy to re-review once the CI change is split out.

PR #282, opened 2026-08-17
feat(menus): session-row ↔ canvas-node parity + Transfer-with-model nested submenu

dkattan:
## What

Two related parity fixes for the context menus:

1. **Full session-row ↔ canvas-node parity.** Right-clicking a session in the sidebar list now builds the **same** single-node menu as right-clicking the node on the canvas — Color, Group, Duplicate, Branch, Collapse, Markdown view, Refresh terminal, Restart agent, Restart agent and shell, Reopen session as, Switch model, and Transfer — instead of the old short list (Go to / Rename / Duplicate / Transfer / Close). The session-list-specific **Go to** and **Rename** stay on top; the canvas's destructive **Delete** is swapped for the non-destructive **Close** (hide the tab, keep the tmux session). Non-active projects keep the narrow set, since the shared rows read the active canvas's live nodes + per-node registered closures, which don't exist for a non-active project.

   The homogenize-menus work's stated goal was "full parity everywhere", but only **Transfer** had been ported to the session-row; the rest stayed canvas-only. This closes the gap by having `onRowContextMenu` reuse `selectionItems([id])` directly — **one builder, two surfaces**, so they can't drift again.

2. **Transfer becomes a nested submenu mirroring Switch model.** Each transfer target that is model-discovery-capable (`MODEL_SWITCH_CAPABLE`, base-resolved — so a claude-base custom agent qualifies) opens a sub-submenu of the agent's **default model** plus one row per discovered gateway model. Picking a model opens the new agent node with `--model <id>`. Non-capable targets (gemini, grok, baseless customs) stay flat rows; a relay source hides the model submenus (its gateway is on another machine); an unconfigured gateway degrades to flat rows (transfer-without-model is still useful).

## How

- **`createAgentNode`** (`state/workspace.ts`) gains a trailing `model?: string`, threaded into `assembleLaunchCommand` (`withAgentModel` appends `--model` for a capable agent, no-ops otherwise) and persisted as `data.agentModel` so cold-restore/restarts keep it.
- **`transferConversation`** threads `model?` through to the new node.
- **`transferConversationItems`** (`lib/transferItems.tsx`) becomes nested: `TransferConversationArgs` gains `gatewayModels` + `relaySession`; capability is resolved from the passed-in `customAgents` (not the runtime's injected resolver) so the builder stays pure and testable — mirroring `reopenVariants`.
- **`onRowContextMenu`** (`Canvas.tsx`) reuses `selectionItems([id])` for the active project (full parity), swapping the trailing Delete for Close; keeps the narrow set for non-active. Both surfaces pass the new Transfer args.

## Critical files

- `src/renderer/lib/transferItems.tsx` + `.test.tsx` — nested Transfer submenu; 13 unit tests (capable target → submenu of default + models; non-capable → flat; relay → flat; empty gateway → flat; model threaded to handler).
- `src/renderer/state/workspace.ts` + `.test.ts` — `createAgentNode` `model` param; 3 tests (stamps `agentModel` + `--model` for capable; omits when none; drops `--model` for non-capable).
- `src/renderer/canvas/Canvas.tsx` — `transferConversation` `model?`; `onRowContextMenu` parity reuse.

## Verification

- `npm run typecheck` ✅
- `npm test` — 174 passed across the touched areas (`transferItems`, `workspace`, `agent-restart`, `reopenVariants`). The only failures are pre-existing on `origin/main` and environmental (real-tmux version behavior, xterm dep-version drift, flaky fs.watch) — confirmed by stashing my changes and reproducing them on clean main.

🤖 Generated with [Claude Code](https://claude.ai/code)
via [Happy](https://happy.engineering)

Co-Authored-By: Claude <noreply@anthropic.com>
Co-Authored-By: Happy <yesreply@happy.engineering>

### DIFF
--- a/src/renderer/canvas/Canvas.tsx
+++ b/src/renderer/canvas/Canvas.tsx
@@ -4930,7 +4930,15 @@ export function Canvas() {
   // agent's native transcript to a handoff file (main) and open a target node that reads it
   // and continues. The source node stays. Mirrors branchClaude's placement.
   const transferConversation = useCallback(
-    async (sourceNodeId: string, targetAgentId: AgentId, at?: { x: number; y: number }) => {
+    async (
+      sourceNodeId: string,
+      targetAgentId: AgentId,
+      at?: { x: number; y: number },
+      /** A model chosen from the Transfer submenu's nested model list. Applied to the NEW node's
+       *  launch (`--model <value>`) when the target is MODEL_SWITCH_CAPABLE; silently dropped
+       *  otherwise (the target's own default model). Persisted as `data.agentModel`. */
+      model?: string
+    ) => {
       const source = nodesRef.current.find((n) => n.id === sourceNodeId) as CanvasNode | undefined
       if (!source) return
       const sourceAgentId = source.data.agentId
@@ -4988,7 +4996,9 @@ export function Canvas() {
         source.data.accountId,
         // The mode belongs to the node being OPENED, so it is gated on the TARGET agent — a
         // handoff into grok must not inherit claude's version gate.
-        activePermissionMode(targetAgentId)
+        activePermissionMode(targetAgentId),
+        // A model picked from the Transfer submenu (only offered for switch-capable targets).
+        model
       )
       node.selected = true
       const placed = placeSpawned(node, at ?? besideNode(source))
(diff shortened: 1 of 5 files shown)


eneskirca (approved):
Reviewed in depth (merged onto current main in a scratch worktree: clean merge, typecheck green, 89/89 on the touched suites) — approving. Verified no new IPC surface, nothing on argv (model reaches the launch line only through the existing validated+quoted `withAgentModel`; gateway creds stay on the env route), and no reverts of anything merged in #244 or later.

Non-blocking nits for a follow-up:
1. The reused `selectionItems([id])` leaves the "1 node" label header in the sidebar row menu ("Go to / Rename / 1 node / …") — filtering it like Delete would read cleaner.
2. Delete is removed by label string match (`it.label === 'Delete'`); a shared constant or id would keep the destructive row from silently reappearing if it's ever relabeled.
3. `transferItems` passes raw `gatewayModels` instead of `modelsForAgent(models, tg.id)` — identical today, but future per-agent filtering there would drift Transfer from Switch model.
4. `workspace.ts` stamps `data.agentModel` even for non-capable agents (documented by your own test) — harmless since every consumer re-gates, but inconsistent with the Switch-model path.

Thanks — nice follow-up to the menu homogenization.

PR #283, opened 2026-08-17
feat(accounts): per-account default node color

Senzaiken:
## What

A managed account — **Claude or Codex** — can carry a **default node color**. Set one in Settings → Accounts and every node opened under that account takes it instead of the agent's brand color, so a second login is recognizable on the canvas at a glance rather than being one more clay terminal.

*(The sections below are in the order they happened; Codex accounts arrived on `main` mid-review and are covered in the last section, which supersedes the Claude-only wording earlier on.)*

## Why this shape

**Resolved in `createAgentNode`.** Every agent-node creation path already funnels through it — context menu, command palette, dock, kanban column, canvas-control `open-agent`/`verify`/`spawn-team`, transfer, explain-commit — so the feature lands everywhere without touching a single call site. It reads off the same `boundAccountId` that stamps `data.accountId`, extracted to one local so the color and the binding cannot drift apart.

**Applied at creation, not derived per render.** The color is baked into `data.color` like any other node color. That means a hand-picked node color is never fought, and editing the account later repaints nothing. The trade-off, named explicitly: changing an account's color does **not** recolor existing nodes. Deriving it live would give that, but then a manual per-node recolor needs an extra override flag to survive — a worse deal for the common case.

**Nothing else moves.** An unset color, an account id that no longer resolves, and any non-Claude agent all produce a node byte-identical to before. `accountNodeColor` treats a whitespace-only color as unset, because an empty `data.color` renders a node transparent.

## Review catch, fixed in 9cee281e

`project-node-append.ts` honored `input.accountId` but minted the node with `agent?.color`. A session started **from the phone** under a colored account therefore landed on the desktop canvas in the agent's color — the account-at-a-glance distinction silently missing on one surface, while the adjacent comment claimed the node "looks exactly like one minted by `createAgentNode`".

The color is host-derived (the phone must not get to choose it), so `main` resolves it from `settingsStore` and passes it through `WorkspaceStore` into the pure append. `accountNodeColor` moved to `shared/agents/` so core, main and the renderer share **one** definition — a second copy of the rule is exactly the drift CLAUDE.md warns about.

## Second review catch, fixed in 47e81b62

`accountNodeColor` reached for `.color?.trim()`. `claudeAccounts` is never validated field-by-field on load (`mergeSettings` merges, it does not check), so a hand-edited `"color": 123` arrives verbatim and throws a `TypeError` — **inside `createAgentNode`**, meaning every new node under that account stops opening, with nothing in the failure pointing back at the file that was edited.

Guarded with `typeof color !== 'string'`, the same discipline `appendProjectNode` already applies to a phone-supplied account id one file over. A whitespace-only color stays unset as before. Test written first and watched fail on the real `TypeError: accounts.find(...)?.color?.trim is not a function`.

The other nit from that review — the claude-only gate living only at the renderer call site — is **#319**, which takes a different shape than proposed and explains why there.

## Three surfaces

- **Desktop** — full.
- **Server Edition** — works unchanged. Pure renderer + settings store, no new IPC, no bridge member.
- **Mobile** — the *picker* stays desktop/server only, but a session the phone registers now gets the account color via the fix above. Surfacing or editing the color on the phone would need the companion UI (`nodeterm-ios`) — follow-up there, not this PR. @eneskirca

## Testing

16 new tests, written before the code and each watched fail for the right reason: `accountNodeColor is not a function`, the node opening `#d97757` instead of `#0a84ff`, and the swatch group not existing.

- `workspace.account-color.test.ts` — precedence, blank color, stale id, non-Claude agent, account-less node.
- `AccountsSection.color.test.tsx` — swatch writes the color, ✕ clears it, `aria-pressed` reflects state, one account's color doesn't touch another's.
- `project-node-append.test.ts` — the phone path takes the host-resolved account color, and falls back to the agent color without one.
- `workspace.account-color.test.ts` — a non-string color in a hand-edited settings.json yields no color instead of throwing.

`npm run typecheck` clean. Full suite: **6978 pass**.

## Not verified

- **The phone path end-to-end.** It needs the iOS companion, which I can't run. The behavior is covered by unit tests on `appendProjectNode`, not on a device.
- **`local-send-keys.realtmux` and `tmux-paste.realtmux` fail in my environment** *(the first is a 103-character `sun_path` limit, root-caused and fixed in **#320**; the second is a low-rate flake, still open there)* — but they fail identically on a clean `main` (confirmed by stashing), and the second fails a *different* test run-to-run. They drive a real tmux server that can't spawn in this sandbox. No claim that they pass; CI is the judge.
- **Manual testing was against seeded accounts** in an isolated `NT_MULTI` instance, plus the real Settings → Accounts flow. Remote (SSH-host) accounts were not exercised — they carry `color` like any other account and take the same code path, but that is reasoning, not a measurement.

## Merged with `main`, and Codex accounts now carry a color too

`main` moved twice through this feature's own surface, so both conflicts were real.

**`createAgentNode` (35750011).** Managed accounts now bind to **Codex as well as Claude** (S6), and `agentLaunchOverride` gained a `projectId`. Both kept: `boundAccountId` is now `claude | codex` — the same single local the account color reads from, so the binding and the color still cannot drift — and main's binding rule is pinned by a test (`data.accountId` lands on a Codex node), because a silently reverted binding is not something a comment catches twice. The other conflict, `AccountsSection` imports, kept both sides.

**`CodexAccount.color` (f4f6db35).** The reviewer's design question — the color reaching only half the account-bound nodes — is answered by extending it rather than by declaring the feature Claude-only. A half-covered version is worse than either alternative: the panel would show swatches on one kind of managed account and not the other, with nothing on screen saying why. And a machine with two Codex logins is exactly the case the feature exists for.

- **Which list answers is ONE function**: `agentAccountColor(agentId, accountId, { claude, codex })` in `shared/agents/account-color.ts`, shared by `createAgentNode` and the phone-registered node path in `src/main`. The two lists are keyed **independently** — nothing stops the same id appearing in both — so reading a Codex node's color out of `claudeAccounts` would repaint it from an unrelated row. Pinned by a test where a Claude and a Codex account deliberately share `a1` and each node takes its own. Every other agent takes no managed account, so nothing answers for it; no nearest guess.
- `accountNodeColor` is now **structural** (`AccountColorSource`: `id` + `color`) — the two account records share nothing else, and a helper naming one of them could only be copied for the other.
- The swatches are **one component** (`AccountColorSwatches`) rendered by both row kinds, not a second copy of 30 lines of JSX. The Codex row's left column now stacks the way the Claude row already did.
- `color` survives the pending-account reconcile untouched: `applyResolvedCodexAccounts` spreads the existing row, so a late Codex login promotion cannot drop it.

### Test changes

- `AccountsSection.color.test.tsx` — the section now also mounts the Codex accounts panel, which hydrates `ssh.list()` and `codexAccounts.systemIdentity()` on mount. Unstubbed, that effect **threw and React unmounted the whole section**, so the failure read as "the color group does not exist" while the test was really reporting on a page that was gone. Both stubbed. Plus three Codex cases: a swatch writes `codexAccounts`, ✕ clears it, and coloring a Codex account leaves a same-id Claude account alone.
- `workspace.account-color.test.ts` — `agentAccountColor` per agent, and `createAgentNode` for a colored Codex account, an uncolored one, and the shared-id cross-list case. The old "accounts are Claude-only" case is gone; a gemini case pins that an agent taking no managed account is never colored and never bound.

### Verification after the merge

`npm run typecheck` clean. Full suite **8808 pass, 8 fail** — the same 8 that fail on a clean `origin/main` in this environment, verified by running exactly those three files in a detached worktree at `origin/main`: `session-host/hello-probe`, `main/remote-atomic-write`, `core/session-host-client`. None are touched by this branch. (The two `.realtmux` suites named under *Not verified* above are no longer a standing failure: #320 — the `sun_path` fix — is on `main`. `tmux-paste` still flakes at a low rate, which that PR already tracks.)

### Still not verified

- The Codex row's **new stacked layout** was not looked at in a running app. It mirrors the Claude row's existing structure and is covered only by the jsdom tests above.
- The phone path end-to-end, unchanged from the note above — it needs the iOS companion.


eneskirca (approved):
Reviewed in depth — approving. Storage/restart, precedence (explicit node color always wins), all creation paths funneling through createAgentNode, theme-token usage, and the phone-registration leg all check out; the 15 tests are behavioral (they fail under real mutations), and CI is green.

Two small latent nits, fine as a follow-up:
1. The claude-only gate lives only in the renderer call site (`boundAccountId = accountId && agentId === 'claude' ? …`), not in the shared `accountNodeColor` helper — if the phone path ever registers a non-claude node with an `accountId`, it would take the account color. Encoding the rule in the helper would make it truly one definition.
2. `accountNodeColor`'s `.color?.trim()` throws on a hand-edited non-string color in settings.json (e.g. `"color": 123`), which would break every new node creation for that account. A `typeof color === 'string'` guard closes it.

Thanks — nice, well-tested feature.

eneskirca:
Heads-up before this can merge — it went `CONFLICTING` while open, and one of the conflicts is **semantic, not textual**.

First, thanks for the follow-up commit: `typeof color !== 'string'` with a test and a CLAUDE.md line is exactly the nit I raised, done properly.

**The blocker:** `main` merged the Codex managed-accounts series (#334–#342), so managed accounts are no longer Claude-only:

```ts
// origin/main:src/renderer/state/workspace.ts:630
// Managed accounts bind to the builtin Claude and Codex agents (S6) …
...(accountId && (agentId === 'claude' || agentId === 'codex') ? { accountId } : {}),
```

Your branch still has the older `agentId === 'claude'` form in the same hunk, so resolving that conflict by taking your side would **silently revert Codex per-node account binding** — a Codex node would lose its account and spawn under the system `~/.codex`. Git will offer you both sides; please make sure the merged rule is the claude|codex one.

Conflicts to expect:
- `src/renderer/state/workspace.ts` — two hunks: the ternary above, plus `agentLaunchOverride(agentId)` → main's `agentLaunchOverride(agentId, projectId)` (`createAgentNode` gained a `projectId` parameter).
- `src/renderer/components/settings/sections/AccountsSection.tsx` — import collision with main's `useSystemCodexAccount`.
- `AccountsSection.color.test.tsx` also needs its mock extended for main's new dependencies (`codexAccounts.systemIdentity`, `sshServers.hydrate`/`ssh.list`) — it currently throws `Cannot read properties of undefined (reading 'systemIdentity')` on the merged tree.

One design question the rebase forces, worth answering in the body: **`ClaudeAccount.color` has no `CodexAccount` counterpart**, so on current main the feature reaches only half the account-bound nodes. Extending `color` to Codex accounts (or stating plainly that it's Claude-only for now) would close that.

Everything else from my review stands — the approval holds once the rebase keeps main's binding rule. #319 is stacked on this, so this one goes first.

Senzaiken:
Merged with `main` in 35750011, and the design question is answered by **extending the color to Codex accounts** in f4f6db35 rather than by declaring it Claude-only.

**On the semantic conflict — you were right to flag it, and it is the claude|codex rule that survived.** `boundAccountId` (the local this branch introduced so the binding and the color read the same value) is now:

```ts
const boundAccountId =
  accountId && (agentId === 'claude' || agentId === 'codex') ? accountId : undefined
...
...(boundAccountId ? { accountId: boundAccountId } : {}),
```

so a Codex node keeps its per-node account binding. There is a test on it (`workspace.account-color.test.ts`) asserting `data.accountId` lands on a Codex node, because "reverted silently" is exactly the failure mode you named and a comment does not catch it a second time. The other hunk took main's `agentLaunchOverride(agentId, projectId)`, and the `AccountsSection` import collision kept both sides.

**On `codexAccounts.systemIdentity` / `ssh.list`** — confirmed exactly as you described. Worth recording *why* it was more than a mock gap: the unstubbed effect **threw**, React unmounted the whole `AccountsSection`, and the failure surfaced as the color group "not existing" — the test was reporting on a section that was no longer on the page. Both stubbed.

**On `CodexAccount` having no `color`.** Extending it was small, and the half-covered version is worse than either alternative: the accounts panel would show swatches on one kind of managed account and not the other, with nothing on screen saying why. So:

- `CodexAccount.color` mirrors the Claude field.
- **Which list answers is one function** — `agentAccountColor(agentId, accountId, {claude, codex})` in `shared/agents/account-color.ts`, used by `createAgentNode` **and** the phone-registered node path in `src/main`. This matters more than it looks: the two lists are keyed independently, so nothing stops one id appearing in both, and reading a Codex node's color out of `claudeAccounts` would repaint it from an unrelated row. Pinned by a test where a Claude and a Codex account deliberately share `a1` and each node takes its own. Any other agent takes no managed account at all, so nothing answers for it — no nearest guess.
- `accountNodeColor` is now structural (`AccountColorSource`: `id` + `color`), because the two account records share nothing else and a helper naming one could only be copied for the other.
- The swatches are one component (`AccountColorSwatches`) rendered by both row kinds, not a second copy of the JSX.
- `color` survives the pending-account reconcile: `applyResolvedCodexAccounts` spreads the existing row, so a late Codex login promotion does not drop it.

**Verification.** `npm run typecheck` clean; full suite **8808 pass, 8 fail** — the same 8 that fail on a clean `origin/main` in this environment (`session-host/hello-probe`, `main/remote-atomic-write`, `core/session-host-client`), verified by running exactly those files in a detached worktree at `origin/main`. The `.realtmux` suites are no longer a standing failure now that #320 is on `main`; `tmux-paste` still flakes at a low rate, which that PR already tracks.

**Not verified, so you may want to eyeball it:** the Codex row's new stacked left column (label row, then swatches) was not looked at in a running app — it mirrors the Claude row's existing structure and is covered only by the jsdom tests. The phone path stays untested on a device for the same reason as before.

eneskirca:
Re-reviewed the rebase — you did this properly rather than mechanically, and the design gap I raised is closed. Nice work.

**The semantic conflict is resolved the right way, and I verified it by running main's own test on the trial merge rather than reading the diff:**

```
✓ accountId on Claude node factories > stamps accountId onto a Claude agent node
✓ accountId on Claude node factories > stamps accountId onto a Codex agent node (S6 per-node account picker)
✓ accountId on Claude node factories > does not stamp accountId onto a non-account agent node
✓ accountId on Claude node factories > omits accountId when none is given
```

Better than a resolution, actually: hoisting the rule into one `boundAccountId` local that feeds both the color read and the `data.accountId` stamp means the two facts can no longer drift apart — which was the whole worry.

The other three items are handled (`agentLaunchOverride(agentId, projectId)` intact, the `useSystemCodexAccount` import clean, the color test's `codexAccounts.systemIdentity` + `ssh.list` stubs present), and there is **no contact with #352** — your merge-base already contains it. Worth noting your `agentAccountColor` internalizes exactly that PR's lesson by dispatching on `agentId` rather than on an account id the two lists share, and you have a test naming it (*"reads the list that owns the agent when both hold the same account id"*). That's the trap #345 was.

And you extended `color` to `CodexAccount` instead of leaving it half a feature — one swatch component, one ownership rule, three tests on the hazard. That closes the design question.

My original approval re-verified on the new code: explicit per-node color still wins (baked once into `data.color`, nothing re-derives on load, and `reopenNode`'s `COSMETIC_KEYS` copies the snapshot's color), it survives restart through ordinary serialization, `styles.css` is untouched so the theme guard passes, and the `typeof color === 'string'` guard is still there with its test.

**One line owed before merge.** `AccountsSection.color.test.tsx`'s `window.nodeTerminal` mock omits `settings.save`. Every swatch click goes `useSettings.update()` → `scheduleSave()` → a 300 ms `SAVE_COALESCE_MS` timer that calls it. Run the file alone and the process exits before the timer fires, which is presumably how it was validated; in any longer run the timer fires and vitest reports an unhandled `TypeError: Cannot read properties of undefined (reading 'save')` and **exits 1**. Measured:

- merged tree, this file + `src/renderer/state`: `46 passed / 571 passed` but `Errors 1 error` → **EXIT=1**
- `origin/main` baseline: no `Errors` line at all — so it is PR-introduced, not this box's drift
- adding `settings: { save: async () => {} },` after line 50: no Errors, **EXIT=0**

CI happens to be green on the current shard, which is why it hasn't bitten yet — but it's timing-dependent, and this repo has enough CI flakes already. I'd rather not add another.

Two nits, take or leave: `workspace.ts:574-575` reads `settings.claudeAccounts`/`codexAccounts` raw while `src/main/index.ts:3276` guards the same two with `?? []` — a hand-edited `"claudeAccounts": null` survives `mergeSettings` and would throw on `.find` one level *above* your `typeof color` guard, i.e. the exact failure that guard exists to prevent (pre-existing pattern elsewhere, so only the asymmetry between your two new call sites is yours). And the swatch `aria-label={c}` is a bare hex where the Accent picker uses `Accent ${c}`.

Separate follow-up, not for this PR: `project-node-append.ts:128` still stamps `accountId` for *any* agentId on the phone path, while the canvas gates on claude|codex — your PR correctly gates the *color* there, which just makes the pre-existing looseness visible.

Ping me when `settings.save` is in and I'll merge it.

Senzaiken:
`settings.save` is in (36e6b330), and both nits with it. Also merged your `main` merge (c99c36fe) forward, so the branch is current.

**On `settings.save` — you were right, and I reproduced it before fixing rather than taking it on faith.** It did not fall out of any shard I ran, exactly as you predicted (`Test Files 48 passed / EXIT=0` on this file + `src/renderer/state`), so I forced it: a throwaway test that clicks a swatch and then awaits 400 ms — past `SAVE_COALESCE_MS` — inside the same file.

```
Unhandled Errors
TypeError: Cannot read properties of undefined (reading 'save')
Tests  9 passed (9)
Errors  1 error
```

With `settings: { save: async () => {} }` in the mock: no `Errors` line. The throwaway test is **not** kept — it would trade a timing-dependent failure for 400 ms of sleep on every run — but the stub carries a comment saying what it is standing in for, so the next person who adds a click to this file does not rediscover it. Your shard now: `602 passed`, no `Errors`, `EXIT=0`.

**Both nits taken, neither was cosmetic in the end:**

- **`?? []` on both lists.** You called the asymmetry between my two new call sites; the sharper version is that `settings.claudeAccounts.find` throws one level *above* the `typeof color` guard, so a hand-edited `"claudeAccounts": null` produces exactly the failure that guard exists to prevent — every new node under any account stops opening, with nothing pointing at the edited file. It now has its own test, watched fail on the real `TypeError: Cannot read properties of null (reading 'find')`.
- **`aria-label={`Node color ${c}`}`**, matching `Accent ${c}` in `AppearanceSection`. Test call sites name the full label rather than hiding the convention behind a helper, so the next rename breaks the test loudly.

**On the `project-node-append.ts:128` follow-up** — agreed, and agreed it is not this PR. Worth writing down while it is fresh: the phone path stamps `accountId` for any `agentId`, so a phone-registered node of some third agent can carry a binding the canvas would never have given it. My change gates the *color* through `agentAccountColor` (claude → `claudeAccounts`, codex → `codexAccounts`, everything else → nothing), which is why the looseness became visible — the two now disagree in the same file. The fix is to gate the stamp on the same claude|codex rule the canvas uses, and the interesting question is what to do with nodes already on disk carrying a stray binding. Happy to take it as a separate PR if you'd rather not.

Verification on the merged tip: `npm run typecheck` clean, full suite **9039 pass, 8 fail** — the same 8 that fail on a clean `main` in this environment (`session-host/hello-probe`, `main/remote-atomic-write`, `core/session-host-client`), all environment-bound, none touched here.

Still unverified, unchanged: the Codex row's stacked layout has not been looked at in a running app, and the phone path has not been exercised on a device.

Senzaiken:
Ready to merge. `settings.save` is in (36e6b330) with both nits, the branch is level with `main` (0 behind), and all four checks are green — CodeQL, Dependency review, quality, quality-windows.

Local tip: `npm run typecheck` clean, full suite **9055 pass / 8 fail** — the same 8 environment-bound failures that fail on a clean `main` here. Your shard (`AccountsSection.color.test.tsx` + `src/renderer/state`): 602 passed, no `Errors`, `EXIT=0`.

All yours whenever you want to land it, and #319 behind it.

eneskirca:
Re-verified `36e6b330` — the owed line and both nits are in, and I measured rather than read:

- **`settings.save` stub**: ran your shard on the branch tip (`AccountsSection.color.test.tsx` + `src/renderer/state`): 602 passed, **no `Errors` line, exit 0**. The comment on the stub saying what it stands in for is exactly right.
- **`?? []` on both lists**, with its own test asserting the node still opens (and takes the agent color) on a hand-edited `"claudeAccounts": null` — the failure lands one level above the `typeof color` guard no more.
- **`aria-label={\`Node color \${c}\`}`** matching the Accent convention, with the test call sites naming the full label.

`npm run typecheck` clean on the tip.

One logistics item, not your fault: `main` moved again since the 25th (the relay End-session work among others) and the branch has gone CONFLICTING once more — CLAUDE.md prose, plus the `registerNode` hunk in `src/main/index.ts` where `main` added `destroyNode` right beside your color-passing `appendRemoteNode` call. The resolution is keep-both (main's `destroyNode` + your call with the color argument); no semantic trap like last time. One more merge of `main` and I'll land it — sorry for the moving target, this one has been chasing `main` for a while.

Senzaiken:
Merged `main` again — d8c46e9a. `MERGEABLE`, checks running.

One conflict, keep-both exactly as you called it: `src/main/index.ts`, main's `destroyNode` beside this branch's color-passing `appendRemoteNode` call — no overlap, one registers a phone node and the other ends a session. `CLAUDE.md` also conflicted (prose): took main's rewritten `data.accountId` bullet with the two #419 resolver rules, kept the account-color bullet after it. #419 doesn't touch the color path — it changes which account id reaches `createAgentNode`, and the color is read off whatever `boundAccountId` ends up being.

`npm run typecheck` clean. Full suite **9327 pass / 11 fail**, byte-identical to a clean `origin/main` here — I ran the same five files in a detached worktree at `origin/main` and got the same 11. Beyond the usual three, `main` has since added `control-master.realsh` and `pane-cwd.realtmux`, which need a real ssh/tmux this sandbox can't give them.

No apology needed for the moving target — yours to land whenever.

eneskirca:
`d8c46e9a` resolves the conflict exactly as hoped — keep-both verified on the tip: main's `destroyNode` sits beside the color-passing `appendRemoteNode` call, all four checks are green on the merge commit, and the trial-merge story is over. Addressed, ready — thank you for chasing main through three rounds of it. (#319 queues up behind this one.)

PR #288, opened 2026-08-18
fix(kanban): preserve spaces when editing sticky notes in card modal

### LINKED ISSUE #285: Cannot type spaces in the note editor in Kanban view
Space key not working in the note editor (Kanban view)

Version: v0.3.2

Steps to reproduce
1. Manually create a note in the Kanban view
2. Open it for editing
3. Try to type a space

Expected: The space is inserted, as it is in the Canvas view.

Actual: Nothing happens; the space key is ignored in the editor.

Notes: The same editor works correctly in the Canvas view, so the issue seems specific to Kanban (likely the space key being captured as a shortcut there).

Image

Abhirup0:
### Summary

Fixes #285: Cannot type spaces in the note editor in Kanban view.

### Problem

When editing a sticky note in the Kanban view (`CardModal`), pressing the Space key did not insert a space (or trailing spaces immediately vanished). 

The root cause was in `toKanbanSession`: when mapping sticky nodes to `KanbanSession` cards for the board, `(n.data.text as string).trim()` was assigned directly to `session.text`. On every keystroke, `onEditSticky` updated the node with the typed text (e.g. `"Hello "`), but `toKanbanSession` trimmed it back to `"Hello"`. When React re-rendered `<textarea value={session.text}>`, the trailing space was stripped.

### Changes

1. **Preserve raw text in `toKanbanSession`** (`src/renderer/lib/kanbanSession.ts`):
   - Keep `rawText` as `session.text` so `<textarea>` in `CardModal` receives the exact characters typed (including spaces, newlines, and indentation).
   - Use `rawText.trim()` only when deriving `session.title` for the card header label.
2. **Guard `spacePan` against active Kanban view** (`src/renderer/canvas/Canvas.tsx`):
   - Added `isKanbanOpen(...)` guard to the capture-phase `spacePan` `onKeyDown` listener, matching the house rule that canvas shortcuts refuse when the board covers the canvas.
3. **Tests**:
   - Added `src/renderer/lib/kanbanSession.test.ts` to test whitespace preservation in note sessions.
   - Added `src/renderer/components/kanban/CardModal.test.tsx` to verify sticky note editing in `CardModal` preserves space characters and handles Esc cancellation.

### Verification

- `npm run typecheck` (tsc node & web)
- `npx vitest run src/renderer/components/kanban src/renderer/lib/kanbanSession.test.ts`
- `npm run build`

### DIFF
--- a/src/renderer/components/kanban/CardModal.test.tsx
+++ b/src/renderer/components/kanban/CardModal.test.tsx
@@ -1,0 +1,352 @@
+// @vitest-environment jsdom
+import { act } from 'react'
+import { createRoot } from 'react-dom/client'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import type { ProjectKanban } from '@shared/types'
+import { CardModal } from './CardModal'
+import type { KanbanSession } from './KanbanView'
+
+vi.mock('../../session/session', () => ({
+  useSession: () => ({
+    api: {
+      pty: { generateName: vi.fn(async () => ({ ok: true, message: 'AI Name' })) },
+      shell: { openExternal: vi.fn(async () => {}) }
+    }
+  })
+}))
+
+// Mock ModalTerminal and BrowserSurface to keep tests lightweight and focused on CardModal
+vi.mock('./ModalTerminal', () => ({
+  ModalTerminal: ({ nodeId }: { nodeId: string }) => (
+    <div className="kanban-modal__term" data-node-id={nodeId} tabIndex={0}>
+      Terminal Mock
+    </div>
+  )
+}))
+
+vi.mock('../../nodes/BrowserSurface', () => ({
+  BrowserSurface: ({ nodeId }: { nodeId: string }) => (
+    <div className="browser-surface" data-node-id={nodeId}>
+      Browser Mock
+    </div>
+  )
+}))
+
+;(globalThis as { IS_REACT_ACT_ENVIRONMENT?: boolean }).IS_REACT_ACT_ENVIRONMENT = true
+
+const board: ProjectKanban = {
+  columns: [{ id: 'col-1', title: 'To Do', color: '#3b82f6' }],
+  assignments: []
+}
+
+describe('CardModal', () => {
+  let host: HTMLDivElement
+
+  beforeEach(() => {
+    host = document.createElement('div')
+    document.body.append(host)
+  })
+
+  afterEach(() => {
+    document.body.innerHTML = ''
+  })
+
+  it('allows editing sticky note text and preserves space characters', () => {
+    const session: KanbanSession = {
+      id: 'node-sticky-1',
+      title: 'First line',
+      color: '#ffd60a',
+      kind: 'sticky',
+      text: 'First line',
+      spawn: {}
+    }
+
+    const root = createRoot(host)
+    const onEditSticky = vi.fn()
+    const onClose = vi.fn()
+
+    act(() =>
+      root.render(
+        <CardModal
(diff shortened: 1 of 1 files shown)


eneskirca:
Thanks — reviewed in depth. The root-cause fix is correct (the over-eager `.trim()` in the node→card mapping reverting the controlled textarea on every keystroke), and the new `kanbanSession.test.ts` is a real regression guard (mutation-checked).

One blocking item before this can be approved: the branch is based on v0.3.2, ~57 commits behind main, and conflicts with main's `toKanbanSession` — which has since gained a `partition` field for browser cards (guarded by `browser-partition-parity.test.tsx`). Please rebase onto current main and carry `partition: n.data.partition as string | undefined` into the browser branch of the extracted `src/renderer/lib/kanbanSession.ts`, otherwise browser-modal session-jar parity silently breaks (the field is optional, so typecheck won't catch it).

Happy to approve once rebased — CI has been enabled on the PR now.

Abhirup0:
Rebased onto current `main` and carried `partition: n.data.partition as string | undefined` into `toKanbanSession`'s browser card mapping in `src/renderer/lib/kanbanSession.ts`. 

Verified with `browser-partition-parity.test.tsx`, `kanbanSession.test.ts`, `CardModal.test.tsx`, and `npm run typecheck`. Ready for review!

eneskirca:
Update before you spend time on the rebase: the same root cause was independently fixed on main today (PR #314, merged 10:27 UTC — it removes the over-eager `.trim()` and extracts `toKanbanSession` with tests, keeping the `partition` field). Apologies for the race — the repo moves fast. If you'd like, the parts of your PR that still add value on top of main are the CardModal edit/Esc-wiring tests; feel free to rebase down to just those, otherwise we'll close this as superseded. Either way, thanks for the correct root-cause diagnosis on #285.

Abhirup0:
No worries at all,, glad the fix landed quickly!

I’ve rebased onto latest main and trimmed the PR down to just the CardModal integration tests (covering sticky note edits/whitespace preservation and Escape-key handling). All tests and typechecks are passing green.

Ready for another look whenever you have time!

eneskirca (approved):
Thanks for narrowing this down exactly as suggested — approving.

Reviewed in depth: the file is tests-only now, and it is genuinely not a duplicate of the `toKanbanSession.test.ts` that came with #314. That one tests the pure projection; this one tests modal wiring that had **zero** coverage on main — most valuably the three-tier Escape precedence (title-edit → note-edit → terminal-focused → close), which is exactly the kind of ordering logic that silently regresses when someone adds a fourth branch.

I ran a mutation sweep against CardModal.tsx: 12 mutations, 10 turned a test red (Escape branches, the write-through, click/Enter-to-edit, scrim close, rename commit). Nothing passed vacuously. Applies clean on main, typecheck clean, kanban suites green.

Non-blocking follow-ups if you feel like it:
1. Stub `BoardLogPanel`/`CardMetaBar`/`ContextMeter` like `browser-partition-parity.test.tsx` does — `BoardLogPanel` really mounts and calls `useSession()`, surviving only because `projectId` is empty; the day it calls `api` unconditionally, all five tests break for unrelated reasons.
2. Two guards stay unpinned: the sheet's `stopPropagation` (an inside mousedown must not close the card) and the `isTopDialog` ownership check.
3. The first test's name mentions preserving spaces, but what it actually pins is 'the textarea's write-through passes the raw value' — the space bug itself lived in the projection.

I retitled the PR to `test(kanban): …` so the history doesn't claim a fix #314 already shipped.

PR #293, opened 2026-08-18
feat(canvas): give a session an emoji or picture icon

alblor:
## What

A session node can carry an icon — one emoji, or a picture from disk.

Twelve terminals on one canvas are currently told apart by a colour dot and a title. This adds a
stronger mark, drawn by a single `NodeIconView` on **all four surfaces that list a session** — the
canvas node header, the kanban card, the card modal, and the sessions sidebar row — so one session
can't read as four different things depending on where you're looking at it.

Set it from the node right-click menu ("Set icon…", hideable like Colors), by clicking the icon in
the node header, or from the card modal's header slot.

Terminal (session) nodes only. The menu row is gated on the kind on purpose: offering it on an
editor or a group frame would persist a value nothing draws — a row that looks like it worked.

## Why the two non-obvious decisions

**The value is validated at *both* serializer seams.** `.nodeterm/project.json` is git-shared,
hand-editable, and on an SSH project a file on someone else's host. So `normalizeNodeIcon` runs in
`nodeStatesToFlow` (a cloned file becoming live state) *and* in `flowToNodeStates` (live state
becoming the next reader's file — live node data is reachable by a peer canvas mutation, and what
we write is what the next machine trusts). Validating one side passes every round-trip test while
leaving the other open; both seams are mutation-pinned in `workspace.test.ts`.

Three rules the validator enforces, each answering a specific failure:

- **An emoji is ONE grapheme** (`Intl.Segmenter`; the UTF-16 cap is only a fallback, never the
  primary rule — slicing units cuts a ZWJ sequence into a fragment). Uncapped, a shared project
  file could put a 40 kB "emoji" into every header, card and sidebar row.
- **An image path must look like an image** (extension → MIME). That gate is what stops a
  hand-edited project file from aiming `fs.readBinary` at `~/.ssh/id_rsa`. It isn't a full jail —
  the path can still name any `*.png` on the machine, exactly as an editor node's `filePath`
  always could — but the bytes only ever become an `` under a `'self'` CSP with no network,
  so the reachable outcome is "an icon fails to draw".
- **A `./` path may not traverse** — the same rule as `isSafeQuickOpenRelPath`.

**Picture bytes are copied, not referenced.** The picker writes through `files.saveCanvasImage` —
the seam canvas image nodes already use — so the file lands in the project's git-shared
`.nodeterm/images/`, and a path inside the project cwd is stored `./`-relative and resolved on
read. That's the convention `toPortableNodes` already set for node cwds, applied so the icon
travels with the canvas that names it. A cwd-less canvas, an SSH project and the app-local
fallback keep an absolute path and simply don't travel — a degrade, not an error. A relay tab is
refused outright with the same message as canvas image import, because the write is this machine's
preload while the read is the peer's core.

**Trade-off I'll name rather than hide:** canvas *image nodes* still store their `filePath`
absolutely, so today their file travels with the repo but the node naming it doesn't. This PR
doesn't fix that — it just doesn't repeat it. Happy to follow up if you want them aligned.

Image reads are cached per `(projectId, absPath)`: four surfaces mount independently, so a
thirty-card board would otherwise re-read the same bytes thirty times per open.

## Surfaces

- **Desktop** — full.
- **Server Edition** — full, and **no new IPC was added**. Every leg is already core
  (`fs.readBinary`, `files.saveCanvasImage`) or has a real browser implementation
  (`dialog.selectFile` → the web picker). Nothing is stubbed.
- **Mobile** — N/A for v1. The transport protocol carries no per-node icon concept; surfacing one
  means extending it. Flagging for @eneskirca rather than assuming.

## Testing

`npm run typecheck`, the full `vitest` suite, and a production build all pass.

I also **ran the app** (Electron under an isolated `--user-data-dir`, so no real workspace or tmux
server was touched): set an emoji, watched it reach the header, the sidebar row, the kanban card
and the card modal, and confirmed it landed in `project.json` on disk.

I mutation-tested every guard added here — the extension gate, the traversal check, the grapheme
cap, and both validation seams — and confirmed each one turns a test red when removed.

**Not verified:** the Server Edition and relay paths. Both are argued from the code (no new IPC;
reads go through the project's session api rather than `window.nodeTerminal`), not exercised on a
running server or a real peer.

Three `local-send-keys.realtmux` tests fail on my machine and fail identically on a clean `main` —
macOS's temp path exceeds the unix socket length limit. Unrelated to this change.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -171,7 +171,8 @@ Fire-time `TriggerArmStore.isArmed` re-ask everywhere; every rule test-pinned. T
 machine-local, content-bound `core/trigger-arm-store.ts` (a spec that arrives or CHANGES via git
 reads as disarmed until armed on this machine). A node's `data`
 carries `title, color, group, tags, collapsed, expandedHeight, shell, cwd, text,
-initialCommand, filePath, diffStaged`, `agentId` (which agent CLI a terminal node runs —
+initialCommand, filePath, diffStaged`, `icon` (a user-chosen emoji or picture — see **Node icons**
+below), `agentId` (which agent CLI a terminal node runs —
 persisted), and `accountId` (which managed Claude account a terminal node runs under — immutable,
 resolved at creation, persisted; see **Managed Claude accounts**). `nodeStatesToFlow` defaults a
 missing `kind` to `terminal` for backward compat and migrates the legacy `tags:['claude']` marker
@@ -1988,6 +1989,56 @@ disappearing" rather than as an occasional cull. The `vm_stat` reader is what ma
 again; the grace window was never the thing that was wrong.
 
 
+## Node icons (emoji or picture)
+
+A node may carry `data.icon` (`NodeIcon` in `@shared/node-icon`): `{type:'emoji', value}` or
+`{type:'image', path}`. Absent = the node draws exactly as it did before the feature, which is the
+degrade every failure path falls back to. Set from the node right-click menu ("Set icon…", hideable
+like Colors — id `icon`), from the icon itself in the terminal node header, and from the kanban card
+modal's header slot; drawn by the one `NodeIconView` on all four surfaces that list a node (canvas
+header, kanban card, card modal, sessions sidebar row), because a session seen in four places must
+not look like four sessions. **Terminal (session) nodes only in v1** — the menu row is gated on the
+kind, deliberately: offering it on an editor or a group frame would persist a value nothing draws,
+which is the "looks like it worked" failure this file warns about elsewhere. Extending it to sticky
+or browser nodes means adding the draw and the set together, in one change.
+
+- **`.nodeterm/project.json` is hostile input, so the icon is validated at BOTH serializer seams.**
+  `normalizeNodeIcon` runs in `nodeStatesToFlow` (a cloned file becoming live state) *and* in
+  `flowToNodeStates` (live state becoming the next reader's file — live node data is reachable by a
+  peer canvas mutation, and whatever we write is what the next machine trusts). One-sided validation
+  passes every round-trip test while leaving the other direction open; both seams are mutation-pinned
+  in `workspace.test.ts`.
+- **An emoji is ONE grapheme** (`Intl.Segmenter`, with a UTF-16 cap as the fallback, never as the
+  primary rule — slicing units cuts a ZWJ sequence into a fragment). Uncapped, a shared file could
+  put a 40 kB "emoji" into every header, card and sidebar row.
+- **An image path must LOOK like an image** (extension → MIME). That gate is what stops a hand-edited
+  project file from aiming `fs.readBinary` at `~/.ssh/id_rsa`. It is not a full jail — the path can
+  still name any `*.png` on the machine, exactly as an editor node's `filePath` always could — but
+  the bytes only ever become an `<img>` under a `'self'` CSP with no network, so the reachable
+  outcome is "an icon fails to draw". A `./` path may not traverse (same rule as
+  `isSafeQuickOpenRelPath`).
+- **The bytes are COPIED, not referenced.** The picker reads the chosen file and writes it through
+  `files.saveCanvasImage` — the same seam canvas image nodes use — so it lands in the project's
+  git-shared `.nodeterm/images/`. A path inside the project cwd is then stored `./`-relative
+  (`portableIconPath`) and resolved on read (`resolveIconPath`), which is the convention
+  `toPortableNodes` already set for node cwds. **This is the one place that convention is applied to
+  a `filePath`-like field**: canvas image nodes still store theirs absolutely, so their file travels
+  with the repo while the node naming it does not — an existing gap, not one this introduced.
+  A cwd-less canvas, an SSH project (its cwd is on the host; the image is written app-locally) and
+  the app-local fallback all keep an absolute path and simply do not travel. Not an error.
+- **A relay tab is refused** (`canvasImportRefusal`, the same message and the same reason as canvas
+  image import): the write is this machine's preload while the read is the peer's core, so the node
+  would name a file only this machine has. Reads otherwise go through the PROJECT's session api, not
+  `window.nodeTerminal` — which is what makes a peer-authored `./` icon resolve on the peer.
+- Image reads are cached per `(projectId, absPath)` in `lib/nodeIconImage.ts`, because four surfaces
+  mount independently and a thirty-card board would otherwise re-read the same bytes thirty times per
+  open. Caching by path is safe: `saveCanvasImage` creates exclusively, so re-picking yields
+  `logo (2).png` rather than overwriting.
+- **Surfaces.** Desktop: full. **Server Edition**: full — every leg is already core (`fs.readBinary`,
+  `files.saveCanvasImage`) or has a real browser implementation (`dialog.selectFile` → the web
+  picker), so no new IPC was added and nothing is stubbed. **Mobile**: N/A for v1 — *nodeterm mobile*
+  attaches to tmux sessions over the transport protocol and carries no per-node icon concept;
+  surfacing one means extending that protocol (follow-up in the iOS repo).
 ## Keybindings (registry, overrides, dispatch)
 
 Every user-facing chord is a registry command, and the whole engine is **one module**:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2017,6 +2017,42 @@ or browser nodes means adding the draw and the set together, in one change.
   the bytes only ever become an `<img>` under a `'self'` CSP with no network, so the reachable
   outcome is "an icon fails to draw". A `./` path may not traverse (same rule as
   `isSafeQuickOpenRelPath`).
+- **Two dialects in, one dialect out — and the traversal guard splits on BOTH separators, always.**
+  The value is written by one machine and read by another, so `normalizeNodeIcon` ACCEPTS a Windows
+  absolute (`C:\…`, `C:/…`) and a POSIX one wherever the check runs, while everything STORED is
+  POSIX-separated. Both halves are load-bearing. Refuse `C:\…` on a mac and a mac user merely
+  opening the shared canvas and saving it **silently strips a Windows teammate's icons** from
+  `project.json` — such a path does not resolve on a mac, but not-drawing is a degrade and a degrade
+  is not a reason to destroy the value on the way past. Store `.\a\b.png` and it means a file
+  called `a\b.png` on POSIX and `b.png` inside `a` on Windows, so a relative path is canonicalized
+  to `./` with forward slashes (the same way an emoji is canonicalized to its first grapheme).
+  **`isSafeRelIconPath` splits on `[\\/]` on every platform**, because splitting on `/` alone made
+  `./a\..\..\secret.png` ONE segment — neither `''`, `.` nor `..` — so it passed the guard
+  everywhere and escaped the project root the moment a Windows reader resolved it; a segment may
+  also not contain `:` (a drive qualifier, or an NTFS alternate data stream). **UNC is refused**,
+  matching `renderer/terminal/file-links.ts`, which consumes UNC specifically so it can refuse it:
+  reading one reaches another machine over SMB.
+- **`localIconCwd` is the ONE definition of which cwd a `./` icon may resolve against**, asked by
+  the picker's write side and by `useNodeIconSrc`'s read side. It was written twice and drifted: an
+  SSH project's `cwd` is a path on the REMOTE host while the icon is read through the LOCAL `api.fs`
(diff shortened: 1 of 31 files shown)


eneskirca:
This closed without ever getting the review it deserved — that's on us, and I'm sorry: you closed it after ten days of silence, and having now read every file, it was one of the best-shaped external PRs this repo has received. Validating at *both* serializer seams with mutation-pinned hostile-input tests, the grapheme-not-units emoji cap via `Intl.Segmenter`, the `null`/`undefined` choice semantics pinned in their own module, one `NodeIconView` across all four surfaces with the kanban board treated as first-class in the same change, real Server Edition legs, zero new dependencies, token-only CSS. CI was fully green.

**We'd genuinely like this reopened.** Three things it would need on the way in:

1. **`src/renderer/lib/nodeIconImage.ts` contains a raw NUL byte** (the cache-key separator in `keyFor`), which makes git classify the file as *binary* — it renders as "Binary files differ" in every diff surface and grep skips it. By bad luck that's the one new file performing `fs.readBinary` reads driven by persisted data, i.e. the file a reviewer most needs to see. The logic inside is fine once visible — please just write the escape: `\u0000`. (We've been bitten by NUL-hidden files before.)

2. **Windows:** `normalizeNodeIcon` requires absolute paths to start with `/`, so the `C:\...` path `saveCanvasImage` returns on Windows is refused — and `chooseImage` then misdiagnoses a valid PNG as "that file type cannot be used". Windows desktop is being brought up as first-class right now, so the path seams (`portableIconPath`/`resolveIconPath`, `picked.split('/')`) want to be dialect-aware or at least fail with the honest message.

3. **Rebase:** the branch is now ~2 weeks behind and conflicts with main in 9 files (Canvas.tsx and TerminalNode.tsx moved substantially).

Smaller, non-blocking if you do pick it back up: Escape only closes the picker while the input has focus (`useDialogStack()`'s is-top answer is discarded — owning window Esc when top, like ConfirmDialog, would fix it); consider downscaling the picked image to a thumbnail at pick time (it draws at 13–16 px but a full-size photo is cached in renderer memory AND committed into the git-shared `.nodeterm/images/`); and for SSH projects the read side could skip `./` resolution the way your write side already does.

If you'd rather not carry it further, no hard feelings — but the door is open, and the next review will not take two weeks.

alblor:
Thanks, that's a generous read, and it was specific enough to just act on. No hard feelings about the wait. I closed it instead of nudging, which was clearly the worse of the two options.

Reopened and rebased onto today's `main`. All three blockers are done, plus all three of the smaller ones. One commit per point, so re-reviewing is just reading the deltas instead of the whole thing again.

## 1. The NUL byte

Confirmed: one NUL at byte 2446, and `git diff` printed `Bin 0 -> 4869 bytes`. It's `\x00` now, same byte at runtime, and the file shows up as 112 readable lines in this PR's diff.

I put the escape **in the feature commit itself** rather than in a follow-up, so the file is never binary anywhere in the branch's history. Someone reading just that commit can still see it.

The separate commit is a guard. `src/shared/source-hygiene.test.ts` fails if any tracked `.ts`/`.tsx` under `src/` has a raw NUL. It scans via `git ls-files` so build output and gitignored files stay out with no hand-maintained list, and it asserts it actually looked at something, because a guard that silently scans nothing passes forever. All 1424 tracked sources are clean today. You said you'd been bitten by this before, so catching the class seemed better than the instance.

## 2. Windows, and it's worse than you found

You're right about `normalizeNodeIcon`. Chasing it turned up a security bug underneath.

**`isSafeRelIconPath` split on `/` only.** So `./a\..\..\secret.png` was one single segment. Not `''`, not `.`, not `..`, so it sailed through the traversal guard **on every platform**, then walked out of the project root as soon as a Windows reader resolved it. One machine writes this value and another one reads it, so the rule has to hold where it's *checked*, not just where it happens to be dangerous. Both separators count everywhere now, and a segment can't contain `:` either (drive qualifier, or an NTFS alternate data stream).

For the rest I went with **two dialects in, one dialect out**:

* **Validation accepts both** (`C:\…`, `C:/…`, `/…`) no matter where the check runs. This is the bit I'd most like you to sanity-check, because the alternative is quietly destructive. If a mac's `flowToNodeStates` refused `C:\…`, then a mac user just opening the shared canvas and saving it would **wipe a Windows teammate's icons out of `project.json`**. That path doesn't *resolve* on a mac, sure, but it just doesn't draw, same as a POSIX absolute path from someone else's machine already doesn't. Not drawing isn't a good reason to delete the value on the way past.
* **What gets stored is POSIX-separated.** That's the whole reason the `./` form is portable: a stored `.\a\b.png` means a file called `a\b.png` on POSIX and `b.png` inside `a` on Windows. Relative paths get canonicalized, same as an emoji gets canonicalized to its first grapheme.
* **UNC stays refused**, following `renderer/terminal/file-links.ts`, which consumes UNC specifically so it can refuse it. Reading one reaches another machine over SMB.

`portableIconPath` compares separator-insensitively now. It never matched on Windows before, so the icon silently stayed absolute and travelled nowhere with nothing said about it. And `picked.split('/')` is now a shared `iconFileName` that basenames either dialect.

**One extra bug, closed by the same fix.** `chooseImage` copied the picked file into the git-shared `.nodeterm/images/` *before* checking its extension. So every refusal left an orphan file sitting in the user's repo, and nothing ever removes it (`saveCanvasImage` creates exclusively, so the next attempt lands next to it as `photo (2).heic`). `dialog.selectFile` applies no filter, so a `.heic` is one click away. The check runs first now and the bytes aren't even read. That's also what fixes the misdiagnosis you spotted, since the post-save error stops blaming the file type.

## 3. Rebase

Done, onto `main` at `dba7ce40`. Nine files conflicted. The branch is purely additive so they were context conflicts, not semantic ones. Two worth flagging:

* `toKanbanSession` moved out of `Canvas.tsx` into its own module, so the one-line `icon:` addition went there instead of resurrecting a duplicate definition.
* `CardModalProps.onSetIcon` is required, and `main` has since added `CardModal.test.tsx` and `browser-partition-parity.test.tsx` that build it without the prop. I added it at those call sites rather than making the prop optional, since an optional one lets a surface silently forget to wire it, which is exactly the "looks like it worked" failure the rest of this PR is trying to avoid. Say the word if you'd rather have it the other way.

## The three smaller ones, all taken

**Escape.** Spot on, `useDialogStack()`'s answer was just being thrown away. A window listener owns the key for the whole dialog now, gated on `isTop()` **alone**, deliberately matching `confirmKeyAction` where `inDialog` guards Enter and never Escape. I tried the symmetric version first and it was wrong: requiring focus inside the box just reproduces the original bug for anyone whose focus is on the body. The input's own Escape branch is gone too, since two handlers for one keypress is the double-answer the stack exists to prevent.

**Downscale.** Taken, bounded at 256px on the long edge (16x the drawn size). Measured in a real Chromium against real PNGs:

| input | output | size | time |
|---|---|---|---|
| 1200x900, 0.49 MB | 256x192 | 3 KB, **99.4% smaller** | 7 ms |
| 4000x3000, 2.32 MB | 256x192 | 37 KB, **98.4% smaller** | 38 ms |

`thumbnailPlan` is pure so it tests under vitest's default `node` environment (jsdom has no canvas, and a rule you can only test in a browser doesn't get tested). The decode/encode is injected.

It deliberately leaves alone: **SVG** (rasterizing it makes it worse at *every* size, not just smaller), anything **already small** in both dimensions and bytes (a canvas round-trip can make a hand-made 32px PNG *bigger*), and any **re-encode that came out larger**. An animated GIF becomes a static PNG, which is the right call at 16px and is stated in the module.

It fails open in every direction, including a decode that never settles. `chooseImage` awaits this before writing, so a hanging promise would leave the button stuck on "Copying…" with no way out but closing the dialog. 3s bound, about 80x headroom over the measured worst case.

**SSH `./` resolution.** Taken, and the interesting part is *why* it was broken. The write side already knew (`project.ssh ? undefined : project.cwd`). The read side didn't. The rule was written twice and drifted. So it's written once now, `localIconCwd(project)`, asked by both sides. Undefined means the icon doesn't draw, which is the honest answer for a file on a filesystem this reader can't see. Absolute paths are unaffected, and absolute is what the write side stores for SSH anyway. Relay tabs are untouched, since there the api and the cwd both belong to the peer and `./` resolution is already correct.

## Verification, and what I couldn't verify

One thing needs you: the force-push put both workflow runs into **`action_required`**. GitHub gates fork-PR runs behind a maintainer click, so CI is waiting on "Approve and run" rather than failing. Everything below is my machine, not yours.

* `npm run typecheck`, production build, and the full vitest suite all pass. **9771 tests, and the 13 failures are byte-identical to the 13 on a clean `main`.** I ran `main` in a parallel worktree to check rather than just asserting it. They're environment failures here (real-tmux, real-sh, localStorage in my sandbox), not code.
* Every guard added is **mutation-tested**. Removing the backslash split, the colon rejection, the UNC refusal, the drive-absolute acceptance, the pre-save extension check, the dialect-aware basename, the `isTop()` gate, the Escape listener, the SSH exclusion, the SVG passthrough, the already-small skip, the larger-re-encode guard, the `.png` rename, or the decode timeout each turns a test red.
* Ran the app under an isolated `--user-data-dir` and confirmed it boots clean on this branch.
* The downscale numbers above come from a real headless Chromium against real PNGs. That's the one piece the unit tests inject *around*, so I wanted it actually exercised rather than argued.

**Not verified, and I'd rather just say so:** I don't have a Windows machine. The Windows behaviour is correct by construction and by unit test, not observed on Windows. The traversal case and the `C:\` round-trip are pinned in `node-icon.test.ts`, but a real Windows pick-and-save is unexercised. Server Edition and relay paths are still argued from the code, same as in the original description.

Also flagging: I promoted the CLAUDE.md section from `###` to `##`. Sitting as a `###` right after "Session memory" it read like a subsection of it, which it isn't. Just an artifact of where the branch originally appended. Two entries went into `CONTRIBUTING.md` as well, per the two-docs convention: the NUL trap, and treating `\` as a separator wherever a persisted path gets split.

Happy to split any of this up differently, or pull the downscale into its own PR if it widens the diff more than you want here.

alblor:
Rebased again onto `2864b596`, since main moved another 11 commits while I was writing the above. Green again.

Heads up on the one recurring collision: it's always `styles.css`, and always the same cause. Everybody appends their block at the end of the file, so any two branches that add CSS conflict there by construction. If this goes red again before you get to it, that's all it is, and it's a 10 second resolve.

Worth knowing if you hit it yourself though: twice now git split that conflict *inside* a rule, so the two sides shared a single trailing `}`. Naively keeping both sides leaves the first rule unclosed. Tests and typecheck both stay green on that (CSS isn't typechecked), only `npm run build` catches it. So if you resolve it, run a build rather than trusting the suite.

I also re-checked we haven't drifted from main while doing all this rebasing:

* Zero merge commits, `origin/main` is a straight ancestor.
* 31 files changed, 2170 insertions, and only **5** deletions. Every one of those 5 is a line replaced by a longer version of itself (adding `icon` to the hideable list, `onSetIcon` to the CardModal signature, `IconSmiley` to an import, and so on). No content of yours is dropped.
* `styles.css` specifically: 123 added, 0 removed, and 0 of main's 11752 lines missing from the branch.
* No file deleted or renamed, no `package.json` or lockfile change.

eneskirca:
Verified the rebased branch (`daaa9653`) in an isolated worktree rather than reading it: `npm run typecheck` clean, **full suite green — 717 files, 9796 passed / 40 skipped, 0 failed** (so the 13 environment failures you saw are your sandbox, not the branch). CI ran after approval and is green on every job, `quality-windows` included. The NUL byte is gone (`nodeIconImage.ts` diffs as 116 text lines) and `source-hygiene.test.ts` guards the class.

On the `styles.css` hazard you flagged: I could not run `npm run build` here (this box's `node_modules` predate the monaco 0.56 bump — unrelated to you), so I checked the CSS directly instead: your `styles.css` and the three-way merge with today's `main` (`git merge-file`, clean) both parse under esbuild's CSS loader with balanced braces (1812/1812 and 1824/1824). No unclosed rule on either side.

The three blockers are closed as described, and the extra finds hold up on reading: `isSafeRelIconPath` splitting on both separators plus the `:` refusal is the right fix for the cross-platform traversal, "two dialects in, one dialect out" is the correct call (refusing `C:\…` on a mac would have deleted a teammate's icons on save), `localIconCwd` as the single definition for the `./` resolution rule closes the read/write drift, the pre-copy extension check removes the orphan-file leak, and the Escape listener gated on `isTop()` alone matches `confirmKeyAction`'s asymmetry. `downscaleIconImage` fails open in every direction with a bounded decode, and `thumbnailPlan` being pure is why it tests under node. Requiring `onSetIcon` on `CardModalProps` (rather than optional) is the choice we'd make too.

From my side this is ready; the merge call is @eneskirca's. Windows pick-and-save stays device-unverified, as you said.

PR #294, opened 2026-08-18
feat(canvas): file-manager node — a directory listing on the canvas

alblor:
## What

A `files` node: **one directory listing** (`data.cwd`, persisted), pinned to the canvas beside the
terminals working in it. Navigate (clickable breadcrumbs, up), filter, open, create a file or
folder, copy a path, reveal, and open a terminal in the folder shown.

## Why not just the Explorer drawer

The drawer is a single tree rooted at the project cwd, and it's modal in practice — it covers the
canvas and you close it to get back to work. That gives you one cursor and a lot of scrolling.
Several of these give you `src/renderer/nodes` open next to the agent editing it and another on
`docs/` — which is the same argument the app already makes about terminals versus tabs.

## It adds NO new IPC

Everything runs on the existing `FsApi` (`list` / `mkdir` / `exists` / `write`). That's why
Desktop, the Server Edition, SSH projects and relay tabs all work on day one, with nothing stubbed.

**Rename / move / delete are a deliberate v1 gap, not an oversight.** Each needs a leaf in
`core/fs-ops`, an IPC channel, preload, the ws-bridge, `main/ssh-fs` (remote quoting) and the relay
host-service, plus confirm dialogs and dangerous-path guards to meet the destructive-safety rules
in `CONTRIBUTING.md`. That's a separate change with separate risk, and I'd rather propose it on its
own than bury it in this one. Happy to write it as a follow-up if you want it.

## Decisions worth the review time

- **Which filesystem** is `EditorNode`'s decision, read the same way: `data.sshFs` → the project's
  host over the ControlMaster; otherwise the node's **own session api**. Reading it off
  `useSession()` rather than `window.nodeTerminal` is the entire reason a relay tab browses the
  machine its terminals are actually on.
- **Opening is delegated, never reimplemented.** A file dispatches `nodeterm:open-file` — the event
  `TerminalNode`'s Cmd+click links already use — so editor/video/image routing stays in Canvas's
  one `openFile`, and this node never grows a second opinion about what a `.png` is.
  `fileOpenTarget` decides only canvas-vs-OS, and a **remote listing never reaches the OS branch**:
  `shell.openPath` opens a path on *this* machine, so handing it a host path either does nothing
  or — if that path happens to exist locally too — opens an unrelated local file.
- **Four empty states stay four**: loading / could not read / the folder is empty / the filter
  matches nothing. An empty filter matches everything, so select-all + delete can't blank the
  listing.
- **The title tracks the folder only while `titleAuto` is unset** — the same contract an agent
  node's session name uses, so navigating never overwrites a name you typed.
- Creation is gated on `hasCwd` on every surface (pane menu, Dock, sidebar `+`, ⌘K, group menu).
  Inside a group frame it inherits a bound **worktree's** cwd via `cwdForNewNodeIn`, so a frame per
  branch also means a file tree per branch.

## One bug fixed that this would otherwise have shipped

A files node inside a **removed worktree** kept a dead path in `project.json` and showed "Could not
read this folder" forever — the same persisted-dead-path failure the terminal branch of
`displacedByWorktree` exists to prevent.

It's now caught **by path wherever it sits**, like an editor and unlike a terminal: the `under`
restriction on terminals is there because displacing one *respawns* it, and a session you rooted
somewhere by hand shouldn't restart under you. A files node has no session. And it gets its `cwd`
re-pointed rather than an editor's `fileMissing`, because a directory can be re-pointed and a dead
file can't.

## Surfaces

- **Desktop** and **Server Edition** — full, no new IPC.
- **Mobile** — N/A. *nodeterm mobile* attaches to tmux sessions over the transport protocol and has
  no canvas or file-browsing concept; adding one means extending that protocol. Flagging for
  @eneskirca rather than assuming.

## Testing

`npm run typecheck`, the full `vitest` suite, and a production build all pass.

I also **ran the app** (Electron under an isolated `--user-data-dir`, so no real workspace or tmux
server was touched): created the node from the pane menu, listed the project, navigated in and out,
filtered to a miss and back, opened a file (an editor node appeared), read the entry context menu,
and confirmed the node and its cwd survive a restart. **That run found a real defect** — a doubled
`/ /` in the root breadcrumb — which is fixed here.

The pure logic (`lib/filesNode.ts`) has its own tests, and I mutation-tested the new
`displacedByWorktree` branch to confirm it turns tests red when removed.

**Not verified:** the SSH and relay paths, and the Server Edition. All three are argued from the
code (no new IPC; the filesystem is picked exactly as `EditorNode` picks it), not exercised against
a real host, peer, or browser session.

Three `local-send-keys.realtmux` tests fail on my machine and fail identically on a clean `main` —
macOS's temp path exceeds the unix socket length limit. Unrelated to this change.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -161,12 +161,12 @@ separate store mirroring node state — earlier dual-source designs caused sync 
 `src/renderer/state/workspace.ts` holds only pure helpers: the color palette, the node
 factories (`createTerminalNode`, `createSshTerminalNode`, `createAgentNode(agentId, …)`,
 `createAccountLoginNode`, `createStickyNode`, `createGroupNode`, `createEditorNode`,
-`createDiffNode`, `createVideoNode`, `createWebNode`, `createBrowserNode`, `createDinoNode`,
-`createTriggerNode`), the
+`createDiffNode`, `createVideoNode`, `createWebNode`, `createBrowserNode`, `createFilesNode`,
+`createDinoNode`, `createTriggerNode`), the
 group transforms (`groupSelectedNodes`, `ungroupNodes`, `duplicateNode`), and the
 `nodeStatesToFlow` / `flowToNodeStates` serializers. Node kinds (`NodeKind` in
 `src/shared/types.ts`): `terminal | sticky | group | editor | diff | video | web | browser |
-subagent | loop | dino | trigger` — `subagent` and `loop` are render-only (ephemeral hook-driven
+files | subagent | loop | dino | trigger` — `subagent` and `loop` are render-only (ephemeral hook-driven
 viz) and never persisted. `trigger` (issue #493, all four
 phases landed) is a first-class PERSISTED kind. The whole host-side
 machine is composed ONCE in `core/trigger-service.ts` (`startTriggerService`) and booted
@@ -850,6 +850,44 @@ session.
 - **browser** (`BrowserNode.tsx`) — a navigable Chromium browser wrapping the shared
   `BrowserSurface` (webview + toolbar); the last top-level URL persists to `data.url`, and the same
   surface backs the kanban card modal's browser popup.
+- **files** (`FilesNode.tsx`) — a file-manager node: ONE directory listing (`data.cwd`, persisted),
+  pinned to the canvas beside the terminals working in it. Deliberately not a second Explorer: the
+  drawer is a single tree rooted at the project cwd that covers the canvas, so it gives you one
+  cursor and a lot of scrolling; several of these give you `src/renderer/nodes` next to the agent
+  editing it and another on `docs/`. Navigate in place (breadcrumbs collapse deep paths but every
+  crumb stays clickable), filter, create a file/folder, copy a path, reveal, and open a terminal in
+  the folder shown (a `nodeterm:open-terminal` event, the sibling of `nodeterm:open-file`).
+  - **It adds NO new IPC.** Everything runs on the existing `FsApi` (`list`/`mkdir`/`exists`/
+    `write`) — which is why it works on Desktop, the Server Edition, an SSH project and a relay tab
+    on day one. **Rename/move/delete are the deliberate v1 gap**: each needs a new leaf in
+    `core/fs-ops`, an IPC channel, preload, the ws-bridge, `main/ssh-fs` (remote quoting) and the
+    relay host-service, plus confirm dialogs and dangerous-path guards. That is a separate change
+    with separate risk, not a corner to cut inside this one.
+  - **Which filesystem** is `EditorNode`'s decision, read the same way: `data.sshFs` → the SSH
+    project's host over the ControlMaster; otherwise the node's own SESSION api, which is the local
+    core for a local project and the PEER's for a relay tab. Reading it off `useSession()` rather
+    than `window.nodeTerminal` is the entire reason a relay tab browses the right machine.
+  - **Opening is delegated, never reimplemented**: a file dispatches `nodeterm:open-file`, so
+    editor-vs-video-vs-image routing stays in Canvas's one `openFile` and this node never grows a
+    second opinion about what a `.png` is. `fileOpenTarget` decides only canvas-vs-OS, and a
+    REMOTE listing never reaches the OS branch — `shell.openPath` opens a path on THIS machine, so
+    handing it a host path either does nothing or, if that path exists here too, opens an unrelated
+    local file.
+  - **Four empty states stay four** (loading / could not read / the folder is empty / the filter
+    matches nothing). Collapsing any pair into one blank pane is the failure this file warns about
+    elsewhere, and an empty filter matches everything so select-all+delete cannot blank the list.
+  - The title tracks the folder only while `titleAuto` is unset — the same contract an agent node's
+    session name uses, so navigating never overwrites a name the user typed.
+  - A files node inside a REMOVED worktree is displaced like an editor, not like a terminal
+    (`displacedByWorktree`): it has no session to disturb, so it is caught by path wherever it sits,
+    and it gets its `cwd` re-pointed rather than an editor's `fileMissing` — a directory can be
+    re-pointed, a dead file cannot. Left out, it would show "Could not read this folder" forever
+    AND keep the dead path in `project.json`.
+  - Creation is gated on the project having a directory (`hasCwd`) on every surface — pane menu,
+    Dock, sidebar "+", ⌘K, and the group-frame menu, where it inherits a bound **worktree's** cwd
+    via `cwdForNewNodeIn`, so a frame per branch also means a file tree per branch.
+  - **Mobile**: N/A — *nodeterm mobile* attaches to tmux sessions over the transport protocol and
+    has no canvas or file-browsing concept; adding one means extending that protocol.
 - **dino** (`DinoNode.tsx`) — a small self-contained T-Rex-style runner on a canvas (no PTY);
   high score persists via `data.highScore`.
 - **trigger** (`TriggerNode.tsx`) — a canvas-owned schedule (cron / interval / once) that
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -891,7 +891,14 @@ session.
     disambiguated by probing the PARENT's listing instead (`classifyEmptyListing`, pure,
     `lib/filesNode.ts` — the idiom `SshProjectDialog.tsx:112-125` and `file-links.ts`'s
     `makeDirListingLookup` already use), which answers `missing` / `empty` / `unknown`: an unreadable
-    parent answers `unknown`, and we keep saying "empty" rather than claim a deletion we can't prove.
+    parent answers `unknown`, and we keep saying "empty" rather than claim a deletion we can't
+    prove. **`unknown` also covers every path whose parent CANNOT answer** — a non-`/`-absolute cwd
+    (an SSH project's `remoteCwd` defaults to **`~`**, where `parentDir` is `/` and `/` has no entry
+    named `~`, so an empty remote HOME reported a deletion), a `.git` cwd (both listing legs strip
+    it on purpose), and `.`/`..` segments; a case-folded match counts as present, for the
+    case-insensitive filesystems where `readdir` answers with the on-disk spelling. `missing` must
+    be the conclusion we are SURE of. Nothing is published until the verdict is in, so a deleted
+    folder never flashes "empty" on its way to the error.
     (b) Nothing reset the list on navigation, so "Loading…" was reachable only on the very first
     mount — every later directory change showed the PREVIOUS directory's rows. Fixed by storing the
     listing WITH the cwd it belongs to, so a cwd change IS the loading state by construction; a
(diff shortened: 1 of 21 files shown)


eneskirca:
Thanks for this — and apologies it sat two weeks without a review. This is an unusually careful external PR: the no-new-IPC design is the right call, the `sshFs`/`useSession()` filesystem selection is byte-for-byte the `EditorNode` pattern (I checked), delegation through `nodeterm:open-file` keeps file-type routing in one place, and the pure-lib + mutation-tested `displacedByWorktree` branch is exactly the discipline this repo asks for. I verified the load-bearing claims rather than trusting the body: persistence round-trips for free (`cwd`/`sshFs`/`titleAuto` are already generic serializer fields), `resetDisplacedCwd`'s else-branch really does re-point a displaced files node, `newEntryPath` refuses traversal, and typecheck plus the full shared/lib/state sweep (1871 tests) are green on your branch. I was fully prepared to find nothing — most of what I looked for wasn't there to find.

Asks, in order:

1. **Rebase onto current main.** The branch now conflicts in `types.ts` / `workspace.ts` / `styles.css` / `CLAUDE.md` (main gained a `trigger` node kind and reworked `workspace.ts` since Aug 18). Resolution should be mechanical. After rebasing, please run two guard tests that landed after your branch point: `src/renderer/lib/nodeterm-events.test.ts` (your `nodeterm:open-terminal` pairing should pass — both ends exist) and `src/renderer/styles.theme.test.ts` (your CSS is token-only, so I expect a pass, but that's read, not run).

2. **The "Could not read this folder" state is unreachable on local and SSH projects** (measured): `core/fs-ops.listDir` and `SshFs.listDir` both fold every error into `[]` by contract, so your `.catch` only fires when the IPC itself rejects. A dead directory — including the removed-worktree case your PR body describes — actually renders "This folder is empty." Suggestion that needs no new IPC: when `list` resolves `[]`, probe `fs.exists(cwd)` (already in `FsApi`) and show the error state on `false`. (The displacement fix itself is correct and stays regardless — this is about the fourth state actually existing.)

3. **Server Edition: opening a `.zip`/`.dmg` is a silent dead click** (measured): in a browser tab the session source is `'local'`, so `fileOpenTarget` can return `'os'`, and `api.shell.openPath` is a documented no-op in the ws-bridge stubs. You caught exactly this for Reveal with `isBrowserRuntime()` — the `'os'` branch needs the same gate (open on canvas anyway, or toast).

4. One sentence each in the CLAUDE.md bullet for two decided-but-unstated calls: relay tabs can create files but "New folder…" always fails there (relay `mkdir` degrades to `false` by contract), and the kanban board deliberately doesn't surface files nodes (cards derive from terminal nodes).

Nits, take or leave: a worktree-displaced node keeps its stale title (the cwd rewrite bypasses `navigate`, so `titleAuto` tracking never fires); `createFilesNode` inlines `folderTitle`; the `/`-separated path assumption in `lib/filesNode.ts` is a known shared gap with `explorerCreate` that Windows bring-up will eventually hit — fine to inherit, worth a comment.

CI never ran on this branch (external PR, needs workflow approval) — we'll kick it on your next push.

alblor:
Thanks — this was specific enough to act on directly, and the verification you did up front saved a round.

I've worked through all four asks and the three nits. Before the code lands, three things from your comment worth correcting now rather than after, since two of them change what the fix should be.

**Ask 4(a) — relay `mkdir` doesn't degrade.** "New folder…" works on a relay tab: `relay-api.ts:86` wires `fs: files.fs`, `buildFilesApi` routes `mkdir` → `IPC.fsMkdir`, and `fs-handlers.ts:43` maps that to the real `fsOps.makeDir` on the peer's core. The "degrades to `false`" line in `CLAUDE.md:2260` describes `src/renderer/terminal/remote-fs.ts`, which was deleted in `fc1b2bc1` — the file is gone and the sentence outlived it. What still lacks `fs.exists`/`fs.mkdir` is the legacy phone vocabulary in `main/remote/host-service.ts:376-390`, and that one *rejects* rather than degrading. So I'm fixing the stale line instead of propagating it into the files bullet.

**Ask 2 — taking it, but not via `fs.exists`.** The diagnosis is exactly right; the probe doesn't quite close it. `fsOps.pathExists` is `stat`-based, so a directory you can stat but not `readdir` still renders "empty", and on SSH `false` can't separate "absent" from "the ControlMaster died" — which is the case `ssh-fs.ts:152-172` singles out with *"a failed read is never evidence of absence."* So I'm using the parent-listing idiom instead, the one `SshProjectDialog.tsx:112-125` and `file-links.ts`'s `makeDirListingLookup` already use for exactly this. Same no-new-IPC constraint. Say the word if you'd rather have the simpler version.

**CI has since run.** You approved it about four minutes after posting — `quality`, CodeQL and Dependency review all passed at 18:35Z. Those results are stale now (Aug-18 merge base, and GitHub reports the PR dirty), and `quality-windows` will apply for the first time after the rebase, since the branch's own `ci.yml` predates that job.

**Ask 3, taken as written.** You're right that the file states the rule and then doesn't apply it — `canReveal` sits three lines above the ungated `openPath`. It's going behind a pure `{browser, ssh, source}` predicate shaped like `canRevealLocally`, with the refusal surfaced through the `toast()` helper the file already has. Worth knowing the same hole already exists at `Canvas.tsx:3851-3865` (`openProjectFile`, the ⌘K path), which also reads `window.nodeTerminal` rather than the session api — I'd rather send that as its own PR than widen this one.

**Ask 1 and the nits, all taken.** The rebase is 11 hunks over 6 files (`Canvas.tsx` and `Dock.tsx` too — `trigger` landed where `files` did), all mechanical, resolved by union rather than by side: taking a side on `Canvas.tsx` drops `addFiles` while the dispatch survives, which turns `nodeterm-events.test.ts` red.

**One thing the conflict list doesn't show.** `lib/reopenNode.ts` didn't exist at my branch point. `'files'` is in neither `UNRESTORABLE` nor `buildBase`'s switch, so deleting a files node records a ⇧⌘T entry that can never be restored — compiles, typechecks, passes everything. Your own comment at `:55-57` is what flagged it. I'm adding the `case 'files'` rather than the exclusion, since `createFilesNode` already has the shape `buildBase` wants.

Four smaller things the same pass turned up, also fixed: **"Loading…" was unreachable after mount** (`entries` never resets, so navigation shows the previous folder's rows — my "four states stay four" claim was really two); **"New terminal here" discards the folder on SSH and hands a peer path to a local terminal on relay** (same class as ask 3 — `createTerminalNode`'s `cwd: ssh ? ssh.remoteCwd : cwd` means `cwdOverride` is already broken for every SSH caller, not just mine); a displaced node whose `fallbackCwd` is undefined **silently browses `/`**; and Escape out of a rename **permanently detaches `titleAuto`**.

One commit per point, so re-reviewing is reading deltas. I'll push once the full suite is compared against a clean `main` in a parallel worktree and a build passes — `styles.css` conflicts have twice resolved into an unclosed rule that only `npm run build` catches. Relay, a real SSH host and Windows stay unverified from here, same as the original description.

alblor:
Thank you by the way for inspecting my pull request, happy to help. The notes were specific enough to just act on, which is the best kind of review to get.

All four asks and the three nits are done, one commit per point, so re-reviewing is reading the deltas rather than the whole thing again. CI is green on `0980973b`, `quality-windows` included, which has never run on this branch before: its own `ci.yml` predated that job. So nit 3 has a real Windows signal behind it now.

## 1. Rebase

Done, three times, because main moved twice while I was working and #293 landed in the middle of it. Zero merge commits, `origin/main` is a straight ancestor, 20 files, +1513/-22.

Both guard tests pass. The one that nearly caught me was `addMenuSpec.test.tsx`, and not for the reason you'd expect: #621 landed while I was rebasing, and it **reverses the rule this branch was built on**. My files row hid itself on a cwd-less project, with a comment saying it did that "for the same reason 'New file…' is". You'd just made that the wrong reason. It's a disabled row carrying `FILES_NO_CWD_HINT` now, on the pane menu, Dock and sidebar. ⌘K still hides it, matching what your own "New file…" does there, since a disabled palette entry is a search result that does nothing.

Resolving that conflict mechanically would have compiled, passed, and shipped a row contradicting a rule you'd written down the same day.

Heads up on the recurring one: it's `styles.css` again, same cause as last time. And on the first pass git split it **inside a rule**, so both sides shared one closing brace. Tests and typecheck stay green on that. Only the build catches it.

## 2. The empty state

You're right, and the measurement is right. Taken, but through the parent's listing rather than `fs.exists`: `pathExists` is `stat`-based, so it answers true for a directory you can stat but not `readdir`, and on SSH its `false` can't separate "gone" from "the master died", which is the conflation `readTextChecked` refuses on the grounds that a failed read is never evidence of absence. `SshProjectDialog` and `makeDirListingLookup` already ask the parent for exactly this question.

Then my own version turned out worse than what it replaced. `remoteCwd` defaults to `~`. `ls ~` works fine, but `parentDir('~')` is `/`, and `/` has no entry called `~`, so an **empty remote home reported a deletion**. Same for a `.git` cwd, since both listing legs strip it on purpose, and on case-insensitive filesystems. Anything whose parent can't answer is `unknown` now, and a case-folded match counts as present. "Missing" has to be the answer we're sure of.

While I was in there: "Loading…" was only ever reachable on the first mount. Nothing reset the list, so every later navigation showed the previous folder's rows. My "four empty states stay four" was really two.

## 3. The OS open

Spot on, and the file states the rule three lines above the call that ignores it. It's behind `canUseLocalShell` now, one predicate for every `shell.*` path action, with `canRevealLocally` kept as its older name so `ExplorerPanel` is untouched.

Worth knowing the same hole already exists in `Canvas.tsx`'s `openProjectFile`, the ⌘K path, which also reads `window.nodeTerminal` rather than the session api. I'd rather send that as its own PR than widen this one.

## 4. The docs

Both facts are in. On the relay one though, I don't think it's true any more: a relay tab's `fs` goes `relay-api.ts:86` to `buildFilesApi`'s `IPC.fsMkdir` to `fs-handlers.ts:43` to the real `fsOps.makeDir` on the peer. The "degrades to `false`" line in CLAUDE.md is describing `renderer/terminal/remote-fs.ts`, deleted in `fc1b2bc1`. The sentence outlived the file. What genuinely still lacks those verbs is the phone vocabulary in `host-service.ts`, and that rejects rather than degrading. So I fixed the stale line instead of copying it into the files bullet. Push back if I've misread it.

Kanban is in too, phrased off what the code actually does: `toKanbanSession` maps browser, sticky and terminal, and returns null for everything else. A files node has no session to co-attach and no text to edit, and the whole point of it is sitting next to the terminal working in that directory, which a column layout throws away.

## The nits, and four more

All three taken. Escape restores `titleAuto` as well as the text, since `onChange` set it false on the first keystroke and cancelling a rename you never committed detached the title for good. `folderTitle` moved down into `explorerCreate`, which imports nothing, because `filesNode.ts` imports `isVideoFile` from `workspace` and importing back would close a cycle. The `/` assumption is written down now, naming `explorerCreate` as the shared owner and `file-links.ts`'s core-owns-the-dialect rule as the thing to copy when it's closed.

Four others turned up on the way.

`reopenNode.ts` didn't exist at my branch point, and `files` was in neither `UNRESTORABLE` nor `buildBase`. So deleting one recorded a ⇧⌘T entry, plus its persisted `closedSessions` twin, that `default: return null` could never restore. Compiles, typechecks, passes everything. Your `trigger` comment at `:58-60` is what made it findable, and that's two kinds now.

"New terminal here" was broken on both remote kinds. `createTerminalNode` does `cwd: ssh ? ssh.remoteCwd : cwd`, so on SSH the folder on screen was thrown away and the terminal opened at the project root. That's a pre-existing hole in `addTerminal`'s `cwdOverride` contract rather than something this node introduced. On relay there's no `ssh` to rebind, so the peer's path went to a plain local terminal, and I've withheld the row there rather than fake it.

I'd written `sshBindingForCwd` for that. `nodeSshFor` already did the job, already had tests, and was already imported into `Canvas.tsx`. Deleted, call site switched to yours.

And the node ignored snap to grid: `createFilesNode` used `placeAt` while every peer uses `placeNode`, which snaps size as well as position. `placeNode` didn't exist when this branch was cut, and `workspace.placement.test.ts` only covered terminal and sticky, so it stayed green throughout. It covers this kind now.

## Testing, and what I couldn't test

Full suite against a clean `main` worktree: 17 failures each side, same test names, zero delta. Re-baselined after every rebase rather than reusing the numbers. `board-log-handlers` failed once mid-run and had me worried for a minute; it flakes on main too, 3 runs here and 5 there.

Mutation tested: the `case 'files'`, the parent comparison, the `cwdOverride` rebinding and the placement each turn a test red when reverted.

**Not verified, and I'd rather just say so:** relay tabs, a real SSH host, the Server Edition browser, and Windows. The QA pass drove the node in jsdom, but the fs api was simulated at the component boundary, so the `~` case above is proven against a fake host rather than a real one. Nothing visual has been looked at either, and `styles.css` gained about 170 lines I've never seen rendered.

## Three gaps left in on purpose

Flagging these because they'll otherwise cost someone an hour:

1. An unreachable filesystem still reads as "This folder is empty." The `unknown` verdict is right to refuse to claim a deletion, but that fallback sentence is too reassuring for what's probably the commonest cause of an empty remote listing. A fifth state felt like its own change.
2. On relay the promise can just stay pending, since `RpcClient` has no request timeout, so the node sits on "Loading…" with nothing retrying it.
3. `sshFs` is read off `window.nodeTerminal` rather than the session api. Inherited from `createEditorNode`/`createVideoNode`, but it bites harder here, because `sshFs` rides the git-shared `project.json`: clone the repo locally and you get a node flagged remote with a perfectly valid local cwd, listing nothing.

Happy to split any of this up differently, or to take the first of those in a follow-up if you'd rather it didn't sit.

eneskirca:
Re-verified `0980973b` in an isolated worktree, delta by delta: `npm run typecheck` clean, **full suite green — 733 files, 10043 passed / 40 skipped, 0 failed** (so your 17 environment failures are your sandbox), and the guards this branch is measured against (`nodeterm-events`, `styles.theme`, `addMenuSpec`, `reopenNode`, `workspace.placement`, `download`, `worktree`, `fs-atomic.guard`, both `no-electron`) 170/170. `main` has moved 69 commits since your push; a trial-merge is clean, and `styles.css` parses under esbuild's CSS loader both on the branch and merged — no unclosed rule this time.

All four asks and the nits are closed as described, and the calls you made on the way hold up:

- **The cwd-less row is consistent with #621**, which is the thing I most wanted to see: `disabled: !ctx.hasCwd` with `FILES_NO_CWD_HINT = NEW_FILE_NO_CWD_HINT` on the pane menu and the Dock, hidden in ⌘K exactly as "New file…" is — and `addFiles` still refuses with a sentence if it is reached some other way. Resolving that conflict mechanically would indeed have shipped a row contradicting the rule; thanks for reading the rule instead.
- **`classifyEmptyListing` is the right shape.** `unknown` for `~`, `.`/`..`, `.git` and an unanswerable parent, a case-folded match counting as present, and "missing" only when the parent listing is positive evidence — this is the "a failed read is never evidence of absence" rule applied where it was about to be broken.
- `canUseLocalShell` as the one predicate for every `shell.*` action, the relay withholding of "New terminal here" (a peer path must not reach a local terminal), `nodeSshFor` reused instead of a second binding helper, `placeNode` for snap parity, and the `case 'files'` in `reopenNode` are each correct. Your relay-`mkdir` correction is right too — that CLAUDE.md line described a file deleted in `fc1b2bc1`.

One follow-up worth logging, not blocking: `newEntryPath` still splits on `/` only, so on Windows a name like `..\x` is one segment and passes the traversal check. It is the pre-existing `explorerCreate` gap you documented, and it is a UX-correctness issue rather than a security one (the user is writing into a filesystem they already own; the relay fs is core-jailed) — but #293 just added a CONTRIBUTING rule that `\` counts as a separator wherever a persisted or typed path is split, so that function is the next site for it. Your three "left in on purpose" gaps read the same way to me: the too-reassuring "empty" sentence on an unreachable fs is the one I would take first if you do send a follow-up.

From my side this is ready; the merge call is @eneskirca's. Relay, a real SSH host, Server Edition and Windows stay device-unverified, as you said.

alblor:
Thanks for merging, and for the re-verification. Good to know the 17 failures were my sandbox; I had been reading them as environment noise from the delta against main, but it is better to know they are not there at all on a clean machine.

I have opened #662 with the two follow-ups, both of them one commit each.

The `newEntryPath` one is yours: the `..` guard splits on `[\\/]` now, and a name that is absolute in the Windows dialect is refused the way a leading `/` already was. I left the construction on `/` on purpose, since CONTRIBUTING is explicit that the separators are not interchangeable and a backslash is ordinary filename text on POSIX, so `weird\name.txt` is still one legal file. Guard on both dialects, construct in one.

The other one is the "empty" sentence you said you would take first. The parent listing turned out to be the evidence I was throwing away: it contains the directory we are standing in, so it cannot be childless, and an empty listing there means we could not see it. That is an `unreachable` verdict now with its own sentence, and `unknown` is left to the paths whose shape means no parent could ever answer, so an empty remote home does not start claiming the host is down.

No rush on it from my side.

PR #298, opened 2026-08-18
feat(usage): add token-count display mode for context meters

gor3a:
## Summary
- `settings.usagePercentMode` gains a third value, `'tokens'`, alongside `used`/`remaining`
- Context-window surfaces (canvas node pill, session sidebar chip) render it as a raw `48k/200k` fraction instead of a percentage, since "% left" was being read ambiguously as context-left vs some other left
- Provider quota pill (subscription limits, no token counts) falls back to `used` display in this mode
- Bar fill and severity color untouched — still keyed to used-percent

## Test plan
- [x] `npm run typecheck`
- [x] `npx vitest run src/renderer/lib/usageFormat.test.ts src/renderer/components`
- [ ] Manual: Settings → Usage → "Usage percentages" → Tokens, check canvas node pill + sidebar chip

### DIFF
--- a/src/main/notch-hud.ts
+++ b/src/main/notch-hud.ts
@@ -103,7 +103,7 @@ export interface NotchHudTunables {
   /** Expand the panel on hover (else click-only). */
   hoverExpand: boolean
   /** settings.usagePercentMode — how the rows' context percentages render ("42% used" / "58% left"). */
-  percentMode: 'used' | 'remaining'
+  percentMode: 'used' | 'remaining' | 'tokens'
 }
 
 /** Clamp a hand-editable width to something that can't push the capsule off the display. */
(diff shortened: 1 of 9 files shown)


eneskirca (approved):
Reviewed in depth — approving. Verified: defaults unchanged ('remaining' stays default, the rewritten ternaries are exactly equivalent for the two existing modes), 'tokens' round-trips through settings persistence, usedTokens/windowTokens are always present (no undefined/undefined pill risk), typecheck passes at the head and on a clean trial merge with current main, and bar fill / severity colors stay keyed to used-percent in every mode.

Non-blocking nits, ideally a small follow-up:
1. No tests for the new logic: `formatTokensShort` / `contextPillText` / the 'tokens'→'used' fold are uncovered, and the usageFormat.test.ts invariant loop ("bar agrees with the number next to it") still iterates only ['used','remaining'].
2. The notch HUD only carries `contextPercent`, so in Tokens mode it silently shows "42% used" while the canvas pill shows "48k/200k" — worth a sentence in the settings description.
3. `ROWS.percentMode.keywords` in UsageSection doesn't include "tokens", so settings search misses the row that now owns that option.
4. Stale two-mode doc comments in notch-hud.ts / preload hud.ts next to the widened type.

Thanks!

PR #302, opened 2026-08-18
Windows packaging groundwork: electron-builder NSIS target, .ico pipeline, fresh-machine bootstrap (phase 4 of the Windows-support extraction from #276)

eneskirca:
**Phase 4 of the Windows-support extraction from #276** (credit: @cafepromenade — the fork's packaging work was the reference and the icon approach is adopted from it). Stacks on #300 (`fix/atomic-writes-windows`): this branch is based on it, so its two commits appear here until #300 merges.

## ⚠️ Groundwork, not a shippable app

This adds the Windows **packaging** substrate only. A Windows build produced from this branch is not usable — terminals need the session-host phase (ConPTY-side work from #276) which has not been ported yet. Nothing here publishes, signs, or feeds the updater.

## Decision: electron-builder NSIS, not Squirrel.Windows

The fork stayed on electron-builder but switched the Windows target to **Squirrel.Windows** (`electron-builder-squirrel-windows` dependency + an ~800-line `windows-installer.mjs` wrapper + its own release/update-feed tooling). We deliberately go with **NSIS** instead:

- Our release pipeline is electron-builder end-to-end (`.github/workflows/release.yml`, mac dmg/zip + the self-hosted generic update feed at `nodeterm.dev/updates`). NSIS is electron-builder's native Windows target — zero extra dependencies, one config block, and it is what electron-updater's **generic provider** expects on Windows (`latest.yml` + NSIS differential updates), so the future auto-update leg slots into the feed we already run.
- Squirrel.Windows would bring a second installer framework, an extra build dependency, and the fork's wrapper existed largely to compensate for Squirrel-specific issues (nuspec source-SHA identity, ExecutionStub PE re-editing via `resedit`). No concrete blocker for NSIS was found in the fork's history — their Squirrel choice is coupled to their replaced release pipeline (release-on-every-push, dim-sum code names, own feed), which is explicitly out of scope here.

## What's in

- **`build.win` + `build.nsis`** in `package.json`: NSIS installer + zip, x64, `build/icon.ico`, artifact `nodeterm-Setup-<version>.exe`, one-click per-user install (Electron convention). App identity stays the existing global `appId com.nodeterm.app` / `productName nodeterm` → `nodeterm.exe`.
- **`npm run dist:win`**: `make-icon` → `electron-vite build` → `electron-builder --win --x64 --publish never`. Unsigned by design (no cert env ⇒ electron-builder skips signing).
- **`scripts/make-icon.mjs`** now also emits a multi-resolution `build/icon.ico` (16–256px, PNG-compressed frames in a hand-rolled ICO container — Vista+ accepts PNG frames directly, so no new dependency; NSIS requires the 256px frame). Adopted from #276. Gitignored like `build/icon.png` — every dist script regenerates it, unlike the fork we don't need a committed derivative (that was a Squirrel nuspec-SHA requirement).
- **`bootstrap-windows.bat`**: fresh-Windows-machine bootstrap — verifies Node ≥ 20, VS Build Tools C++ workload (via vswhere), Python 3, refuses to run elevated, then `npm ci` (postinstall patches node-pty + rebuilds native modules against Electron's ABI). It never installs machine-wide tools itself; each missing tool prints the exact `winget` command. A deliberate ~100-line distillation of the fork's 1,500-line bat suite (elevation guard and the `NoDefaultCurrentDirectoryInExePath` invocation note kept; their auto-installing toolchain machinery not adopted — installs that touch Program Files are the human's call).
- **`.github/workflows/win-package-smoke.yml`**: `workflow_dispatch`-only job on `windows-latest` that runs `npm run dist:win` and asserts the installer + zip exist. **Build only, never publishes**; the runner image already carries the toolchain (the Phase-1 `quality-windows` CI job proves `npm ci` green there).
- CLAUDE.md packaging section updated with the decisions above.

## Deliberately NOT taken from #276

- Squirrel target + `electron-builder-squirrel-windows` + `windows-installer.mjs` (see decision above).
- Their release tooling (release-on-push, `release-notes.mjs`/`release-assets.mjs`, code names, own update feed) — out of scope.
- `before-pack.cjs`/`after-pack.cjs` — those stage/verify their codex-relay session-host artifact; they belong to the session-host phase, not packaging.
- `windows-pe-identity.mjs` (`resedit` afterSign hook fixing the empty `OriginalFilename` electron-builder writes) — load-bearing only for Squirrel's ExecutionStub, cosmetic for NSIS. Noted as follow-up polish.
- Windows **auto-update wiring** — follow-up, blocked on code signing (an unsigned auto-update downgrades trust): electron-updater NSIS leg + `latest.yml` on the nodeterm.dev feed + a release.yml Windows job.

## Verification

- `npm run typecheck` green.
- Full `vitest` suite green.
- `npm run make-icon` run locally; the emitted ICO's container validated frame-by-frame (7 PNG frames, correct ICONDIR/ICONDIRENTRY offsets).
- Workflow YAML parse-validated. The packaging smoke job itself is Windows-only and is meant to be dispatched once this merges — that dispatch is the real end-to-end proof.


HopeCollector:
I reproduced a false-positive in `bootstrap-windows.bat` on a native Windows machine.

`vswhere` returns exit code 0 when the query succeeds but matches no Visual Studio installation. The current bootstrap discards stdout and checks only `ERRORLEVEL`, so it prints `[OK] Visual Studio C++ build tools`; `node-gyp` then fails later with `Could not find any Visual Studio installation to use`.

I prepared a one-commit fix stacked directly on this PR's head (`589fff5a`):

- commit: https://github.com/HopeCollector/nodeterm/commit/ca95d210
- distinguish query failure from a successful empty result
- require a non-empty, existing installation path before printing `[OK]`
- add real `cmd.exe` behavior tests for empty output, nonzero exit, and a valid path containing spaces
- wire the test into `quality-windows`

Validation: `npm run typecheck` passes; the Windows-targeted CI set passes 44/44; the new guard was mutation-checked by inverting it and confirming the test failed; the full local bootstrap now stops at the VS probe before Python/`npm ci`.

Would you prefer to cherry-pick this commit, or should I open a small stacked PR targeting `feat/windows-packaging`?

PR #319, opened 2026-08-19
fix(accounts): one binding rule behind data.accountId and the node color

Senzaiken:
Review follow-up to **nit 1** on #283 — the claude-only account gate lived as a ternary inside `createAgentNode`, so it was enforced on the canvas and nowhere else. **Stacked on #283**: until that merges this PR carries its commits too, and the change here is the single commit `fix(accounts): one binding rule behind data.accountId and the node color`. Nit 2 (the `typeof` guard on `color`) is pushed to #283 itself, being a one-liner in code that PR introduces.

`shared/agents/account-binding.ts` → **`boundAccountId(accountId, agentId)`**, asked by both node-mint sites: `createAgentNode` (canvas) and `appendProjectNode` (the phone's `projects.registerNode`). In the registrar it gates the `accountId` stamp **and** the account color off one local, so the two cannot disagree.

## Why not the shape the review asked for

The nit proposed encoding the rule in `accountNodeColor`. Rejected, for two reasons:

- **It fixes the color and leaves the actual breach.** `appendProjectNode` stamps `node.accountId = input.accountId` with **no agent gate at all** — that predates #283. The binding is not cosmetic: it becomes a `CLAUDE_CONFIG_DIR` path segment and scopes every account-aware reader (transcript, context meter, find-bar index). Gating only the color leaves a codex node carrying a Claude account it is not painted for.
- **It would invert #283's own design.** That PR's argument is that the color reads off the SAME `boundAccountId` that stamps `data.accountId`, "so the color and the binding cannot drift". Gate the color while the stamp stays ungated and they drift by construction, in the mirror image of the bug fixed in 9cee281e.

So the rule moved up one level, to what it is actually about: whether the node is account-bound at all.

`main` keeps resolving the color from the RAW account id and lets the registrar refuse it. Deliberate — the alternative is re-deriving the gate at the call site, which is the second copy that drifts. A discarded color costs nothing.

## The judgement call: an unstated agent keeps its binding

`boundAccountId` refuses a **known** other agent, not an unknown one — `agentId: undefined` keeps the binding. The strict reading (`agentId === 'claude'`) is the obvious one and is wrong here, because the failure directions are not symmetric:

- **Drop a real Claude binding** → the wrong-identity bug `accountId` exists to prevent: an off-LAN session started under account X registers as the system account, every scoped reader resolves against the wrong root, and a cold restore resumes the conversation as the wrong identity. Silent, and it corrupts state.
- **Keep a stray binding on an agent-less node** → a `CLAUDE_CONFIG_DIR` on a plain shell that nothing reads.

And the premise for the strict form is not established: `handleRegisterNode` reads `agentId` and `accountId` as two independent optional wire fields, and whether the phone always sends the first alongside the second is an **open question** — `docs/ios-protocol-migration.md` §6, "confirm which the phone controls vs. inherits". `nodeterm-ios` is not readable from here, so this is designed for the uncertainty rather than assuming it away.

A **custom** agent never binds either, even one based on claude — `createAgentNode`'s existing comment already holds that a custom agent inheriting the claude harness is still its own agent. A non-string `agentId` reads as "no agent stated", the same way the `agentConfig` lookup on the next line has always treated it: garbage must not be mistaken for a known *other* agent and cost a node its binding.

`createAgentNode` takes `agentId` as a required argument, so the canvas path is **bit-for-bit** what it was.

## Three surfaces

- **Desktop** — full; both mint sites go through the helper.
- **Server Edition** — canvas path identical (pure renderer). The phone-registration leg is not served there at all: `registerNode` / `appendRemoteNode` have no `src/server` or bridge caller, the standing relay host lives in `src/main`. Nothing to degrade.
- **Mobile** — no phone-side change; this is host-side validation of what the phone already sends. **@eneskirca** two to carry over: (a) if `nodeterm-ios` ever registers a non-Claude agent with an `accountId`, that binding is now dropped by design; (b) the §6 question above — confirm the phone always sends `agentId` and the gate tightens to the strict form, deleting the special case.

## Mutation checks (4 introduced / 4 caught)

- `boundAccountId` stops gating (always binds) → **5/97 red**, including the pre-existing `accountId on Claude node factories` tests — which is what proves the refactor is genuinely wired to the shared rule, not just sitting beside it.
- The `accountId` stamp ignores the gate (`if (input.accountId)`) → **2/19 red**.
- The color ignores the gate (`accountColor ?? agent?.color`) → **1/19 red**.
- `boundAccountId` uses the strict `agentId === 'claude'` → **1/5 red** — the unstated-agent case discriminates the design decision above rather than restating the code.

## Gates

`npm run typecheck` clean. Full `npm test`: **6986 passed / 24 skipped**, plus the 3 pre-existing `local-send-keys.realtmux` failures below.

8 new tests, written first and each watched fail for the right reason — `boundAccountId is not a function`, then `expected 'acct1' to be undefined`:

- `src/shared/agents/account-binding.test.ts` (5) — claude binds; codex/gemini do not; a custom agent does not; an unstated agent does; no account id.
- `src/core/project-node-append.test.ts` (+3, 19 total) — a non-Claude registration takes neither the binding nor the account color; a custom agent takes neither; an agent-less registration keeps both.

## Not verified

- **The phone path end-to-end** — same limit as #283, it needs the iOS companion. Covered by unit tests on `appendProjectNode`, not on a device. That a non-Claude registration carrying an `accountId` is something `nodeterm-ios` can actually produce is itself unconfirmed; this closes the hole host-side either way.
- **`local-send-keys.realtmux` (3 tests) fails in my environment**, on this branch and on an untouched `feat/account-default-color` (checked by stashing): the macOS temp root makes the tmux socket path 106 characters against a 103-character `sun_path`, so it never binds — `error connecting to …: File name too long`. Unrelated to this change, and now **fixed separately in #320**; CI is the judge here either way.
- **No manual run.** Two pure functions with no UI surface; exercised through the tests and the mutation table, not by driving the app.

## Docs

`CLAUDE.md` — the binding rule and the asymmetry as their own bullet under managed accounts. `CONTRIBUTING.md` — the generalisable house rule this is an instance of: *a rule enforced at one mint site is enforced nowhere.*

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1524,6 +1524,21 @@ else, and its context links must keep classifying across restarts).
     the project-default account), and validation runs against `accountsForProject`, not the raw
     list, so a **pending** account or one **pinned to another machine's host** is never stamped
     onto a node it cannot run on (both used to reach the missing-dir fallback at spawn).
+  - **`boundAccountId(accountId, agentId)` (`shared/agents/account-binding.ts`) is the ONE rule for
+    whether a node is account-bound at all**, and it feeds `data.accountId` *and* the account color
+    from a single decision — split them and a node carries an account it is not painted for, or is
+    painted for one it does not carry. Two surfaces mint nodes and both ask it: `createAgentNode`
+    (canvas) and `appendProjectNode` (the phone's `projects.registerNode`, which used to write
+    whatever the wire sent, so a gemini node could come back bound to a Claude account). Managed
+    accounts belong to the builtin **claude and codex** (S6); a **known** other agent — builtin or
+    custom, since a custom agent inheriting one of those harnesses is still its own agent — never
+    binds. **An UNSTATED agent keeps its binding** — the asymmetry is deliberate: the phone chooses
+    `agentId` and `accountId` independently and is not known to always send the first
+    (docs/ios-protocol-migration.md §6), dropping a real binding is the wrong-identity bug the
+    field exists to prevent, while a stray one on an agent-less node only sets a config-home
+    variable nothing reads. On the canvas `agentId` is always stated, so that path is bit-for-bit
+    what it was. `main` resolves the color off the RAW id and lets the registrar refuse it, rather
+    than re-deriving the gate at the call site.
   - **Account default node color (`ClaudeAccount.color` / `CodexAccount.color`, optional)** — a
     per-account default node color (Settings → Accounts) that beats the agent's own brand color in
     `createAgentNode`, so a second login is recognizable on the canvas. Read off the SAME
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -162,6 +162,15 @@ feature, and the boundary tests can only tell you an import is wrong, never that
 The same applies to any hook-server signature change; this repo has shipped one to a single shell
 three times.
 
+**A rule enforced at one mint site is enforced nowhere.** Nodes are created on two surfaces — the
+canvas (`createAgentNode`) and the phone's `projects.registerNode` (`appendProjectNode`) — and a
+constraint spelled out inline at one of them silently does not exist at the other. "Which agents
+bind a managed account" lived as a ternary in the renderer while the phone leg wrote whatever the
+wire sent.
+Put the rule in one predicate under `src/shared` and have every mint site ask it, and derive the
+things that follow from it (a node's color, say) from that same call rather than re-deriving the
+condition per caller.
+
 **Do not take scrolling away from tmux.** It owns the mouse, the scrollback and the alternate
 screen. A previous design moved that into the emulator and failed structurally; `CLAUDE.md` explains
 why in detail.
--- a/src/core/project-node-append.test.ts
+++ b/src/core/project-node-append.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect } from 'vitest'
 import { appendProjectNode, removeProjectNode } from './project-node-append'
+import { agentConfig } from '../shared/agents/config'
 
 const NOW = new Date('2026-07-16T10:00:00.000Z')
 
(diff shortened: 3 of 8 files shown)


eneskirca:
Reviewed in depth. **Your counter-argument is correct and my nit was the weaker fix — I concede it.** All three legs check out against the code:

- `project-node-append.ts:121` really does `if (typeof input.accountId === 'string') node.accountId = input.accountId` with no agent condition anywhere in the function.
- The binding really is load-bearing: `claude-accounts-core.ts:25` turns it into `CLAUDE_CONFIG_DIR`, and `pty-manager.ts:2189` sets that env **and deletes `AUTH_ENV_STRIP`** keyed on `accountId` alone, no `agentId` in sight.
- So gating `accountNodeColor` would have fixed the cosmetic half and left the state corruption, and it would have inverted #283's own "the color reads off the same decision that stamps the binding" design. One shared predicate asked by every mint site is the better shape.

Your tests prove the wiring too — under the "always binds" mutation a **pre-existing** `workspace.test.ts` case goes red, which is exactly the evidence you claimed it was. Measured counts came out slightly *higher* than you reported (6 red, not 5), i.e. conservative rather than inflated.

**What blocks it is timing, not reasoning.** `main` merged the Codex managed-accounts series (#334–#342) hours after you branched, and accounts are no longer Claude-only:

```ts
// origin/main:src/renderer/state/workspace.ts:630
...(accountId && (agentId === 'claude' || agentId === 'codex') ? { accountId } : {}),
```

So `boundAccountId`'s `agentId !== 'claude'` now **drops the Codex per-node binding**. A pre-existing main test fails on the merged tree:

```
FAIL workspace.test.ts:559 > stamps accountId onto a Codex agent node (S6 per-node account picker)
AssertionError: expected undefined to be 'a1'
```

Failure scenario: user picks Codex account B → binding dropped → `resolveCodexSessionScope(dir, undefined)` resolves to the **system** `~/.codex` → the node silently runs as the wrong Codex login. That's the wrong-identity class your own body argues against.

Two more things to re-derive:

1. **The agent-less asymmetry no longer holds.** `needsCodexAccountScope(agentId?, accountId?)` is `agentId === 'codex' || !!accountId`, so *any* node with an accountId enters the fail-closed Codex path. I probed it: an agent-less node carrying an account id returns `{unavailable:'codex-account'}` and **spawns zero processes** — it refuses to open, not "sets a CLAUDE_CONFIG_DIR nothing reads". Both directions now cost something.
2. **Citation:** the "open question" text is `docs/ios-protocol-migration.md` **§8 item 6**, not §6, and it concerns `PtyCreateOptions` rather than `projects.registerNode`. The premise still holds by better evidence — `host-service.ts:453-458` reads `agentId` and `accountId` as two fully independent optional `str()` fields — so I'd just cite that instead.

Also worth noting in the body: `ClaudeAccount.color` has no Codex counterpart, so #283's headline promise doesn't reach half the account-bound nodes on current main.

**Bonus:** chasing this turned up two live bugs on `main` that are nothing to do with your PR — #345 (the Claude `claude /login` terminal is refused by that same over-broad `|| !!accountId` gate) and #346 (`nodeterm:add-codex-account-login` is dispatched with no listener). #345 is prerequisite reading for getting item 1 right.

Merge order: #283 first (it also needs a rebase — conflicts in `workspace.ts` and `AccountsSection.tsx`, and its `AccountsSection.color.test.tsx` mock needs `codexAccounts.systemIdentity` + `ssh.list` stubs for main's new deps), then this rebased on top. The architecture stays; only the predicate's contents change.

eneskirca:
Heads-up — #283 ahead of you is now merge-ready, so once it lands this branch will want a rebase onto main; ping us when it's up and we'll re-review promptly.

Senzaiken:
#283 landed, so this is now a real one-commit branch: **rebased onto `main`** (19da3a02), no stack, `MERGEABLE`.

**It was not a mechanical rebase — the rule this PR extracts had to change.** When I wrote it, managed accounts were Claude-only. The Codex series (#334–#342) made them claude|codex, and #283 merged with that rule. Replaying `boundAccountId` as `agentId !== 'claude' → undefined` would have reverted Codex per-node binding at *both* mint sites — the exact trap you caught on #283, this time hidden inside a helper whose whole point is to be the single definition. So:

- `boundAccountId` refuses a **known agent that takes no managed account** (gemini, grok, any custom agent — including one inheriting claude or codex), binds claude and codex, and still keeps the binding for an **unstated** agent. The asymmetry argument is unchanged and is now the only interesting thing in the helper.
- `createAgentNode` drops its inline ternary and asks the helper. Canvas path stays bit-for-bit: `agentId` is always stated there, so every stated value resolves exactly as main's ternary did — `main`'s own `accountId on Claude node factories` suite passes untouched.
- The color stays `agentAccountColor(agentId, bound, …)` from #283 — it already routes claude→`claudeAccounts`, codex→`codexAccounts`. The binding feeds it, so the two still cannot drift.
- `appendProjectNode` gates the stamp *and* the handed-in color off that one call. This is the hole you filed as a follow-up on #283 (`project-node-append.ts:128` stamping for any `agentId`), closed here rather than separately.

Tests updated to match, not just re-pointed: the old "codex takes neither binding nor color" case is now **gemini** (an agent that genuinely takes no account), plus a new case asserting a Codex registration *does* bind, color and all. `account-binding.test.ts` likewise gained the codex-binds case.

`npm run typecheck` clean. Full suite **9657 pass / 11 fail** — the same 11 that fail on a clean `main` in this environment (`hello-probe`, `remote-atomic-write`, `session-host-client`, `control-master.realsh`, `pane-cwd.realtmux`), all needing a real ssh/tmux this sandbox cannot give them; verified against `origin/main` in a detached worktree.

Unverified, unchanged from before: the phone path end-to-end (needs the iOS companion), and whether `nodeterm-ios` can even produce a non-binding agent carrying an `accountId` — §6 of the protocol migration doc is still the open question the unstated-agent case is designed around.

eneskirca:
Re-verified the rebased `19da3a02` in an isolated worktree on `main` (8be08bc1): typecheck clean, full `npx vitest run` **9674 passed / 40 skipped / 0 failed**. The `stamps accountId onto a Codex agent node` case that failed on the previous revision passes now, and the "always binds" mutation still reddens `workspace.test.ts`. Merge-tree against `main` and against #612 (the other open accounts PR): no conflicts. CI is green.

The re-derived asymmetry holds on current main: since #345 `needsCodexAccountScope` asks an `isCodexAccount` resolver, so an agent-less node carrying a **Claude** account id no longer enters the fail-closed Codex path — "sets a `CLAUDE_CONFIG_DIR` nothing reads" is accurate again.

One leftover from the first review, non-blocking: the citation is still `docs/ios-protocol-migration.md §6` in `account-binding.ts`, its test, `project-node-append.test.ts` and the CLAUDE.md bullet. §6 is "Desktop-side lockstep work"; the open question is **§8 item 6**, and the stronger evidence is `host-service.ts` reading `agentId` / `accountId` as independent optional fields. Fine as a follow-up commit or fixed on merge.

Merge-ready from my side; the decision is @eneskirca's.

PR #320, opened 2026-08-19
fix(test): give the realtmux suites a socket path that fits sun_path

Senzaiken:
Every test in `local-send-keys.realtmux.test.ts` failed on macOS with `error connecting to …: File name too long` — which reads like a broken test and is really a **path-length limit**. Three tests, dead on every contributor's Mac, silently carried as "environmental" in #283's *Not verified* section. They pass now.

A unix socket is bound into a fixed `sockaddr_un.sun_path`, declared `char sun_path[104]` on macOS — 103 usable characters. The suite put its socket under `os.tmpdir()`, and **tmux resolves that symlink before it binds**:

```
/private/var/folders/_f/…/T/ntsendkeys-XXXXXX/tmux-501/nt-sendkeys-test-38663   = 106   OVER by 3
/private/var/folders/_f/…/T/ntpb-XXXXXX/tmux-501/nt-paste-test-38663            =  97   ok, by 6
```

The 8 characters that overflow are the `/private` prefix, which is why measuring the unresolved `/var/folders/…` path says 98 and looks fine. The sibling suite passes with **six characters of margin** — so shaving three characters off a name fixes this only until someone adds a word to one.

## What this adds

`src/core/tmux-test-socket.ts` makes the choice instead of hard-coding a dir: resolve the candidate temp roots, keep the first whose *eventual* socket path fits, and **refuse rather than bind something truncated**. On macOS that falls back from `/private/var/folders/…` (56 chars) to `/private/tmp` (12). It predicts the six characters `mkdtemp` appends, and re-checks the directory it actually got — so a miss fails naming the number, not the errno.

Production is not exposed to any of this: `PtyManager` binds `-L node-terminal` under the default `/tmp/tmux-<uid>/`, short on every platform we ship to. The module is test-only and says so in its header.

## It also closes a socket leak in the third suite

`session-env.realtmux.test.ts` never set `TMUX_TMPDIR` at all, so its socket went to the **shared** `/tmp/tmux-<uid>/`. `kill-server` stops the server but never unlinks the socket file — the exact hazard `local-send-keys`'s own comment describes and defends against, on the one file that did not apply the discipline. Measured before this change: **14 stale `nt-envtest-*` sockets** on one dev machine, one per run since August 17.

Routing it through the same helper makes `afterAll`'s `rm -rf` the one cleanup path. That includes the two tmux invocations wrapped in `/bin/sh -c`, which inherit their env — miss those and the file runs **two servers for one socket name**, the private one gets killed and the shared one keeps leaking. Verified by counting: 16 sockets before four runs, 16 after.

## Mutation checks (3 introduced / 3 caught)

- `MKDTEMP_SUFFIX_LEN` → 0 (ignore the six characters mkdtemp appends) → **2/5 red**.
- `SUN_PATH_MAX` → 107 (Linux's limit instead of the smaller macOS one) → **1/5 red**, the regression case.
- `pickTmuxTmpdirBase` returns `bases[0]` unconditionally → **3/5 red**.

## Gates

`npm run typecheck` clean. `local-send-keys.realtmux` **3 failed → 3 passed**, including inside the full suite. Full `npm test`: **7275 passed / 24 skipped**, with the one flake below.

New: `src/core/tmux-test-socket.test.ts` (5), including the real `/private/var/folders/…` shape and this suite's own prefix and socket name as the regression fixture.

## The one thing I did NOT fix, and why

`tmux-paste.realtmux.test.ts > CONTROL: the same attack DOES run when the payload is not sanitized` is a **pre-existing low-rate flake**. #283's author reported this same file failing run-to-run before any of this. Measured here:

| condition | branch | main |
|---|---|---|
| that test alone, ×10 | 0 failures | — |
| whole file, ×8 | 0 | 0/8 |
| 3-file combo, ×6 | 0 | 0/6 |
| full suite | 2 failures in ~4 runs | 0/4 |

It only appears under full-suite parallelism. **This branch plausibly makes it more likely to be seen**, and I want that stated plainly rather than buried: on `main` those three `local-send-keys` tests abort instantly instead of driving a real bash pane, so `main` is not a fair control — this branch genuinely adds concurrent tmux work to that window.

I did not patch it, because **I could not establish its mechanism**. My hypothesis was that `bashDrain`'s opening `C-c` flushes the pasted keys as typeahead before bash runs them. I tested it directly — 12 paste-then-`C-c` cycles against a real tmux and a real bash — and the marker was created **12/12 with the `C-c` and 12/12 without**. The theory is disproved. Changing the timing of a security-control assertion on a theory that just failed its own experiment is worse than reporting it, so this is a flagged follow-up, not a silent tweak.

## Not verified

- **Linux and CI.** Everything here was measured on macOS, which is where the bug is; `sun_path` is 107 there and `os.tmpdir()` is already `/tmp`, so the helper should be a no-op — reasoned, not run.
- **The flake's real cause**, as above.
- The 14 pre-existing stale sockets in `/tmp/tmux-501/` on my machine are left alone — this stops new ones, it does not clean up old ones, and deleting files outside the repo is not this PR's business.


eneskirca (approved):
Reviewed in depth — approving. Every number in the body checked out when measured independently (the `/private` prefix really is the 8 characters that overflow; 106 vs the 103 usable bytes of macOS's `sun_path[104]`), all three claimed mutations reproduce at exactly the stated red counts, the trial merge onto current main is clean, both typechecks pass, and all 2949 `src/core` tests are green.

Confirmed test-only: bundled both real entry points with esbuild metafiles — `tmux-test-socket` appears in neither, and `electron-builder` ships only `out/**`, so it cannot reach an artifact. `mkdtemp` under `/private/tmp` is not hijackable either (mkdir is atomic, EEXIST on a pre-existing dir or symlink throws rather than borrowing it).

**The leak half is the stronger half, and it reproduces on Linux too** — which your body listed as unverified. On our own dev box `/tmp/tmux-0/` held **220 stale `nt-envtest-*` sockets**, in the same directory as the live production `node-terminal` socket. One run on main: 220 → 221. Four runs on this branch: 223 → 223, with no leftover temp dirs. Good catch.

Non-blocking follow-ups:
1. The `makeTmuxTmpdir` doc says macOS falls back to `/private/tmp`. Measured with the new `nts-` prefix the path is 99/103 — it fits, so the picker keeps `/private/var/folders/…` and the fallback never fires. The prefix rename is what actually fixes the suite; the fallback is the safety net behind it. Worth one sentence so a future reader doesn't reason about a 55-char path that will never exist.
2. `realpathOrSelf` swallows the error and returns the unresolved path — in both the predict and the verify step, so a realpath failure makes them agree with each other and the original 8-char undercount returns as the same obscure errno. Letting it throw (or dropping that base) would keep 'a failed read is never evidence of absence'.
3. The regression fixture still pins `'ntsendkeys-'` while the suite now uses `'nts-'`, so nothing goes red if someone lengthens `SOCKET` back toward the limit (99/103 is 4 chars of margin — less than the 6 you flag in the sibling suite).
4. A test-only module in `src/core` has nothing keeping production out of it: a future import would compile, pass `no-electron.test.ts`, and ship. Either move it under `test/` or add one assertion that no non-test file imports it.

Thanks — this is exactly the kind of 'reads like a broken test, is really a platform limit' diagnosis that saves everyone else the afternoon.

PR #321, opened 2026-08-19
feat(canvas): per-node toggle to hide subagent/loop fan-out cards, plus tidy

gor3a:
## Summary
- Persisted `data.hideFanout` flag on agent nodes (`CanvasNodeState`/`NodeData`), threaded through both serializers.
- New header eye toggle on subagent/loop-capable agent nodes hides that node's ephemeral fan-out cards + edges. Render-only: subagent/loop tracking, `agentStatus.loop`, and the hibernation guard's evidence keep working underneath.
- New header "tidy" button (shown when a node has 2+ subagent cards) resets dragged position/size/expanded overrides back to the default packed grid. Loop cards excluded — one per parent, nothing to pack.
- Both buttons registered as user-hideable in Settings → Appearance (`HIDEABLE_HEADER_BUTTONS`).

## Test plan
- [x] `npm run typecheck`
- [x] `npx vitest run src/renderer/state/agentNodes.test.ts src/renderer/lib/ui-visibility.test.ts` (new/updated tests pass)
- [x] Full `npx vitest run` — 32 pre-existing failures, none touching changed files (node-pty rebuild patch, real-tmux socket path length, board-log broadcast — all environment-dependent, unrelated)

### DIFF
--- a/src/renderer/canvas/Canvas.tsx
+++ b/src/renderer/canvas/Canvas.tsx
@@ -1315,7 +1315,7 @@ export function Canvas() {
       // lives here, in the render layer, and nowhere else.
       if (!st.loop || st.loop.dismissed) continue
       const parent = nodes.find((n) => n.id === pid)
-      if (!parent) continue
+      if (!parent || parent.data.hideFanout) continue
       const ph = parent.measured?.height ?? (parent.height as number) ?? 400
       const accent = agentConfig((parent.data.agentId as string) ?? 'claude')?.color ?? '#d97757'
       const lid = `loop-${pid}`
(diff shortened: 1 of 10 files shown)


eneskirca (approved):
Reviewed — approving. Verified: the render-layer guard skips both cards and their edges in the same loop pass (tracking stores, `agentStatus.loop`, and the hibernation guard's evidence untouched, as claimed); `hideFanout` is threaded through both serializers and persists (no field whitelist in core strips it); `tidyFanout` correctly excludes the loop card and the doc comment about the localStorage mirror hazard is exactly right; the hover rgba matches existing house style in styles.css; tests are behavioral (override-drop per parent, no-op identity). CI green.

One non-blocking perf nit: the `fanoutCount` selector runs `Object.values(s.byId).filter(...)` on every agentNodes store update (including per-chunk `appendActivity`) for every fanout-capable node. It returns a primitive so re-renders are cheap, but with many streaming subagents the selector work itself is O(nodes×subagents); a per-parent count map in the store would make it O(1) if this ever shows up in a profile.

PR #327, opened 2026-08-19
feat(explorer): pin the file tree so it stays open on the canvas

### LINKED ISSUE #297: featReq: Add ability to pin file explorer in the view
Would it be much to add for a pin icon here to allow this panel to always appear 

Image

dev-krish-xyz:
## What

Explorer can be pinned so it stays open while you work on the canvas.

It was a click-outside overlay, so opening a file or touching the canvas dismissed the tree ([#297](https://github.com/eneskirca/nodeterm/issues/297)). The header pin docks it like the sessions sidebar: no scrim, pointer-events pass through to the canvas, × hides without clearing the pin, and the preference persists in `localStorage` (`nodeterm.explorerPinned`).

Default **off**. Explorer is a modal today, and flipping the default would dock it on every existing user's next launch. Unpinning a visible drawer keeps it as the modal so the tree does not vanish (there is no hover-peek). Source Control is unchanged.

## Why the overlay change is two halves

A pinned drawer that still sat under a full-screen overlay would **steal canvas clicks** even if the click handler no-op'd. So:

1. Overlay click-outside closes the **modal only** (`explorerOverlayClickCloses`).
2. A pinned overlay is `pointer-events: none`; only the drawer captures events.

Both are load-bearing. The docked card sits below the controls cluster (`top: 96px`) so Settings / SCM / Explorer stay reachable, at z 26 so it remains visible on the kanban board.

## Surfaces

- **Desktop** — full.
- **Server Edition** — full (renderer `localStorage`; no new IPC).
- **Mobile companion** — N/A. There is no explorer there. Flagging for @eneskirca rather than assuming.

Kanban: the pinned tree is a panel, not a node action, so there is nothing to mirror on cards. It stays visible on the board with the controls cluster.

## Testing

`npm test -- src/renderer/lib/explorerPin.test.ts` — 12/12 pass. The four load-bearing guards were mutation-tested (default-on parse, overlay always closes, unpin dropping a launch-pinned drawer, dismissed pin still showing) and each turned a test red.

`npm run typecheck` passed.

**Not verified:** clicking through it in a running app.

Full `npm test` on this machine still has pre-existing failures in `pty-shadow` / `pty-background-write` / `pty-idle-reap` (tmux control-client / unix socket path) and a few jsdom `localStorage` store loads. None of those files are in this change; they fail the same way on a clean tree here.

Closes #297.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1693,7 +1693,13 @@ the Settings section and ShortcutsPanel start disagreeing about what a chord mea
   opens an editor node). These use `mkdir` + `exists` added to `FsApi`/`SshFsApi` across
   desktop/server/SSH (`core/fs-ops.ts`, `main/ssh-fs.ts`; relay remote-fs degrades to `false`).
   Expanded dirs **persist per project** across drawer close + app restart (`state/explorer.ts`
-  zustand store, localStorage `nodeterm.explorerExpanded`).
+  zustand store, localStorage `nodeterm.explorerExpanded`). The header pin docks it like the
+  sessions sidebar (`lib/explorerPin.ts`, `nodeterm.explorerPinned`, default off): overlay
+  click-outside closes the modal only, and a pinned overlay is `pointer-events: none` so it
+  cannot steal canvas clicks. × is a transient hide and does not clear the pin. Pinned z-index
+  is 26 so the tree stays visible on the kanban board with the controls cluster. Desktop +
+  Server Edition (personal `localStorage`). Mobile companion: N/A — no explorer there. Source
+  Control stays a modal.
 - **Source Control** (`main/git-service.ts` system `git` + `gh`, `SourceControlPanel.tsx`,
   ⎇): file-level **stage/unstage** (+/−), **discard**, click a file → **diff node**,
   **branch switch/create**, commit (message box at top) + push / sync / publish, **gh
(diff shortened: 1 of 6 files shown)


eneskirca (approved):
Reviewed in depth (trial-merged onto current main in a scratch worktree: clean merge, typecheck green, 1384 renderer tests across 122 files pass, all five styles.css guard suites included) — approving. Verified both load-bearing halves: a pinned overlay is `pointer-events: none` with no click handler (canvas clicks pass through; the context menu is body-portaled so it's unaffected), and the unpinned modal keeps the scrim + click-to-close exactly as before. A pinned-but-dismissed explorer unmounts entirely, so nothing invisible can steal clicks. localStorage is the right store — `nodeterm.sessionsPinned` set the precedent and copy-feedback.ts documents the family. z/top math checks out (controls cluster 54+34+8 → 96px, z 26 over kanban's 25, under the tab bar's 30), theme-safe (var tokens + --shadow-k pattern), and ⌘⇧E toggling the transient `dismissed` without clearing the pin matches the sessions-sidebar semantics.

Non-blocking nits: `ExplorerShowAction`'s `'open'` variant is exported+tested but unused; the toolbar Explorer button silently became a toggle (necessary for a docked panel, but worth a line in the body); `explorerOverlayClickCloses` is a one-line wrapper with a near-tautological test.

Nice first contribution — the pure state machine + mutation-tested guards made this easy to verify. Merging; closes #297.

PR #353, opened 2026-08-20
test(kanban): stub child panels and pin CardModal stopPropagation and dialog stack guards

Abhirup0:
### Summary
Follow-up to [#288](https://github.com/eneskirca/nodeterm/pull/288) addressing the non-blocking suggestions from the review:

1. **Stub child panels for test isolation**:
   - Explicitly stub `BoardLogPanel`, `CardMetaBar`, and `ContextMeter` in `src/renderer/components/kanban/CardModal.test.tsx` (mirroring `browser-partition-parity.test.tsx`).
   - Prevents store/context subscriptions from mounting unnecessarily and keeps modal wiring tests decoupled from subcomponent API changes.

2. **Pin remaining unpinned guards**:
   - **Sheet `stopPropagation`**: Added test verifying `mousedown` inside the `.kanban-modal` sheet does not reach the outer `.kanban-modal-scrim` to trigger `onClose`.
   - **`isTopDialog` ownership**: Added test verifying `CardModal` ignores `Escape` when an overlay dialog is on top (`pushDialog`), and resumes handling `Escape` once the overlay is dismissed (`popDialog`).
   - Added `resetDialogStack()` in `beforeEach` / `afterEach` for test isolation.

3. **Refine test description**:
   - Updated the sticky note edit test title to accurately describe raw textarea write-through behavior and added assertions covering multiline/indented text write-through.

### Verification
- `npm run typecheck` (tsc node & web clean)
- `npx vitest run src/renderer/components/kanban src/renderer/nodes/browser-partition-parity.test.tsx src/renderer/canvas/toKanbanSession.test.ts` (all 21 tests pass)


eneskirca (approved):
Approving — this is an unusually clean follow-up, and every one of the three nits is addressed on the merits rather than gestured at.

I mutation-checked the two newly claimed guards, since that was the crux:
- breaking the sheet's `stopPropagation` (`onMouseDown={() => {}}`) reddens `does not close modal on mousedown inside the card sheet`;
- dropping `|| !isTopDialog(id)` reddens `respects isTopDialog ownership`, and separately removing `pushDialog(id)` reddens its *positive* half too — so the ownership test isn't passing by merely asserting nothing happened;
- and the original `.trim()` trap still reddens the renamed write-through test.

The stubs faithfully mirror `browser-partition-parity.test.tsx`, and you were right to keep the richer `ModalTerminal` mock instead of `() => null` — the `.kanban-modal__term` class it renders is what the Escape guard at `CardModal.tsx:102` keys on. I checked they're prophylactic rather than compensating: the suite is still 7/7 with all three stubs removed, so they aren't masking a failure, they just remove the latent "survives only because projectId is empty" fragility.

Nice touch on nit 3 — you strengthened the payload (newline + leading/trailing whitespace) instead of only relabelling the test.

Trial-merged onto current main: clean, kanban suites green. No nits from me this time.

PR #356, opened 2026-08-20
fix(hooks): shell-quote endpoint env values — a spaced path no longer kills the source

### LINKED ISSUE #351: Generated hook-endpoint.env is invalid shell syntax when path contains spaces
What did I do?

Created a Codex node and attempted to use Nodeterm context links.

What did I expect?

The generated hook endpoint file should source successfully so Codex can load its per-node token and reach Nodeterm’s local context bridge.

What happened instead?

The generated file:

~/Library/Application Support/node-terminal/hook-endpoint.env

contained this unquoted value:

NODETERM_NODE_TOKEN_DIR=/Users/work/Library/Application Support/node-terminal/node-tokens

Because the path contains a space, sourcing the file exits with status 127. Codex then fell back to plain mode and context reads failed with:

Running plain codex — nodeterm's hook server was unreachable

and:

Could not read linked context (nodeterm unreachable).

Quoting the path locally fixed the endpoint loading:

NODETERM_NODE_TOKEN_DIR='/Users/work/Library/Application Support/node-terminal/node-tokens'

---
```
nodeterm: v0.3.2
surface: desktop
os: macOS
electron: 42.8.1
```

eneskirca:
Fixes #351.

## The bug

The endpoint env files — `hook-endpoint.env` locally, and the SSH host's remote twin — are **sourced by POSIX sh** in three readers: the managed hook script's failover (`nt_adopt`), the codex identity proxy, and `context.sh`. `NODETERM_NODE_TOKEN_DIR` was written unquoted, and on macOS its default value lives under `~/Library/Application Support/…` — the space makes the assignment parse as a command prefix, the source exits 127, and the var is never set. Downstream:

- the codex identity proxy treats the failed source as `broker-unreachable` → **"Running plain codex"**, exactly as reported;
- `context.sh` loses the token dir → **"Could not read linked context (nodeterm unreachable)"**;
- the managed script's failover candidate walk rejects the file outright.

This hits **every macOS install** on the default data dir, not just unusual paths.

## The fix

- Extract the local writer into an exported `endpointFileContents()` and quote values through the existing `shellQuoteIfNeeded` (`@shared/shell-quote`): space-free values stay **bare**, so the historical format is byte-identical everywhere it already worked; anything with a space is single-quoted, which POSIX `.` reads natively.
- Same treatment in `remoteEndpointFileContents` (SSH host, unix-socket transport).
- The opencode plugin is the one **JS** reader of these files — teach its line parser the quoted grammar so it keeps resolving the token dir.

## Tests

- New `hook-endpoint-file.test.ts` proves against the **real reader**: it writes the generated contents and sources them with `sh`, asserting the vars round-trip — the spaced-dir cases fail with the exact 127 on the old format. A byte-identical pin covers the space-free path.
- New opencode test feeds the plugin the writer's own (quoted) output and asserts the per-node token still resolves.
- `src/core` suite green (2977 passed).

### DIFF
--- a/src/core/agents/hook-endpoint-file.test.ts
+++ b/src/core/agents/hook-endpoint-file.test.ts
@@ -1,32 +1,66 @@
-import { describe, it, expect, beforeAll, afterAll } from 'vitest'
-import fs from 'node:fs'
-import os from 'node:os'
-import path from 'node:path'
-import { hookServer } from './hook-server'
-import { nodeTokenDir } from './node-token-files'
-import { initPlatform, resetPlatformForTests } from '../platform'
-import { fakePlatform } from '../platform-fake'
+import { execFileSync } from 'child_process'
+import fs from 'fs'
+import os from 'os'
+import path from 'path'
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+import { endpointFileContents } from './hook-server'
+import { remoteEndpointFileContents } from '../remote-ssh/control-master'
 
-let dir = ''
-beforeAll(async () => {
-  dir = fs.mkdtempSync(path.join(os.tmpdir(), 'nodeterm-ep-'))
-  resetPlatformForTests()
-  initPlatform(fakePlatform({ userDataDir: dir }))
-  await hookServer.start()
+// Issue #351: the endpoint env file is SOURCED by a POSIX shell (managed script, codex identity
+// proxy, context-link). An unquoted value with a space — macOS's default
+// `~/Library/Application Support/node-terminal/node-tokens` — makes `.` exit 127 and the var is
+// never set, so Codex fell back to "plain codex" and context links died. The proof here is the
+// real reader: write the file, source it with sh, read the var back.
+
+let tmp: string
+beforeEach(() => {
+  tmp = fs.mkdtempSync(path.join(os.tmpdir(), 'nt-ep-'))
 })
-afterAll(() => {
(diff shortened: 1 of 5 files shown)


eneskirca:
Superseded: main landed the same fix independently and more completely — d840f63a quotes the writer's values (posixQuote, both the local and the remote endpoint file) and 11a30fc6 adds the shared quote-aware `parseEndpointEnv` for the two TS readers. Nothing in this branch is left to merge. Closing; #351 is covered by those commits.

eneskirca:
Closing as superseded — the fix is already on `main`, and issue #351 closed with it on 2026-08-20.

Main covers all three legs this PR proposed:
- **local writer** — `hook-server.ts:944` writes `NODETERM_NODE_TOKEN_DIR=${posixQuote(nodeTokenDir())}`
- **remote writer** — `remoteEndpointFileContents` quotes each value, with `control-master.test.ts:399` pinning the spaced-remote-path case
- **the JS reader** — `hook-endpoint-parse.ts`, a shared quote-aware parser for both TS readers (landed as `11a30fc6`)

It also kept the discipline this PR argued for: `hook-endpoint-file-sh.test.ts` sources the generated file with a real `sh` and asserts the space survives, which is the test that actually proves the 127 is gone.

This branch is `CONFLICTING` against those same files, so there is nothing left to rebase toward. Reopen if I've missed a leg.

PR #357, opened 2026-08-20
fix(settings): capability toggles persist — the setter now schedules the workspace save

### LINKED ISSUE #318: AgentsSection capability toggles never persist (setProjectCapability writes no disk)
Found while building the Project Settings panel (PR #317) and mirroring existing persistence precedents.

`setProjectCapability` (src/renderer/state/projects.ts) mutates the store only; its sole call site — the AgentsSection per-project capability toggle (src/renderer/components/settings/sections/AgentsSection.tsx:470) — calls it without any follow-up save. Unlike Canvas's rename/color callbacks (`setter + persist()`), nothing schedules a workspace save, so toggling `agentBrowserControl` / `agentMessaging` from Settings is lost on restart unless some unrelated canvas edit happens to dirty the workspace afterwards.

Cheap fix: call `markWorkspaceDirty()` (state/workspaceDirty.ts) next to the toggle — the same seam PR #317 uses for identity/defaults edits — or add the persist inside `setProjectCapability` itself.

eneskirca:
Fixes #318.

`setProjectCapability` mutated the store only; its call sites — the AgentsSection per-project capability toggle and the clone notice's decline — schedule no save of their own, so toggling `agentBrowserControl` / `agentMessaging` from Settings was lost on restart unless an unrelated canvas edit happened to dirty the workspace afterwards.

The same gap exists in `recordProjectCapabilityAck`: the clone notice's **kept** answer rides `IndexEntryV3.capabilityAck` on the next workspace save — which was never scheduled, so a confirmed notice could return after restart.

Per the issue's suggestion, the persist lives **inside the setters** (covering every call site) via `markWorkspaceDirty()` — the seam PR #317's identity edits use, so the write inherits Canvas's 800ms debounce and the external-change conflict gate.

Tests: two new store tests pin that each setter rings the workspace-dirty seam (both failed before the fix); the existing capability suites stay green (43 passed).


eneskirca:
Reviewed — correct and well-targeted. Merging.

I checked the thing that actually mattered here: whether the same gap exists in the other nine project setters. It does not, and for a reason worth writing down — on `main` **every** setter in `projects.ts` relies on its caller for the persist, and every other caller does it: `Canvas.tsx` wraps `setProjectColor` / `setProjectDefaultAccount` / `setProjectDefaultPermissionMode` in `void persist()`, `ProjectSettingsSection` calls `markWorkspaceDirty()`, the breadcrumbs site calls `markDirty()`, and `setDinoHighScore` rides the `updateNodeData` that precedes it. The two setters this PR fixes are the only ones whose call sites (`AgentsSection:475`, `CapabilityNotice:89/94`) schedule nothing — so the bug is exactly as scoped, and there is no double-persist introduced either.

One thing to keep in mind rather than change now: this makes two setters self-persisting while the other nine defer to callers, so the file no longer has one rule. If a third setter ever needs it, that's the moment to push the persist into all of them (or into a shared wrapper) rather than growing a third variant.

PR #359, opened 2026-08-20
fix(codex): a shared-identity node keeps a per-project unique node id

### LINKED ISSUE #350: Context links reuse a Codex node identity across projects
What did I do?

1. Opened two separate local projects, Project A and Project B.
2. Created a Claude node and a Codex node in each project.
3. Drew one green context-link edge in each project: Claude ↔ Codex.
4. In each Codex node, ran:

   printf 'current-node=%s\n' "$NODETERM_NODE_ID"

5. In the Project B Codex node, ran:

   sh "/Users/work/Library/Application Support/node-terminal/context-links/context.sh" list

What did I expect?

Each Codex node should receive a unique NODETERM_NODE_ID and resolve only its own green-linked Claude node.

What happened instead?

Both Codex nodes printed the same NODETERM_NODE_ID:

term-mt0uxowx-52e6e71d

The Project B Codex node listed the Project A Claude node, and messages sent from it routed to Project A. Closing/deleting and recreating the Project B Codex node did not assign a distinct identity.

The canvas still displayed separate green context edges, so the visual links and effective context/message routing disagreed.

---
```
nodeterm: v0.3.2
surface: desktop
os: macOS
electron: 42.8.1
```

eneskirca:
## Root cause

The shared Codex app-server is a **persistent daemon** — the first codex node to launch starts it (via `codex app-server daemon start` inside the launcher's `nt_preflight`) and every later node reuses it, the #167/S6 RAM saver. But that command runs in the **node's pane**, so the daemon inherited that node's per-pane `NODETERM_NODE_ID`. Every tool shell the daemon later spawns for **every** node then inherited the **first** node's id.

The thread→node resolver (`codex-thread-identity-sh.ts`) only recovers the correct node when a tool shell has **no** `NODETERM_NODE_ID` — its outer guard is `[ -z "$NODETERM_NODE_ID" ]`. The leaked value made that guard no-op, silently collapsing every codex node's effective identity onto whichever node started the daemon.

## Cross-project routing impact (the confused deputy)

Two codex nodes in **different projects** reported the *same* `NODETERM_NODE_ID` (identical prefix **and** 8-hex suffix). Project B's codex listed Project A's context links and messages **routed to A** — even though the canvas showed separate green edges. Deleting/recreating B's node didn't help because the poisoned daemon kept running with A's env. This matches the report in #350 exactly, including the id being stable across recreate and the visual links disagreeing with routing.

It is **not** a node-id generation collision: node ids are CSPRNG (`nextId` in `workspace.ts`) and would essentially never collide. The collision is the daemon leaking one node's id into every tool shell.

## Fix

Start the daemon in a **subshell with the per-pane `NODETERM_*` scrubbed**, so its tool shells carry no `NODETERM_NODE_ID` and the resolver recovers each thread's real node from `CODEX_THREAD_ID`. `NODETERM_CODEX_ACCOUNT_ID` is **deliberately kept** — one daemon serves one account and the resolver reads it to pick the record scope — so the shared app-server / per-node thread model (#167/S6) is **untouched**. No changes to thread minting, record writing, ownership, or the resolver itself.

## Tests (both mutation-checked)

- `codex-launcher-sh.test.ts`: runs the real generated launcher under `/bin/sh` with a start command that dumps the daemon's environment; asserts `NODETERM_NODE_ID` and the hook endpoint are **empty** for the daemon while `NODETERM_CODEX_ACCOUNT_ID` is **preserved**. Red before the fix (`NID=[node-1]`), green after. Mutations killed: removing the scrub (bug returns), and also scrubbing the account (breaks scope preservation).
- `codex-thread-identity-sh.test.ts`: cross-project isolation — a **clean** daemon-spawned tool shell for Project B's thread resolves node-B and **never** node-A (and symmetrically). Vacuity-checked.

Full suite: 594 files / 8344 tests green; `tsc --noEmit` clean.

Fixes #350

### DIFF
--- a/src/core/codex-identity-proxy.ts
+++ b/src/core/codex-identity-proxy.ts
@@ -602,7 +602,20 @@ nt_preflight() {
   # <CODEX_HOME>/packages/standalone/current/codex". Caps normally keeps that second case away from
   # here entirely, but the pane resolves CODEX_HOME from its OWN environment (§8.5) and an install
   # can be removed after boot, so the launcher still has to be able to say which it hit.
-  if ${appServerStartCommand}; then
+  #
+  # The app-server is a PERSISTENT DAEMON: the FIRST codex node to reach here starts it and every
+  # later node reuses it. Start it in a SUBSHELL with THIS node's per-pane NODETERM_* scrubbed, or
+  # the daemon — and every tool shell it later spawns for EVERY node — would inherit this one node's
+  # NODETERM_NODE_ID. The thread → node resolver (codex-thread-identity-sh.ts) only recovers the
+  # right node when a tool shell has NO NODETERM_NODE_ID; a leaked value makes its guard no-op and
+  # silently collapses every codex node's identity onto whichever node started the daemon, so
+  # Project B's codex lists Project A's links and routes to A (#350). NODETERM_CODEX_ACCOUNT_ID is
+  # the ONE var deliberately KEPT: one daemon serves one account, and the resolver reads it from the
+  # daemon's env to pick the record scope. The unset runs BEFORE the (test-injected) start command.
+  if ( unset NODETERM_NODE_ID NODETERM_HOOK_ENDPOINT NODETERM_HOOK_PORT NODETERM_HOOK_TOKEN \\
+    NODETERM_HOOK_SOCK NODETERM_AGENT_ID NODETERM_CANVAS_CONTROL NODETERM_NODE_TOKEN_DIR \\
+    NODETERM_PERM_WAIT_SECS
+    ${appServerStartCommand} ); then
     return 0
   fi
   if [ -x "\${CODEX_HOME:-$HOME/.codex}/packages/standalone/current/codex" ]; then
--- a/src/core/codex-launcher-sh.test.ts
+++ b/src/core/codex-launcher-sh.test.ts
@@ -185,6 +185,40 @@ describe('generated Codex launcher', () => {
     expect(fallbacks).toEqual([])
   })
 
+  // #350 (the confused deputy): the shared app-server is a PERSISTENT DAEMON started by the FIRST
+  // codex node to launch, and reused by every later node. If the launcher starts it while THIS
+  // node's per-pane NODETERM_NODE_ID is in the environment, the daemon — and every tool shell it
+  // later spawns for EVERY node — inherits that one node's id. The thread→node resolver only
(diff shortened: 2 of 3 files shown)


eneskirca:
## Alternate hypotheses examined (and why the daemon-env leak is the mechanism)

**1. Not an id-generation collision.** Node ids are CSPRNG-tailed (`nextId` in `workspace.ts`); the identical base36 *timestamp* prefix in the report proves the second pane *adopted* an existing id rather than minting a colliding one.

**2. Not the sh resolver's scan.** Verified against both `v0.3.2` and current `main`: the thread→node resolver is **thread-scoped in both versions** — it only ever opens the record named by this shell's own `CODEX_THREAD_ID` (`<root>/<threadId>`, plus per-account scopes for the *same* thread id on main, failing closed on ambiguity). There is no machine-global "adopt the only candidate regardless of thread" path. A tool shell with a clean env resolves its *own* thread's record, i.e. the correct node — now pinned by the new cross-project isolation test. The existing "inert without `CODEX_THREAD_ID` / unknown thread" tests pin the no-adoption cases.

**3. Not downstream of #351.** The launcher sources the endpoint file *before* the daemon start (same order at v0.3.2), so a #351-style sourcing failure falls back to plain codex **without ever starting the shared app-server** — and plain-codex tool shells inherit the pane's own (correct) env. #350 requires the shared path to have *proceeded*. The two shipped bugs are independent, not a chain; commit `dd58fe9f` pins this with a real-`/bin/sh` test (mutation-checked by moving the daemon start ahead of the endpoint checks).

**What remains is the daemon env leak fixed here:** the first codex node's pane env (`NODETERM_NODE_ID=term-mt0uxowx-52e6e71d` in the report) rode into the persistent shared app-server, and every tool shell it spawned — in *any* project — inherited it, no-opping the resolver's `[ -z "$NODETERM_NODE_ID" ]` guard. That reproduces every symptom: identical full id in both projects, B listing A's links and routing to A, delete/recreate not helping (the poisoned daemon persists), and green edges disagreeing with routing.

**SSH remote path:** it shared the leak — the remote launcher is built with `remoteCodexAppServerStartCommand` and the remote pane carries `NODETERM_NODE_ID` via tmux `-e`. The subshell scrub wraps whatever start command was injected, so it covers the remote daemon too, and deliberately leaves `CODEX_HOME` intact for the remote `nohup` fallback.

PR #362, opened 2026-08-20
feat(control): the open-project grant ledger and targeting gate (inert)

### LINKED ISSUE #338: Canvas control: a verb to open a project (`open-project --cwd`)
### What is missing

Canvas control can open nodes, but not projects. There is no verb that creates or opens a project, and no way to aim a node at a project other than the caller's.

### Why it bites

An agent that orchestrates work across several repositories wants one session per repository, each on its own canvas. Today every session it opens lands in the project that owns the calling node — `routeControlSource` in `src/renderer/lib/controlRouting.ts` resolves the owning project and travels there — so the sessions pile up on whichever canvas the caller happens to sit on, and the app jumps to that project each time.

Creating the projects by hand up front does not help either: a freshly created project has no nodes, so there is no node to call `open-agent` from. A project can only receive nodes once it already has one, which makes "one project per repository" reachable by hand but not by an agent.

`createProject` already exists as a renderer store action (`src/renderer/state/projects.ts`); the gap is that no verb reaches it.

### Proposal

```
open-project --cwd P [--name N] [--color C]     # returns the new/existing project id
```

Plus an optional `--project <id>` on `open-terminal` / `open-claude` / `open-agent`, so a node can be opened in a named project instead of the caller's.

Idempotent on `--cwd` would be the most useful shape: an existing project for that directory is returned rather than duplicated.

### Things worth deciding explicitly

- **Consent.** Creating a project touches a new directory, so it probably belongs behind project-capability consent, and/or the confirm dialog that `write` and `close` get.
- **Other surfaces.** What it means on SSH projects and on the Server Edition — including "refused here", if that is the answer.
- **Travel.** Whether `open-project` should activate the new tab or leave the current one in front. Leaving it alone seems friendlier for an agent opening several in a row.

### Related

#284 asks for something adjacent from the user side — reaching sessions in other projects from one place. Same boundary, different half: that one is about linking across projects, this one about creating and targeting them.

eneskirca:
PR 1 of 2 for issue #338 (`open-project --cwd` + project-targeted opens). This half ships the grant ledger, the targeting gate and the verb model — **inert**: there is no renderer dispatch yet, so a live `open-project` parses, passes the gates, and is answered by the renderer's `default:` case with `unknown verb: open-project` (`ok: false`, so no grant is ever minted). PR 2 wires the dispatch, the confirm dialogs, the store-path opens and the agent-facing docs, and closes #338.

## Tasks → commits

| Task | Commit | What |
|---|---|---|
| 1.1 verb model + parse | `a223ac3e` | `'open-project'` in `ControlVerb` + `VERBS`; `parseControlRequest` requires `--cwd` |
| 1.2 verified-only (P1) | `100f8353` | `requiresVerified` + `'Project open refused.'` in `verifiedRefusalFor`; wire tests on the real `/control/` route |
| 1.3 grant ledger (P2/P3) | `6e01845b` | `src/main/project-grants.ts`: pure `Map<caller, Set<projectId>>`, cap 64, `grant`/`isGranted`/`atCap`/`clearCaller`/`clearAll` |
| 1.4 targeting gate + cwd (P7/P8-ssh) | `6e01845b` + `7e6a18f5` | `gateProjectTarget` + `validateOpenProjectCwd` (pure) and the `setControlHandler` wrapper wiring; `workspace-store.projectMetaFor` |
| V-2 teardown anchor | `7e6a18f5` | Resolved: `corePlatform.on(IPC.ptyDestroy/ptyRecycle)` — the same anchor the browser leases die on. **Both** legs clear the caller's grants (a recycle mints a fresh identity that must re-earn its rights); app restart clears the in-memory ledger by construction. |
| 1.5 Server Edition pin (P8) | `c6db741f` | `serverEditionControlHandler({ verb: 'open-project' })` answers `control-unsupported-on-this-edition`; direct call + a wire leg past identity. No server code changes. |

## Security properties shipped here

- **P1 verified-only.** `open-project` joins `requiresVerified`, decided on the VERDICT before any handler — unreachable by `hookIdentityStrict: false` and the dated window (the #212 send/reply pattern). Legacy/forged/foreign-kid callers get 403 + the flat sentence `Project open refused.` (no token/restart advice). The `--project` arg half: `gateProjectTarget` refuses `verified !== true` with `Project targeting refused.` before anything is forwarded.
- **P2 per-caller grants.** Caller B presenting an id granted to caller A is refused — pinned against the real ledger + gate composition.
- **P3 session-scoped.** In-memory only; the module is pure (no `fs`, no electron, no `child_process` — asserted structurally); cleared per-caller on `ptyDestroy`/`ptyRecycle`, wholesale on restart. Recording happens ONLY in the wrapper on `result.ok && verified && non-empty string projectId` (the open-browser ledger pattern at the same call site).
- **Targeting gate fails closed.** Unknown/deleted target ids refuse with the SAME sentence as ungranted ids (`project-target-refused: …`), so the gate is not a project-existence oracle. SSH targets (other than the caller's own) refuse `project-target-ssh-unsupported` — and that check runs BEFORE the granted allow, as the belt for "a granted id can never be SSH". Target existence/SSH-ness comes from `projectMetaFor` (main's own store), never from the request.
- **cwd validation (P7).** Absolute required, `path.resolve` once, ONE stat on the resolved path (single-fd, no TOCTOU re-walk; the path never leaves the process). The resolved form replaces the raw argument in the forwarded args — PR 2's dialog, dedupe and storage all see it. SSH-project callers get `open-project-local-only`; an unresolvable caller fails closed with `open-project-caller-unresolved` (transient: a just-created node inside the save debounce).
- **Cap.** 64 grants/caller; the 65th refuses `open-project-grant-cap` (no silent eviction), checked BEFORE forwarding so no dialog can be consented to and then dropped.
- **P8 Server Edition.** Pinned unreachable by name.

## Mutation checks (introduced/caught: 9/9)

| # | Mutation | Result |
|---|---|---|
| 1 | verified-only decided on `decision === 'refuse'` instead of `verdict !== 'verified'` | red (2 tests) |
| 2 | `isGranted` keyed globally instead of per-caller | red (2 tests) |
| 3 | `clearCaller` made a no-op | red (1 test) |
| 4 | gate: `verified` check dropped | red (1 test) |
| 5 | gate: unknown target allowed | red (1 test) |
| 6 | cap removed | red (1 test) |
| 7 | granted checked before the SSH refusal | red (1 test) |
| 8 | wrapper records a grant without `verified` | red (1 test) |
| 9 | `ptyRecycle` teardown leg dropped | red (1 test) |

## Deliberate deferrals (per the plan)

- **`DESTRUCTIVE_VERBS` membership moves to PR 2 (Task 2.2):** `control-destructive.test.ts` structurally requires every member's `Canvas.tsx` case to read `isDestructiveVerb(verb)`, and that case only exists in PR 2 — adding the membership now trips the set↔dispatch drift alarm with no dispatch to agree with. A tripwire test (`open-project is NOT in DESTRUCTIVE_VERBS yet`) goes red the moment PR 2 adds it.
- **No skill/instructions text:** spec §8 — docs land in the PR that makes the verb reachable. The parity test walks a named verb list, so it stays green without a half-documented stub.

## Verification

`npm run typecheck` green; per-task suites green (`canvas-control-core` 39, `messaging-verified-only` 6, `project-grants` 21, `workspace-store` 92, `control-unsupported` 11); full `npx vitest run`: **598 files passed | 1 skipped, 8393 tests passed | 12 skipped**.

Refs #338 (PR 2 closes it).

### DIFF
--- a/src/core/agents/hook-server.ts
+++ b/src/core/agents/hook-server.ts
@@ -175,11 +175,22 @@ export interface HookEventMeta {
  * protect — the routes are new — which is the one place in the whole control surface where
  * fail-closed from day one costs nobody anything.
  *
+ * `open-project` (issue #338) is here for the same class of reason as `sticky`: the main-side
+ * grant ledger (src/main/project-grants.ts) mints per-caller targeting rights off a successful
+ * `open-project`, and a grant recorded for an unverifiable caller would authorize whoever can
+ * name that caller's node id. NEW verb, so fail-closed from day one strands nobody.
+ *
  * Consulted in the `/control/` route BEFORE `identityGate`'s decision is, so no future change to
  * the policy table can widen it; `messaging-verified-only.test.ts` drives the route on both sides
  * of every hatch and is the test that fails if either half of this comment stops being true.
  */
-export const requiresVerified: ReadonlySet<string> = new Set(['send', 'reply', 'notify', 'sticky'])
+export const requiresVerified: ReadonlySet<string> = new Set([
+  'send',
+  'reply',
+  'notify',
+  'sticky',
+  'open-project'
+])
 
 /**
  * The refusal for a messaging verb: one sentence, no diagnosis, no hint about tokens or restarts —
@@ -192,9 +203,15 @@ export const MESSAGING_CONTROL_REFUSAL = 'Agent messaging refused.'
  *  because "agent messaging refused" answering a note write is a diagnosis-delaying lie. */
 export const STICKY_CONTROL_REFUSAL = 'Sticky write refused.'
 
+/** Same posture again for `open-project` (issue #338): one sentence naming what was refused, no
+ *  diagnosis, no token or restart advice — a designed refusal, not a rollout accident. */
+export const OPEN_PROJECT_CONTROL_REFUSAL = 'Project open refused.'
+
 /** The verified-only refusal, worded for the verb that was refused. */
 export function verifiedRefusalFor(verb: string): string {
-  return verb === 'sticky' ? STICKY_CONTROL_REFUSAL : MESSAGING_CONTROL_REFUSAL
+  if (verb === 'sticky') return STICKY_CONTROL_REFUSAL
+  if (verb === 'open-project') return OPEN_PROJECT_CONTROL_REFUSAL
+  return MESSAGING_CONTROL_REFUSAL
 }
 
 class HookServer {
--- a/src/core/agents/messaging-verified-only.test.ts
+++ b/src/core/agents/messaging-verified-only.test.ts
@@ -112,7 +112,7 @@ describe('send/reply require `verified` — and controlPolicy is NOT the decider
     // from day one — there is no upgrade population to protect.
     hookServer.clearNodeAuthSecretForTests()
     try {
-      for (const verb of ['send', 'reply'] as const) {
+      for (const verb of ['send', 'reply', 'open-project'] as const) {
         const res = await post(verb, 'n-src')
         expect(res.status, verb).toBe(403)
       }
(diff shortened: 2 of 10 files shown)


eneskirca:
## Review fix round 1 — commit `2796f6a6`

Addresses the adversarial review's 3 Important + 2 Minor findings:

- **I1/I2 (unpinned SSH-caller + unresolved-caller refusals):** the whole `open-project` pre-forward gate is now the pure `gateOpenProject` (project-grants.ts) — caller-unresolved → local-only → cap → cwd, resolved once — with a behavioral test per branch (deleting either previously-unpinned branch is now red), plus a structural pin that the wrapper routes every open-project through it and forwards only its resolved cwd.
- **I3 (SSH-kind oracle):** every stranger id — unknown, local-ungranted, SSH-ungranted — now refuses with **byte-identical** `project-target-refused` (Buffer-equality test). The `project-target-ssh-unsupported` wording survives only inside the granted branch as belt (a granted SSH id cannot exist; if that invariant ever breaks the target is refused, not opened, and only a grant holder — who was handed the id by open-project — sees the kind named).
- **M1 (discarded `'cap'`):** the record decision is the pure `recordOpenProjectGrant` (`'recorded' | 'not-applicable' | 'cap'`); on the cap race the wrapper logs and replaces the success reply with `open-project-grant-cap` — a caller never hears ok while holding no targeting right (idempotency means nothing is lost). A structural-only pin proved insufficient here (a `=== 'cap' && false` defusal survived it), which is what forced the pure extraction.
- **M2 (branch order):** `gateProjectTarget` allows OWN before the unknown-target refusal — no spurious refusal if the two index read paths ever stagger mid-debounce; still fail-closed.

**Mutation checks this round: 11 introduced / 11 caught** (branch deletions ×2, oracle reintroduction, granted-SSH belt allow, wrapper cap defusal, cap→recorded, verified drop, ok drop, order regressions ×2, unknown-but-granted allow).

Verification: typecheck green; project-grants suite 31/31; full `npx vitest run` 598 files / 8403 tests passed | 12 skipped.

PR #366, opened 2026-08-20
feat(canvas): breadcrumb trail — back/forward navigation + resume card

gor3a:
## Summary
- Records a breadcrumb (with an auto-generated note) on every deliberate node landing (`goToNode`), stored machine-locally on `IndexEntryV3` — never in the git-shared `.nodeterm/project.json`.
- Browser-style back/forward camera navigation (`Cmd+[` / `Cmd+]` + two new Dock buttons), skipping deleted nodes.
- A dismissible "resume where you left off" card shown once per project activation, listing the last few distinct places you were.
- Design spec + implementation plan were brainstormed, Fable-reviewed, and executed task-by-task via subagent-driven development, each task independently reviewed plus a final whole-branch review and fix pass.

## Test plan
- [x] `npx vitest run` — clean aside from the repo's known pre-existing real-PTY/tmux-spawn/real-`sh`/server-e2e sandbox flakiness (confirmed unrelated to any file this PR touches)
- [x] `npm run typecheck` — clean
- [ ] Manual verification in a real Electron window (Cmd+[/Cmd+] firing, Dock button enable/disable at trail boundaries, resume card placement) — not yet performed, no GUI access in the environment this was built in


eneskirca (changes requested):
# Review — PR #366 `feat(canvas): breadcrumb trail — back/forward navigation + resume card`

**Verdict: Request changes** (small — all three Important findings are mechanical; nothing here is a
persistence or correctness defect, and the highest-stakes surface in the diff is done right).

**Gates, run in this worktree:** `npx vitest run` → **634 test files passed / 1 skipped, 8840 tests
passed / 27 skipped**; `npm run typecheck` → **clean**. Both fully green — the PR body's hedge about
"known pre-existing sandbox flakiness" was not needed here, nothing was red.

---

## What the author did well

This is a careful, house-style branch, and the part that mattered most is the part that is most
correct:

- **The persistence boundary is exactly right, including the sites people forget.** `breadcrumbs` was
  threaded through *all eight* index-entry sites in `src/core/workspace-store.ts` — the same eight
  `capabilityAck` occupies (lines 285, 305, 834, 971, 1289, 1352, 1459, 1474), **including the
  `unavailable`-window restore at `workspace-store.ts:834`**, which is the one a first pass at this
  always drops. `projectToFile` is an allowlist builder, so nothing can leak into the shared file by
  construction, and `splitWorkspace`'s `localState` block (`workspace-files.ts:414-421`) puts it in
  the same tier as `viewport`/`defaultAccountId`/`capabilityAck`. A pre-change build reading a v3
  index written by this build is unaffected (field-pick readers, unknown key ignored); it drops the
  trail on its next write, which is the right degrade.
- **The tests for that boundary assert the real facts, not the shape.** `workspace-files.test.ts:143`
  round-trips *and* proves a file-borne `breadcrumbs` (the forgery case) is never read;
  `workspace-store.test.ts:73` goes through a real `WorkspaceStore` → real `project.json` on disk →
  a *fresh* store's `load()`, and greps the committed bytes for both the key and the note text.
- **`recordBreadcrumb` returning the same object on a dedupe no-op**, with identity as the caller's
  skip test (`breadcrumbs.ts:1155`, pinned at `breadcrumbs.test.ts:939`) — a nice, cheap contract.
- **The `e.key` vs `e.code` bracket test** (`keybindings.test.ts:359`) is a real test with its
  reasoning attached, not a restatement of the registry.
- **Excluding ephemeral `subagent`/`loop` nodes from recording** (`Canvas.tsx:5675`) is a genuinely
  non-obvious call that a hasty implementation gets wrong.
- **`ResumeCard.test.tsx` asserts behavior**: newest-first ordering, filter-before-cap, de-dupe at the
  newest occurrence, React key uniqueness (via a `console.error` spy), and that a dismissal survives
  a parent re-render carrying a new breadcrumb. That last one is the bug this component would
  actually have shipped with.
- The `key={resumeProject.id}` on `ResumeCard` with the "dismissed → straight to another qualifying
  project never passes through null" reasoning (`Canvas.tsx:10809-10812`) is the kind of comment this
  repo wants.

---

## Findings

### Important

**1. `stepAndFrame` copy-pastes `goToNode`'s framing block, so the "Go to node" invariant now has two
implementations and neither copy is pinned.**
`src/renderer/canvas/Canvas.tsx:5766-5783` (diff lines 415-431) is a verbatim copy of
`Canvas.tsx:5716-5735` (measured ⇒ `fitView` with `solveFitPadding`; unmeasured ⇒ `nodeFitRect` +
`viewportForRect`; unknowable ⇒ stand still). The author names this in the comment above it
(`Canvas.tsx:5745-5747`) and defers the extraction — but CLAUDE.md's **"Go to node"** bullet ends
with *"never fall back to a bare `fitView` there, that IS the origin jump"*, and a duplicated copy
with no test is precisely how that invariant regresses: the next person fixing the measured rule
fixes one of two call sites and the bug reappears only on back/forward. The extraction is safe here
because the *only* difference between the two paths is the recording, which already sits above the
framing block: pull `Canvas.tsx:5716-5735` out as `frameNode(node: Node)` (a `useCallback` over
`[fitView, setViewport, getInternalNode]` — identical deps to what both already close over) and have
`goToNode` = record + `frameNode`, `stepAndFrame` = step + `frameNode`. The reasoning the comment
worries about losing lives in the comment block, which moves with the code.

**2. The Dock's back/forward enabled state does not agree with what a step will do.**
`Canvas.tsx:10879-10880` (diff lines 479-480):
```
canGoBack={navRef.current.index > 0}
canGoForward={navRef.current.index < navRef.current.list.length - 1}
```
`stepBreadcrumb` (`breadcrumbs.ts:1174-1186`) skips stops whose node is gone and returns `null` when
*every* stop in that direction is dead — the case `breadcrumbs.test.ts:1044` explicitly covers. So a
trail whose earlier entries all point at deleted nodes renders an **enabled ← that does nothing when
clicked** (and a Cmd+[ that is swallowed by `preventDefault` with no effect). This is the same shape
as the session-memory row that "promised a kill it could not perform". The fix is one expression, and
it is cheap (≤20 entries, and Canvas already re-renders on both `bumpNav` and `nodes`):
```
const exists = (id: string) => nodesRef.current.some((n) => n.id === id)
canGoBack={!!stepBreadcrumb(navRef.current, 'back', exists)}
canGoForward={!!stepBreadcrumb(navRef.current, 'forward', exists)}
```
Bonus: it then also stays honest as nodes are deleted, which the raw index comparison never notices.

**3. None of the Canvas-level wiring the new CLAUDE.md bullet calls load-bearing is pinned by a
test.** `breadcrumbs.ts` and `ResumeCard.tsx` are well covered, but everything in `Canvas.tsx` is
not: the `subagent`/`loop` record exclusion (`Canvas.tsx:5675`), the "back/forward must NOT record"
rule (the entire reason `stepAndFrame` exists rather than calling `goToNode`), and the
once-per-activation resume gate (`Canvas.tsx:2011-2021`). CLAUDE.md states all three as facts, and
this repo's own rule is *"a doc line with no such test is a plan, not a fact"*. The house pattern for
exactly this is already in the tree — `src/renderer/canvas/canvas-wiring.test.tsx` reads
`Canvas.tsx` as source and asserts on it (e.g. line 58's `expect(CANVAS_SRC).toContain(...)`). Three
source-level pins there would cost a few lines and would red on the regressions that matter (a
refactor that routes `stepAndFrame` through `goToNode`, or that drops the node-type guard).

### Minor

**4. Two shipped source comments cite a spec file that is not in the repo.**
`Canvas.tsx:603` and `src/shared/types.ts:659` both point at
`docs/superpowers/specs/2026-08-20-breadcrumb-trail-design.md`; `docs/superpowers/specs/` contains no
such file (verified). Either commit the spec — the eight siblings in that directory are all committed
— or drop the two references.

**5. The new CLAUDE.md bullet overstates the dirty claim.** Diff lines 30-32 / `CLAUDE.md`
"Breadcrumb trail" → *"stepping back and forth writes NO project state and marks nothing dirty"*. The
breadcrumb code marks nothing dirty (correct, and `Canvas.tsx:5759-5761` says it well), but the step
**does** move the camera, and CLAUDE.md's own Zoom-chords bullet states *"a camera move here is not
read-only — `onMove` → `markDirty` persists the viewport"*. As written, the two bullets contradict
each other. Suggest: *"a step records no breadcrumb and rewrites no `project.json` — only the ordinary
`onMove` viewport persist, which is machine-local."*

**6. `buildNote`'s generic fallback is unreachable, and its test does not test it.**
`breadcrumbs.ts:1135` — `KIND_LABEL[target.kind ?? 'terminal'] ?? 'node'`. `KIND_LABEL`
(`breadcrumbs.ts:1086-1096`) covers every `NodeKind` except `subagent`/`loop`, which `goToNode`
refuses before calling in. So `'node'` is dead for any real node type. Meanwhile
`breadcrumbs.test.ts:923` is named *"an unknown node kind degrades to a generic label rather than
inventing one"* but passes `kind: undefined` and asserts `'terminal · Untitled'` — it exercises the
`?? 'terminal'` default, not the branch its name describes. Either delete the `?? 'node'` arm or
rename the test to what it actually pins (`a missing kind defaults to terminal`).

**7. A project that opens on the kanban board burns its once-per-run resume slot on an invisible
card.** `Canvas.tsx:2016-2018` spends `resumeCardShown` without asking `isKanbanOpen(project.id)`,
and `.resume-card` is `z-index: 11` (styles.css, diff line 1290) under the opaque
`.kanban-overlay` at `z-index: 25` (`styles.css:8186`). This is the *same* failure mode the author
deliberately guarded against for all-dead breadcrumbs two lines earlier ("`once` is only spent on a
card that could actually render"). Adding `&& !isKanbanOpen(project.id)` to that condition closes it
(the helper is already imported and used at `Canvas.tsx:5754`).

**8. Unrelated scope creep in `.gitignore`.** Diff line 9 adds `.worktrees/`, which has nothing to do
with breadcrumbs. Harmless, but it belongs in its own change.

**9. The three-surfaces call is not stated anywhere.** Per CLAUDE.md's Conventions, each surface
needs a decision even when it is "N/A". For the record, and worth putting in the PR body / the
CLAUDE.md bullet: **Server Edition** works as-is (the whole feature is shared renderer code plus
`WorkspaceStore`, which both shells boot — no bridge member added, nothing to stub); **mobile** is
N/A (no canvas, no camera); and the **kanban board** — which CLAUDE.md insists is a first-class
surface, not an afterthought — is likewise N/A for the same reason. Saying so costs two sentences and
stops the next reader from re-deriving it.

**10. `resumeCardShown` (`Canvas.tsx:580`) is never pruned** when a project is deleted or closed. It
is in-memory and per-app-run, so this is a handful of strings at worst — noted only so nobody later
mistakes it for a leak worth "fixing" with persistence, which the comment there already argues
against.

**11. `Ctrl+[` is `ESC` in readline/vim** — worth one clause in the registry comment
(`keybindings.ts:1426-1429`), since `both('Cmd+[')` makes the non-mac binding `Ctrl+[`. It is safe in
practice: `canvas.goBack` is `scope: 'canvas'` without `allowWhileTyping`, so
`resolveCommandForKeyEvent` refuses it whenever a terminal or input has focus. Stating that in the
comment is cheaper than the next person re-deriving it.

---

## Things I checked that are fine (recorded so they are not re-litigated)

- **No breadcrumb-shaped field can reach the git-shared file.** `projectToFile`
  (`workspace-files.ts:219-250`) is an explicit allowlist; the SSH `cache` goes through the same
  function (`workspace-files.ts:449`); inline (cwd-less) projects keep the whole `Project` verbatim in
  the machine-local index, so their breadcrumbs never touch disk outside `workspace.json`.
- **`toWorkspace()` passes projects wholesale** (`state/projects.ts:642`) and `commitCanvas` spreads
  `{...p}`, so the field survives the debounced persist round trip. Verified end to end by the new
  store test.
- **markDirty on record is not new I/O pressure.** `goToNode` already moves the camera, which marks
  dirty via `onMove`; and since `projectToFile` output is unchanged, `sameProjectContent` suppresses
  the `project.json` write and the SSH mirror. Only the index is rewritten.
- **The cross-project focus path does not lose its breadcrumb**: the `pendingFocusRef` `goToNode` at
  `Canvas.tsx:2034` runs *after* `loadingRef.current = false` (line 2003), so `markDirty` is live.
- **Keybinding registry compliance**: one registry entry per command with sane `scope`/`group`, the
  whole-table pin in `keybindings.test.ts:84-89` updated in place, dispatch through the single
  handlers map (`Canvas.tsx:5907-5908`), no second listener, and no count constants to update
  (`ShortcutsSection.test.tsx` derives from `COMMAND_DEFINITIONS.length`). `MAIN_INTERCEPTED_COMMAND_IDS`
  correctly untouched — no menu accelerator claims `Cmd+[`/`Cmd+]` (checked `buildAppMenu`).
- **Server Edition browser hijack**: `dispatchGlobalKeydown` calls `e.preventDefault()` whenever the
  handler returns true (`globalKeybindings.ts:93`), and both handlers return `true`
  unconditionally — so Cmd+[ / Cmd+] never reach Chrome's own history navigation.
- **No origin-jump path was introduced**: both framing blocks gate on `isMeasured(getInternalNode(id))`
  and fall back to `nodeFitRect`/`viewportForRect`, never a bare `fitView`; unknowable size stands
  still.
- **Store/render hygiene**: `navRef` is a ref with an explicit `bumpNav` counter (not a zustand
  subscription), so nothing here re-renders Canvas per hook event or per camera frame — the
  `armedDepSig`/`loopSig` discipline is respected. `ResumeCard` holds a snapshot `Project` rather than
  subscribing to `useProjects`, which the comment at `Canvas.tsx:1307-1313` explains correctly.
- **`ResumeCard` z-order and placement**: mounted as a sibling of `<ReactFlow>` inside `.flow-wrap`
  (so it is not trapped in React Flow's inline `z-index: 0` stacking context), `z-index: 11` — above
  the canvas, below the sessions sidebar (12) and the kanban overlay (25); `top: 100px` clears the
  controls cluster. `data-canvas-chrome` is a real fitView opt-in (`canvas/fit-view.ts:18`), correctly
  used. It does not block input: it is a small fixed-width card with real buttons and a dismiss.
- English-only throughout; no competitor product names; no dead exports beyond finding 6.

eneskirca (approved):
Took the findings into the branch myself (maintainer edit) rather than round-tripping — all three Importants and the cheap Minors are closed in `fix(breadcrumbs): close the review findings…`:

- **frameNode** is now the one framing implementation behind both `goToNode` and `stepAndFrame` (the reasoning comments moved with it), with a wiring pin asserting exactly one `isMeasured` copy exists.
- Dock ←/→ enabled state now derives from `stepBreadcrumb` itself — an arrow is never enabled when a click would do nothing.
- The resume slot is not spent while the kanban board is open (same rule as the all-dead-stops guard beside it).
- Wiring pins added for the subagent/loop exclusion, step-never-records, and the resume-slot condition; dead `?? 'node'` arm removed and its test renamed to what it pins; the two uncommitted-spec references dropped; CLAUDE.md's dirty claim reconciled with the Zoom-chords bullet and the three-surfaces call stated; Ctrl+[ = ESC safety argued in the registry comment.

Not touched, deliberately: the `.gitignore` line (harmless), `resumeCardShown` pruning (correctly argued in-code as not worth persistence). Full suite green on the branch (the one red was a real-tmux negative-control test that also fails/passes identically on clean HEAD under this machine's shared socket — environmental, not this PR).

Genuinely nice work — the machine-local persistence boundary (the highest-stakes surface here) was clean on first read, the `capabilityAck`-parity threading is exactly the house pattern, and the e.key-vs-code bracket test saved the next person a silent dead chord. Approving; auto-merge armed.

PR #368, opened 2026-08-20
fix(hooks): a sandboxed Codex gets an actionable error, and the hook server listens on a unix socket

### LINKED ISSUE #367: Codex workspace sandbox blocks local context-link and canvas-control shims
### What did I do?

On macOS, I opened a local Codex CLI node and a Claude Code CLI node, context-linked them, and asked Codex to send a task to Claude Code.

The endpoint-file quoting problem from #351 was already worked around: `hook-endpoint.env` sources successfully, contains a valid port/token/token-directory, and the nodeterm hook server is listening.

Codex then ran the installed shims:

```sh
sh "$HOME/Library/Application Support/node-terminal/context-links/context.sh" list
sh "$HOME/Library/Application Support/node-terminal/canvas-control/nodeterm.sh" send --node "Claude Code" --text "..."
```

### What did I expect?

The documented context-link and canvas-control workflows should work from a default local Codex node. If local loopback requires leaving Codex's command sandbox, Codex should receive actionable instructions that cause one narrow approval/escalation retry.

### What happened instead?

Both shims reported that nodeterm itself was unreachable:

```text
Could not read linked context (nodeterm unreachable).
Could not reach nodeterm (control endpoint unreachable).
```

The Codex tool environment includes:

```text
CODEX_SANDBOX=seatbelt
CODEX_SANDBOX_NETWORK_DISABLED=1
```

For local nodes, both generated shims use `curl` against `http://127.0.0.1:$NODETERM_HOOK_PORT`. Inside the normal `workspace-write` sandbox, curl returns HTTP code `000`. Running the exact same shim commands with command-specific `require_escalated` access reaches nodeterm immediately and returns semantic application responses instead of the transport error.

This means the server can be healthy and the nodes can be correctly configured while the UI/agent advice incorrectly recommends relinking or restarting nodeterm.

### Reproduction

1. Use Codex with `sandbox_mode = "workspace-write"` and `approval_policy = "on-request"`.
2. Open local Codex and Claude Code nodes in nodeterm.
3. Context-link the two nodes and enable agent messaging.
4. Ask Codex to list its linked nodes or send a message to Claude Code.
5. Observe the `nodeterm unreachable` error.
6. Retry the identical command outside the network-disabled sandbox and observe that it reaches nodeterm.

### Suggested resolution

At minimum:

- Include Codex-specific guidance in the installed AGENTS instructions: when either exact transport error occurs, retry the same command once with command-specific `require_escalated` access and a narrow loopback justification.
- Do not recommend relinking/restarting until that escalated retry also fails.
- Make the shim's error actionable when `CODEX_SANDBOX_NETWORK_DISABLED=1`, so a sandbox denial is distinguishable from a dead hook server.

A transport that works inside Codex's default sandbox would be even better, if one is available. Globally enabling network access or requiring `danger-full-access` would be an undesirable workaround.

### Environment

```text
nodeterm: v0.3.2 desktop
Codex CLI: 0.148.0
macOS: 15.7.7 (24G720)
Codex sandbox: workspace-write / Seatbelt, network disabled
Codex approval policy: on-request
```

Related but separate: #351 fixed invalid shell syntax in the endpoint file. This issue occurs after the endpoint file sources successfully.

eneskirca:
Fixes #367.

## The probe (what the Codex sandbox actually does)

Measured live against codex 0.146.0 (`codex sandbox`, sandbox-state JSON) plus a static read of the seatbelt generation in openai/codex:

- **Linux: BLOCKED, unconditionally.** With network disabled, a seccomp filter EPERMs `connect()` (and accept/bind/listen/getsockname/sendto/…) for **every address family** — unix sockets included; only `socket(AF_UNIX)`/`socketpair` creation is allowed. The Linux sandbox has **no unix-socket allowlist**. `curl --unix-socket` dies with "getsockname() failed … Operation not permitted", loopback TCP with "Failed to connect … after 0 ms" — both HTTP 000.
- **macOS: BLOCKED by default, ALLOWLISTABLE.** The seatbelt base profile is `(deny default)` with no network allowance, but `seatbelt.rs` generates `network-bind`/`network-outbound` allows from `network.allow_unix_sockets` in `~/.codex/config.toml` **even when network is disabled** — so a user can permanently re-open exactly the nodeterm hook socket inside the sandbox.
- Both platforms export `CODEX_SANDBOX_NETWORK_DISABLED=1` into every sandboxed command, which is what makes self-diagnosis possible at the exact moment of failure.

So the shims' old failure text — "nodeterm unreachable" / "control endpoint unreachable" — was a misdiagnosis that sent agents (and users) off to relink/restart a perfectly healthy server.

## The three parts

**1. Self-diagnosing shim error (both platforms — the primary fix).** Both generated sh shims (canvas-control + context-link) share one fragment (`hook-sandbox-hint-sh.ts`) that fires only on a dead transport (empty/000 status) with `CODEX_SANDBOX_NETWORK_DISABLED` present: it names the sandbox as the blocker, instructs retrying the **same** command once with command-specific escalated permissions, and says plainly not to relink or restart nodeterm. On Darwin, when `NODETERM_HOOK_SOCK` is advertised, it prints the exact `network.allow_unix_sockets = ["<sock>"]` line for `~/.codex/config.toml`. Without the env var, the original messages survive to the byte; a genuine HTTP error (the server answered) never triggers it. Proven against real `/bin/sh` with `uname` faked on PATH so the Darwin-only line is deterministic on any CI.

**2. Unix-socket listener on the hook server (the macOS remedy enabler).** A second listener on `<userDataDir>/sock/hook.sock` — SUN_LEN-guarded with a `~/.nodeterm/sock/hook-<digest>.sock` fallback (the SSH ControlMaster discipline; home rather than tmpdir so the predictable name cannot be squatted by another local user). Same HTTP handler as TCP, so bearer + per-node token auth run identically on both transports; 0600 file in a 0700 dir; stale socket unlinked before bind (the StreamLocalBindUnlink lesson) and on shutdown. Advertised as `NODETERM_HOOK_SOCK` in `buildPtyEnv` and — posixQuoted per #358 — in the endpoint file, only when it actually bound. Every generated client was already sock-first, so no client change was needed beyond Part 1's error text. **TCP stays** for back-compat (existing tmux panes, and the Linux sandbox can't use the socket anyway). This is also a step toward **#195**: local traffic now has a filesystem-permissioned path (0700 dir, 0600 socket), which a defense-in-depth follow-up can make the only door. It does not close #195 on its own — the TCP listener stays for back-compat, so the (bearer-gated) port is still reachable by any local user until that port is retired (the tracked follow-up).

**3. Agent-facing guidance.** The canvas-control skill + AGENTS.md instructions and the get-linked-context skill + instructions all teach the same recovery: on either exact transport error, one escalated retry; never advise relink/restart first; on macOS the config.toml allowlist is the permanent fix. Rendered from the same constants the sh fragment embeds, with parity tests walking the constants against all four bodies.

## Verification

- `npm run typecheck` clean; full `npx vitest run`: **8824 passed, 27 skipped, 0 failed** (633 files).
- Real-sh / real-curl coverage: shim sandbox matrix (7 new cases, canvas-control) + context-link leg (3), unix-socket listener suite (7: TCP/socket parity, auth 403s, 0600/0700, stale rebind, verified-only gate over the socket), SUN_LEN path chooser (4), doc parity (4).
- **Mutations:** neuter the env-var branch → **4 tests red**; always-true bearer check (socket auth bypass) → **1 test red** (the dedicated 403 guard); drop the stale-socket unlink → **3 tests red**. All restored and re-verified.

### DIFF
--- a/src/core/agents/codex-thread-id-route.test.ts
+++ b/src/core/agents/codex-thread-id-route.test.ts
@@ -75,8 +75,10 @@ describe('/codex-thread/bind refuses a path-unsafe thread id', () => {
     }
     expect(bound).toEqual([])
     // Nothing anywhere: not in the store, and — the `..` row — not in the store's parent either.
+    // (`sock/` is the hook server's own unix-socket listener dir, issue #367 — always present,
+    // never a record.)
     expect(existsSync(codexThreadIdentityRoot())).toBe(false)
-    expect(readdirSync(dir)).toEqual(['hook-endpoint.env'])
+    expect(readdirSync(dir).sort()).toEqual(['hook-endpoint.env', 'sock'])
   })
 
   it('still binds a thread id the app-server would actually mint', async () => {
--- a/src/core/agents/hook-identity-enforcement.test.ts
+++ b/src/core/agents/hook-identity-enforcement.test.ts
@@ -361,7 +361,10 @@ describe('the latch is memory, not state', () => {
     const walk = (root: string): string[] =>
       readdirSync(root).flatMap((entry) => {
         const full = join(root, entry)
-        return statSync(full).isDirectory() ? walk(full) : [full]
+        const st = statSync(full)
+        // Regular files only: the hook server's own unix socket (issue #367) lives under the data
+        // dir too, and a socket is not readable state — opening it is ENXIO, not evidence.
+        return st.isDirectory() ? walk(full) : st.isFile() ? [full] : []
       })
     for (const file of walk(dir)) {
       // A persisted latch is a node that one filesystem accident can brick forever. Nothing on
--- a/src/core/agents/hook-sandbox-hint-sh.ts
+++ b/src/core/agents/hook-sandbox-hint-sh.ts
@@ -1,0 +1,60 @@
+/**
+ * The Codex-sandbox self-diagnosis both generated sh shims share (issue #367).
+ *
+ * Codex's command sandbox denies network connect() outright when its network policy is disabled —
+ * on Linux a seccomp filter EPERMs `connect()` for EVERY address family (unix sockets included,
+ * with no allowlist), and on macOS the seatbelt base profile is `(deny default)` with no
+ * network-outbound allowance. Either way curl dies with exit != 0 / HTTP 000 while the nodeterm
+ * hook server is perfectly healthy, and the shims' generic "nodeterm unreachable" message sent
+ * agents (and users) off to relink or restart a server that was never the problem.
+ *
+ * Codex exports `CODEX_SANDBOX_NETWORK_DISABLED=1` into every sandboxed command on both
+ * platforms, so the shims can tell the two failures apart at the exact moment they happen. The
+ * remedies differ by platform:
+ *  - both: retry the SAME command once with command-specific escalated permissions (the sandbox
+ *    is per-command; an escalated retry runs outside it);
+ *  - macOS only (codex >= 0.146): `network.allow_unix_sockets = ["<hook sock>"]` in
+ *    ~/.codex/config.toml permanently re-opens exactly the hook socket inside the sandbox, no
+ *    escalation needed. Linux has no such allowlist, so that line would be a false promise there
+ *    — the fragment gates it on `uname -s` = Darwin as well as on the socket being advertised.
+ *
+ * The message text lives in exported constants so the agent-facing docs (skill bodies + AGENTS.md
+ * instruction blocks) can carry the same sentences and the parity tests can hold the two ends
+ * together; the sh fragment is exercised against real /bin/sh in the shim test suites.
+ */
+
+/** Both platforms, line 1: what actually failed, and what NOT to do about it. */
+export const CODEX_SANDBOX_BLOCKED_LINE =
+  "Codex's sandbox blocked this connection to nodeterm. nodeterm itself is running — do not relink or restart it."
+
+/** Both platforms, line 2: the one action that works everywhere. */
+export const CODEX_SANDBOX_RETRY_LINE =
+  'Retry the SAME command once with command-specific escalated permissions (approve the escalation prompt for this command).'
+
+/** macOS only, wrapped around the advertised hook socket path. */
+export const CODEX_SANDBOX_MACOS_BEFORE =
+  'Permanent fix on macOS: add network.allow_unix_sockets = ['
+export const CODEX_SANDBOX_MACOS_AFTER =
+  '] to ~/.codex/config.toml — the sandbox then allows this hook socket with no escalation.'
+
+/**
+ * POSIX-sh fragment defining `nt_codex_sandbox_hint`: prints the actionable error to stderr and
+ * returns 0 when the codex sandbox is the diagnosis, returns 1 (printing nothing) otherwise —
+ * so call sites read `nt_codex_sandbox_hint || echo "<the old generic message>" >&2` and the
+ * genuine-unreachable case keeps its original text to the byte.
+ */
+export const CODEX_SANDBOX_HINT_SH = `# Codex-sandbox self-diagnosis (issue #367). Codex exports CODEX_SANDBOX_NETWORK_DISABLED=1 into
+# every sandboxed command; when it is present, a dead transport (curl exit != 0 / HTTP 000) means
+# the SANDBOX denied connect() — Linux seccomp EPERMs every address family, macOS seatbelt is
+# deny-by-default — and the generic "unreachable" message would misdirect the agent into
+# relinking/restarting a healthy server. Prints the actionable error and returns 0 only in that
+# state; callers fall back to their own generic message on 1.
+nt_codex_sandbox_hint() {
+  [ -n "$CODEX_SANDBOX_NETWORK_DISABLED" ] || return 1
+  echo "${CODEX_SANDBOX_BLOCKED_LINE}" >&2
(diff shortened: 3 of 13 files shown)


eneskirca:
Reviewed in depth. The security design is sound and the #195 walk-back in `a7158cc6` is honest — the code matches what the docs now say, which is the opposite of this repo's "claims ahead of mechanism" failure mode. Measured live: socket **0600**, dir **0700** (chmod'd *before* bind, which is the right ordering and the comment explains why), primary path 29 bytes against a 103 budget. The handler is built once and passed to both `createServer` calls and nothing in it reads `remoteAddress`, so bearer + per-node auth are genuinely transport-agnostic — the socket is a strictly **narrower** door than the port beside it, not a second weaker one. Unlink-before-bind is fine too: `unlinkSync` on a symlink removes the link, not the target, and the 0700 dir means nothing hostile can be planted there.

**One blocking behavioural bug, in the headline feature itself.**

`hook-sandbox-hint-sh.ts:27,53` gates the hint on `CODEX_SANDBOX_NETWORK_DISABLED` being non-empty plus a dead transport. But **`stop()` unlinks the socket and never deletes `hook-endpoint.env`** (`hook-server.ts:1076-1090`) — it is written at start and never removed. So in a routine state:

1. user quits nodeterm;
2. a tmux pane outlives the app (this repo's core design);
3. its codex node sources the stale endpoint file, gets a dead sock *and* a dead port, curl returns `000`;
4. the shim prints **"Codex's sandbox blocked this connection to nodeterm. nodeterm itself is running — do not relink or restart it."**

Both halves are false, and it explicitly forbids the one remedy that works. I reproduced this under real `/bin/sh` + real curl. Before this PR that same state printed the accurate "Could not reach nodeterm (control endpoint unreachable)" — so it is strictly worse in a state users will hit often. The agent-facing docs carry the same over-claim ("while nodeterm itself is healthy") in `canvas-control-core.ts:526` and `context-link-core.ts:109/249`.

Decisive one-liner: gate on `[ -S "$NODETERM_HOOK_SOCK" ]`. That's a `stat`, not a `connect`, so it works inside both sandboxes — **socket present + connect fails = sandbox; socket absent = the server really is gone.**

**CodeQL red is this PR's and will not clear itself.** Two new high alerts, `js/insecure-temporary-file` + `js/file-system-race`, both at `hook-server.sock.test.ts:116` (`fs.writeFileSync(sock, '')` on a path tainted from `os.tmpdir()`). It is a test-only false positive — the parent is a `mkdtempSync` 0700 dir — and the pattern is endemic (105 test files use `mkdtempSync`; 23 alerts for these two rules are already open on main). But it's new code in this diff, so it's what's holding `BLOCKED`. Your call: dismiss as "won't fix" (keep it ≤280 chars, per the #244 experience) or restructure the test.

Two smaller ones:

- **`hook-server.ts:792-798` leaks a live listener.** If `chmodSync(p, 0o600)` throws after a successful `listen`, the catch nulls `unixServer`/`sockPath` but never `srv.close()`. The socket stays bound at the umask default (0755, saved only by the 0700 dir), `stop()` won't unlink it because `sockPath` is `''`, and a stop→start cycle leaves two listeners racing the path. One `try { srv.close() } catch {}`.
- **`hook-sock-path.ts:23-29` — the fallback isn't SUN_LEN-checked.** The primary is budget-checked, the fallback is returned unconditionally. Measured: a 76-char home gives a 118-byte fallback against the 103 budget. It fails safe (bind throws → TCP-only) but silently, and the macOS codex remedy quietly becomes unavailable with no diagnostic.

One thing worth an explicit decision rather than a fix: local hook traffic now silently migrates TCP → socket with **no same-instance fallback**. `nt_request_post` is `if sock; elif port`, and `nt_send_request` only fails over to *other* endpoint files (`nt_adopt` clears both vars first). Before this PR local sessions had no sock and always used the port; now anything that breaks the socket while TCP is healthy loses hook events — badges, context meter — instead of degrading to the live port sitting right there.

Repo-convention nit: per CLAUDE.md's own "two docs, two audiences" rule, the new facts (local sessions now advertise `NODETERM_HOOK_SOCK`; 0700 dir / 0600 socket; the socket *path* is deliberately argv-safe where the token was not) live only in source comments. The argv-safety reasoning belongs in the "a credential never rides argv" section.

Verification: trial-merged onto current main — clean. Typecheck clean. PR-touched suites 214/214; `src/core/agents` + context-link 691/691; `src/main`+`src/shared` 2139 pass with 3 failures all in `node-pty-patch.test.ts` (this box's node_modules drift, not yours). `quality` and `quality-windows` are **green on the current head** — the earlier reds were a pre-merge SHA and don't reproduce.

Want me to push the `[ -S ... ]` gate and the `srv.close()` fix? The branch isn't maintainer-modifiable from here, so I'd need to do it as a follow-up PR onto your branch.

eneskirca:
Test-only false positive. Line 116 (`fs.writeFileSync(sock, '')`) deliberately fabricates a crash-left stale socket file to prove that `startUnixListener` unlinks-then-rebinds (the StreamLocalBindUnlink lesson). The path is the server's OWN socket under a test-fixture userDataDir — not untrusted input — and production has no TOCTOU here: `startUnixListener` unlinks unconditionally then binds, a lost race just fails bind with EADDRINUSE (fail-safe, TCP stays), and the socket dir is 0700 so no other local user can create there. Same class as the already-dismissed test-fixture alerts #65/#66/#70. Resolving.

PR #370, opened 2026-08-21
fix(windows): reject empty vswhere results in bootstrap

HopeCollector:
## Why

PR #302 added a fresh-machine Windows bootstrap, but its Visual Studio probe trusted only `vswhere`'s exit code. A valid `vswhere -property installationPath` query exits 0 even when it matches no installation, so the bootstrap could print `[OK]` and defer the real failure to `node-gyp` / `electron-rebuild`.

I reproduced that false-positive on a native Windows checkout before #302 merged: https://github.com/eneskirca/nodeterm/pull/302#issuecomment-5336573972.

## What changed

- capture the first matching Visual Studio path before deciding success
- when capture is empty, retry only to classify `[FAILED]` query execution versus `[MISSING]` C++ workload
- validate that a reported installation path exists before printing `[OK]`
- use `-latest` to choose the newest matching installation deterministically
- expose a probe-only mode so Windows CI can exercise the real batch script without elevation checks or `npm ci`
- gate fixture injection on both probe-only mode and `NODETERM_BOOTSTRAP_TESTING=1`, while clearing inherited internal override state before dispatch
- add real `cmd.exe` coverage for missing vswhere, override rejection, empty output, stable and stateful query failures, nonexistent paths, multi-line first-wins output, and spaced valid paths
- document the probe-only exception in `CLAUDE.md` and run the suite in `quality-windows`

The healthy path invokes `vswhere` once. Only an empty capture triggers a status-only retry, keeping "query failed" distinct from "query succeeded with no match" without a temporary file. `call` is intentional because production invokes `vswhere.exe`, while the test fixture is a `.cmd` file.

Surface decision: this affects the Windows desktop bootstrap only. Server Edition and the mobile companion are not applicable.

## Verification

- `npm run typecheck`
- `npm run host:build`
- `npm run build`
- current `quality-windows` Vitest target list: 27 files, 478 passed, 6 skipped
- focused bootstrap suite: 9/9 passed under real `cmd.exe`
- nine restored mutation checks each made the suite fail, covering error classification, empty/path guards, missing-vswhere handling, first-wins capture, inherited state clearing, exact sentinel gating, dispatch placement, and installation-variable reset
- full local bootstrap stops at the VS probe with `[MISSING] No Visual Studio installation with the C++ build tools component was found.` instead of continuing to `npm ci`
- latest `main` merged at `4b6cfd94`; the sole `CLAUDE.md` conflict was resolved by retaining both the current release-wiring documentation and this PR's probe-only safety documentation

## Not verified

- successful full bootstrap on a machine with a real Visual Studio C++ workload
- native dependency rebuild, `dist:win`, or installer smoke (this machine does not have the required VS workload)
- the full cross-platform `npm test` suite locally on Windows; the repository still has POSIX-only fixtures, so the Ubuntu `quality` job remains authoritative for that suite


eneskirca:
Reviewed this closely — the diagnosis is exactly right and the batch is well-written CMD. `vswhere -property installationPath` really does exit 0 with empty stdout when nothing matches, so the baseline's errorlevel-only check was dead code for the condition it claimed to guard (the vswhere-absent case is already handled separately one branch up), and your #302 reproduction lines up with the deferred node-gyp failure.

Things I verified rather than took on trust: quoting is correct at every site including the `Program Files` case; `for /f` + `if not defined` first-wins needs no delayed expansion; `exit /b 0` fences the subroutines; `setlocal EnableExtensions` is present. The security question was the crux for me and it holds — `_NODETERM_VSWHERE_OVERRIDE` has exactly one writer, double-gated on the argv flag *and* `NODETERM_BOOTSTRAP_TESTING`, with an unconditional clear above the branch inside `setlocal`, so an inherited variable cannot redirect the production probe (and probe-only mode `exit /b`s before `npm ci` regardless). The test is a real one — real script, real `cmd.exe`, fake vswhere, spaced paths — and it skips **visibly** on non-Windows (I ran it here on Linux: 4 reported as skipped, no silent pass). CI wiring is right, and putting the probe-only branch *before* the elevation refusal is what makes it work on GitHub's elevated runners.

Three asks before I approve:

**1. (the one I'd push on) The second vswhere call's own failure is laundered into "no installation found".** `for /f` discards the inner exit status and `2^>nul` swallows stderr, so if call #1 succeeds and call #2 fails — AV scan, the VS updater swapping the `Installer` directory mid-run — `VS_INSTALLATION` stays empty and the script prints `[MISSING] No Visual Studio installation…` plus a winget command, on a machine that has one. That is the same "a read that could not run is not evidence of absence" error this PR exists to fix, moved one call later; CLAUDE.md states that rule explicitly for git reads and the session-memory sweep.

Suggested restructure, which also drops the extra call from the happy path: run the `for /f` capture **first**; only if `VS_INSTALLATION` is empty, run the status-only probe to decide `[FAILED]` vs `[MISSING]`. One invocation on a healthy machine, two only when something is already wrong, both failure modes classified honestly. It flips the invocation counts your tests pin (1 for the happy path, 2 for the empty one), which is arguably the better pin anyway.

**2. Pin the override-clearing.** The safety argument in your description is "the full bootstrap always clears the internal override" — but no test asserts it, so a refactor that moves that `set` below the `goto` leaves the suite green. One negative case: spawn with `_NODETERM_VSWHERE_OVERRIDE` preset and `NODETERM_BOOTSTRAP_TESTING` unset, assert the fixture was never invoked. Since it's the reason the mode is safe, it should be the best-tested line in the diff. (Small companion: `readFileSync(invocationLog)` throws ENOENT when the fixture is never invoked — default it to `''` so that case is expressible.)

**3. A line in CLAUDE.md.** The paragraph documenting `bootstrap-windows.bat` says it "refuses to run elevated", and `--check-vs-build-tools` deliberately bypasses that *and* is now the only thing CI exercises. Per the repo's Conventions, that belongs in the docs too.

Smaller notes, take or leave: `-latest` is a behavior change worth a bullet in the description (good change — deterministic newest-install pick); the `call` on what is an `.exe` in production exists only so the fixture can be a `.cmd`, which is worth a clause in the `rem` above it or someone will "clean it up"; and two cheap cases are missing — the vswhere-absent branch, and multi-line vswhere output pinning the first-wins.

One thing for your device checklist: the happy path (a machine with a real VS C++ workload) is the one path nobody has run, and it is the first time the reported path gets inspected — worth confirming a real install still prints `[OK]`.

Nice first contribution — the mutation runs in your description are exactly the house discipline.

HopeCollector:
Thanks for the close review — all three asks are addressed in `fdc76b6b` (and the branch is rebased onto current `main`).

1. The probe now captures first and runs the status-only retry only after an empty capture. A healthy path invokes `vswhere` once; empty/error paths invoke it twice. The stateful regression fixture returns rc 0 with empty stdout first, then rc 7, and asserts that the later failure is `[FAILED]`, never laundered into `[MISSING]`.
2. The negative override case now preloads both the inherited internal override and the public fixture path, with the sentinel unset and set to `0`; the fixture must have zero invocations. Missing invocation logs default only on `ENOENT`. An additional narrow source-order guard pins the clear above the check-only dispatch and keeps fixture assignment inside the check-only boundary.
3. `CLAUDE.md` now documents that check-only branches before the elevation refusal, runs only the VS probe, exits before Node/Python/`npm ci`, and requires the explicit testing sentinel for fixture injection.

I also took the smaller suggestions: the PR body calls out `-latest`, the batch explains why `call` is intentional, and tests cover absent `vswhere` plus multi-line first-wins output. The helper preloads `VS_INSTALLATION` as hostile inherited state as well.

Local verification after restoring every mutation:

- focused real-`cmd.exe` suite: 9/9
- exact `quality-windows` Vitest set: 27 files, 473 passed, 6 skipped
- typecheck, host build, and production build pass
- nine mutation variants were each killed, including weakened sentinel gating and moving the override clear below dispatch

I still cannot check the real-workload happy path or native rebuild / `dist:win` on this device because it has no Visual Studio C++ workload; that remains explicitly listed under **Not verified**.

eneskirca:
Re-reviewed `fdc76b6b` — all three asks are addressed, and I verified them on the diff and by measurement rather than from the reply:

1. **Capture-first restructure**: the `for /f` capture now runs first and the status-only probe only after an empty capture, so a healthy machine invokes vswhere once and a later query failure is classified `[FAILED]`, never laundered into `[MISSING]`. The stateful fixture test (`rc 0 empty → rc 7`) pins exactly the laundering case I described, and the invocation-count pins flipped to 1/2 as predicted.
2. **Override safety is now the best-tested line in the diff**: the negative case preloads the inherited internal override with the sentinel unset/`0` and asserts zero fixture invocations; the source-order guard pins the clear above the check-only dispatch; the log read defaults only on `ENOENT`.
3. **CLAUDE.md** documents the `--check-vs-build-tools` exception to the elevation refusal, the sentinel gate included.

The smaller notes are all taken too — `-latest` in the PR body, the `call` rem, the vswhere-absent and multi-line first-wins cases.

Measured on the branch tip: `npm run typecheck` clean; the suite skips **visibly** on Linux (9 reported skipped, no silent pass); `quality-windows` ran the real-`cmd.exe` suite green on CI.

One logistics item before merge: `main` moved and the branch is now CONFLICTING on a single CLAUDE.md prose paragraph (the Windows release wiring bullet landed on `main` in the same paragraph your docs line touches). It resolves textually — both edits keep. One more merge of `main` and this is ready.

HopeCollector:
Merged current `main` (`4b6cfd94`) in `32243e92` and resolved the single `CLAUDE.md` conflict.

The resolution keeps both sides as requested: `main`'s current Windows release/signing and no-update-feed documentation, plus this PR's `--check-vs-build-tools`, elevation-bypass boundary, early exit before Node/Python/`npm ci`, and exact testing-sentinel documentation. I also dropped the stale wording that still listed the now-landed Windows release job as a follow-up.

Relative to the updated `main`, the PR still contains only the intended four files:

- `.github/workflows/ci.yml`
- `CLAUDE.md`
- `bootstrap-windows.bat`
- `src/main/bootstrap-windows.test.ts`

Post-merge local verification:

- focused real-`cmd.exe` bootstrap suite: 9/9
- exact current `quality-windows` Vitest command: 27 files, 478 passed, 6 skipped
- `npm run typecheck`
- `npm run host:build`
- `npm run build`

The branch was pushed normally (no history rewrite).

eneskirca:
The merge in `32243e92` resolves the CLAUDE.md conflict exactly right — both prose edits kept (main's Windows-release wiring paragraph and this PR's `--check-vs-build-tools` exception). Verified on the tip: `npm run typecheck` clean, and the branch is mergeable against current main. CI hasn't run on the new merge commit yet (fork workflow awaiting approval) — we'll kick it. Addressed, ready — thanks for the quick turnaround.

PR #371, opened 2026-08-21
feat: reopen last closed project/node with Cmd+Shift+T

gor3a:
## Summary

- Adds `Cmd+Shift+T` (`app.reopenLastClosed`) to reopen the most recently closed project tab, or recreate the most recently deleted node batch (terminal, agent, sticky, editor, video, diff, web, browser, dino) — whichever happened more recently. Repeated presses walk further back through a single chronological history stack shared by both kinds of "close" events.
- History is in-memory only per app run (capped at 10 entries), never persisted — matches a browser's own "reopen closed tab" convention.
- A recreated node is always a fresh node/session (new id), rebuilt through the app's existing `create*Node` factories rather than resurrecting the killed tmux session — so an agent node gets a correctly assembled launch command and permission mode, an SSH node's trust provenance is preserved verbatim (never re-derived), etc.
- The reopen decision logic (which branch to take, staleness checks, active-vs-non-active-project routing) is extracted into a pure, unit-tested `planReopen` function rather than living untestable inside `Canvas.tsx`.

Known limitation: `Cmd/Ctrl+Shift+T` is a browser-reserved chord (Chrome's own "reopen closed tab"), so it does not reach the page in Server Edition — this PR documents that as Desktop-only rather than claiming full coverage.

## Test plan

- [x] `npm run build` succeeds
- [x] New unit tests: `reopenHistory.test.ts`, `reopenNode.test.ts`, `reopenPlan.test.ts` (covers recency interleaving across entry kinds, stale-entry skip, batch restore, cross-project routing)
- [x] `keybindings.test.ts`, `keybindingOverrides.test.ts`, `canvas-wiring.test.tsx`, `ShortcutsSection.test.tsx` all green
- [x] Full suite run; all failures traced to pre-existing, environment-dependent real-tmux/pty/session-host integration tests unrelated to this change (none touch `reopen*`, `Canvas.tsx`, or `keybindings.*`)

### DIFF
--- a/src/renderer/canvas/Canvas.tsx
+++ b/src/renderer/canvas/Canvas.tsx
@@ -346,6 +346,8 @@ import { buildBackgroundLinkMaps, buildContextLinkNote, buildLinkMap, buildNoteP
 import { dependencyEdges, launchesToFire, unmetDeps, type ArmedNode } from '../lib/pendingLaunch'
 import { freeSpot } from '../lib/placement'
 import { pushSessionRename } from '../lib/sessionRename'
+import { useReopenHistory } from '../state/reopenHistory'
+import { snapshotNode, recreateNodeFromSnapshot } from '../lib/reopenNode'
 import { oneLine } from '@shared/one-line'
 import { parseLenses, verifyLensPrompt, verifySynthesisPrompt } from '../lib/verifyPanel'
 import { useSettings } from '../state/settings'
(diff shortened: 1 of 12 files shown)


eneskirca:
Keybinding slice reviewed: clean. `Cmd+Shift+T` collides with nothing in the registry, the dispatch-map source pin and the applyNodeMutation race pin are exactly the house discipline, and the test ripples (computed unassigned count, whole-table pin) are handled.

One product-level suggestion, non-blocking: for an **agent** node, the reopen plan could carry the node's last `sessionId` (it's in the agentStatus localStorage at delete time) and rebuild `initialCommand` as that agent's resume command instead of a bare launch. Reopening a deleted Claude node would then bring the *conversation* back, not just an empty CLI — closer to the browser's "the tab comes back with its state" promise this feature borrows. Fine as a follow-up.

PR #376, opened 2026-08-21
Directional node focus: move between nodes with Cmd and an arrow key

EliJaghab:
## Summary

`⌘←` `⌘→` `⌘↑` `⌘↓` hand the keyboard to the node in that direction. It frames that node and drops you straight into typing, so you can cross a canvas of terminals without touching the mouse. Off macOS the default is `Ctrl+Shift+arrow`.

Getting around the canvas is genuinely challenging once you have more than a few terminals open, and that's what this is for. I'm trying to get nodeterm closer to how Ghostty navigation feels, where you move between splits with the keyboard and never reach for the pointer.

## Why this matters

Reaching another terminal is a mouse job today: click it, or hover it for ~0.6s. If it's off screen you pan there first, and panning is the part that hurts. The xterm container carries `nodrag nowheel` (`TerminalNode.tsx:4703`), so over a terminal the wheel is that terminal's scrollback and a drag is not a pan. On a busy canvas you're hunting for a patch of empty space to drag from before you can even see where you're going. Multiplexers solved this years ago (`goto_split`, `select-pane -L`) and the canvas already has the geometry to do the same.

## How the direction gets resolved

Nodes aren't a tree of splits, so "the node to the left" has to be inferred. Nearest-by-distance is the obvious rule and the wrong one, because the closest node is very often the one diagonally up and left. Candidates are ranked in two tiers instead:

- **Tier 0**: the candidate's perpendicular span overlaps the origin's, so a user would call it "the same row" (horizontal move) or "the same column" (vertical). Nearest along the axis wins.
- **Tier 1**: no overlap, reachable only diagonally. Scored by axial distance plus a weighted sideways penalty, so a nearly aligned node beats a closer but far off-axis one. Without this tier a node sitting alone on its row is a dead end in every direction, which reads as the feature being broken rather than as a boundary.

Ties break on node id, so one canvas always answers the same way. There's no wrap-around, matching `goto_split`.

Every rect goes through `nodeFocus`'s `absolutePosition` rather than `node.position`, because a grouped node's position is relative to its frame. Read raw, a terminal at (24, 60) inside a frame parked at (4000, 0) looks like the leftmost node on the canvas and `⌘←` flies to it from anywhere.

The origin is the node the keyboard is actually in, read off the focused element rather than off `selected`. Hover-dwell focus hands the keyboard to a terminal without selecting it, so a selection-only origin would walk from whichever node was last clicked. Selection is the fallback for a canvas driven with no terminal focused, and the node nearest the view center seeds a canvas with neither. Group frames are never destinations, since a frame's rect contains the nodes you actually want and would win every direction from inside itself.

## Why the non-mac default isn't Ctrl+arrow

`allowInTerminal` has to be on, because the hand that wants to move is already typing in a terminal. That means the chord has to be one no shell wants. `Ctrl+arrow` is readline's word jump. `Ctrl+Alt+arrow` is out for a different reason: the `speech.dictation` default is the modifier-only chord `Cmd+Alt`, which resolves to Ctrl+Alt held on those platforms, so the hold listener would start recording on the way to the arrow. `Ctrl+Shift+arrow` is what's left that nothing else owns.

## Surfaces

- **Desktop**: the feature. `⌘arrow` reaches the page: `keydown-intercept.ts` describes itself as the closed list of main-intercepted chords (`⌘M`, `⌘W`, `⌘0`), and `buildAppMenu`'s only explicit accelerators are `CmdOrCtrl+,` and `CmdOrCtrl+Shift+B`, the rest being standard roles.
- **Server Edition**: same renderer, same dispatcher, so it works there, with one thing I couldn't settle. On macOS browsers `⌘←` and `⌘→` are history back and forward, and I don't know whether the page can `preventDefault` them the way it can `⌘0`. If it can't, the fix looks like the existing "Jump to project" precedent in `ShortcutsPanel` (drop the row under `isBrowserRuntime()`) plus a browser default that isn't `⌘arrow`. Glad to add that if you know the answer.
- **Mobile companion**: not applicable, there's no canvas to walk. Flagging @eneskirca in case you want an equivalent there.

## Tests

New unit tests cover the aligned grid, the diagonal tier and its sideways penalty, group frames, grouped children in their parent's coordinate space (both directions), a `parentId` cycle, unmeasured nodes, non-finite positions, the not-on-this-canvas case, tie-breaking and the seed. The registry pin test carries the four new rows, and `FLOW_NODE_CLASS` is pinned against the installed xyflow dist the same way `keyContext.ts` pins xterm's helper-textarea class.

`npm run typecheck` is clean and the renderer and shared suites pass.

**Guards are mutation-tested.** Seven mutations, each killing at least one test: group exclusion removed, non-finite center guard removed, `AXIS_EPSILON` to 0, tiers collapsed, sideways penalty removed, absolute position bypassed, id tie-break removed.

Two fixtures did not discriminate on the first pass, which is worth naming since it is the trap the guide warns about. The NaN candidate sat in a different tier from the good node, so `tier` decided before `cost` ever ran and the guard looked unnecessary. The grouped node sat to the right of the top-level node in both the raw and the absolute reading, so the coordinate-space case passed with `absolutePosition` bypassed. Both fixtures now disagree between the two readings.

The one test that reads source text rather than behaviour is the `FLOW_NODE_CLASS` pin against the installed xyflow dist. That is deliberate, and modeled on `keyContext.test.ts`, which pins xterm's helper-textarea class the same way.

The two refusals every canvas chord here needs both come from the registry rather than from new code: `scope: 'canvas'` is what the resolver refuses while the kanban board is open (`keybindings.ts:517`), and leaving `allowWhileTyping` off is what refuses it while the user is typing (`keybindings.ts:512`).

## What I did not verify

- Nobody has driven this in a running build. The geometry is unit-tested and typecheck is clean, but I have not pressed the chord in the app.
- Whether a macOS browser lets the page `preventDefault` on `⌘←` and `⌘→`, which is the Server Edition question above.
- A fully green suite. Five files under `src/core` and `src/main` fail on my machine with and without this change (real tmux, node-pty spawn, remote shell writes), so I could not use the suite as evidence on its own and have treated those as local environment rather than related.


eneskirca:
Reviewed the keybinding dimension — this is exactly how a new chord should enter the app post-registry: proper `CommandId` + definitions with the flags argued in a comment, no second dispatcher, and the off-mac default deliberately routed around both readline's Ctrl+Arrow and the dictation hold-chord. Nice.

Two questions before merge:

1. **mac, from inside a focused terminal** — the feature's home case. Did you verify on-device that `⌘+Arrow` keydowns actually *bubble out of a focused xterm* to the window dispatcher? xterm.js consumes some combos in its own keydown handler before they ever bubble; if it `preventDefault`s meta+arrows, `allowInTerminal` never gets a chance and the gesture works everywhere except the one place it's for. (If you tested it live on a Mac, saying so in the PR body is enough.)
2. **off-mac, `Ctrl+Shift+Arrow` + `allowInTerminal`** — this steals the `CSI 1;6 A-D` modified-arrow sequences from every TUI in the pane (some editors and agent CLIs use shifted/ctrl arrows for selection). Probably the right trade for a multiplexer-style gesture, and it's remappable — but worth one sentence in the PR body so it reads as a decision rather than an accident.

CI note: first-contribution workflows needed maintainer approval — kicked them off, the `quality` check should report shortly.

eneskirca:
Took the open questions into the code and answered both from xterm's own source, then pushed a fix to this branch (maintainer edit):

**Q1 (mac, from inside a terminal): confirmed WORKING.** xterm's arrow cases open with `if (e.metaKey) break` — meta+arrow is never processed or cancelled, so ⌘+Arrow bubbles to the window dispatcher exactly as this PR needs.

**Q2 (off-mac, from inside a terminal): confirmed BROKEN before the pushed fix.** For modified arrows xterm produces `CSI 1;N x`, writes it to the pty and CANCELS the keydown — and the dispatcher bails on `defaultPrevented` — so `Ctrl+Shift+Arrow` never reached it from the one place directional focus targets. Not a flaw in this PR's own code: the collision lives in the terminal key path, which this branch didn't touch.

**The pushed fix** (`fix(terminal): registry chords xterm would consume now bubble to the dispatcher`): `terminalKeyAction` gains a trailing `registryOwns` param + a `'bubble'` outcome (return false from the custom key handler with NO preventDefault), decided by a new `terminalChordBubbles` helper that asks the live resolver itself — so terminal-first policy (chord stays with the shell), kanban context (modal terminals keep their bytes) and user remaps all apply for free, and terminal-scope commands are excluded so copy/find keep their owners. Copy chords, Shift+Enter and the project-jump swallow are checked first. Also merged current main (clean). Full suite 8833 + typecheck green.

This generalizes beyond this PR: any future allowInTerminal command whose chord xterm's keymap consumes now works from terminals by construction. Thanks for the unusually well-argued chord selection — the registry comment explaining why Ctrl+Arrow and Ctrl+Alt+Arrow were rejected is exactly the house style. Auto-merge armed; @eneskirca can disable if he wants a device pass first.

EliJaghab:
been using directional focus since the merge and it works great, cmd+arrow is already muscle memory. thanks for the quick review

PR #378, opened 2026-08-21
feat(tabbar): pin the New project + so it cannot scroll away

### LINKED ISSUE #375: Tab bar: pin the New project (+) button instead of letting it scroll out of view
## The problem

I work across a lot of projects — 10 open tabs is a normal day here. Once the tab strip fills the width of the title bar, the `+` button for a new project ends up at the far right, past every tab, and in practice it scrolls out of view entirely.

It is not just "far away": the button lives *inside* the scrolling container, so it moves with the tabs.

```jsx
<div className="tabbar__tabs" onWheel={...}>
  {projects.map(p => <div className="tab" ... />)}
  <button className="tab__add" title="New project" onClick={onOpenWelcome}>+</button>
</div>
```

```css
.tabbar__tabs {
  display: flex;
  overflow-x: auto;
  scrollbar-width: none;   /* and ::-webkit-scrollbar { display: none } */
}
```

So with enough projects, creating a new one means: scroll the strip right, hunt for the `+`, click it. And because the scrollbar is deliberately hidden (understandably — a native scrollbar would look foreign in a title bar), there is no affordance telling you the button is off to the right at all. It reads as "the button is gone".

## Suggestion

Pin the `+` outside the scroller: make it a sibling of `.tabbar__tabs` rather than its last child, with `flex-shrink: 0`, so it keeps a fixed position at the right edge of the bar no matter how many projects are open. The tabs keep scrolling underneath it; the button stops moving. That is a small change and it makes the target position predictable, which is most of the value — you stop hunting for it.

A couple of things that could go with it, in rough order of how much they'd help:

- **Make the overflow visible.** A subtle fade or chevron at the edge of the strip when it is scrollable, so it is clear there are tabs (and previously the button) beyond the edge.
- **Give "New project" a keyboard shortcut.** There is a `new-project` entry in the ⌘K palette already, which is a good escape hatch, but nothing on the keyboard directly and nothing that hints the palette has it.

## Possibly unintended: the two entry points do different things

While looking at this — the `+` button runs `onOpenWelcome`, while the ⌘K palette entry labelled "New project" runs `addProject()` directly. Two paths with the same name that don't land in the same place. I don't know which is the intended behaviour for each, so flagging rather than proposing a fix; if they are meant to differ, the palette label could say so.

## Environment

nodeterm v0.3.2, macOS. Happy to attach a screenshot of a full tab bar if useful.

dev-krish-xyz:
## What

The New project `+` no longer lives inside the scrolling tab strip.

With enough open projects it sat at the far right of `.tabbar__tabs` (`overflow-x: auto`, scrollbar hidden) and scrolled out of view with no hint it was still there ([#375](https://github.com/eneskirca/nodeterm/issues/375)). It is now a sibling of the pill, parked on the right of a `.tabbar__projects` row. Tabs keep scrolling; `+` does not.

`min-width: 0` on the group and the pill is load-bearing: without it the flex item grows with the tabs and shoves `+` off the window — the same bug, one layer out. Dropping a tab on `+` still means "move to the end", as it did when the button lived in the scroller.

Not in this PR (flagged in the issue as follow-ups): overflow fade/chevron, a dedicated New-project shortcut, and the Welcome vs ⌘K `addProject()` mismatch.

## Surfaces

- **Desktop** — full.
- **Server Edition** — full (pure renderer).
- **Mobile companion** — N/A. There is no tab bar there. Flagging for @eneskirca rather than assuming.

## Testing

`npm test -- src/renderer/styles.tabbar.test.ts` — pass. Removing `min-width: 0` from `.tabbar__projects` turned that test red.

A TabBar jsdom mount was not kept: TabBar now pulls `lucide-react` via `ProjectGlyph`, which this vitest setup cannot resolve.

**Not verified:** clicking through a full tab strip in a running app.

Closes #375.


eneskirca:
Reviewed this in depth — the diagnosis is right and so is the direction. I reproduced the bug: baseline at 1280 px with 12 projects puts the `+` at x=1374, i.e. **116 px past the window edge**, invisible; with your change it sits at x=1240 and is on screen. #375 is genuinely fixed.

Two things need changing before it can go in, both measured in headless Chrome against the real `styles.css` and the real markup rather than reasoned about:

**1. `flex: 1` on `.tabbar__tabs` inflates the pill for everyone with few projects.** `flex: 1` implies `flex-basis: 0`, so the bordered rounded pill stops hugging its tabs and eats all free space:

| window / projects | pill width (main) | pill width (this PR) |
|---|---|---|
| 1280 px, 2 projects | 236 px | **1071 px** |
| 1400 px, 3 projects | 335 px | **1191 px** |

A user with two projects gets a nearly-empty 1071 px capsule where a 236 px pill used to be. The CSS comment two lines above still says "Tabs as a single rounded pill of segments."

**2. The macOS window drag region is destroyed.** `.tabbar__projects` is both `flex: 1` and in the `-webkit-app-region: no-drag` list, so it covers every pixel right of the brand. Sampling `-webkit-app-region` at y=22, 1400 px, 3 projects:

- main: x=640/1020/1160/1240 → `drag` — about **900 px of draggable title bar**
- this PR: those points are `no-drag`; what's left is the bar's **12 px** right padding

On macOS (`titleBarStyle: 'hiddenInset'`) the window can then only be dragged by the wordmark and a 12 px sliver. That contradicts the very CLAUDE.md bullet this PR edits — "the tab bar is the drag region" — which the PR doesn't update either. (The wrapper's `no-drag` is also redundant: its only children are already `no-drag`.)

**Both fix with two deletions**, and I measured the result rather than guessing: drop `flex: 1` from `.tabbar__tabs` and drop `.tabbar__projects` from the `no-drag` list. Then at 1280 px / 2 projects the pill is 206 px and hugs as before, the `+` sits adjacent at x=376 and is visible; at 12 projects the pill fills and `+` is at x=1240; and x=640/1160 report `drag` again. As a bonus `flex-shrink: 0` on `.tab__add` becomes genuinely load-bearing that way (at a 300 px bar it stops the button squeezing 28→22 px), whereas as written it is dead code.

**3. The new test doesn't guard what you think.** `styles.tabbar.test.ts` greps rule bodies for declaration text, so it goes red for any deletion and blind to behavior. I deleted each declaration and measured:

| deleted | test | measured effect |
|---|---|---|
| `min-width: 0` on `.tabbar__projects` | RED | **none** at any width |
| `min-width: 0` on `.tabbar__tabs` | RED | **none** (a scroll container's `min-width: auto` already resolves to 0) |
| `flex-shrink: 0` on `.tab__add` | RED | **none** (basis 0 ⇒ never shrinks) |
| **`flex: 1` on `.tabbar__tabs`** | **GREEN** | pill 1071→206 px, drag region restored |

So "`min-width: 0` is load-bearing: without it the flex item grows with the tabs and shoves `+` off the window" is measurably untrue — and the test is green for the one declaration that actually changes the layout.

**4. A real test is available — `TabBar.test.tsx` already mounts the full TabBar in jsdom on `main`.** The `lucide-react` resolution failure is this repo checkout's `node_modules` drift, not a vitest limitation; it's a declared dep and CI runs `npm ci`, which is why that suite is green in CI. I wrote the test you said wasn't possible, and it took two assertions:

```ts
expect(document.querySelector('.tabbar__tabs')!.contains(document.querySelector('.tab__add'))).toBe(false)
```

— passes on your branch, fails on main (`expected true to be false`). Plus a `dragstart` on tab 4 → `dragover`/`drop` on `.tab__add` asserting `onReorder('p3', null)`, which incidentally **confirms your "drop on + still means move to the end" claim empirically**. Both ran in ~1.6 s. That pair belongs here in place of the CSS grep.

Verified-good, no action needed: reorder drag-and-drop still works (per-tab handlers still `stopPropagation`, so extracting the end-zone is behavior-preserving); traffic lights are clear (`.tabbar__projects` starts at x=165, after the brand, and `padding-left: 86px` is untouched); the caret menu is portaled off `getBoundingClientRect` so it follows the tab; theme guards pass — the diff adds no color or size literal. Trial-merge onto current main is clean and the web typecheck is clean.

Tiny nit: the new wrapper `div` didn't re-indent its subtree, so the closing tag reads as mismatched.

Happy to re-review as soon as the two declarations come out — the rest of this is good work.

Abhirup0 (commented):
I pulled this branch and tested it locally, the fix works really well! The `+` button stays visible even with lots of tabs open.

Just noticed one small thing while testing drag and drop: if I drop a tab in the small 4px gap between the tab strip and the `+` button, the drop doesn't trigger because the parent container doesn't have the drag handlers. Putting `onDragOver` and `onDrop` on the `.tabbar__projects` wrapper as well might make dropping at the end feel even smoother.

Also thought it could be nice to add a quick test in `TabBar.test.tsx` to check that clicking `+` opens the welcome screen.

Really clean fix, thanks for working on this!

eneskirca:
Gentle nudge — the branch has since merged `main`, but the two declarations from my review are still in the diff, so the measured regressions stand:

- `flex: 1` on `.tabbar__tabs` still inflates the pill for anyone with few projects (1071 px capsule where a 236 px pill used to be at 1280 px / 2 projects), and
- `.tabbar__projects` is still in the `-webkit-app-region: no-drag` list, which leaves macOS with only the wordmark + a 12 px sliver to drag the window by.

Both fix with the two deletions (drop `flex: 1` from `.tabbar__tabs`, drop `.tabbar__projects` from the no-drag list) — the wrapper's `flex: 1` + `min-width: 0` stay, and the pin keeps working at every width I measured. And I'd still swap `styles.tabbar.test.ts` for the two `TabBar.test.tsx` assertions from the review (sibling-of-the-scroller + the drop-on-`+` reorder), which test the behavior instead of the declaration text — the CSS grep is green for the one declaration that actually changes the layout and red for three that don't.

The core of the PR is good and #375 is genuinely fixed — happy to re-review the moment those two lines come out.

eneskirca:
Re-measured `2fe90bd7` in headless Chrome against the real `styles.css` — the two deletions and the test swap are in, and the few-projects case is now exactly right: at 1280 px / 2 projects the pill hugs at **196 px**, the `+` sits at x=401, and x=640/1020/1160/1240 all report `drag` again. The jsdom tests are also the right kind (the sibling-of-the-scroller assertion + the welcome click), and moving the end-drop handlers to the wrapper covers the 4 px gap @Abhirup0 flagged. Trial-merge onto current `main` is clean.

Two things block it, both measured:

**1. CI is red on a one-line type annotation.** `TabBar.test.tsx:146` — `let onOpenWelcome: ReturnType<typeof vi.fn>` isn't assignable to the `() => void` prop under this repo's vitest types. `ReturnType<typeof vi.fn<() => void>>` (the pattern the caret-menu describe above already uses) clears `npm run typecheck`; I verified locally.

**2. The original bug regresses at many tabs.** The commit also deleted both `min-width: 0` lines, and at 1280 px / **12 projects** the `+` is off-window again — its right edge lands at **x=1382**. My earlier table said each `min-width: 0` was individually inert, but that was measured against the then-present `flex: 1` + `min-width: 0` on the pill; deleting **both** together re-opens the bug (the wrapper's `min-width: auto` now resolves against the pill's full content width, so the wrapper refuses to shrink and shoves the `+` out — issue #375 at a different layer). Restoring **`min-width: 0` on `.tabbar__projects`** fixes it, measured: 12 projects → pill 1036 px, `+` visible at x=1240; 2 projects unchanged (196 px, drag intact). So that line — comment and all — was load-bearing after all; my apologies for the over-broad table row.

Nice-to-have, not blocking: an end-drop reorder test (`dragstart` on a tab → `dragover`/`drop` on `.tabbar__projects`, asserting `onReorder(dragId, null)`) would pin the behavior that just changed surfaces. And one for the device checklist: the end-drop zone now lives on an element that is also the macOS title-bar **drag region** — jsdom can't tell whether Electron's drag-region hit-testing lets an HTML5 drop land on the wrapper's empty area (a drop on the `+` itself bubbles up from a no-drag child, so that path is safe either way). Worth one try on a real Mac.

With the type fix and the one restored line this is ready.

eneskirca:
Gentle ping — this is one type annotation and one restored `min-width: 0` away from ready (both in the measured comment above); happy to re-verify the moment they land.

eneskirca:
Thanks for the fix and for sticking with the measured version — to get this over the line we pushed the two remaining review items to your branch directly (`maintainerCanModify`), rather than asking a fourth time:

- `85303d9c` — restores `min-width: 0` on `.tabbar__projects` (with a comment saying why it is load-bearing: at 1280 px / 12 projects the `+` was back off-window at x=1382 without it) and fixes the CI typecheck failure in `TabBar.test.tsx` (`ReturnType<typeof vi.fn<() => void>>`, the pattern the caret-menu describe already uses).
- `a16b5b21` — merges current `main` (clean).

Locally: `npm run typecheck` clean, `TabBar.test.tsx` + the theme and event guards green (19/19). Nothing of yours was changed otherwise — the pill hugging, the drag region, the sibling-of-the-scroller test and the wrapper end-drop all stand as you wrote them. Once CI is green, @eneskirca makes the merge call.

PR #380, opened 2026-08-22
feat(agents): vanilla-env 'Restart on subscription' — relaunch an agent without gateway/inherited provider env

dkattan:
Gateway failures or inherited provider overrides can leave an agent pointed at an unavailable provider. Adds a per-node “Restart on subscription” action (Copilot: “Restart on Copilot defaults”) that recycles the session with gateway injection disabled and matching inherited provider variables removed. A global “Launch on subscription” setting applies the same behavior to fresh agent sessions.

Provider stripping requires an explicit agent identity. Plain terminals preserve their inherited credentials even when the global setting or `clearEnv` is enabled. Claude account directories and Copilot home directories also survive. Patterns match variable names only and are compiled behind `try/catch`; credentials are never added to command arguments.

The restart checks foreground agent identity before termination and recycles the shell so its environment can change. It clears the gateway-specific model override so the resumed agent uses its own default. The shared core covers desktop and Server Edition; mobile has no dedicated UI change.

Known behavior: inherited-environment stripping runs locally. SSH sessions drop nodeterm's gateway injection, but provider variables exported by the remote host can still apply. Explicit custom-agent environment values are merged after stripping and retain precedence.

Validation: the two plain-terminal credential regressions failed before the guard and pass afterward. All 39 PTY/vanilla tests pass on this prerequisite; the dependent three-way launch-mode branch passed 67 launch/settings tests and the combined stack passed 83 focused tests plus typecheck. Rebased onto upstream `main` at `e2e3e3479f24`; fresh hosted CI validates the published head.

---
This is **part 1 of 7 in a stack** made with GitButler:
- <kbd>&nbsp;7&nbsp;</kbd> #726
- <kbd>&nbsp;6&nbsp;</kbd> #725
- <kbd>&nbsp;5&nbsp;</kbd> #724
- <kbd>&nbsp;4&nbsp;</kbd> #723
- <kbd>&nbsp;3&nbsp;</kbd> #722
- <kbd>&nbsp;2&nbsp;</kbd> #715
- <kbd>&nbsp;1&nbsp;</kbd> #380 👈

### DIFF
--- a/src/core/pty-manager.ts
+++ b/src/core/pty-manager.ts
@@ -2776,7 +2776,9 @@ export class PtyManager {
     // already-running session — only the next fresh launch.
     const stripRe =
       options.clearEnv || this.getSettings().vanillaLaunchDefault
-        ? vanillaEnvStripPattern((options.agentId ?? 'claude') as AgentId)
+        ? options.agentId
+          ? vanillaEnvStripPattern(options.agentId)
+          : null
         : null
     const gatewayEnv =
       options.agentId && !stripRe
--- a/src/core/pty-manager.ts
+++ b/src/core/pty-manager.ts
@@ -99,7 +99,7 @@ import {
 } from './codex-identity-proxy'
 import { ensureNodeToken, ensureRemoteNodeToken, sweepNodeToken } from './agents/node-token-service'
 import { clearNode as clearNodeAgentStatus } from './agent-status-mirror'
-import { hasSharedIdentity, setCustomAgentBaseResolver, type AgentId } from '../shared/agents/config'
+import { hasSharedIdentity, setCustomAgentBaseResolver, vanillaEnvStripPattern, type AgentId } from '../shared/agents/config'
 import { findCustomAgent } from '../shared/agents/custom-agent'
 import { applyCustomAgentEnv, customAgentEnvArgs } from './custom-agent-env'
 import {
@@ -2767,17 +2767,34 @@ export class PtyManager {
     // A plain terminal has no agentId and must never receive provider credentials. The hook env's
     // historical Claude fallback does not apply here: gateway access is an explicit agent
     // capability, not a terminal default.
-    const gatewayEnv = options.agentId
-      ? modelGatewayEnv(
-          this.getSettings().modelGateway,
-          options.agentId,
-          options.agentModel,
-          process.env as Record<string, string | undefined>,
-          this.getModelGatewaySecret()
-        )
-      : {}
+    //
+    // "Restart on subscription" / `vanillaLaunchDefault`: strip the gateway + inherited provider env
+    // so the agent runs against its OWN default provider (Claude's subscription, Copilot's GitHub
+    // routing). `vanillaEnvStripPattern` resolves through the base harness; null ⇒ the agent has no
+    // strip set ⇒ no-op (gemini/grok/opencode are left alone). `buildPtyEnv` runs only in `spawnNew`
+    // (a fresh session), never on a warm reattach, so toggling the setting never strips an
+    // already-running session — only the next fresh launch.
+    const stripRe =
+      options.clearEnv || this.getSettings().vanillaLaunchDefault
+        ? vanillaEnvStripPattern((options.agentId ?? 'claude') as AgentId)
+        : null
+    const gatewayEnv =
+      options.agentId && !stripRe
+        ? modelGatewayEnv(
+            this.getSettings().modelGateway,
+            options.agentId,
+            options.agentModel,
+            process.env as Record<string, string | undefined>,
+            this.getModelGatewaySecret()
+          )
+        : {}
     if (!options.sshRemote) {
       for (const [k, v] of Object.entries(gatewayEnv)) env[k] = v
+      // Strip inherited provider vars so a vanilla session does not fall back to a LaunchAgent-set
+      // ANTHROPIC_BASE_URL instead of the subscription. Local only — see the note above.
+      if (stripRe) {
+        for (const k of Object.keys(env)) if (stripRe.test(k)) delete env[k]
+      }
     }
 
     // The OWNING project's env (`.nodeterm/settings.json`, local overlay + TRUSTED shared half —
--- a/src/core/pty-single-user.test.ts
+++ b/src/core/pty-single-user.test.ts
@@ -392,6 +392,41 @@ describe('SINGLE-USER REGRESSION: co-attach must not change the solo path', () =
     }
   })
 
+  it.each([
+    { vanillaLaunchDefault: true, clearEnv: undefined },
+    { vanillaLaunchDefault: false, clearEnv: true }
+  ])('preserves a plain terminal environment in subscription mode: %j', async (mode) => {
+    const inherited = {
+      ANTHROPIC_API_KEY: process.env.ANTHROPIC_API_KEY,
+      ANTHROPIC_BASE_URL: process.env.ANTHROPIC_BASE_URL
+    }
+    process.env.ANTHROPIC_API_KEY = 'plain-terminal-key'
+    process.env.ANTHROPIC_BASE_URL = 'https://plain-terminal-provider.example.test'
+    try {
+      const { PtyManager } = await import('./pty-manager')
+      const m = new PtyManager()
+      m.init(() => ({
+        ...DEFAULT_SETTINGS,
+        vanillaLaunchDefault: mode.vanillaLaunchDefault,
+        modelGateway: { baseUrl: 'https://bifrost.example.test', apiKey: 'vk-gateway' }
+      }))
+      m.registerIpc()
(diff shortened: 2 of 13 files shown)


eneskirca:
Reviewed in depth. This is a marked step up from the earlier stack — the security-sensitive core is genuinely right, and I verified the load-bearing parts rather than reading them.

**First, the CI red is not yours.** `quality-windows` failed in the `npm ci` step, before typecheck or a single test ran — `smart-whisper`'s `node-gyp rebuild` hitting the known undici crash (`AssertionError [ERR_ASSERTION]: assert(!this.paused)` in `client-h1.js`). Your diff touches zero dependency files, and `quality-windows` passed on the last four `main` runs. I've re-run the failed jobs.

**What I confirmed by testing, not by eye:**
- The regex only ever reaches `regex.test(k)` and a truthiness check — never eval'd, never interpolated into a command, compiled once behind `try/catch` with a malformed literal degrading to no strip.
- **The `CLAUDE_CONFIG_DIR` exclusion holds end-to-end.** Your own test sets an *inherited* one, but the load-bearing case is the **managed-account** path, where `pty-manager` writes `env.CLAUDE_CONFIG_DIR = accountDir` above the strip. I built that case (real `accountId`, account dir on disk, `clearEnv: true`): the account dir survives while `ANTHROPIC_BASE_URL`/`ANTHROPIC_AUTH_TOKEN` are gone. No wrong-identity risk.
- **`clearEnvEligibility` permitting `working`/`blocked` is defensible, not sloppy.** `terminateForeground` reads the foreground group's full argv and requires `isAgentPane(...) === 'agent'`, refusing on both `not-agent` and `unknown` — fail-closed, so the SIGTERM cannot land on vim or a build. It SIGTERMs with an ESRCH-polled grace rather than SIGKILLing, and the branch sits inside the same `guardConcurrentRestart` lock hibernation uses, so no new concurrency path opens.
- `vanillaLaunchDefault` defaults off and an untouched setting emits the byte-identical pre-feature env. The one-shot `clearEnv` is absent from the serializer field lists, so a stuck `true` cannot survive a reload.
- No new IPC surface, nothing on argv, and every `-` line is replaced by its extended version — no silent reverts.

**One blocking defect** — `src/core/pty-manager.ts:2631`:

```ts
? vanillaEnvStripPattern((options.agentId ?? 'claude') as AgentId)
```

A plain terminal has no `agentId`, so that `?? 'claude'` applies **claude's strip set to every non-agent shell** whenever the global setting is on. I proved it: spawning with no `agentId` and `vanillaLaunchDefault: true`, `ANTHROPIC_API_KEY` came back `undefined` where I'd set `'sk-users-own-key'`. It contradicts the settings copy ("every fresh **agent** session") and sits three lines under this file's own comment — *"A plain terminal has no agentId and must never receive provider credentials."* Scenario: user turns on "Launch on subscription", opens a plain terminal to run a script needing `ANTHROPIC_API_KEY`, and gets an auth failure with nothing pointing at the cause.

```ts
? options.agentId ? vanillaEnvStripPattern(options.agentId) : null
```

(For what it's worth the `?? 'claude'` is a copied local idiom from two other sites in the same file, so this reads as inherited rather than careless — but those sites aren't deleting the user's credentials.)

**Two honest-but-unsurfaced limitations I'd want a hint for:**

1. **SSH is half-functional.** The inherited strip is inside `if (!options.sshRemote)`. `gatewayEnv` correctly empties so nodeterm's injected vars drop, but the **host's own** exported `ANTHROPIC_BASE_URL` — the zshrc case the PR title advertises — survives. The menu row isn't gated on SSH (`session.source` is `'local'` for an SSH-project node; only `'relay'` is refused), so the action can SIGTERM a working turn, change nothing, and report "Restarted on subscription — conversation resumed."
2. **A claude-based custom agent declaring `ANTHROPIC_BASE_URL` gets the gateway back** after the strip. You document that precedence as intentional and I agree with it — but the same silent no-op-that-killed-your-turn applies.

Nits: `clearEnvEligibility(...)` is called three times per menu render in `Canvas.tsx:6487` (hoist to a const); the `!ok` hint always says "Nothing to resume yet…" though the function can also return `not-resumable` (unreachable only because of an enclosing early return); and the PR body states Desktop + Server Edition but not Mobile — genuinely N/A, but CLAUDE.md asks for the explicit call.

Trial-merged onto current main: clean. 411 tests across the targeted suites pass, 5441 in the broad run (the 2 failures are this box's known `node_modules` drift). Happy to approve once F1 is in.

eneskirca:
Checked the branch after the force-push + `cf3e0645`. The new commit is a genuine catch — dropping `agentModel` on the clearEnv recycle is right (a gateway-routed model id is meaningless once the gateway env is stripped, and the failed relaunch would have been hard to diagnose), and the `clearEnvEligibility` tests are welcome.

But the **blocking item from my review is still in place** — `src/core/pty-manager.ts:2631` still reads:

```ts
? vanillaEnvStripPattern((options.agentId ?? 'claude') as AgentId)
```

so with the global "Launch on subscription" setting on, a **plain terminal** (no `agentId`) still gets claude's strip set and loses the user's own `ANTHROPIC_API_KEY`/`ANTHROPIC_BASE_URL` — the case I reproduced in the review, and the settings copy still promises "every fresh **agent** session". The fix is the one-liner:

```ts
? options.agentId ? vanillaEnvStripPattern(options.agentId) : null
```

(plus a test with no `agentId` + `vanillaLaunchDefault: true` asserting the inherited key survives).

The two unsurfaced limitations also still stand as described — the menu row still refuses only `'relay'`, so on an SSH-project node the action can SIGTERM a working turn while the host's own exported `ANTHROPIC_BASE_URL` survives; a hint (or an `isRemoteSessionNode`-style gate) would close that honestly.

Happy to approve as soon as the `?? 'claude'` fallback is out — everything else I verified in the review still holds on the rebased branch.

eneskirca:
Gentle ping — the one-liner at `pty-manager.ts:2631` (`options.agentId ? vanillaEnvStripPattern(options.agentId) : null`) is the only thing between this and approval.

dkattan:
The [blocking plain-terminal case](https://github.com/eneskirca/nodeterm/pull/380#issuecomment-5498535595) is fixed in 2642b1d6d897980206678ac1b357dcdf6a00fb72. `PtyManager` now requires an explicit `agentId` before selecting a provider strip pattern, so a plain terminal retains its inherited `ANTHROPIC_API_KEY` and `ANTHROPIC_BASE_URL` with the global subscription default enabled, or with `clearEnv` set. Agent sessions keep the existing stripping behavior.

Both new regression cases failed before the fix and pass afterward. The prerequisite’s PTY/vanilla suites pass all 39 tests; the dependent three-way launch mode carries the same guard and passed 67 launch/settings tests. The combined stack passed 83 focused tests and typecheck. The branch is also rebased onto current upstream `main` (`e2e3e3479f24`).

dkattan:
@eneskirca I think this is good to go now.

eneskirca:
## Verdict: STILL VALID — MERGE-READY (only CI stands in the way)

This was opened in August, so the first question was whether `main` has since solved it. It has not.

**Measured against `main` (422767af):**

- `grep -rn "clearEnv\|vanilla" src/` → **zero hits**. No per-node or per-launch gateway opt-out exists anywhere.
- `ModelGatewaySettings` (`src/shared/agents/model-gateway.ts:6`) is `{baseUrl, apiKey}` only — global and all-or-nothing. `modelGatewayEnv` injects `ANTHROPIC_BASE_URL`/`ANTHROPIC_AUTH_TOKEN` for **every** claude node once a gateway is set; the model-less escape hatch exists only for copilot (`:296-301`).
- The inherited-env half is real too: `src/core/pty-manager.ts:2740-2743` strips `AUTH_ENV_STRIP` **only inside `if (accountDir)`**, so a system-account claude node still inherits a LaunchAgent-set `ANTHROPIC_BASE_URL`.

**And it is not stale code.** The branch head is dated 2026-09-09 and its merge-base is only 12 commits behind `main` — it has been actively rebased, and it is already written against main's model-gateway / `terminateForeground` / `restartShell` recycle machinery (`clearEnvEligibility` in `agent-restart.ts` explicitly reuses the model-switch interruption contract).

**Verification on the merged tree:** `git merge pr380` onto `main` is **clean** (13 files, no conflicts); `npm run typecheck` passes for both projects; its five focused suites pass **164 tests**.

The only blocker is CI: `quality-windows` fails, while `quality` (Linux), CodeQL and dependency-review all pass. That is what the UNSTABLE status is. Worth a look before merging, but it is not a review finding.

### One thing to decide first

Three PRs currently carry this branch's commits: **#715** (#380 + one commit), **#756** (#380 + #715 + one), and the **#722–#726** stack (rebased copies). Merging any of those merges #380 by side effect. Cleanest is to merge this one explicitly first, then the rest reduce to their own small deltas.

Issue #383, opened 2026-08-22
Ctrl+W closes the window instead of deleting a word in the terminal (macOS)


keif:
## Environment
- macOS, Apple Silicon (M5) MacBook Pro
- nodeterm, `main` branch
- Reproduced with a Claude Code agent terminal node open

## Steps to reproduce
1. Open a terminal (or agent) node on the canvas.
2. Click into the terminal and press **Ctrl+W** (not Cmd+W) — the standard readline/shell "delete previous word" (unix-word-rubout) binding used in bash/zsh.

## Expected
Ctrl+W deletes the word before the cursor in the shell. **Ctrl+U (kill line) works correctly today** — only Ctrl+W is affected.

## Actual
Ctrl+W closes the window instead, with no way to reopen it.

## What I checked before filing
- The suspect was the `node.close` keybinding (default `Cmd+W`), but its matching logic (`src/shared/shortcut.ts` `matchesShortcut` / `resolvedModifiers`) is **exact on all four modifier flags**, and on macOS `Cmd` resolves to `metaKey`, never `ctrlKey`. A literal Ctrl+W (ctrlKey, no metaKey) should not match a `Cmd+W` binding on macOS — this is true both in the main-process intercept (`src/main/keydown-intercept.ts`) and the renderer dispatcher (`src/renderer/lib/globalKeybindings.ts`).
- I checked my machine's `settings.json`: `keybindings` overrides are empty (`{}`) and `terminalShortcutPolicy` is unset (default `app-first`), so this isn't a user override rebinding `node.close` to Ctrl+W.
- So whatever intercepts Ctrl+W here doesn't appear to go through the documented keybinding registry path on the code's own terms. This is apparently [default Electron behavior](https://stackoverflow.com/questions/60350520/electron-browser-window-prevent-controlw-closing-window).
- Once *something* does route this to "close", `onCloseNode` in `src/renderer/canvas/Canvas.tsx` closes the selected node(s) if any are selected, otherwise calls `window.nodeTerminal.closeWindow()` — closing the entire window. That matches "no way to reopen it" (the app has no title-bar/close-role recovery when nothing else is open).

coljac:
Linux user here - same behaviour. I came to post this exact issue.

eneskirca:
Thanks — this is an unusually well-diagnosed report, and your reading of the matching code is correct. Splitting the two platforms, because the mechanisms are different:

[...]

PR #387, opened 2026-08-23
fix(keybindings): off-mac Ctrl+W reaches the shell while a terminal has focus

eneskirca:
Addresses the Windows/Linux half of #383 (confirmed by a Linux user there; the macOS report is still under diagnosis with the reporter — our code provably does not route Ctrl+W to close on darwin, so this PR deliberately does not auto-close the issue).

## The bug

Off-mac, `node.close`'s default chord is **Ctrl+W** — readline's kill-word in every shell. Two owners above the page claimed it even while the user was typing into a focused terminal: the Window ▸ Close menu role (its accelerator) and the main-process intercept. Result: delete-a-word closed the node, or with nothing selected the whole window, with no reopen affordance.

## The fix

One new pure predicate, `closeStandsDownInTerminal(isMac, terminalFocused)` — off-mac + terminal focused, **policy-independent** — consulted by both consumers:

- the **intercept** lets a resolved `close-node` decision fall through **untouched** (no `preventDefault`), so the keystroke reaches the page → xterm → the pty as kill-word;
- **`syncMenuForStandDown`** disables the Close menu item on top of the shared suspension list (a disabled item suppresses its accelerator).

Deliberately narrow: mac's ⌘W is not a shell key and is unaffected; ⌘/Ctrl+M and ⌘/Ctrl+0 keep firing under app-first; focus leaving the terminal restores Close immediately (same fail-safe mirror as the policy leg — a dead report resolves to "close works"). Closing while typing off-mac: the node's ×, or focus away first.

Pinned in `keydown-intercept.test.ts`: the truth table, the fall-through seam behavior (no prevent, no send; other intercepts unaffected; restores on focus loss), the byte-identical default for pre-existing callers, and a source-level pin that index.ts wires the predicate into BOTH consumers. CLAUDE.md's stand-down and window-chrome bullets updated. 8914 tests + typecheck green.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1684,6 +1684,14 @@ the Settings section and ShortcutsPanel start disagreeing about what a chord mea
     suspends the same items, so ⌘M / ⌘⇧B / ⌘, / off-mac Ctrl+W reach the recorder instead of the
     menu item that owns them; `menuStandsDown(false, …)` is `policyStandsDown(…)` by construction.
     The two INTERCEPT thunks stay independent parameters — only the menu ORs them.
+    **The CLOSE leg has one extra, policy-independent stand-down** (issue #383, off-mac only):
+    `closeStandsDownInTerminal(isMac, terminalFocused)` — off-mac `node.close`'s default chord is
+    Ctrl+W, readline's kill-word, so while a terminal has focus the close intercept lets the chord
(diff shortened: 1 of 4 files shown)


coljac:
The current version 0.3.3 still has this issue on Linux. Other ctrl combinations seem to go to the shell, ctrl-w now (thankfully) pops up a "Quit Nodeterm?" dialog instead of just closing.

PR #388, opened 2026-08-23
fix(packaging): hand-pack the macOS .icns — wrong-sized frames painted rainbow noise in the app launcher

### LINKED ISSUE #369: MacOS corrupted app icon
Image

Minor issue - missing app icon in the APp Launcher

eneskirca:
Fixes #369.

## Diagnosis (measured on the shipped artifact)

I unzipped the released `nodeterm-0.3.2-arm64-mac.zip` and parsed `Contents/Resources/icon.icns`. The frames electron-builder derived from `build/icon.png` are internally valid PNGs, **but two are in the wrong slots**:

| type | nominal size | shipped payload |
|------|--------------|-----------------|
| `ic13` | 128pt@2x = **256px** | the **512px** frame |
| `ic14` | 256pt@2x = **512px** | the **1024px** frame |

A consumer that trusts the icns *type's* nominal size and decodes the PNG into a buffer of that stride paints exactly the rainbow noise in the report's screenshot (I cropped and enlarged it — it's a stride-mismatch pattern, not a missing icon).

## Fix

`make-icon.mjs` now packs `build/icon.icns` itself — the same hand-rolled treatment the Windows `.ico` already gets in that script — using Apple's iconutil frame table (`icp4/icp5` + `ic07`–`ic14`, PNG payloads, retina variants included). The packer (`scripts/icon-pack.mjs`) **refuses** any frame whose PNG dimensions disagree with its type, so this class of skew fails the build instead of shipping. `mac.icon` points at the icns and electron-builder embeds it as-is.

## Tests

`src/main/icon-pack.test.ts`: pins the type→size table, packs real sharp-rendered frames and re-parses the container asserting every payload matches its type, and pins both refusals (mismatched `ic13` — the shipped bug — and a missing frame). All watched red first against a stub.

Generated icns verified locally: all 10 frames present, each at its nominal size; the small frames render the mark correctly.

Note: the fix reaches users with the next release build — existing installs keep the old icns until they update. If the corrupted icon lingers after updating, that's macOS's icon cache (re-registering the app / a reboot clears it).


eneskirca:
Reviewed this without taking the diagnosis on trust — and the mechanical half holds up completely. But there's a reason I'd hold the `Fixes #369` before merging.

**The ic13/ic14 claim is confirmed, and root-caused upstream.** I pulled electron-builder 26.15.3's actual icns toolset (`icons@1.1.0`, checksum verified) and ran it on the same `build/icon.png`. `ic13` gets a 512px payload, `ic14` a 1024px one — exactly as you measured. The bug is verbatim in `toolset/icons-bundle/icon-tool.js:687-689`:

```js
{ osType: "ic13", size: 512 },  // 256@2x   ← ic13 is 128@2x = 256px
{ osType: "ic14", size: 1024 }  // 512@2x   ← ic14 is 256@2x = 512px
```

They shifted the retina mapping by one slot.

**Your table is right on every entry**, cross-checked three ways: Apple's OSType table, electron-builder's own (correctly-labelled) 1x rows, and a real Apple-toolchain `.icns` sitting in this repo's `node_modules` (`app-builder-lib/templates/icons/proton-native/`). I also wrote an independent parser — not yours — and the packed output passes on all ten slots: magic, container length, per-chunk framing, PNG signature, IHDR CRC, IEND, and payload w×h against my own table. Refusals fire before any write (I mutated `ic13` to 512 and got `packIcns: ic13 must carry 256x256, got 512x512`, EXIT=1, no file). Six mutations of the implementation are all caught by your tests. And `release` **does** run `make-icon` in every mac path including CI — the highest-consequence failure mode isn't there.

**What I'd change before merge:**

**1. The causal claim isn't established, and there's a better candidate the PR preserves.** `icp4`/`icp5` carrying PNG is separately documented as a cause of *"completely scrambled"* icons **specifically when used inside a .app** — and the note adds that it bites on non-Retina displays, where the 16×16 slot is what gets used. That matters because the reporter's screenshot is a launcher **list row** at roughly 22–26px, and in the shipped file **every small slot was correctly sized** (`icp4`=16 ✓, `icp5`=32 ✓, `ic11`=32 ✓, `ic12`=64 ✓, `ic07`=128 ✓) — only 256pt and 512pt were skewed. For ic13/ic14 to explain a ~24px render, macOS would have to pick a 256/512pt representation for a list row *and* mis-decode a well-formed PNG that states its own size in its IHDR.

Corroborating: the Apple-toolchain reference icns in our own `node_modules` uses **`ic04`/`ic05` (ARGB)** for 16/32 @1x and emits no `icp4`/`icp5` at all.

I can't settle it without a Mac, and I'll be honest that the reporter's tile shows more horizontal row coherence than my raw-bytes simulation of the scramble did — so I'm not claiming certainty. But the assertion currently appears in **four places** (PR body, `icon-pack.mjs:4-10`, `make-icon.mjs:108-113`, `icon-pack.test.ts:5-10`) plus CLAUDE.md, and `Fixes #369` will close the issue on merge. That's the house rule about not naming a cause without measuring it, pointed at us.

Ask: drop `icp4`/`icp5` from `ICNS_FRAMES` (macOS then downsamples 16/32pt@1x from `ic11`/`ic07` — worst case slightly soft, strictly safer than a documented scramble) or emit `ic04`/`ic05`; soften the four claims to "a" cause; and keep #369 open until the reporter confirms. Worth asking them whether they were on a non-Retina display.

**2. The refusal can leave a tree that silently falls back to the buggy converter.** `mac.icon: "build/icon.icns"` is not a hard requirement — `iconConverter.js`'s `buildSourceCandidates` appends `icon.png`, and `roots` includes `build/`. So **png-present + icns-absent** silently resolves `build/icon.png` and runs it through the exact converter this PR exists to avoid: green build, broken artifact, no warning. And your own refusal creates that state — `make-icon` writes png and ico *first*, then throws in `packIcns`. Every scripted path is `&&`-protected so it only bites someone running `npx electron-builder --mac` directly, but that's the same trap `release.yml:46-49` already documents for tmux. Fix: write the icns first (or `rm -f build/icon.icns` at the top), and put the fallback in the CLAUDE.md packaging paragraph.

Smaller: `pngDims` reads offsets 16/20 with no length check, so a truncated PNG throws Node's `RangeError` instead of a `packIcns:` message; an extra frame in the map is silently dropped since `packIcns` iterates the table (a `frames.size !== ICNS_FRAMES.length` guard would make it symmetric with the missing-frame refusal); and `CLAUDE.md:2088-2092` now contradicts itself — it still says "**both** gitignored" with three outputs, re-mentions `icon.ico` in the appended sentence, and the new line overruns the file's wrap. Worth rewriting the paragraph rather than appending.

No regressions: clean trial-merge onto main, typecheck clean apart from this box's known dep drift, and the 11 failing files are identical on the unmerged baseline.

Not verifiable from here: whether the packed icns renders correctly in Dock/Finder/Cmd-Tab, whether the `icp4`/`icp5` scramble reproduces on the reporter's macOS + display, and whether dropping those two slots leaves 16pt surfaces acceptably sharp.

Issue #412, opened 2026-08-24
Pin the Node version: Node 25 breaks 12 test files with a misleading localStorage error


ezwep:
## What happened

Fresh clone, `npm install`, `npm test` — 12 test files failed to load and 10 tests failed. Nothing was wrong with the tree: my system Node was **v25.6.1**, and CI pins **22**.

Switching to Node 22.23.2 and re-running the same checkout: **3 files failed instead of 12**, 8980 tests passing, and every renderer suite green.

There is no `.nvmrc`, no `.node-version`, and no `engines` field in `package.json`, so nothing said which Node to use before the failures did.

## Why the failure is misleading

The renderer suites die on:

```
TypeError: localStorage.getItem is not a function
 ❯ src/renderer/state/viewMode.ts:63
```

at this guard:

```ts
// localStorage guard: this module is also imported by node-environment vitest suites.
viewByProject: parseViewMap(
  typeof localStorage === 'undefined' ? null : localStorage.getItem(PROJECT_VIEW_KEY)
),
```

The comment shows the case was anticipated. What defeats it is that recent Node ships a *partial* `localStorage` global: `typeof localStorage` is no longer `'undefined'`, so the guard falls through to `.getItem`, which is not there. Because `viewMode` is imported widely, everything downstream collapses with it — including `src/renderer/components/TabBar.test.tsx`, `zoomShortcut.test.ts`, `SharedGlyphLayer.test.ts` and `browser-partition-parity.test.tsx`. None of those names hint at a Node version, so the natural conclusion is that your checkout is broken.

`npm install` does emit a clue, but it scrolls past among the deprecation warnings:

```
npm warn EBADENGINE Unsupported engine {
npm warn EBADENGINE   package: 'jsdom@30.0.1',
npm warn EBADENGINE   required: { node: '^22.22.2 || ^24.15.0 || >=26.0.0' },
npm warn EBADENGINE   current: { node: 'v25.6.1' }
```

Note the range excludes 25 specifically — an odd-numbered release nobody targets, and exactly what `brew install node` gives you right now.

## Suggested fix

- Add `.nvmrc` with `22`, matching `ci.yml`. One line, and version managers pick it up without anyone reading a doc.
- Add `engines: { "node": ">=22 <23 || >=24" }` (or whatever range you actually support) so `npm install` refuses or warns loudly, up front, instead of failing later inside a test that names something else.
- Optionally harden the guard so a partial global cannot slip through:
  ```ts
[...]

PR #413, opened 2026-08-24
chore(repo): pin the Node version for contributors with .nvmrc

ezwep:
Adds `.nvmrc` with `22`, matching the Node version `ci.yml` pins for the `quality` job.

Refs #412. Deliberately **not** "Closes": this is a signpost, not a guard. It only helps people whose version manager reads `.nvmrc`, and it does nothing for anyone who installs Node globally — which is how I hit this. The actual fix for #412 is hardening the check in `src/renderer/state/viewMode.ts`, which I have left out of here to keep this reviewable in one glance. Happy to follow up with that plus a regression test if you want it.

## Why

Nothing in the repo states which Node to use, so the first signal you get is the test suite failing. On Node 25 the partial `localStorage` global slips past

```ts
typeof localStorage === 'undefined' ? null : localStorage.getItem(PROJECT_VIEW_KEY)
```

and takes 12 renderer test files down with a `TypeError` that names none of them — `TabBar.test.tsx`, `zoomShortcut.test.ts`, `SharedGlyphLayer.test.ts` and friends. Same checkout on Node 22: 3 failures instead of 12, every renderer suite green. Details and the full comparison are in #412.

## One thing worth deciding, because I had to guess

The repo is not consistent about Node, so `22` is my read of your intent rather than something the tree told me:

- `ci.yml` and `win-package-smoke.yml` → `22`
- `release.yml` → `20` for the macOS job, `22` further down
- `README.md:217` → "Requires Node.js 20+", while `README.md:231` says the Server Edition "needs Node 22 + tmux"

I went with `22` because that is what gates pull requests. If `20` is genuinely still the floor, that is worth knowing for a different reason: `jsdom@30`'s declared range is `^22.22.2 || ^24.15.0 || >=26.0.0`, which excludes 20 as well — so the README's "20+" may be stale for development even if it holds for running the app. Say the word and I will change the pin, or widen this into a PR that makes the four places agree.

## Not verified

I did not test Node 20, 24 or 26 — only that 25 breaks and 22 is good on macOS (Darwin 25.5.0, Apple Silicon). `npm run typecheck` is green; `npm test` leaves 3 files failing for me (`remote-atomic-write`, `tmux-paste.realtmux`, `session-host-client`), all of which spawn a real shell or real tmux and which I believe are my sandboxed environment rather than the change — this PR touches no code, so it cannot plausibly affect them.


eneskirca (approved):
Verified independently before approving — your macOS finding is cross-platform:

- Linux, node **25.6.1** (official tarball): `typeof localStorage` → `'object'`, `localStorage.getItem` → `undefined` (the global exists but is method-less without `--localstorage-file`). Running `src/renderer/lib/zoomShortcut.test.ts` under it reproduces your exact failure: `TypeError: localStorage.getItem is not a function` at `viewMode.ts:63`.
- Same checkout under node 22.22.2: green.

So the pin is right, and your read of the intent is right too: **22 is the version that gates PRs**, and `jsdom@30`'s floor (`^22.22.2 || ^24.15.0 || >=26.0.0`) excludes 20 for development regardless of what `README.md:217` says — that line and `release.yml`'s macOS job (`node-version: 20`) are stale, not counter-evidence. Making the four places agree is worth doing, but separately from this one-liner.

On the follow-up you offered (guard hardening + regression test): yes please, and one scoping note from reading the five `typeof localStorage` sites — they are not all equally broken:

- **Genuinely broken under the partial global** (typeof guard with no try/catch, module-level): `state/viewMode.ts`, `state/cardPanel.ts`, `state/cardModalSize.ts`.
- **Already safe**: `terminal/useCopyFeedback.ts` and `canvas/SharedGlyphLayer.tsx` wrap the same pattern in try/catch, so the method-missing TypeError lands in the catch — as does the try/catch-only style in `state/explorer.ts` and friends.

The shape we'd want is one shared helper (e.g. `readLocal`/`writeLocal` with `typeof localStorage?.getItem === 'function'` inside a try/catch) used by all five, plus a source-scan guard test banning the bare `typeof localStorage` pattern so it cannot come back — same discipline as `src/core/fs-atomic.guard.test.ts`. An `engines` field (mirroring the jsdom range is the honest choice, since it is the strictest dep and specifically excludes 25) would fit the same PR.

Your three remaining failures (`remote-atomic-write`, `tmux-paste.realtmux`, `session-host-client`) are indeed environment: all three spawn a real shell or real tmux. For what it's worth, `remote-atomic-write`'s real-shell test also just flaked with a timeout on our own windows-latest CI runner today — treating those as environment was the right call.

Issue #160, opened 2026-08-13
npm test is flaky under parallel execution (pty-single-user.test.ts and others); green single-threaded


dcrjodle:
## Summary

`npm test` intermittently reports test failures (in one run, 44 of them, all in `src/core/pty-single-user.test.ts`) that **do not reproduce in isolation and disappear entirely when file parallelism is disabled**. The failures are timing/resource-contention flakiness under full-suite parallel execution, not real regressions — the suite is green single-threaded.

I'm filing this so the flakiness is tracked; it can make `npm test` look broken to someone setting the project up for the first time.

## Environment

- HEAD: `a72732a` (Merge PR #125)
- macOS 15.7.4 (24G517), Apple Silicon, 10 cores
- Node v24.19.0, npm 11.17.0
- tmux installed at `/opt/homebrew/bin/tmux`
- Fresh `npm install` (native rebuild of node-pty/smart-whisper succeeded)

## Reproduction & evidence

| Configuration | Result |
| --- | --- |
| `npm test` (full, parallel) — run A | **44 failed**, all in `pty-single-user.test.ts` |
| `npm test` (full, parallel) — run B | **0 failed** (5119 passed) |
| `npm test` (full, parallel) — 3 further runs | 2× green; 1× a single file-level failure in an **unrelated** file (`test/server/presence-e2e.test.ts`) |
| `vitest run src/core/pty-single-user.test.ts` alone (×3) | **always green** (20/20) |
| `vitest run src/core/pty` (all 18 pty files, ×5) | **always green** (224/224) |
| `vitest run --no-file-parallelism` (full) | **green — 396 passed, 1 skipped, 0 failed** |

The tell-tale signature of flakiness rather than a real bug:
1. Never reproduces in isolation.
2. The failing file **moves between runs** (pty tests one run, a WebSocket presence e2e the next).
3. **Vanishes completely with `--no-file-parallelism`.**

## Why `pty-single-user.test.ts` is the usual victim

Every failing assertion in the 44-failure run was of the form "expected 1 `kill-session` / `capture-pane` call, got 0" — i.e. `PtyManager` took the **plain-shell fallback** (no tmux) for that run, so no tmux commands were issued.

`tmuxManager()` → `init()` → `ensureTmux()` → `findTmux()` depends on a chain that must finish before the assertions run, and none of it is mocked away:

[...]

PR #418, opened 2026-08-24
OSC 8 hyperlinks: Cmd+click opens links whose URL hides behind a label

EliJaghab:
## Summary

Cmd/Ctrl+click now opens OSC 8 hyperlinks, the kind where only a label is printed and the URL rides in an escape sequence. This is what Claude Code emits for every markdown link, and what gh, systemd and delta emit too. Parity with Ghostty and iTerm2, which open these natively.

## Problem

An agent CLI printing `[#376]` as an OSC 8 link was a dead end in nodeterm twice over. Inside tmux (every nodeterm session), tmux strips the escape because our generated conf never declares hyperlink support for the outer terminal, so only the label text survives. And even when the escape reaches xterm.js, the renderer sets no `linkHandler`, and the existing URL/file link providers only regex the visible text, which for a labeled link does not contain the URL.

## Change

Three pieces, one per layer the link travels through:

```
tmux conf (pty-manager.ts, ssh.ts)     terminal-features ",*:hyperlinks" -> escape passes through
createOsc8LinkHandler (file-links.ts)  options.linkHandler on both terminal surfaces
├─ modifier-gated like the providers, so a plain click stays a selection
└─ http(s) only: the URI is invisible text, a javascript:/file: link never reaches openExternal
osc8UrlAt (file-links.ts)              hit-test for installLinkClickFallback
   under tmux mouse-reporting xterm's own activate never fires, so the existing
   capture-phase mouseup fallback now checks the clicked cell for an OSC 8 link first
```

`osc8UrlAt` reads `CellData.extended.urlId` plus `_core._oscLinkService`, private API, because the public buffer API exposes no hyperlink data. Same pattern as the existing `_core._renderService` uses; degrades to "no link" if internals move.

Set on both surfaces of a session: the canvas node and the kanban card modal.

## Testing

`file-links.test.ts` covers the modifier gate, the scheme allowlist and the private-API shape degrading to null; `tmux-conf.test.ts` and `ssh.test.ts` pin the new conf line next to the RGB precedent. Typecheck green, touched suites green, full suite green single-threaded (the parallel flakes are #160). Verified by hand in a packaged build on macOS: Claude Code markdown links inside tmux-backed sessions open in the default browser on Cmd+click, plain click still selects, and the live tmux server picked up the feature via `set -as` without a server restart.

Mobile note for @eneskirca: the remote tmux conf change applies to SSH sessions the iOS app attaches too, so OSC 8 escapes start reaching SwiftTerm; worth checking whether it opens them or needs its own handler.

### DIFF
--- a/src/core/pty-manager.ts
+++ b/src/core/pty-manager.ts
@@ -275,6 +275,8 @@ set -as terminal-features ",*:clipboard"
 # known to speak RGB. xterm.js does, so declare it — via terminal-features like the clipboard
 # entry above, never terminal-overrides (see the MIGRATION note). Issue #78.
 set -as terminal-features ",*:RGB"
+# OSC 8 hyperlink passthrough — tmux strips the escape unless the outer terminal declares it.
+set -as terminal-features ",*:hyperlinks"
 # Mouse copy: on release tmux copies to its buffer AND (thanks to the two lines above) emits OSC 52,
 # which the client writes to the system clipboard. No pipe-to-a-local-command here, deliberately.
 bind -T copy-mode    MouseDragEnd1Pane send-keys -X copy-pipe-and-cancel
--- a/src/core/tmux-conf.test.ts
+++ b/src/core/tmux-conf.test.ts
@@ -35,6 +35,12 @@ describe('tmuxConf', () => {
     expect(c).not.toMatch(/set -a[gs]? terminal-overrides/)
   })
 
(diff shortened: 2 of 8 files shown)


EliJaghab:
marked this ready. been testing it in a packaged build today, claude code links inside tmux sessions open with cmd+click and plain clicks still select. pretty clean so far. would love eyes when you get a chance

PR #421, opened 2026-08-25
feat(status): restore state across restarts + sidebar workflow-state grouping

dkattan:
## Summary

Two related changes to agent-status / sessions-sidebar behavior:

### fix(sidebar): group by workflow state
Keep unread as a row-level notification affordance instead of a workflow bucket. Finished unread sessions remain under **Done**, and unknown unread sessions remain under **Unknown** — rather thanunread pulling rows into a separate bucket.

### feat(status): restore state across restarts
Keep a lifecycle-bound display snapshot while operational evidence still expires. Reconcile Claude/Gemini transcripts and Codex app-server status without letting restored state drive notifications or process control.

---

Draft — opening for early visibility. Builds on the current `main` (`6476c894`).


eneskirca:
Checking in on this draft — a quick status of where main has moved since your base, so effort doesn't go where it's already spent:

- **The grouping half is now main's model.** The sidebar settled on three workflow-state sections (Waiting for your response / Running / Unknown), with unread as a row-level affordance that NEVER changes the bucket — CLAUDE.md pins it in exactly your framing ("a click may clear the glow but cannot move the row"). So that half of the draft describes behavior main already ships.
- **The restore-state-across-restarts half is still open and still interesting** — nothing on main persists a display snapshot across app restarts (live `state` is deliberately transient), and the constraint you set (restored state must never drive notifications or process control) is the right one. It also touches a documented gap: Eco hibernation is structurally inert for sessions with no turn in the current app run, and a reconciled snapshot is one of the plausible fixes.

Are you planning to carry this forward? If so, trimming it to just the restore/reconcile half and rebasing onto current main would make it reviewable — happy to give it a full pass then. If it's parked, no problem, just say so and we'll treat it as a design note.

eneskirca:
## Verdict: recommend CLOSE — main solved the durable half, and kept the other half deliberately

Thanks for this, and sorry it sat so long. Reviewing it now against today's `main` (422767af), I think the honest answer is that it has been overtaken rather than that it was wrong.

**The restore-across-restarts half is in `main`.** `src/core/agent-status-mirror.ts:111` carries `restored?: true` — persisted state with a 6h expiry, never pushed to a listener, refused by the messaging gate (`src/core/agents/agent-message-decide.ts:229,297`), cleared by the first live event (`:464`). That design is deliberately more conservative than this PR's, and the conservatism is the point: `src/renderer/state/agentStatus.ts:52,59,70` documents `state`/`lastEventAt`/`stateVerified` as never-persisted, with the reasoning attached.

**The sidebar half is a design disagreement, not a gap.** This PR deletes `sessionStatusGroup`, `STATUS_GROUP_ORDER` and the `'unread'` bucket. `main` still has all three (`src/renderer/lib/sessionList.ts:54-78`) with a doc comment defending the Unread section and its membership priority. That is a call for @eneskirca, but it is not something a merge should decide silently.

**And the branch no longer builds.** `npx tsc -p tsconfig.web.json` on the PR head gives ~19 errors: `Canvas.tsx` imports `contextLink`, `dependencyLink`, `lineageLink` and `nodeEndpoints` from `../lib/noteLink`, plus `Link` from `@shared/types` — **none of which exist on that branch** (`git grep -l nodeEndpoints` finds only `Canvas.tsx`). It also calls `GitApi.syncBranch` and passes props that do not exist there. It looks like a mis-rebase that picked up half of #422's links substrate. Merging it would mean 9 conflicted files (`Canvas.tsx` alone has 10 hunks) *and* repairing a branch that does not compile.

### The part that is genuinely still unmet

Worth keeping, but as a fresh issue on top of main's mirror rather than as this branch:

- Main's `restored` flag is consumed only by the messaging gate and the phone mirror — **never by the sidebar** — so a desktop restart still shows Unknown rows. Your original complaint is real.
- `agent-status-recovery.ts` here (transcript-tail inference for claude/gemini, plus codex app-server thread status at boot) has no equivalent in `main`, which uses `decideFromTranscriptTail` only for SSH reconnect resync (`src/main/remote-ssh/agent-resync.ts`).

If you are up for it, a small PR that teaches the sidebar to render main's existing `restored` state would land the user-visible win without re-opening the persistence design. I would review that quickly.

eneskirca:
## Review — take with named fixes

Reviewed against `origin/main` (`e43d3a54`); this branch is freshly rebased and 0 commits behind. Not part of the #722→#756 stack — it stands alone.

### What I ran

Isolated worktree at `52812a2e`:

- `npm run typecheck` — clean (both tsconfigs).
- `vitest run` on `agent-status-recovery`, `agent-status-mirror`, `agent-status-mirror.restart`, `agent-status-handlers`, `sessionList`, `agentStatus`, `codex-session-name`, `SessionRow`, `hibernationCandidates`, `hibernation-policy` — **10 files, 325 tests, all passing.**

### The good part: this fixes a real CLAUDE.md divergence

CLAUDE.md is explicit that read state and workflow state are independent:

> Selecting, focusing, dwelling into, or opening a session card clears `unread` … **This NEVER changes the workflow bucket: read state is independent from agent state.** … A click may clear the glow but cannot move the row.

`origin/main` doesn't honour that in status-grouping mode: `sessionStatusGroup(kind, unread)` moves an unread row into a dedicated `unread` bucket, so reading a session *does* move it between sections. Removing that and making `unread` pure row metadata (`statusRestored` / `unread` on the VM, sections keyed on `StatusKind` alone) brings the code back in line with the documented contract. That's the right direction and I'd like it to land.

The `restored` flag threaded through core → mirror → VM is also well done — distinguishing "this is the last value we knew" from "this is a live observation" is exactly the `ok:false` ≠ `ok:true with no rows` discipline this repo applies elsewhere.

### Fix 1 — the new section order puts Running last, with no rationale

```ts
const STATUS_ORDER: StatusKind[] = ['attention', 'done', 'unknown', 'working']
```

Sections render in that order, so **Running sits at the bottom of the sidebar, below Unknown.** Your live agents — the ones actually doing work right now — are the hardest to find.

The old constant carried its reasoning:

> Section display order in status-grouping mode: what needs you first, then what is new for you, then what is live, then the settled ones.

The replacement deleted that comment and didn't write a new one. If `working` last is deliberate, the reasoning needs to be in the code; if it's incidental, `['attention', 'working', 'done', 'unknown']` matches both the old intent and CLAUDE.md's ordering ("Waiting for your response" → "Running" → "Unknown").

### Fix 2 — the new visible "Done" section contradicts CLAUDE.md

CLAUDE.md §Status-grouped sessions:

> three always-visible sections: **Waiting for your response** maps internal `done`, `waiting`, and `blocked` together (a completed turn, question, or approval all need the user); **Running** maps `working`; **Unknown** means no live hook state is available. **There is no Done bucket**: a normal `done` hook means the turn ended and the agent is waiting for another user prompt.

This PR surfaces `done` as its own section labelled "Done" (via `STATE_LABEL`), which is the bucket that paragraph says should not exist, and for the stated reason — a `done` agent *is* waiting for you, so splitting it from "Waiting for your response" hides work that needs attention.

`main` was already drifted here (it had an `idle` bucket), so this isn't a regression you introduced — but the PR is the right place to settle it. Either fold `done` into `attention` as CLAUDE.md describes, or keep the split and **update CLAUDE.md in the same change**, per the repo's convention that a change other developers must know about belongs in both docs.

### Smaller notes

- `sessionList.ts` nets −64/+? with `StatusGroup`, `STATUS_GROUP_ORDER`, `STATUS_GROUP_LABEL` and `sessionStatusGroup` all removed. Please confirm nothing outside this file imported `sessionStatusGroup` — typecheck passing says yes, just flagging it as a public-ish helper being deleted.
- `agent-status-recovery.ts` (+194) is new core logic behind `CorePlatform` and wired in **both** `src/main/index.ts` and `src/server/index.ts`, with a real ws-bridge member rather than a stub. That's the shell-parity rule correctly applied — noting it because this is the failure mode CLAUDE.md warns about most often.

### Recommendation

Take once fix 1 and fix 2 are settled. Neither is deep work, and the underlying change is a correctness improvement over `main`. This is the second-strongest PR of the nine I reviewed.

PR #444, opened 2026-08-26
docs: fix dictation shortcut in alt text and remove duplicated feature bullet

chiang21fcb:
﻿## Summary

Two small documentation fixes in README.md around the voice dictation feature.

## Changes

- The image alt text on the Talk to your terminal feature row said hold cmd-shift-D. The actual shortcut is hold `⌘⌥` which matches the body text and the keyboard shortcuts table.
- The More section listed Talk to your terminal as a bullet even though the feature already has its own row in the feature table above. Removed the duplicate.

## Verification

Documentation only. No code or behavior changed.


eneskirca (commented):
Verified against the keybindings registry: `speech.dictation`'s default binding is `Cmd+Alt` as a hold chord on both platforms (`shared/keybindings.ts` — `defaultBindings: both('Cmd+Alt')`, `allowHoldChord: true`), so `⌘⌥` is correct. The alt text was the last place still saying cmd-shift-D — the body text (line 113), the shortcuts table (line 246) and the feature row all already say hold `⌘⌥`. The removed "Talk to your terminal" bullet is a genuine duplicate of the feature-table row above it. Docs-only, consistent with the surrounding text.

PR #451, opened 2026-08-27
feat(canvas): per-burst wheel-zoom cap + zoom speed setting

Rick-VIPER:
## Why

With **Scroll wheel zooms** enabled, one click of a high-resolution ratchet wheel (Logitech MX Master 3S) jumps the canvas from 200% to 84%.

The handler in `Canvas.tsx` clamped each wheel **event** to ±50 deltaY — correct for a classic notched wheel, where one click is one event. High-res wheels deliver one physical detent as **several pixel-mode packets a few milliseconds apart**, and each packet passed the clamp independently. Two maxed packets compound to exp(1) ≈ 2.72×, which is exactly the observed jump (2.0 × 0.42 ≈ 0.84).

## What

**1. `WheelZoomBurstLimiter` (`src/renderer/canvas/wheel-zoom.ts`)** — the cap becomes a shared ±50 budget per 40 ms burst window instead of per packet, so a detent zooms the same amount whether the OS delivers it as one packet or four.

Trade-offs considered:
- **Window length**: 40 ms covers a detent's packet spread but stays well under a fast scroller's ~125 ms click cadence, so deliberate successive clicks each get a fresh budget. I rejected a longer window (risks merging real clicks) and a per-animation-frame cap (rAF timing varies with load).
- **Pinch**: the limiter applies to all zoom packets for simplicity, but a trackpad pinch never feels it — its packets are small, and the ceiling is exp(0.5) per 40 ms ≈ 25×/second, far beyond any physical pinch.
- **Reversal inside an exhausted burst** returns 0 rather than a free counter-step — that's device jitter, not intent.

**2. `settings.wheelZoomSpeed`** (0.2–2, default 1 = bit-for-bit historical feel) — a slider under the *Scroll wheel zooms* toggle multiplying the zoom exponent. Deliberately applied to the **plain-wheel path only**: Cmd/Ctrl+wheel and pinch keep the fixed step, so tuning a chunky mouse down never slows the trackpad. Validated at point of use (`clampWheelZoomSpeed`), per the hand-editable-settings rule.

## Tests

- 12 unit tests in `wheel-zoom.test.ts`, including the MX-detent burst case, budget refill, pinch-stream passthrough, and hand-edited settings values.
- Mutation-tested per CONTRIBUTING: reverting the limiter to per-packet clamping fails 3 tests; `npm run typecheck` clean; the full suite shows the same 4 pre-existing environment-dependent failures on a clean checkout of `main` (session-host / remote-atomic-write / cardModalSize), none touched here.

## Three surfaces

- **Desktop**: the change.
- **Server Edition**: shares the renderer wheel handler and Behavior settings UI — behaves identically, nothing extra needed.
- **Mobile companion**: not applicable (no mouse wheel) — noting it per the contract, @eneskirca.

## Not verified

- Feel of the 40 ms window on real MX Master hardware — the value comes from packet-timing reasoning, not measurement on the device. Happy to adjust if a tester's detent still double-spends.
- Windows/Linux (no path-shaped code involved; the handler is shared).


eneskirca (commented):
Deep review from an isolated worktree (branch fetched via `refs/pull/451/head`). Summary: the diagnosis is correct, the fix is the right shape, defaults are genuinely bit-for-bit, and every test claim in the description reproduced on my machine. Two small nits, one CI note — merge-ready shape from my side.

**What I verified locally**

- `npm run typecheck` clean.
- `wheel-zoom.test.ts` 12/12 green; `src/renderer/canvas` + `src/shared` suites → 1213 green; the only `src/main` failures are the pre-existing `node-pty-patch` environment ones (identical on a clean `main` checkout — unpatched `node_modules`).
- The mutation claim reproduces exactly: reverting the limiter to per-packet clamping fails 3 tests (detent burst, split budget, reversal).
- The compounding math in the description checks out: two maxed packets → `exp(1)` ≈ 2.72×, and 2.0 × 0.42 ≈ 0.84 matches the reported 200% → 84% jump.
- Default-path equivalence: with `wheelZoomSpeed` absent/1 and a classic notched wheel (one packet per click, ≥40 ms apart), the emitted zoom steps are identical to the old inline clamp — the behavior change is confined to multi-packet bursts, which is the bug. `clampWheelZoomSpeed` at point of use follows the hand-editable-settings rule, and `nextWheelZoom`'s bounds mirror the ReactFlow `minZoom`/`maxZoom` props.
- Composition with the wheel handler is right: pan-routed packets return before the limiter, so a trackpad pan never spends the zoom budget; the speed knob applies to the plain-wheel path only, and the pinch ceiling argument (exp(0.5) per 40 ms) is sound — no physical pinch gets near it.

**Nit 1 — the disabled slider is still keyboard-operable**

The wrapper uses `pointer-events-none opacity-40` + `aria-disabled`, but the `<input type="range">` itself is not `disabled`: a keyboard user can Tab into it and change `wheelZoomSpeed` while it reads as disabled to assistive tech. Setting `disabled={!settings.wheelZoom}` on the input (keeping the wrapper styling) closes the gap.

**Nit 2 (non-blocking)** — `MIN_ZOOM`/`MAX_ZOOM` are duplicated from the `<ReactFlow>` props with a "keep in sync" comment. The old inline code had the same duplication, so this is not a regression, but since the module now exists it could export the pair and let Canvas.tsx consume it — one constant instead of a comment promising sync.

**CI note for the maintainer**: this PR's workflow runs are sitting in `action_required` (first-contribution approval), so no checks show on the PR. I ran typecheck + the suites locally in the meantime — green as above.

**Interaction with #452**: both PRs touch the same effect; I applied them in both orders mentally and the conflict really is trivial (import line + the two-line handler body). One thing to preserve on whichever merges second: the limiter must stay *after* the router's pan decision, exactly as composed here.

PR #452, opened 2026-08-27
feat(canvas): route wheel-vs-trackpad by main-process gesture facts, not delta guessing

Rick-VIPER:
## Why

With **Scroll wheel zooms** and **Trackpad scroll pans** both on, a precise-pixel mouse (MX Master 3S, Magic Mouse) never zooms: its wheel packets are indistinguishable from a trackpad's at the DOM level, so `MacWheelGestureRouter` classifies it as a trackpad and routes every scroll to pan. The settings hint currently tells those users to turn trackpad panning off — trading one input device's behavior for the other's.

## The measurement this is built on

The main process can tell the devices apart. Probing `webContents.on('input-event')` on Electron 42 / macOS with an MX Master 3S and a MacBook trackpad side by side:

- **Trackpad two-finger scroll** → `gestureScrollBegin` → interleaved `gestureScrollUpdate`+`mouseWheel` pairs → `gestureScrollEnd`, with the momentum glide as its **own** begin/end pair and a few bare `mouseWheel` packets in the touch→momentum gap. Pinch wraps itself in `gesturePinchBegin/End` the same way.
- **MX Master wheel** (detents and free-spin) → bare `mouseWheel` only, never a gesture event.

## What

- **`main/trackpad-gesture.ts`** — `TrackpadGestureLedger`, a pure reducer of the raw input stream to edge transitions ("a trackpad gesture is now open/closed"). Depth-counted so an inner pinch inside an open scroll phase can't report a false close; a stray end (begin predates the listener) is ignored rather than going negative. The 'input-event' hook fires for every ~120 Hz pointer packet, so `observe()` is one Set lookup on the hot path and IPC is a handful of messages per physical gesture, not per packet.
- **One new IPC channel** (`canvas:trackpad-gesture`) → preload `onCanvasTrackpadGesture` → the router.
- **`MacWheelGestureRouter`** gains an opt-in `gestureReporting` constructor flag (desktop only — decided by the shell, never inferred from traffic). With it on, routing is by fact: gesture open or closed <500 ms ago → pan; reported silence → wheel mouse → **zoom**, even for trackpad-smooth deltas. The 500 ms linger is the same window the heuristics already used, and it absorbs the measured stray `mouseWheel` packets in the touch→momentum gap.
- Settings hint + doc comments updated: the trackpadPan off-switch remains the recourse for the **Server Edition only**.

**Trade-off accepted:** a brand-new gesture's first `mouseWheel` precedes its `gestureScrollBegin` (measured), so that packet can theoretically beat the Begin IPC to the renderer and misroute one small leading delta to zoom. Wheel dispatch is rAF-aligned while IPC is a plain task, so the flag almost always wins; I chose this over holding first packets back (adds latency to every gesture) or heuristic fallback in the silent state (which is exactly the misclassification being removed).

## Three surfaces

- **Desktop**: the change.
- **Server Edition**: a browser tab has no raw input stream — the router is constructed without reporting and keeps today's heuristics and escape hatch. Documented degrade in `bridge/stubs.ts`.
- **Mobile companion**: not applicable (no mouse wheel / no canvas wheel routing) — noting per the contract, @eneskirca.

## Tests

- 6 ledger tests (including the measured touch→momentum sequence) and 6 new router tests (the MX fix, notch-packet mid-gesture, momentum-gap linger, guards, legacy default, no mode-inference from traffic).
- Mutation-tested per CONTRIBUTING: deleting the definitive-silence branch fails 1 router test; removing the ledger's depth dedupe fails 1 ledger test.
- `npm run typecheck` clean; full suite shows the same 4 pre-existing environment-dependent failing files as a clean `main` checkout (session-host / remote-atomic-write / cardModalSize), none touched here.

## Not verified

- Magic Mouse: its touch surface presumably emits gesture events like a trackpad (it would keep panning, matching current behavior) — I don't have one to measure.
- Windows/Linux: the router is macOS-gated as before; the main-side listener is attached unconditionally but gesture events simply don't fire there.
- Verified live on macOS hardware (MX Master 3S + MacBook trackpad) for the event-shape measurement; the end-to-end feel test on the same hardware is being done on this branch now.

Related: #451 touches the same wheel handler in `Canvas.tsx` (zoom step size); the two are independent and either merge order leaves at most a trivial conflict.


eneskirca (commented):
Deep review from an isolated worktree (branch fetched via `refs/pull/452/head`, base compared against `origin/main`). Summary up front: the mechanism is sound, the measurements in the description check out against the code, and the test claims are real — I could reproduce all of them. One robustness finding worth addressing before merge, two smaller notes.

**What I verified locally**

- `npm run typecheck` clean.
- Targeted suites: `trackpad-gesture.test.ts` + `wheel-gesture.test.ts` → 19/19 green. Full `src/renderer` suite → 3415 green. `src/main` suite shows only the pre-existing `node-pty-patch` environment failures (identical on a clean `main` checkout here — unpatched `node_modules`, per CLAUDE.md).
- Both mutation claims reproduce exactly: deleting the definitive-silence branch fails 1 router test (`the MX Master fix`); the ledger/depth mutation likewise trips its test.
- Electron 42's `webContents 'input-event'` and the gesture type names match `electron.d.ts` — the API contract is real.
- Invariant cross-checks against the canvas rules: the `'input-event'` listener is purely observational (no `preventDefault`), so it cannot interact with the `before-input-event` chord intercepts or the terminal-first stand-down; the router ↔ `panOnScroll` agreement holds in all four `wheelZoom`/`trackpadPan` combinations with reporting on (pan packets return before the zoom path, zoomed packets are consumed before React Flow sees them); legacy heuristics are byte-identical with the default constructor, and a test pins that `noteGesture` on a non-reporting router cannot flip it. Server Edition degrade is a documented kept-behavior stub, and the mobile note is present. Nice work on all of that.

**Finding 1 — a missed `gestureScrollEnd` desyncs the ledger permanently (suggest a failsafe)**

If macOS ever fails to deliver the End for an open gesture — candidates: ⌘Tab / window blur / minimize / display sleep mid-two-finger-scroll — the ledger's depth sticks at ≥1 for the life of the window. From then on every real gesture nests inside the phantom one (Begin → depth 2, End → depth 1), so **no transition is ever emitted again**: the renderer's router holds `gestureActive = true` forever, every plain wheel packet routes to pan, and wheel zoom is silently dead until app restart. That's the worst failure shape for this feature — silent, permanent, and it reproduces the exact symptom the PR fixes, with no later event that can clear it (same class as the "a badge stuck on a finished node has no later hook to clear it" rule in CLAUDE.md).

A cheap failsafe closes it: reset the ledger and send `false` on `win.on('blur')` (a trackpad gesture can't meaningfully span a focus loss), or add a staleness guard in the ledger itself (real open gestures interleave `gestureScrollUpdate` packets ~120 Hz; a depth > 0 with no update for a couple of seconds is a phantom). Either keeps the hot path at one Set lookup. If you measured that macOS reliably delivers End across blur/app-switch, documenting that measurement in the ledger comment would settle it instead.

**Finding 2 — "gesture events simply don't fire on non-mac" is not quite true (touchscreens)**

Chromium synthesizes `gestureScrollBegin/End` for touchscreen scrolling on Windows/Linux as well, so on a touch-enabled machine the unconditionally attached main-side listener will emit IPC per touch gesture with no subscriber on the other end (the renderer only subscribes when `isMac`). Harmless, but it makes the doc comment's claim false and pays per-gesture IPC for nothing. Gating the attach (or the send) on `process.platform === 'darwin'` makes the stated contract true by construction.

**Note 3 (non-blocking)** — the wheel effect re-runs on `canvasLocked`/`wheelZoom`/`trackpadPan` changes and constructs a fresh router whose state starts at "silence = wheel mouse"; a settings toggle landing mid-gesture can misroute the remaining momentum packets to zoom once. Rare and self-healing on the next Begin — a sentence in the effect comment would do, no machinery needed.

**Trade-off review**: the accepted first-packet race (documented in `shouldPan`) is the right call — both alternatives you rejected are worse, and the loser is one small delta. The clock-domain handling is also correct as written: `noteGesture` and `shouldPan` both stamp with renderer-side `performance.now()`, so IPC latency only ever *extends* the linger, never shortens it.

Verdict from my side: with Finding 1 addressed (failsafe or measurement), this looks merge-ready. Final call is the maintainer's.

PR #499, opened 2026-08-28
fix(kanban): filter disabled agents in New session menu

AlexanderWillner:
Settings > Agents > Disabled correctly filtered Dock, pane menu and palette, but Kanban "+ New session" still showed all agents (e.g. only opencode enabled -> all visible anyway).

Fix: KanbanView createOptions now filters via `disabledAgents` (builtin + custom), like Dock etc. Memo deps updated.

Fixes: Kanban New session ignored disabledAgents
Refs: src/renderer/components/kanban/KanbanView.tsx:210ff

### DIFF
--- a/src/renderer/components/kanban/KanbanView.tsx
+++ b/src/renderer/components/kanban/KanbanView.tsx
@@ -208,18 +208,20 @@ export const KanbanView = memo(function KanbanView({
     if (latest && latest.updatedAt !== modalIssue.updatedAt) setModalIssue(latest)
   }, [github, modalIssue])
   const customAgents = useSettings((s) => s.settings.customAgents)
+  const disabledAgents = useSettings((s) => s.settings.disabledAgents)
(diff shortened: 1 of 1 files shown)


eneskirca (commented):
Verified the parity claim against the code before judging:

- **Dock** filters via `disabledAgents` (Dock.tsx:161 builtins, :108 customs) ✓
- **Pane/group context menu** filters (`agentCreationItems`, Canvas.tsx:6281) ✓
- **Command palette** filters (`buildCommands`, Canvas.tsx:9940) ✓
- **Kanban "+ New session"** was indeed the one surface that did not — this PR closes it with the same `!disabledAgents.includes(id)` pattern the other three sites use, for builtins and customs both, and correctly adds `disabledAgents` to the `useMemo` deps (the stale-closure half of the bug).

The non-agent entries (terminal / browser / sticky) stay unconditional, matching the dock, so a user who disables every agent still gets a usable menu. CI is green (quality, quality-windows, CodeQL, dep review). Looks correct and complete to me.

PR #500, opened 2026-08-28
fix(hibernate): enable hibernation for opencode via /exit

AlexanderWillner:
Eco (Hibernate idle agents) appeared broken when only opencode was enabled.

Root cause: opencode is in RESUMABLE_AGENTS and its TUI documents `/exit` (aliases `/quit`, `/q`, keybind `ctrl+x q`), but `EXIT_SEQUENCES` had no entry. So `restartEligibility` returned `not-resumable` and `planHibernation` always refused opencode nodes. With all other agents disabled, no node ever qualified - the feature did nothing.

Fix: Add `opencode: '/exit'` to EXIT_SEQUENCES and update tests (restartEligibility, bulk restart, hibernation policy) to expect the now-eligible behaviour.

Verified: `npx vitest run src/renderer/terminal/agent-restart.test.ts src/renderer/terminal/hibernation-policy.test.ts` -> 93 passed

Refs: src/renderer/terminal/agent-restart.ts:37, docs: https://opencode.ai/docs/tui/ (/exit)


eneskirca (commented):
Measured this on a real install before judging (opencode 1.18.18 → self-updated to 1.18.25 mid-test, Linux, tmux, isolated socket), since the house rule for `EXIT_SEQUENCES` is "the CLI's documented primary, bare, with no destructive variant" — the gemini `/quit --delete` lesson.

**What checks out**

- `/exit` is opencode's documented primary: the TUI command list shows `/exit — Exit the app` (aliases `/quit`, `/q`). No destructive variant exists: TUI slash commands take no flags, there is no `/delete`-style command in the list at all, and after several `/exit` + resume round-trips `opencode session list` still showed every session intact.
- The resume grammar holds: `opencode --session <sid>` reopened the **same** session — title, prior conversation, and token context (30.9K / 15%) all restored. Matches `resumeCommandWith`.
- The test moves are house-style (pins moved to `custom:unknown` so the "no exit sequence" half of the gate keeps its own coverage). 93/93 pass on the PR head.

**What does not check out — the delivery, not the table**

`performExitPhase` writes the exit blind: `io.write(KILL_LINE)` then `io.write(exit + '\r')` (agent-restart.ts:208–209), so the text and the CR reach the pane in one input burst. Measured against a fully ready TUI ("Ready." on screen, 15 s after launch):

- One-burst `/exit\r` → opencode does **not** submit. The composer is left holding `/exit` with the command popup armed, and `pane_current_command` is still `opencode` at the full 6 s budget → the run reports `exit-timeout`, every time (reproduced 3×).
- Splitting the CR off by even **100 ms** → exits in ~500 ms.

So the failure is specifically "CR arriving in the same read as the text" (opencode's batched-input handling), not launch timing. Consequence: with this PR, Eco/restart for opencode goes from "refuses, does nothing" to "attempts, times out, and leaves `/exit` sitting in the composer of a live session" — where the user's next Enter executes it and quits their session unexpectedly. That is a worse failure mode than the current refusal.

**Suggested direction**

The resume half already delivers safely for exactly this shape: `deliverCommand` writes the text, verifies the pane echo, and only then writes `\r` as a separate write (command-delivery.ts). Routing the exit line through the same echo-verified delivery — or at minimum splitting the `\r` into its own delayed write — would make this entry actually work. The comment at agent-restart.ts:204–207 already notes that each agent added to the table owes exactly this pane test.

Happy to see this land once the delivery half is addressed — the table entry itself and the resume grammar are verified correct.

AlexanderWillner:
Thanks for the detailed measurement (1.18.18→1.18.25, batched-input). Fixed as suggested:

- `EXIT_SEQUENCES` already added `opencode: '/exit'` (documented primary, no destructive variant verified)
- **Delivery:** `performExitPhase` now splits CR for opencode (`/exit` + 150ms + `\r`) - measured fix was 100ms / 500ms exit. Keeps blind-write for other agents. Prevents leaving `/exit` in composer where next Enter would unexpectedly quit (worse than prior refusal).

Ready for re-review.

eneskirca:
Re-measured on opencode 1.18.25 (Linux, real TUI, isolated tmux socket): the defect and the fix both confirmed.

- Old one-burst `/exit\r`: `/exit` left unsubmitted in the composer with the command popup armed, still running past the 6s budget (reproduced again — and the armed popup entry is exactly the "next Enter kills the session" hazard).
- New split write as shipped in `00d92ea8` (including the `KILL_LINE` prefix the code actually sends): opencode exits in ~270–370 ms.

The `agentId === 'opencode'` gate keeps every other agent's exit write byte-identical to the historical one-burst path, and the mid-split `gone()` re-check matches the existing polling-loop semantics — choosing the minimal split over routing through `deliverCommand` to leave the blind-write contract untouched was a reasonable call. Touched test files pass (93/93), typecheck clean on the branch, trial-merge onto current main clean (main hasn't touched agent-restart/command-delivery since your base).

Addressed, measured working — ready. One optional follow-up, fine as a later PR: a unit test pinning the split delivery order (write `/exit`, delay, write `\r`) for opencode, so a refactor can't silently re-batch it — the current tests pin table membership and eligibility but not the mechanism.

AlexanderWillner:
Addressed follow-up from https://github.com/eneskirca/nodeterm/pull/500#issuecomment-5498524737:

- Added unit test pinning the opencode split exit delivery: `performExitPhase` now asserts `['\x15','/exit']` before the 150ms gap and `['\x15','/exit','\r']` after, plus a `gone()` mid-delay case where the CR is never sent.
- Also pins that every other agent stays on the historical one-burst path (`/exit\r` / `/quit\r`), so a refactor cannot silently re-batch either side.

Commit: `81da4c5e` on `AlexanderWillner:fix/hibernate-opencode`
Checked: `typecheck` clean, `agent-restart.test.ts` + `hibernation-policy.test.ts` 95 passed (was 93).

Issue #399, opened 2026-08-24
Canvas: maximize a node to fill the viewport (resize + reflow), not just zoom to it


ezwep:
## The ask

A one-click **maximize** for a node: resize the node itself so it fills the visible canvas, rather than zooming the viewport onto it. Click again to restore its previous size and position.

## Why the existing gesture is not this

Double-click focus (`doubleClickFocus` → `goToNode` → `viewportForRect` → `setViewport`) already frames a single node nicely, and for an editor node reading a Markdown file it is exactly right — bigger text is the whole point there.

For a terminal it is not, because the canvas zoom is a CSS transform. `patchTerminalScale` only rewrites mouse coordinates to compensate for the scale:

```js
if (prop === "clientX") return rect.left + (event.clientX - rect.left) / s;
```

Columns and rows come from the node's own pixel size via `applyFit()`, so zooming in makes the same 80×24 bigger without showing one extra line. When I want to actually read a long build log or watch an agent work, magnification is not what I need — I need more rows.

Dragging the resize handle out does give more rows (it reflows and pushes a `pty.resize`), but it is fiddly, it is destructive to my layout, and putting the node back where it was afterwards is manual.

## Suggested behaviour

- Maximize resizes the node to fill the current viewport bounds, keeping a small margin — same rect `viewportForRect` computes, applied to the node instead of the camera.
- Toggling again restores the previous size and position. Storing one previous rect on the node is enough; no history needed.
- It should reflow, i.e. go through the normal resize path so the pty gets its new cols/rows.
- A titlebar button next to the existing close control would be the discoverable spot, plus a shortcut. Editor and diff nodes would benefit too — a Markdown preview at full width beats one at 320px, even without the reflow argument.

Restoring matters as much as maximizing: what stops me doing this by hand today is not the enlarging, it is putting everything back.

## Relationship to the zones request

This is deliberately the narrow version of #394. That one is about arranging several nodes into regions; this is one node, one click, one toggle. If zones ship with a "fill the viewport" zone, this is subsumed and can be closed — but it is far smaller on its own, so I am filing it separately in case the bigger one does not make the cut.

## Environment

nodeterm v0.3.2, macOS. Desktop canvas.

Issue #394, opened 2026-08-24
Canvas: snap nodes into layout zones (MacsyZones/FancyZones-style) instead of hand-dragging


ezwep:
## The ask

A zones-style layout helper for the canvas — the thing MacsyZones, FancyZones or Rectangle give you at the OS window level, but for nodes: drop a node into a predefined region and it snaps to fill it exactly.

I run a lot of sessions side by side and spend a fair amount of time hand-dragging and hand-resizing nodes into "left half / right half" or a 2×2 so I can watch several agents at once. Every time the arrangement is thrown off — a new node, a resize, switching machines — I redo it by hand.

## What exists today

For completeness, since these are adjacent and I did not want to ask for something already there:

- `snapToGrid` + `gridSize` — snaps to a 24px grid while dragging, but the node keeps its own size, so it aligns without laying out.
- **Tidy canvas** (`arrangeNodes` with `{ layout: "grid" }`) — arranges everything at once, on its own terms. Great for cleaning up, not for "put this one here".
- `NodeResizer` — manual resize per node.
- Double-click focus — zooms the viewport to one node.

What is missing is the middle ground: deliberate placement of *one* node into a *chosen* region, at that region's size.

## Sketch

Roughly in order of how much I would use it:

- **Drag to a zone.** Hold a modifier while dragging (or drag toward an edge) to reveal a zone overlay; dropping snaps the node to that zone's position and size. Halves, thirds and a 2×2 would cover most of it.
- **Keyboard.** Send the focused node to a zone without the mouse — the Rectangle-style `⌃⌥←` / `⌃⌥→` idiom is what my hands already expect.
- **Saved layouts per project.** Name an arrangement and restore it. This is where it would beat OS-level window managers, because the canvas already persists per project.

## One design question worth deciding early

An infinite pan/zoom canvas has no fixed frame, so "left half" is ambiguous: half of the *visible viewport*, or half of some fixed region on the canvas?

Viewport-relative is what feels natural in the moment, but it means the same node lands somewhere different depending on zoom level, and the arrangement stops being meaningful once you pan away. Canvas-relative is stable and persists, but you have to define the frame — possibly reusing **groups**, which already draw a bounded region: zones *within* a group would give the frame for free and make saved layouts obvious.

I lean toward group-relative for that reason, but I am guessing at your intent for groups; you would know whether that fits.

## Environment

nodeterm v0.3.2, macOS. Desktop canvas.

PR #504, opened 2026-08-29
Keep maximize and zone snap clear of pinned side panels

ezwep:
## What I hit

I keep the sessions sidebar pinned — with 16 sessions across my projects it is how I navigate. Maximizing a terminal node then lays it out at full canvas width and the pinned panel simply covers its left edge.

For a terminal that is not a cosmetic clip: the first ~300px of *every line* is gone, which is where the indentation, the prompt and the start of each sentence live. The node title, the model badge in the header and the input field at the bottom-left disappear under it too. Nothing reflows — the node is laid out as if the panel were not there and then painted over.

Zone snap has the same problem, and worse: `left-half`, `left-third`, `top-left` and `bottom-left` all land under the panel, and `left-half` is precisely where it sits.

Unpinned this would not bother anyone — both panels are hover peeks that get out of the way. Pinning is the user saying "this one stays".

## What changed

`maximizeTargetRect` (#399) and `zoneTargetRect` (#394) both derived their rect from the full canvas wrapper inset by a uniform margin. They now also inset by what the pinned panels cover.

- **`lib/pinnedInsets.ts`** (new) — a pure `insetsForPanels(wrap, panels)` plus a thin `measurePinnedInsets` that reads the docked panels out of the document. Measured rather than derived from the CSS widths: the two panels differ in width, the drawer has a wider variant, and `uiScale` changes what any of them render as.
- **`SessionsSidebar`** gains a `sessions-sidebar--pinned` modifier, so the measurement can tell a docked panel from a hover peek. The explorer drawer already had `drawer--pinned`; this just makes the pair consistent. No styles attached to it.
- **`refitMaximizedNode`** (new, in `state/workspace.ts`) — maximize is a *mode*, not a one-shot placement, so a maximized node now follows the usable area when a panel is pinned, unpinned, or changes width. It keeps the restore rect and is idempotent, so a re-measure that lands where the node already is returns the same array and never marks the workspace dirty.
- A **ResizeObserver** drives the re-fit rather than the pin flags alone: the explorer drawer animates in over 160ms (`nt-drawer-in`), so measuring one frame after the state change catches it mid-slide.

Zone snap deliberately stays a one-shot placement — pan away and the node stays put, as documented in `nodeZones.ts`. Only maximize follows.

## Scope I did not take

Window resizes still do not re-fit a maximized node. That is existing behaviour and felt like a separate question rather than something to change under this heading.

## Testing

`npx vitest run src/renderer` — 264 files, 3461 tests green; `tsc --noEmit -p tsconfig.web.json` clean, on Node 22.

New coverage: `pinnedInsets.test.ts` (edge attribution, both panels, widest-per-edge, non-overlapping panels, an offset wrapper, clamping), zone subdivision under insets plus the existing "zones and maximize agree on the frame" invariant re-asserted with panels pinned, maximize with insets and its refusal floor, and `refitMaximizedNode` for the restore rect, idempotence, and leaving non-maximized nodes alone.

Reproduced on v0.3.3, macOS.


eneskirca:
Thanks — this is a well-built change. `insetsForPanels` as a pure, tested helper measuring real DOM rects (rather than trusting the CSS widths) is the right call, and I checked the invariant I care most about carefully: `refitMaximizedNode` keeps `premaxRect` untouched and does all its work in root space through `withNodeRect`/`rootPosition`, so grouped nodes restore correctly. Typecheck and the renderer suites are green on my end too. Four things before merge, the first one blocking:

1. **The header maximize button is not covered.** `MaximizeButton.tsx:25` still calls `maximizeTargetRect(getViewport(), width, height)` with no insets — and that button (terminal/editor/diff headers) is the primary maximize path, i.e. the exact scenario the PR description opens with still reproduces through it. Nothing corrects it afterward either: the re-fit effect and the ResizeObserver only fire on panel changes, not on maximize. The RF store already has `domNode`, so `store.getState().domNode?.getBoundingClientRect()` → `measurePinnedInsets` should slot in with a couple of lines.

2. **Gate the re-fit on the insets actually changing.** `fitMaximizedToUsableArea` recomputes the rect from the *current* viewport and runs on every change of its deps — and `explorerOpen` is also true for the unpinned overlay explorer. So: maximize a node, pan the camera elsewhere, press ⌘⇧E — the node teleports into the current view (and dirties the workspace), even though the insets were 0 before and after. That cuts against the PR's own principle that unpinned panels don't matter, and against the existing choice that pan/window-resize don't re-fit. Keeping the last-applied `ScreenInsets` in a ref and bailing when they're unchanged would confine the behavior to real panel geometry changes.

3. **The idempotence check needs an epsilon.** `refitMaximizedNode` compares with `===`: for a grouped node, `(rect.x - originX) + originX` isn't guaranteed to round-trip in floats, and `nodeW` prefers `measured?.width`, which React Flow repopulates from the DOM (device-pixel-snapped) and at fractional zoom won't equal `innerW / zoom` exactly. Either way each effect tick rewrites the node and marks dirty — on a shared SSH/git project that's a rev bump per panel toggle. The unit tests pass because they use unmeasured nodes and clean numbers; a `< 0.5px` tolerance closes both cases.

4. Small one: `nt-drawer-in` animates `opacity` + `transform`, and ResizeObserver doesn't fire on transforms — so the "observer fires again as it settles" comment describes a mechanism that isn't actually there. The residual error is only ~10px (the `translateX(10px)` start offset) when opening an already-pinned explorer, so functionally it's minor, but I'd either fix the comment or add a one-shot `animationend` re-measure so it does what the comment says. Also, the new `refitMaximizedNode` docblock landed between `restoreMaximizedNode`'s comment ("The toggle's second click…") and the new function, leaving restore's doc attached to the wrong one.

Everything else checked out: zone snap's entry points all funnel through `snapNodeToZone` so they're fully covered, the zones/maximize agreement test with insets is a nice pin, and the decision to keep zone snap one-shot while maximize follows is the right split.

ezwep:
Thanks — all four are fixed in 1489805.

**1. The header button.** You are right, and it was the path the report opens with. It now takes `store.getState().domNode?.getBoundingClientRect()` through the same `measurePinnedInsets`, so the button and the canvas commands compute one usable area.

**2. Gated on the insets.** The last applied `ScreenInsets` lives in a ref and the callback bails when they are unchanged, so the unpinned overlay explorer no longer teleports a maximized node into the current view. Your framing was the useful part: `explorerOpen` is about mounting, not about docking, and only the measured geometry should decide.

**3. Epsilon.** `samePx` with a 0.5px tolerance, applied to all four edges. One caveat I did not resolve: the comparison happens in flow space, so at low zoom the same one-screen-pixel snap is a larger flow-space delta and could still slip past a fixed tolerance. Making it zoom-aware means threading the viewport into `refitMaximizedNode`, which felt like more API than the problem deserves — say the word if you would rather have it.

**4. The animation.** Correct, and the comment was describing a mechanism that was not there. There is now an `animationend` listener on the observed panels, plus `animationcancel` for an animation cut short by a re-toggle. `restoreMaximizedNode`'s docblock is reattached to `restoreMaximizedNode`.

New coverage for the tolerance (sub-pixel drift is a no-op, a real move is not). `npx vitest run src/renderer` — 264 files, 3462 tests green; typecheck clean.

eneskirca:
Re-reviewed `1489805a` — all four are genuinely in, verified on the diff and by running the branch:

1. The header button now routes through the same `measurePinnedInsets(domNode rect)` as the canvas commands — the primary path computes one usable area.
2. The re-fit is gated on the measured insets in `lastInsetsRef`, so the unpinned overlay explorer no longer teleports a maximized node (and "the flags decide what to observe, the geometry decides whether to act" is the right split).
3. `samePx` 0.5 px on all four edges, with the fallback chain for unmeasured nodes. On your flow-space caveat: agreed it's fine as-is — the failure mode of a slip at low zoom is one redundant rewrite, not a teleport, and threading the viewport in would be more API than that's worth. If it ever shows up as rev churn in the wild we know where the knob is.
4. `animationend` + `animationcancel` on the observed panels (the cancel leg is a nice catch — a re-toggle mid-slide would have pinned the halfway measurement), and `restoreMaximizedNode`'s docblock is back on its function.

Measured on the tip: typecheck clean, the maximize/insets suites pass (37 tests), CI green, trial-merge onto current main clean. Addressed, ready.

PR #509, opened 2026-08-29
Add middle-click to end a sidebar session, rename Close to End session

gor3a:
## Summary
- Sidebar session row's × button and right-click menu both said "Close" but already went through `closeSession`'s confirm dialog ("End this session? This stops its tmux session and removes the node from its canvas.") — renamed both to "End session" so the label matches the actual (destructive) behavior.
- Added a middle-click handler on the session row that calls the same `closeSession(projectId, id)` path (× button, right-click menu, and middle-click are now three entry points into one confirmed action — no new confirmation logic needed).

## Test plan
- [x] `npm run typecheck`
- [x] `npx vitest run src/renderer/components/SessionRow.test.*`
- [x] `npx vitest run src/renderer/canvas/canvas-wiring.test.tsx src/renderer/nodes/status-lifecycle.test.ts src/renderer/lib/projectCloseSessions.test.ts`
- [x] Manual: middle-click a sidebar session row → confirm dialog appears → confirming ends the session

### DIFF
--- a/src/renderer/canvas/Canvas.tsx
+++ b/src/renderer/canvas/Canvas.tsx
@@ -9907,21 +9907,22 @@ export function Canvas() {
       // Transfer) place beside the source — the same as the row's existing Transfer behavior.
       //
       // The canvas menu ends in a destructive "Delete" (deleteNodes). The session row's analogue
-      // is non-destructive "Close" (closeSession — hides the tab, keeps the tmux session), so the
-      // trailing Delete is swapped for Close rather than offered beside it.
+      // is "End session" (closeSession — stops the tmux session and removes the node too, just
+      // confirmed via its own dialog rather than the canvas's shared confirm), so the trailing
+      // Delete is swapped for End session rather than offered beside it.
       const body: MenuItem[] =
         projectId === activeProjectId
           ? (() => {
               const full = selectionItems([id])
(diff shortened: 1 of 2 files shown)


eneskirca:
Thanks for this — and welcome! This is an unusually careful first PR: you traced `closeSession` to its actual implementation, noticed the old code comment claimed it was non-destructive when it isn't, and fixed the comment along with the labels. That kind of code-vs-docs cross-check is exactly what we hope for in reviews here.

I verified the safety property: middle-click, the × button, and the context-menu item all route through `closeSession`'s existing confirm dialog, so there's no un-confirmed destructive path — good call keeping all three entry points on one confirmed action rather than adding new logic.

Two small notes, neither blocking:

1. Consider `onAuxClick` instead of `onMouseDown` for the middle-click: it fires on button release (matching browser-tab middle-click semantics) and lets a user bail out by moving the pointer off the row before releasing.
2. If you feel like it, a small event test for the middle-click branch would pin the behavior (the existing `SessionRow.test.tsx` uses `renderToStaticMarkup`, so it'd need a jsdom render — fine to leave for a follow-up).

The "Close" → "End session" rename itself is a product-wording call, so flagging it for @eneskirca to sign off — but for what it's worth, it makes the label match what the action (and its own confirm dialog) already says, and it distinguishes it cleanly from the non-destructive "Close project".

Issue #442, opened 2026-08-26
Closing a project parks its sessions invisibly — offer to end them, and show that they exist


ezwep:
## What happens

Closing a project hides it and leaves its sessions running. `closeProject` sets `closed: true` and touches nothing else — no session is ended, no node is removed.

I am fairly sure that half is deliberate, and I would not change it: `SessionMemoryPanel` calls them "deliberately parked sessions", and surviving a closed tab is the point of tmux-backed sessions. The problem is what happens to my ability to see them.

## Why it costs me

I work across a lot of projects, and closing one is how I tidy up. But a closed project is filtered out of both places I would look:

- `TabBar.tsx:88` — `allProjects.filter((p) => !p.closed)`
- `SessionsSidebar.tsx:74` — the same filter

So the tab is gone, the sessions are gone from the sidebar, and they are still running. Nothing shows a count, and nothing hints they exist. The one place they resurface is the session-memory panel, which deliberately includes closed projects — but that panel is framed around reclaiming RAM, and `onKillSession` is per row, one click each.

The result is that "close" reads like cleanup while actually being "hide, and keep running". With a handful of projects that is invisible; with many it is the thing making the app feel less tidy rather than more.

## What I am asking for

Either of these would solve it for me; the first is the smaller one:

- **An option when closing a project to end its sessions too.** Not the default — parking is right — but offered at the moment I close, e.g. "Close" vs "Close and end N sessions", with the count so I know what I am leaving behind. The desktop already has the per-node confirm wording to borrow (*"This stops its tmux session and removes the node from its canvas."*).
- **A bulk end-all for one project**, reachable from the project menu or the session-memory panel — "End all N sessions in <project>", one confirmation instead of N.

And whichever way it goes, **something that tells me parked sessions exist**. A count on the closed-project entry, or a line in the memory panel grouping by project, would be enough. Right now the only way to find out is to open the panel and read the project column.

## Things you would have to decide

- **What counts toward N.** Working, idle, hibernated, orphaned, already-exited — the number in the button and the set the action ends have to be the same definition, or the count becomes a lie the first time someone checks it.
- **Does "end" mean the node × does?** That stops tmux *and* removes the node from its canvas — destructive in a way closing a project currently is not. That asymmetry is exactly why I think it belongs behind an explicit choice rather than a change to what Close does. Ending a session can also interrupt real work mid-flight; parked means preserved, deliberately, and the confirmation should say what is being given up.
[...]

PR #510, opened 2026-08-30
feat: closed sessions & windows history

gor3a:
## Summary

Adds a persisted, browsable "Recently closed" history to the sessions sidebar:

- Deleting a terminal/agent/sticky/etc. node now logs a git-shared entry (capped at 20 per project, newest-first) into `.nodeterm/project.json` instead of vanishing.
- A new "Recently closed" section in the sessions sidebar merges these closed-session entries with existing closed-project entries, sorted by recency.
- Reopening a session recreates a fresh node/tmux session via the existing `recreateNodeFromSnapshot` machinery (never the original killed session's content); reopening a project reuses the existing `reopenProject`.
- Shares its core snapshot/recreate/plan-execution machinery with the existing in-memory `Cmd+Shift+T` "reopen last closed" feature rather than duplicating it.
- Closed-session entries are sanitized through the same exec-stripping/portability/partition/trigger pipeline live nodes already go through, in both directions — a git-shared `project.json` can never smuggle an executable command via a deleted-session entry.

## Test plan

- [x] `npm run build` succeeds
- [x] `npm test` — same pre-existing, unrelated failure count as the pre-work baseline (generated-shell/real-tmux/real-pty/e2e suites); zero new failures; all new tests pass
- [ ] Manual smoke test in a running Electron app (not performed — no GUI available in this environment): delete a node, confirm it appears in "Recently closed" and reopens correctly with the same cwd; close a project, confirm it appears in the same section; verify both survive an app restart; verify the sidebar's flex sizing holds up with 25+ history rows across two projects

### DIFF
--- a/src/core/workspace-files.closed-history.test.ts
+++ b/src/core/workspace-files.closed-history.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect } from 'vitest'
 import { projectToFile, fileToProject, validClosedSessions } from './workspace-files'
+import { CLOSED_SESSIONS_CAP } from '../shared/types'
 import type { ClosedSessionEntry, Project } from '../shared/types'
 
 const closedEntry = (id: string, over: Partial<ClosedSessionEntry['node']> = {}): ClosedSessionEntry => ({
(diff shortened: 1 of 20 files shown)


eneskirca:
Thank you for this — it's an impressive first contribution. You clearly read the codebase deeply before writing: the `validKanban`-style tolerant reader at the file trust boundary, the cap enforced on read *and* write *and* mutate, running entries through the exec-strip/trigger/partition pipeline in both directions, `lib/uuid` instead of `crypto.randomUUID`, and reusing the ⇧⌘T `snapshotNode`/`recreateNodeFromSnapshot`/`planReopen` machinery instead of duplicating it are all exactly the house style. Typecheck is clean, your 88 new tests pass, and I verified the surrounding workspace/store/reopen suites (533 tests) still pass. The honesty in the test plan about the un-run Electron smoke test is appreciated too.

Two things before this can land, one of which is a maintainer call rather than a flaw in your code:

1. **Where the history lives.** You put `closedSessions` in the git-shared `.nodeterm/project.json` and argue that reopening a teammate's deleted session is intentional. That's defensible (kanban is shared), but it also means every node delete leaves up to 20 full node-state blobs — titles, cwds, positions — in a committed, teammate-visible file, which is the same diff-churn concern that moved `viewport` *out* of the shared file. We need to decide deliberately whether this history is shared content or machine-local (the index entry, like your `closedAt`, or localStorage). Hold off on reworking until @eneskirca makes that call — your `closedAt`-on-`IndexEntryV3` half is already the right shape either way.

2. **The two histories don't consume each other.** A delete now lands in both the in-memory ⇧⌘T stack and the persisted list, but restoring via ⇧⌘T leaves the sidebar row behind (still clickable → a duplicate fresh node), and vice versa. Since both mint fresh ids nothing breaks, but it's confusing — could ⇧⌘T's restore also consume the matching persisted entry (or the sidebar reopen drop the matching in-memory one)?

Smaller notes, none blocking: session rows in the merged list don't show which project they belong to (project rows get a glyph; a small project dot on session rows would help once two projects have history); and since the entry already carries `agentSessionId`, offering a resume on reopen of an agent node would be a lovely follow-up — fine as a separate PR.

Also relevant: #458 (in flight, for #442) covers the close-*project* confirm/end-sessions side. I test-merged the two branches and they don't conflict — they're complementary, and your `closedAt` stamp actually helps that work. So no rebase pain expected there.

Really solid work — looking forward to the next revision.

eneskirca:
Maintainer decision on the storage question: **move `closedSessions` to machine-local storage** (the per-machine index / localStorage tier), not the git-shared `.nodeterm/project.json`. Rationale: whose trash can holds what is a per-machine fact, like the viewport — and 20 node-state blobs per deletion would churn a committed, git-shared document on every delete. Please also make the two ledgers (the in-memory ⇧⌘T stack and the persisted list) consume each other so a deletion can't be reopened twice. Everything else in the review stands — the tolerant reader and destroy semantics looked good. Thanks!

eneskirca:
Re-reviewed `2fca981d` — the storage decision is implemented exactly as asked, and carefully:

- `closedSessions` now rides the machine-local `IndexEntryV3` tier (same as `viewport`/`breadcrumbs`), `projectToFile` never emits it, and the load path treats a file field of that name as a forgery and simply never reads it — the same wording and rule `breadcrumbs`/`capabilityAck` use, which is exactly the precedent to copy. Capped on the way out as well as in, and the tolerant reader covers ref'd projects, inline projects, and the v1/v2 migration.
- The two ledgers consume each other in both directions (`dropClosedSessionRef` ↔ `discardClosedSession`), so one delete can no longer mint two reopened nodes — with tests on both directions.

Measured on the tip: typecheck clean, the six touched/added test files pass (90 tests), CI green on the revision, trial-merge onto current main clean. Addressed, ready — and for a first contribution arc this was handled remarkably well.

PR #511, opened 2026-08-30
feat: open/continue a project by dragging its folder onto the canvas

gor3a:
## Summary

Dragging one or more folders from the OS file manager (Finder/Explorer) onto the nodeterm window now opens each as a project — no need to go through the "Open folder…" dialog. It works from the canvas background, the Welcome screen (including its "Open folder…" card and the other nav cards), and general app chrome.

The drop reuses the exact same rules the "Open folder…" dialog already uses:
- A folder that maps to an already-registered project is reopened (including one that was previously closed).
- A folder carrying its own `.nodeterm/project.json` (a git clone, a synced copy, another machine's project) is adopted.
- A virgin folder becomes a brand-new project.
- Dropping several folders at once opens each as its own project, processed one at a time so they don't race each other; the last one dropped ends up active.

Dropping a folder onto a terminal node still pastes its path as text (unchanged), and dropping image files onto the canvas still creates an image node (unchanged) — the new handling only activates when the drop actually contains a directory, and otherwise gets out of the way.

Desktop only (Electron): resolving a dropped item's real filesystem path uses `webUtils.getPathForFile`, which has no meaningful equivalent for a browser-based drop in the Server Edition, so that surface is an intentional no-op there.

## How it works

- `openOrAdoptFolder(folder)` — the dedupe/reopen/adopt/create decision, extracted out of the existing folder-picker path so both entry points share one implementation.
- `isFolderDropTarget(target)` — decides whether a drop landed somewhere that should open a project (canvas, Welcome screen, general chrome) versus somewhere with its own drop behavior (a terminal, an editor, a dialog, a form control).
- `droppedDirectories(dataTransfer)` — picks out the directories among a drop's items.
- A window-level `dragover`/`drop` listener composes the three: it only intercepts the event when at least one directory is present, resolving each to an absolute path and opening/continuing the corresponding project in order.

## Test plan

- [x] Unit tests for `isFolderDropTarget` and `droppedDirectories` (accepted/rejected drop targets, directory extraction, edge cases)
- [x] Full test suite: no new failures introduced
- [ ] Manual verification in a running build: drag a virgin folder onto the canvas; close it and re-drag the same folder to confirm it reopens instead of duplicating; drag onto the Welcome screen; drag two folders at once; drag a folder onto a terminal node (should still just paste the path); drag an image file onto the canvas (should still create an image node); reorder project tabs via drag (should be unaffected)

### DIFF
--- a/src/renderer/canvas/Canvas.tsx
+++ b/src/renderer/canvas/Canvas.tsx
@@ -10432,39 +10432,52 @@ export function Canvas() {
     void writeDisk()
   }, [commitActiveToStore, writeDisk])
 
+  /** The dedupe/reopen/adopt/create decision for a folder path, shared by the "Open folder…"
+   *  dialog and the drag-and-drop entry point: a folder maps to one project, and this is the
+   *  ONE place that decides whether to reuse/reopen an already-registered project, adopt an
+   *  existing `.nodeterm/project.json` (git clone, synced copy, another machine's project), or
+   *  create a brand-new one. */
+  const openOrAdoptFolder = useCallback(
+    async (folder: string): Promise<void> => {
+      commitActiveToStore()
+      // A folder maps to one project: reuse the already-registered one first…
+      const existing = useProjects.getState().projects.find((p) => p.cwd === folder)
+      if (existing) {
+        useProjects.getState().openFolderProject(folder)
+        // An `unavailable` placeholder never recovers on its own: a save emits a header-only ref
+        // for it (never a file), so a deleted project.json stays deleted and every later load
+        // re-mints the placeholder. Opening the folder is the deliberate act that breaks that
+        // loop — but only on evidence, since clearing the flag lets the next save write this
+        // empty canvas. See #385.
+        const recovery = unavailableRecovery(existing, await api.workspace.projectFileState(folder))
+        if (recovery === 'clear') {
+          useProjects.getState().setProjectUnavailable(existing.id, false)
+        } else if (recovery === 'rehydrate') {
+          // `present` is a stat, not a parse: a corrupt file stats fine, and probeFolder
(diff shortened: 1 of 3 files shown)


eneskirca:
Thank you — this is a really well-built feature PR, especially for a first contribution. The thing we care most about here is that a folder maps to exactly one project (dedupe → reopen → adopt `.nodeterm/project.json` → only then create), and you got the architecture exactly right: extracting the dialog's existing body into `openOrAdoptFolder` verbatim and routing the drop through it means the two entry points can never disagree. The synchronous `webkitGetAsEntry()` call with the comment explaining why it must run before any `await`, the sequential (not `Promise.all`) multi-folder loop, and the Welcome-card `<button>` carve-out with its rationale all show real care. The pure-helper + unit-test split (`isFolderDropTarget` / `droppedDirectories`) also matches how this codebase likes drop logic factored.

Two non-blocking suggestions:

1. Server Edition currently swallows a folder drop silently (the bridge's `getPathForFile` stub returns `''`). Our house rule is "degrade explicitly, never silently" — a small `nodeterm:toast` ("Folder drop needs the desktop app") in the empty-paths case would tell the browser user why nothing happened. Fine as a follow-up.
2. `isFolderDropTarget` only recognizes modals via `[role="dialog"]`, but several of our overlays (Settings, the kanban card modal) don't carry that role, so a folder dropped onto them opens a project underneath. Not dangerous, just surprising — worth a follow-up rather than a blocker.

Manual verification on a running build (your unchecked test-plan item) would be good to complete before merge — particularly the "re-drag the same folder → reopens, doesn't duplicate" case, since that's the invariant this feature lives or dies by. Nice work — merge call is @eneskirca's.

PR #512, opened 2026-08-30
fix(git): stop Source Control panel from opening untracked directories as file diffs

gor3a:
## Summary
- `git status --porcelain` collapses an entirely-untracked directory into a single `?? dir/` entry by default; that path flowed unchanged into the Source Control panel's diff-node click handler, which then tried to diff a directory as if it were a file.
- Pass `-uall` so untracked files are listed individually (still respects `.gitignore`).
- Defensive skip of any leftover directory-shaped (trailing `/`) porcelain entry in the parse loop.

## Test plan
- [x] New regression test `src/core/git-service.untracked-dir.test.ts` (real temp git repo, real `git` commands) asserting individual untracked files are listed instead of the collapsed directory
- [x] Full `git-service` test suite passes (11/11)

### DIFF
--- a/src/core/git-service.ts
+++ b/src/core/git-service.ts
@@ -380,7 +380,10 @@ export class GitService {
         git(cwd, ['rev-parse', '--abbrev-ref', '--symbolic-full-name', '@{upstream}']),
         git(cwd, ['diff', '--cached', '--numstat']),
         git(cwd, ['diff', '--numstat']),
-        git(cwd, ['status', '--porcelain'])
+        // -uall forces git to list every untracked file individually instead of collapsing an
+        // entirely-untracked directory into one `?? dir/` entry (issue: SC panel showed a folder
+        // row that, when clicked, opened a diff node against a directory). Still respects .gitignore.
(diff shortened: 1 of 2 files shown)


eneskirca:
Thanks for this fix — it's a model bug PR: a one-line root-cause fix at the right layer, a defensive guard for the same invariant, and a regression test against a real git repo (which is exactly the testing style this repo uses for anything that shells out — generated commands get run for real, never just string-asserted). Putting the fix in `GitService.status` rather than the panel's click handler was the right call: it also fixes per-file stage/discard for untracked directories and covers both the desktop and Server Edition shells in one place.

One heads-up, no change requested: `-uall` means a repo with a large untracked-and-not-gitignored directory (a stray build output folder, say) now lists every file individually in the Source Control panel where it used to show one collapsed row. That's the more honest rendering and git respects `.gitignore` as you note, so I think it's the right trade — just flagging it in case anyone bisects a "why is my changes list huge" report back to this. Welcome aboard, and hope to see more from you!

Issue #462, opened 2026-08-27
[Feature proposal] Github integration?


Temikus:
@eneskirca thanks for the awesome project!

I would love to see Github PR's / issues as first-class citizens on the canvas, allowing you to easily see what PR(s) the session is working on and what issues it filed or modified.

I am happy to do a small prototype to show my thinking and contribute - would this be something you'd consider?

DingDingChae:
This feature is now being implemented on the downstream fork in a dedicated feature branch. Progress and review links will be posted as separate comments.

DingDingChae:
## 🚀 Implementation started

| Stage | Progress | State |
| --- | ---: | --- |
| Dedicated tracking issue | `[██████████] 100%` | ✅ [downstream #132](https://github.com/Ding-Ding-Projects/material-nodeterm/issues/132) |
| Feature source | `[█░░░░░░░░░] 10%` | 🏃 implementation running |
| Downstream integration | `[░░░░░░░░░░] 0%` | ⏳ pending |
| Owner review PR | `[░░░░░░░░░░] 0%` | ⏳ opens when source is complete |

**Branch:** `feat/github-pr-issue-canvas`  
**Base commit:** `add98488223effb5abf34d0fb0f1dd33656d7f14`

The implementation uses the existing guided GitHub accounts and API services. Session relationships will be explicit or backed by verified app-owned create/modify operations, not inferred from arbitrary transcript text.

People prefer one click to AFK farming browser tabs, so the pull requests and issues are getting their own canvas seats.

No test, build, package, runtime, accessibility, security, or capture verdict is claimed yet.

DingDingChae:
## ✅ Source branch ready

| Stage | Progress | State |
| --- | ---: | --- |
| Feature source | `[██████████] 100%` | ✅ published |
| Downstream pull request | `[██████████] 100%` | ✅ [#170](https://github.com/Ding-Ding-Projects/material-nodeterm/pull/170) open |
| Downstream integration | `[██░░░░░░░░] 20%` | 🏃 queued |
| Consolidated upstream review | `[████░░░░░░] 40%` | 🏃 [#463](https://github.com/eneskirca/nodeterm/pull/463) |

**Feature tip:** `357a5746c6fe11c54b6e2373e526caea8bc849f2`

[...]

PR #513, opened 2026-08-30
refactor(kanban): make card sources a registry

Temikus:
Phase 1 of the #462 plan: internal plumbing only, no behaviour change and no new UI. PR cards land on top of this in the next PR.

**Review decision:** does `lib/kanbanSources.ts` declare the right things about a card source, and does the board still behave identically? The existing kanban tests are unchanged, which is the mechanical half of that answer.

## What changed

`KanbanColumn` stitched two mechanisms together: session cards arrived as `cards` and were placed by the board's own assignments, GitHub cards arrived as eight `github*` props and were placed by the provider. The drag union branched on `kind: 'github'` by hand, and the source filter was three hardcoded strings. PR cards on top of that would have been a third copy of the path.

New `src/renderer/lib/kanbanSources.ts` declares each source once - the same membership-plus-one-leaf shape `AGENT_CONFIG` uses:

- `label` - its button in the source filter
- `placement` - `assignment` (the board's persisted assignments, reorderable in a column) or `provider` (the provider reports the column; a move is its write, not a board edit)
- `lane` - in-column stacking order
- `configured(board)` - whether the source can contribute cards to this board at all

Consumers stop spelling source names:

- `KanbanSourceFilter` renders `all` plus the registry.
- `KanbanColumn` takes one `lanes` prop (`{sourceId, cards, footer?, count}`), places them via `byLane`, and knows nothing about where a card came from - eight props and two hardcoded render blocks gone.
- `KanbanView` builds each source's leaf, and the drop path branches on `placement` via `isProviderDrag` rather than on the string `github`.

Two details worth a reviewer's eye:

- **Two orders are declared deliberately.** Declaration order is the filter's button order (All / GitHub / Sessions); `lane` is the in-column stacking order (sessions above issues). They genuinely differ on the board today, so pinning both is what stops either being re-spelled at a call site.
- **A lane's `count` is passed, not derived from its cards.** A provider reports a column total that can exceed the page fetched so far, which is what the old `displayCount` prop carried.

Out of the registry on purpose, per the scope agreed in #462: the virtual **Ungrouped** column stays board semantics, and `validKanban` stays the single shape gate on every load path - no entry grows its own validation.

## Verification

<details>
<summary>Tests, typecheck, surfaces</summary>

- **Existing kanban tests pass unchanged** - no test file was edited. Kanban suites: 67 passed.
- New tests: `kanbanSources.test.ts` (registry shape, lane order, placement, filter visibility, unknown id) and `KanbanColumn.test.tsx` (lanes placed in registry order regardless of the order handed in, lane footer placement, count sums every lane).
- `npm run typecheck` clean.
- Full suite: 9041 passed, 29 failed - the identical failing set on a clean checkout of `main` here (unpatched `node-pty` in `node_modules`, real-tmux/ssh/session-host suites). None are kanban.
- Surfaces: Desktop and Server Edition share this renderer code, and both are unaffected by construction (pure renderer refactor, no bridge member touched). Mobile: N/A - no board there.

</details>

## Follow-ups (unchanged from the #462 plan)

2. PR cards, read-only - one registry entry plus one list path.
3. Explicit attachment plus the canvas chip.
4. create-from-nodeterm.

Draft while you take a first look at the registry entry's shape - happy to resize it before it carries a second source.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2125,7 +2125,23 @@ the Settings section and ShortcutsPanel start disagreeing about what a chord mea
   issues (`GitHubIssueCardView` via `state/githubIssues.ts`, opened through
   `GitHubIssueSummaryModal`, a column move that closes/reopens the issue confirms first). A
   **source filter** (`KanbanSourceFilter`: All / GitHub / Sessions) and a transient per-board
-  **label filter** narrow what shows. **Labels** are a per-project palette (`ProjectKanban` labels,
+  **label filter** narrow what shows.
+  **Where a card comes from is a registry, not a branch per call site** (`renderer/lib/kanbanSources.ts`,
+  2026-08-30 — the same membership-plus-one-leaf discipline `AGENT_CONFIG` uses): each entry declares
+  its filter `label`, its `placement` (`assignment` = the board's own persisted assignments,
+  reorderable within a column; `provider` = the provider reports the column, the board persists
+  nothing and a move is the provider's write), its in-column `lane` order and whether it is
+  `configured` for a given board. Two orders live there deliberately: **declaration order is the
+  source filter's button order** (All · GitHub · Sessions), **`lane` is the in-column stacking order**
+  (sessions above issues) — they genuinely differ, and pinning both is what stops either being
+  re-spelled elsewhere. `KanbanColumn` therefore takes ONE `lanes` prop (`{sourceId, cards, footer?,
+  count}`) instead of a `cards` + eight `github*` props, places them via `byLane` and names no source;
+  the board builds each source's leaf, and the drag union branches on `placement` (`isProviderDrag`)
+  rather than on the string `'github'`. A lane's `count` is passed rather than derived from
+  `cards.length` because a provider reports a server-side total larger than the page fetched so far.
+  What deliberately did NOT move into the registry: the virtual **Ungrouped** column (board
+  semantics, not a source's concern) and `validKanban`, which stays the single shape gate on every
+  load path — a registry entry must never grow its own parallel validation. **Labels** are a per-project palette (`ProjectKanban` labels,
   edited inline via the Notion-style `LabelPicker`: create/assign/rename/recolor/delete through the
   pure `lib/kanban.ts` transforms) plus each GitHub issue's own labels, both filterable. The canvas stays MOUNTED under the opaque overlay (agent-status
   listeners live in Canvas.tsx; `display:none` would 0×0-resize every terminal into a tmux
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,6 +64,14 @@ The **canvas and the kanban board are two views of the same nodes.** When you ad
 canvas node — a header action, a badge, a menu item — ask whether the board's card and card modal
 need it too, and wire it in the same change.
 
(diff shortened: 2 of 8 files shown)


eneskirca:
This lands the Phase-1 contract from #462 as agreed, and it reads the way a no-behavior refactor should: `KanbanColumn` is net subtraction (eight `github*` props and two hardcoded render blocks gone), the registry entry's shape is the right altitude (`label` / `placement` / `lane` / `configured` — nothing speculative), and pinning the two orders separately with a comment saying *why* they differ is exactly the kind of thing that stops a call-site re-spelling later. The CLAUDE.md + CONTRIBUTING.md updates in the same diff are appreciated.

Verifying against the acceptance criteria from #462:

- **Existing tests unchanged: confirmed.** `git diff --diff-filter=M` over `*.test.*` is empty; both test files in the diff are new. `KanbanSourceFilter.test.tsx` in particular — which pins the All/GitHub/Sessions button order and the onChange values — passes untouched against the registry-driven rendering, which is a nice mechanical proof of the filter's behavior. Kanban suites plus every file importing the touched components pass here (121 + 88 tests), full `renderer/lib` sweep green (976), typecheck clean.
- **Ungrouped stays board semantics: confirmed.** The registry holds exactly `github` and `sessions`; the virtual column remains the `column: null` case and `lib/kanban.ts` is untouched.
- **`validKanban` stays the single gate: confirmed.** Nothing under `src/core`/`src/shared` is in the diff; both call sites (`fileToProject`, `loadV3`'s inline branch) are intact, and `configured` does reads only.

I also traced the old/new expression pairs (visibility gates, header counts, footer gating, drag union order, card props including the `moving`/`issueStatus` optional chaining — safe, the store always initializes both) and found no drift. The persisted board shape is unchanged; the registry is renderer-only.

Nothing blocking. Three notes, two of them for Phase 2 rather than this diff:

1. **`isProviderDrag` will need to change shape when PR cards arrive.** The predicate asserts `drag is Extract<CardDrag, { sourceId: 'github' }>` from a runtime check on `placement === 'provider'`. With one provider source that's exact; with two, a PR drag would satisfy the runtime check, get narrowed to the github member, and `requestGitHubMove(drag.issue)` would receive `undefined` — with TypeScript silent, since the predicate is author-asserted. The provider *write* is inherently per-source, so Phase 2's drop path wants a per-source dispatch (or the write carried on the drag member). Your comment already gestures at this; flagging it so it's a conscious step in the next PR rather than a surprise.
2. **Tiny asymmetry in `lanesFor`:** the github lane consults both `sourceVisible` and `configured`, the sessions lane only `sourceVisible`. Harmless today (`sessions.configured` is constant true), but the registry documents `configured` as deciding whether a source can contribute at all — worth making the sessions branch consult it too, so the invariant holds by construction when a third source copies the pattern.
3. **Cosmetic:** `KanbanColumn` keeps the "identity stability lets memo() skip untouched columns" comment, but `lanes` is a fresh array every render, so the column memo no longer bites (it didn't on main either — the inline `onLoadMoreGitHub` closure was already defeating it — and the leaf card memos still hold via the `dropAtCardFor` cache). The comment could be trimmed or reworded to point at the leaf memos instead.

None of these need to hold this PR — happy for 1 and 3 to ride along with Phase 2, and 2 is your call whether it's a one-liner here or there. Merge itself waits on @eneskirca as usual.

PR #514, opened 2026-08-30
feat(agents): pause a session so it does not auto-resume on reopen

gor3a:
## Summary
- Adds a persisted `paused` flag alongside the existing Eco `hibernated` state. Pausing a node (manually — a shallow "keep tmux" or deep "pause & end session" choice — or via Eco's idle sweep when the new "keep paused after closing" setting is on) exits the agent CLI and marks the node so it does **not** auto-resume on the next reveal or on a cold restart; only an explicit Resume brings the conversation back.
- Scoped to agent (resumable CLI) nodes. State lives in the same machine-local `agentStatus` tier as `hibernated` — never written to `project.json`.
- `agentStatus`: new `paused`/`setPaused`, with the same live-hook-state self-heal as `hibernated`, but (unlike `hibernated`) deliberately not cleared by a `fresh` cold restart. `hibernatedPane` now persists alongside either flag, and a fresh mount of a paused node records the pane it actually lands on, so a later Resume can recognize a shell outside the `isShellCommand` allowlist (nu/xonsh/pwsh).
- `TerminalNode`: cold-restore auto-resume and every auto-wake trigger (mount timer, visibility edge, kanban card modal open) gated on `!paused`; new `pause` closure (`registerAgentPause`) built from the same `performExitPhase` + `transport.recycle` building blocks the existing hibernate/restart closures use — and it skips the exit entirely when the node is already hibernated, so Pause never types `/exit` into a bare shell.
- `Canvas`: "Pause session"/"Resume session" node-menu entries, shared by the canvas right-click and the sessions sidebar (same menu builder); manual Resume routes through the existing wake trigger so it gets the same input-buffer protection and retries as the automatic wake; Eco's sweep can mark a hibernated node paused too, gated on its own setting.
- Settings: "Keep paused after closing" checkbox under Hibernate idle agents. Kanban session cards show a read-only PAUSED badge.
- Command palette and an interactive kanban-card-modal control are deliberately deferred — the node context menu (canvas + sidebar) and the header chip cover the primary flow.

## Test plan
- [x] `npm run typecheck` (both tsconfig.node.json and tsconfig.web.json) — clean
- [x] Targeted suites: `agentStatus.persist.test.ts`, `agentStatus.test.ts`, `agentStatus-session.test.ts`, `hibernation-policy.test.ts`, `hibernationCandidates.test.ts`, `agent-restart.test.ts`, `canvas-wiring.test.tsx`, `status-lifecycle.test.ts` — all passing (219 tests)
- [x] Full `npm test` run — no regressions; remaining failures are pre-existing/environmental (real-tmux/real-shell/native-pty tests that also fail on a clean `origin/main` checkout in this sandbox, confirmed via `git stash` comparison) and touch none of the files in this PR
- [x] Independent code review pass; all High/Medium findings addressed (kanban modal auto-resuming a paused node, manual Resume bypassing the wake input buffer, `hibernatedPane` loss on cold restart, menu/closure eligibility mismatch) plus several Low findings (stale comments, pausing an already-hibernated node typing junk into its shell)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -466,8 +466,9 @@ Lifecycle, by intent:
   process. (2) **Fire-time re-asks**: still-offscreen, remote, eligibility — a plan-time verdict
   is stale by seconds. (3) `hibernated` **self-heals** on live hook states + SessionStart (never
   on `done` — a late Stop POST must not undo a just-performed hibernate); cold restore (`fresh`)
-  clears it and lets the normal auto-resume own the node — **UNLESS the node is also `paused`**
-  (see below), which is the one deliberate exception to that rule. (4) **Ordering with offscreen
+  clears `hibernated` UNCONDITIONALLY and normally lets auto-resume own the node — **`paused` (see
+  below) is what makes that auto-resume itself conditional**, the one deliberate exception: the
+  flag it gates is still cleared, only the relaunch is skipped. (4) **Ordering with offscreen
   release**:
   Eco defers the Phase-2 viewer release until the node hibernates (hard cap idle+offscreen), but
   ONLY when the idle clock is known (`idleKnown` — `lastEventAt` is transient, so after an app
@@ -490,9 +491,14 @@ Lifecycle, by intent:
   which reuses the SAME `wakeHibernatedNode` trigger the SLEEPING/PAUSED chip's click uses, so it
   gets the same `WakeInputBuffer` splice protection and retry budget). Pausing an already-hibernated
   node skips the exit phase entirely (`alreadyExited` in the closure) — asking an idle CLI-less
-  shell to quit would type `/exit` into it as a real command. Node menu only today (canvas
-  right-click + the sessions sidebar row menu, which shares the same `selectionItems` builder, plus
-  a read-only kanban card badge and a clickable one in the card modal); no command palette entry.
+  shell to quit would type `/exit` into it as a real command. `paused` is ALSO excluded from Eco's
+  own candidate plan and its exit closure's fire-time re-ask (`HibernationCandidate.paused`,
+  `hibernationCandidates.ts`) — a deep-paused node has `hibernated` unset (its tmux was recycled,
+  not exited), so `!hibernated` alone would still admit it to a sweep whose dropped SessionEnd hook
+  POST left a stale `done` behind: the same `/exit`-into-a-bare-shell mistake `alreadyExited` closes
+  on the manual path, closed here on the automatic one. Node menu only today (canvas right-click +
+  the sessions sidebar row menu, which shares the same `selectionItems` builder, plus a read-only
+  kanban card badge and a clickable one in the card modal); no command palette entry.
 
 The node id is the `persistKey` (passed to `transport.create`), so it must stay stable.
 If tmux is unavailable, `PtyManager` falls back to a plain shell (no cross-restart
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -466,12 +466,33 @@ Lifecycle, by intent:
   process. (2) **Fire-time re-asks**: still-offscreen, remote, eligibility — a plan-time verdict
   is stale by seconds. (3) `hibernated` **self-heals** on live hook states + SessionStart (never
   on `done` — a late Stop POST must not undo a just-performed hibernate); cold restore (`fresh`)
-  clears it and lets the normal auto-resume own the node. (4) **Ordering with offscreen release**:
+  clears it and lets the normal auto-resume own the node — **UNLESS the node is also `paused`**
+  (see below), which is the one deliberate exception to that rule. (4) **Ordering with offscreen
+  release**:
   Eco defers the Phase-2 viewer release until the node hibernates (hard cap idle+offscreen), but
   ONLY when the idle clock is known (`idleKnown` — `lastEventAt` is transient, so after an app
   restart nothing can hibernate and deferring would make Eco a memory regression). Eco is
   structurally inert for sessions with no turn in the current app run — documented follow-up.
+  The deferral is also unaware of `paused`: a deep-paused node's freshly recycled shell keeps its
+  xterm alive until the hard cap, waiting for a hibernation that (being already exited, or having
+  no CLI to exit) can never come — a second documented follow-up.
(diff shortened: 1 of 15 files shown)


eneskirca:
Thank you for this — it's a genuinely impressive first contribution. You found the right tier for the flag (agentStatus localStorage, never `project.json`), gated every auto-resume path I could find (cold-restore, mount timer, visibility edge, card-modal open, the Eco nudge) while leaving the explicit ways back ungated, and the hibernation interplay is handled with real care: the `fresh && paused` pane re-record for non-allowlist shells, the already-hibernated skip so Pause never types `/exit` into a bare shell, and routing manual Resume through the wake trigger for the input-buffer protection are all exactly the right calls. Typecheck is clean and the targeted suites pass on my checkout too.

A few small things before merge:

1. **Card modal comment vs. reality** — the comment at the modal-open wake (`Canvas.tsx`, "The modal shows the PAUSED chip") isn't true yet: `CardModal`/`ModalTerminal` render no status chip, only the `SessionCard` badge exists. So a paused node opened via its card shows a bare shell with nothing in the modal explaining why. Either add a small PAUSED indicator to the modal header or correct the comment and note the gap — a comment that describes a mechanism that doesn't exist is the kind of thing this repo has been bitten by before.

2. **CLAUDE.md** — the cold-restore section states "fresh=true means… the renderer re-launches the agent CLI", and the Eco section documents `hibernated`'s self-heal rules. Your `paused` flag is a deliberate exception to both, and it's currently documented only in code comments. Per the Conventions section, please add it to CLAUDE.md in the same PR (a sentence in the cold-restore bullet and one beside the `hibernated` rules is enough), so the next person changing that path knows the exception exists.

3. **A test pin on the wiring** — the persist tests are great, but the part most likely to drift is the TerminalNode gate: a test asserting that a `fresh` mount of a `paused` node does not deliver a resume command (and that the mount-timer wake stays quiet) would pin the feature's core promise the way `agentStatus.persist.test.ts` pins the storage half.

4. (Optional follow-ups, fine to defer with a note) "Restart agent" on a paused node still types KILL_LINE+`/exit` into the bare shell before relaunching — your `alreadyExited` skip in the pause closure is the right pattern to share with it. And `offscreen-policy.ts`'s Eco deferral doesn't know about `paused`, so a deep-paused node keeps its xterm until the hard cap waiting for a hibernation that can't come.

None of these are structural — the design is right, and the relay exclusion matching the `restartShell` precedent plus the narrower `st?.sessionId` eligibility for the menu row show you read the surrounding code closely. Looking forward to the next revision — merge call is @eneskirca's.

gor3a:
Thanks for the thorough review — addressed all four points:

1. **Card modal** — added a clickable PAUSED badge to `CardModal`'s header (reuses the same `wakeHibernatedNode` trigger the canvas chip uses, so it gets the same input-buffer protection). The stale comment now matches reality.
2. **CLAUDE.md** — added the `paused` exception to the cold-restore bullet, plus a new "Pause session" bullet beside Eco covering both depths, the shared wake trigger, and the two follow-ups from your point 4 (documented rather than fixed, see below).
3. **Test pin** — extracted `shouldColdResume`/`shouldAutoWake` as pure predicates in `hibernation-policy.ts` and wired every call site (cold restore, mount timer, visibility edge, kanban modal-open) through them instead of inline `!paused` checks. Unit-tested in `hibernation-policy.test.ts`.
4. **Follow-ups** — noted both in CLAUDE.md rather than fixing in this PR: sharing the `alreadyExited` skip with plain "Restart agent", and `offscreen-policy.ts`'s Eco deferral not knowing about `paused`. Happy to pick either up in a follow-up PR if you'd rather they land now.

Typecheck clean, full test suite re-run (same pre-existing/environmental failures as before, none in touched files), targeted suites green (199 tests including the new predicate + persistence tests).

gor3a:
Ran a second independent fable review over the full diff (both commits) to verify the first round's fixes actually hold and catch anything the fix commit introduced.

**All of H1/M1/M2/M3/L2 confirmed genuinely fixed.** Found a few more small things, now addressed in a third commit:

- **The one real gap (worth flagging):** Eco's own idle sweep could still exit a deep-paused node — `hibernated` is unset for that depth (tmux was recycled, not exited) — into a bare shell if its SessionEnd hook POST got dropped somewhere. `HibernationCandidate` now carries `paused` and the plan/fire-time-re-ask both exclude it, same rule `alreadyExited` already applied on the manual path.
- `pauseAgentNode` now shows a notice instead of silently no-op'ing on a missing registration (parity with `resumeAgentNode`, which already did).
- "Pause & end session" is disabled with a named reason on a relay session, instead of a generic refusal that named the wrong cause.
- The live-state self-heal now also drops a paused-only node's `hibernatedPane` (mirrors what resuming already does), closing a small stale-record window.
- Two comments that had gone stale the moment the card-modal PAUSED chip landed, and one CLAUDE.md sentence flagged as ambiguous — reworded.

Typecheck clean, full suite re-run (same pre-existing/environmental failures, none in touched files), targeted suites green (209 tests, several new ones pinning the sweep-exclusion and pane-cleanup fixes).

eneskirca:
Re-reviewed both fix commits (`01466058`, `e0398360`) — the four review items are in, and the second-pass sweep went beyond them:

1. Card modal has a real, clickable PAUSED badge wired through the same wake trigger (so it inherits the input-buffer protection), and the once-stale comment now describes a mechanism that exists.
2. The `paused` exception is documented in CLAUDE.md at both places it bends the rules (cold-restore, beside Eco), with the two deferred follow-ups written down rather than left implicit.
3. The gates are now pure predicates (`shouldColdResume`/`shouldAutoWake`) wired through every call site and unit-tested — better than the pin I asked for, since the predicate can't drift from its call sites the way inline checks could.
4. The Eco-sweep exclusion for deep-paused nodes (`HibernationCandidate.paused`, checked at plan AND fire time) closes a real gap I hadn't flagged — good find.

Measured on the tip: typecheck clean, targeted suites pass (120 tests), CI green, trial-merge onto current main clean. Addressed, ready. The two documented follow-ups (sharing `alreadyExited` with plain Restart, and teaching `offscreen-policy` about `paused`) are fine as a separate PR — happy to see either.

PR #516, opened 2026-08-30
docs(canvas-control): say that --prompt arrives on one line

Temikus:
`assembleLaunchCommand` collapses every whitespace run in the prompt:

```ts
shellSingleQuote(inputs.initialPrompt.replace(/\s+/g, ' ').trim())
```

That is load-bearing, not tidying: the prompt is a positional argument on the agent CLI's launch command line, and that line is *typed into the pane* by `writeWhenShellReady` → `deliverCommand`, which echo-verifies it before submitting. A newline would submit the half-typed line and fail verification.

Neither agent-facing body mentions it. Grepping `canvas-control-core.ts` for `newline` or `single line` returns nothing.

## The failure this cost

The expensive case is a prompt beginning with `/`. Flattened, a leading `/model sonnet` followed by the task reads to the agent as one slash command whose argument is the entire rest of the prompt:

```
> /model sonnet Do not use any tools. Reply with exactly two lines: MODEL: <the
exact model id you are running as> SAW: <yes if a slash command named /model ...
  |- API error: 400
     {"error":{"message":"model: String should have at most 256 characters"}}
```

The command fails, the task is never seen, and the node then sits at an idle prompt looking healthy — **including to `--after`**, which arms every station behind it against an upstream that produced nothing. In a dependency chain one such prompt can start the rest of the pipeline on bad ground.

## What this changes

Documentation only, no behaviour change.

- Both bodies state that `--prompt` arrives on one line and that structure is lost.
- Both name the leading-slash trap and its `--after` consequence explicitly.
- Both point at the two supported alternatives: `--model` for a model, and `send --node <id> --text` for a long or structured brief.
- The collapse is commented at its source, so the next reader knows why it cannot simply be deleted.
- A doc test pins both claims.

## Two things deliberately left out, for you to call

1. **Refusing a prompt that starts with `/`** in `parseControlRequest`. Cheap to add and arguably better than documenting a trap — but it is a behaviour change to a verb agents already call, so I would rather you decide than have it arrive inside a docs PR.
2. **Preserving newlines properly.** Given the above this means not putting the prompt in argv at all: launch the CLI bare, then deliver the body through `sendText`, which already frames multi-line text via tmux `paste-buffer -p`. That changes how every agent node starts and wants its own issue.

## Testing

`canvas-control-core.test.ts` and the `shared/agents` suites pass (316 tests). Same two pre-existing environmental failures as on `main` (`node-pty-patch`, `remote-atomic-write`), verified by stashing.

https://claude.ai/code/session_01HhoRtz59pbXu7XpXfTrH9R

### DIFF
--- a/src/main/canvas-control-core.test.ts
+++ b/src/main/canvas-control-core.test.ts
@@ -292,6 +292,19 @@ describe('parseControlRequest', () => {
     expect(isDestructiveVerb('sticky')).toBe(false)
   })
 
+  it('both agent-facing texts warn that --prompt is one line and must not start with a slash', () => {
+    for (const body of [buildCanvasSkillBody('/x/shim.sh'), buildCanvasControlInstructions('/tmp/nodeterm.sh')]) {
+      // `assembleLaunchCommand` collapses every whitespace run in the prompt, because the prompt
+      // rides argv on a line that is typed into the pane. An agent that does not know this writes
+      // a numbered brief and gets one paragraph.
+      expect(body.toLowerCase()).toContain('one line')
+      // The failure that costs a whole station: flattened, a leading slash command swallows the
+      // task as its argument, and the node then reads as idle to `--after`. Silence here is what
+      // let that ship.
+      expect(body).toMatch(/start a prompt with `\/`|begin a prompt with `\/`/i)
+    }
+  })
+
   it('both agent-facing texts document --model on the open verbs and per-role model on spawn-team', () => {
     for (const body of [buildCanvasSkillBody('/x/shim.sh'), buildCanvasControlInstructions('/tmp/nodeterm.sh')]) {
       // The flag is the only cost lever an orchestrator has: without it every station it opens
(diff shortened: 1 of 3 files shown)


eneskirca:
Verified the whole chain and the claim holds exactly as written: the collapse at `launch.ts` is load-bearing for the argv→typed-line delivery path, both open verbs and spawn-team route through `assembleLaunchCommand`, and the leading-`/` failure mode plus its `--after` consequence are real and previously undocumented. Editing the two generators rather than any derived file is right, and thanks for the test pin — that's the house rule here.

Two small asks, neither blocking:

1. `spawn-team`'s per-role `prompt` fields take the identical funnel (`createAgentNode` → `assembleLaunchCommand`), so a role brief written as a numbered list flattens the same way. Could you add one sentence (or a cross-reference to the `--prompt` note) in the spawn-team section of both bodies, so an agent reading only the team recipe sees the trap?
2. The recommended alternative — `send --node <id> --text` — is gated by the project's agent-messaging switch, which is off by default. Worth a parenthetical pointing at the gate (or naming confirm-gated `write` as the always-available fallback), so the escape hatch you point to doesn't refuse on a stock project.

Also appreciated the restraint on the two deferred items — both are behavior changes and belong in their own issues. Tests green here (75/75 across `canvas-control-core.test.ts` + `launch.test.ts`), typecheck clean. Merge decision is @eneskirca's as usual.

PR #517, opened 2026-08-30
fix(canvas-control): distinguish an unanswered confirm dialog from a denial

Temikus:
A confirm-gated verb (`write`, `close`, `open-project`, `close-worktree --mode remove`) has two ways to come back unsuccessful, and they carry opposite guidance:

- `denied by user` — a decision. The skill text says a denial is final and must not be retried.
- The 120s timeout — nobody reached the dialog. Plainly worth retrying.

The timeout said `timed out (no response / not confirmed)`, which reads as though it covered both. A caller that had merely waited out an unanswered dialog treated it as a refusal and gave up — which is how I read it during a seven-station run, and why I originally filed this as "a timeout and a denial are indistinguishable to the caller". They are in fact distinct; only the wording was.

## Changes

- The timeout names its window and says it is retryable: `no answer within 120s — the confirmation dialog may still be open; safe to retry`.
- `120_000` becomes `CONTROL_REQUEST_TIMEOUT_MS`, so the number in the message cannot drift from the number in the `setTimeout`.
- Both agent-facing bodies now state the two replies and the follow-up each one warrants. Previously they said only "may be denied", which gave a caller nothing to distinguish its own timeout from.

## Testing

`canvas-control-core.test.ts` — 47 passing, including a new pin that both bodies name both replies. `npx tsc --noEmit -p tsconfig.node.json` clean for the touched files.

https://claude.ai/code/session_01HhoRtz59pbXu7XpXfTrH9R

### DIFF
--- a/src/main/canvas-control-core.test.ts
+++ b/src/main/canvas-control-core.test.ts
@@ -292,6 +292,15 @@ describe('parseControlRequest', () => {
     expect(isDestructiveVerb('sticky')).toBe(false)
   })
 
+  it('both agent-facing texts separate a denial from an unanswered dialog', () => {
+    for (const body of [buildCanvasSkillBody('/x/shim.sh'), buildCanvasControlInstructions('/tmp/nodeterm.sh')]) {
+      // The two answers carry opposite guidance — a denial is final, a timeout is retryable — and
+      // a body that names only "may be denied" leaves a caller reading its own timeout as refusal.
+      expect(body).toContain('denied by user')
+      expect(body).toContain('no answer within 120s')
+    }
+  })
+
   it('both agent-facing texts document --model on the open verbs and per-role model on spawn-team', () => {
     for (const body of [buildCanvasSkillBody('/x/shim.sh'), buildCanvasControlInstructions('/tmp/nodeterm.sh')]) {
       // The flag is the only cost lever an orchestrator has: without it every station it opens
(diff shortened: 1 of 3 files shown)


eneskirca:
Confirmed the two replies are genuinely distinct in the code — `denied by user` comes only from the renderer's explicit cancel handlers, the 120s string only from main's `pendingControl` timeout — so this is the right fix, and updating both generated bodies plus the test pin in the same PR is exactly the contract for outcome-meaning changes here. Also agree RETRYABLE is correctly left alone: that table is the messaging-outcome family (`Record<AgentMessageOutcomeKind, boolean>` over send/reply delivery outcomes), and this timeout isn't one of its kinds.

One ask before merge: `CONTROL_REQUEST_TIMEOUT_MS` closes the drift between the `setTimeout` and its own message, but both doc bodies (and the test) still hardcode the literal `120s` — so a future change to the constant would update the runtime string while the agent-facing bodies keep promising 120, with the test still green (it only reads the bodies). Could you move the constant into `canvas-control-core.ts`, export it, interpolate it into both bodies (`no answer within ${CONTROL_REQUEST_TIMEOUT_MS / 1000}s`), and import it in `index.ts`? That's the same derive-don't-retype shape `messagingGuidanceLines` uses, and it makes the number un-driftable in all four places at once.

Tiny non-blocking note: the same timeout fires for non-confirm verbs when the renderer never answers (a wedged window timing out a `list`), where "the confirmation dialog may still be open" doesn't strictly apply — "may" covers it and "safe to retry" is right in both cases, so fine as-is, just flagging in case you want to hedge the phrasing a touch.

Heads-up: this and #516 both add a test block at the same insertion point in `canvas-control-core.test.ts`, so whichever lands second will need a trivial rebase. Tests green here (48/48 + the two renderer source-pin suites), typecheck clean.

PR #522, opened 2026-08-30
fix(keybindings): keep Ctrl+W in focused terminals

coljac:
Follow-up to #383 and #387.

Issue #387 correctly made both main-process owners stand down when an off-mac terminal has focus: the `before-input-event` intercept  leaves `node.close` untouched, and the native Close menu item is disabled.

Under the default app-first policy, `terminalChordBubbles` resolved Ctrl+W to  `node.close` and told xterm to skip its own keymap. Canvas deliberately has no renderer handler for main-intercepted commands, so  `^W` never reached the pty and Electron's fallback close behaviour could still show the quit dialog.

**Change**

- Keep every `MAIN_INTERCEPTED_COMMAND_IDS` chord inside xterm when it reaches the renderer. Main still owns these commands  normally; this branch only matters when main has deliberately stood down.
- Add a regression that follows the actual hand-off: focused-terminal Ctrl+W must not bubble, and `terminalKeyAction` must return  `pass` so xterm sends `^W` to the pty.
- Document the renderer leg alongside the main/menu invariant in `CLAUDE.md`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1888,8 +1888,14 @@ the Settings section and ShortcutsPanel start disagreeing about what a chord mea
     fall through UNTOUCHED and `syncMenuForStandDown` disables the Close menu item on top of the
     shared list. mac's ⌘W is deliberately unaffected (not a shell key), and ⌘/Ctrl+M and ⌘/Ctrl+0
     keep firing — this is one chord whose terminal meaning outranks its app meaning, not a policy
-    change. One predicate, two consumers, pinned in `keydown-intercept.test.ts` (including a
-    source-level wiring pin, since the menu leg lives against a real Menu in index.ts).
+    change. Falling through main is not enough: xterm's custom key handler runs before the Canvas
+    dispatcher, whose main-intercepted command cases deliberately have no renderer handlers.
+    `terminalChordBubbles` must therefore refuse every `MAIN_INTERCEPTED_COMMAND_IDS` command; if
+    it returned true for `node.close`, xterm would withhold `^W` while the unclaimed event bubbled
+    to Canvas. One predicate, two main-process consumers are pinned in `keydown-intercept.test.ts`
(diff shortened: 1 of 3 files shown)


eneskirca:
Thanks — this is the right fix, and the write-up shows you traced the whole hand-off. Verified against the #383 machinery:

- The main-process legs (intercept fall-through + Close menu suspension) already stand down off-mac with a focused terminal, policy-independently, so this renderer exclusion takes nothing away from app-first users — the close was already stood down; the `^W` byte was just being lost between xterm's `'bubble'` and a dispatcher that deliberately has no `node.close` handler. Excluding `MAIN_INTERCEPTED_COMMAND_IDS` in `terminalChordBubbles` is exactly where that belongs, and it stays rebind-safe because both the intercept and this helper resolve the same live registry.
- Where main does NOT stand down (mac ⌘W, terminal unfocused), the intercept preventDefaults before the page sees the key, so this branch is unreachable and the default behavior is byte-identical — confirmed.
- Nice side effect: `node.toggleMarkdown` was equally captured — Ctrl+M (which is CR) in a focused Server Edition terminal used to be withheld from the pty and then dropped by the handler-less dispatcher. It now reaches the shell.

Two tiny optional asks, neither blocking:

1. Could the regression test also assert the `node.toggleMarkdown` case (e.g. `Ctrl+M` → bubbles false)? Right now it pins one member of the list; a second member would pin the *rule*.
2. The `terminalKeyAction` import into `keybindingOverrides.test.ts` is fine since it follows the real call shape in `TerminalNode`, but a one-line comment saying it mirrors `TerminalNode.tsx`'s `attachCustomKeyEventHandler` sequence would keep future readers from "cleaning up" the cross-module import.

Typecheck and the keybinding/terminal suites pass locally (276 + 491 tests). LGTM — merge call is @eneskirca's.

PR #526, opened 2026-08-30
fix(windows): node-pty ConPTY exact-close native patch the session host already relies on

eneskirca:
## What

Adds the **Windows leg** to `scripts/patch-node-pty.mjs` — a version-pinned native patch for node-pty 1.1.0's ConPTY baton/HPCON leak — plus behavior pins for the adapter that was already written against it.

## The gap this closes (found while mining PR #448)

The session-host extraction (#305) shipped `src/session-host/windows-conpty.ts`, whose `closeExactWindowsConpty` refuses anything but a literal `true` from `conpty.kill(id)` — a contract only a **patched** node-pty can satisfy. The patch itself never landed: our script only carried the darwin ptmx-leak fix (microsoft/node-pty#950). So the adapter could never succeed, and the underlying leak stands:

- node-pty's native exit thread deletes its `pty_baton` as soon as the shell process handle signals, **without closing the HPCON the baton owns** (`pty_baton` has no destructor that closes it).
- Our session-host kill path (`src/session-host/host.ts`) terminates the tree with `taskkill /T /F` **first**, then releases the pty — so the exit thread wins that race by construction, the later native `kill(id)` silently finds no baton, and one host-parented conhost stays alive per killed session for the life of the session-host daemon (which is long-lived by design).

## The patch

Adapted from the material-nodeterm fork's PR #448 branch (BUSL-derivative of this repo; upstreaming is fine), 8 anchored edits to `src/win/conpty.cc`:

1. Baton access serialized behind a `std::mutex` (scoped-lock parameters make the ownership precondition impossible to omit), `pty_baton` gains a `closeExactPseudoConsole()` that exchanges the handle before closing so it can never double-close.
2. **The exit thread closes the exact HPCON before deleting the baton** — this alone fixes our existing taskkill-first kill path with zero JS changes.
3. The close primitive is resolved at ConPTY creation (fails the spawn loudly if unresolvable).
4. `kill(id)` returns `true` only as positive proof it found and closed that exact handle — the contract `windows-conpty.ts` requires. Stock callers that ignore the (previously void) return are unaffected.

The darwin leg is kept unchanged (the fork deleted it because it dropped macOS; we ship macOS). Both files are patched on every host — `conpty.cc` only compiles for the win32 native target, so patching on mac/Linux is inert and keeps packaged rebuilds honest.

**Deliberately NOT wired:** `closeExactWindowsConpty` stays out of the ordinary kill path. After `taskkill`, the exit thread usually wins the race and (under the patch) has already closed the HPCON itself, so the strict primitive would report `false` there and fail healthy kills. It exists for a future pre-first-output teardown that bypasses the exit thread; CLAUDE.md now says so.

## Tests

- `src/session-host/windows-conpty.test.ts` (new): behavior pins for the previously untested adapter — exact-close never enumerates console processes or calls the broad JS agent kill; anything but `true` fails closed; drifted private shape refuses before touching any primitive; protocol-crossing errors carry no private process details; release order is input → conout worker → output socket (the output close IS node-pty's public `onExit`).
- `src/main/node-pty-patch.test.ts`: new describe block pins the Windows marker, the close-before-delete ordering in the exit thread, and the Boolean kill result in the patched source.
- Anchor verification is the script itself: every anchor must match exactly once or it exits non-zero; verified against the pristine 1.1.0 `conpty.cc` (all 8 matched), and re-running is a marker-gated no-op.

Locally: typecheck clean; full suite 9259 passed / 1 failed — the failure is `src/core/local-send-keys.realtmux.test.ts` (real-tmux spawn under full-suite load), which passes in isolation on both this branch and main; unrelated (this PR touches no runtime code path).

## Device checklist (what a Linux box cannot prove)

The fork's releases compiled and shipped this patch, but their workflows run no unit tests, and we have not re-measured natively. Owed on a real Windows machine:

1. `npm ci` → postinstall patches + electron-rebuild **compiles** the patched `conpty.cc` (MSVC accepts the mutex/scoped-lock edits).
2. Session create → kill on a **silent** shell (no output yet): confirm no orphan `conhost.exe` remains parented to the session host after the kill (Process Explorer / `Get-CimInstance Win32_Process`), and that node-pty's real `onExit` still arrives (the kill path's acknowledgement depends on it).
3. Same check on a **noisy** session (normal case) and on app-quit-with-live-sessions.
4. Confirm `assert(remove_pty_baton(lock, …))` still executes in the Release build (node-gyp defines `NDEBUG` in Release; this hazard is inherited from stock 1.1.0's identical `assert(remove_pty_baton(id))`, so it is not a regression — but if the assert compiles out, batons accumulate and `kill(id)` on an exited session returns `false`; worth one observation).
5. `win-package-smoke.yml` (workflow_dispatch) on this branch: packaging still green with the patched source in the tree.

I could not dispatch `win-package-smoke.yml` against this branch myself (workflow_dispatch on a non-default branch needs the workflow present there — it is — but the run should be triggered/approved by a maintainer).

## Files touched (for the parallel Windows Faz 3a/4 desks)

- `scripts/patch-node-pty.mjs` (restructured: shared `patchOne`, exported markers)
- `src/main/node-pty-patch.test.ts` (extended)
- `src/session-host/windows-conpty.test.ts` (new)
- `CLAUDE.md` (patch paragraph rewritten for two legs + the do-not-wire note), `CONTRIBUTING.md` (one sentence)

No runtime `src/` code changed; no overlap with terminal-profiles, path-semantics or packaging work.

https://claude.ai/code/session_01BdsRHvW3rT85QD5MV4xACJ

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -53,16 +53,33 @@ npm run rebuild    # re-run electron-rebuild for node-pty if you hit ABI/native 
 ```
 
 **`rebuild` and `postinstall` both run `scripts/patch-node-pty.mjs` first, and that is not
-optional.** node-pty 1.1.0's darwin `pty_posix_spawn` leaks a ptmx device on every SUCCESSFUL spawn
-(an off-by-one in the low-fd cleanup) and master+slave on every FAILED one; on this app's spawn
-churn that exhausts `kern.tty.ptmx_max` within hours, and terminals then simply stop opening. The
-script rewrites `node_modules/node-pty/src/unix/pty.cc` before electron-rebuild compiles it.
+optional.** It carries TWO version-pinned native patches for node-pty 1.1.0, one per platform leg:
+- **darwin** — `pty_posix_spawn` leaks a ptmx device on every SUCCESSFUL spawn (an off-by-one in
+  the low-fd cleanup) and master+slave on every FAILED one; on this app's spawn churn that
+  exhausts `kern.tty.ptmx_max` within hours, and terminals then simply stop opening
+  (microsoft/node-pty#950). Rewrites `node_modules/node-pty/src/unix/pty.cc`.
+- **Windows** — the native exit thread deletes its `pty_baton` without closing the HPCON the baton
+  owns, so the session host's taskkill-first kill path (src/session-host/host.ts) leaves a
+  host-parented conhost alive for the life of the long-lived session-host process, one per killed
+  session, and `conpty.kill(id)` reports nothing. The patch serializes baton access, closes the
+  exact HPCON before every baton deletion, and makes `kill(id)` return `true` only as positive
+  proof — the contract `src/session-host/windows-conpty.ts` was ALREADY written against (it
+  shipped with #305 expecting a patched node-pty that did not exist until this patch; do not
+  wire `closeExactWindowsConpty` into the ordinary kill path — after taskkill the exit thread
+  usually wins the race, has already closed the HPCON itself under the patch, and the primitive
+  would then report `false`; it exists for a pre-first-output teardown that bypasses the exit
+  thread). Rewrites `node_modules/node-pty/src/win/conpty.cc` on every host — the file only
+  compiles for the win32 native target, so patching on mac/Linux is harmless and keeps packaged
+  rebuilds honest.
 
-`src/main/node-pty-patch.test.ts` asserts the marker is present in those sources, so a node-pty
-upgrade that silently drops the patch fails loudly. **If that test is red, your `node_modules` is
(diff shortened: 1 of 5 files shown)


eneskirca:
win-package-smoke passed on this branch: https://github.com/eneskirca/nodeterm/actions/runs/33307392096 (workflow_dispatch, windows-latest, job `package-win` green). That closes device-checklist items **1** (postinstall patches + electron-rebuild compiles the patched `conpty.cc` under MSVC) and **5** (packaging smoke). Items 2–4 — runtime conhost-leak observation on silent/noisy kills, real `onExit` arrival, and the Release-build `assert` check — still need a real Windows machine.

PR #537, opened 2026-08-30
Server Edition: opt-in headless canvas control

### LINKED ISSUE #490: Global agent instruction files are modified at boot with no consent, setting, or uninstall
I’ve really enjoyed playing around with nodeterm, but discovering this was pretty jarring.

On startup, `initCanvasControl()` and `initContextLink()` unconditionally modify a number of global agent config files, including:

* `~/.codex/AGENTS.md`
* `~/.gemini/GEMINI.md`
* `~/.config/opencode/AGENTS.md`
* Copilot’s `copilot-instructions.md`
* Claude skill files
* Claude/Gemini settings for hooks

There’s no prompt, setting to disable it, or uninstall path.

I noticed because my `AGENTS.md` is symlinked into my Claude setup, so the injected block suddenly appeared there too. That part is obviously specific to my machine; the underlying thing that surprised me is that nodeterm is silently editing files in my global agent configuration at all.

The skill installation feels reasonable to me. The `AGENTS.md` / `GEMINI.md` writes feel quite different.

Those files are loaded into ordinary sessions outside nodeterm, so every unrelated Codex session on the machine now gets ~120 lines describing canvas capabilities it doesn’t have. The environment gate makes the implementation inert outside nodeterm, but it doesn’t make the instructions disappear.

More fundamentally: I consider those files user-owned configuration. A terminal app silently adding persistent global instructions to them on boot is a boundary I really wouldn’t expect it to cross.

It also seems unnecessary. Nodeterm launches these agents itself, so the extra instructions could presumably live in nodeterm-owned config and be supplied only to sessions nodeterm launches.

I’m not attached to a particular implementation here. I’d just strongly prefer the invariant to be:

**using nodeterm does not silently modify my global agent instructions.**

If global installation does remain, I think it at least needs to be explicit, optional, and cleanly reversible.

Not raising `.nodeterm/project.json` here — that looks like a deliberate design decision with a threat model behind it.

howieyoungdahl:
## What

Implements a real `/control/*` surface for Server Edition behind a default-off flag (`ServerConfig.canvasControl`, env `NODETERM_SERVER_CANVAS_CONTROL=1`, CLI `--canvas-control`). With the flag off, the existing blanket refusal behavior remains unchanged.

v1 verbs: `open-project` for exact saved local projects, `open-terminal`, `open-agent` (claude|codex|gemini), `send`, `reply`, `notify`, `sticky`, `link`, `group`, `rename`, `color`, `close`, and the `--after` arm. Every other verb keeps a named, permanent `control-unsupported-on-this-edition` refusal.

## Incident hardening

- **Boot is inert.** A service start never adopts a persisted node, attaches or creates its backend, or sends a queued command. Persisted missing backends remain dormant; even a surviving backend is untouched.
- **Creator ownership is fail closed.** Every enabled request requires verified node identity. A process-local ledger permits mutation, messaging, notification, and close only for nodes the caller spawned during the current Server run. The ledger is never rebuilt from project JSON, titles, hooks, or tmux names and clears on restart.
- **No partial destructive requests.** Multi-target operations validate the complete target set before any project write or PTY kill. Closing a frame also refuses if it would reparent an unowned child.
- **Queued messages revalidate ownership.** A delivery that loses caller ownership while waiting is dropped before pane access.
- **Plain terminals stay plain.** Missing `agentId` no longer defaults to Claude. Plain shells receive generic node hook wiring but no `NODETERM_AGENT_ID`, permission wait, shared-identity launcher, or `NODETERM_CANVAS_CONTROL` grant.
- **Cold spawn authority is explicit.** New backends are created only by a verified owner open request or by a browser view.

## Design

- The server mints renderer-format ids, creates PTYs through `PtyManager.createHeadless`, assembles launch commands with the shared builder, persists via `WorkspaceStore.save`, and broadcasts exact mutations.
- Armed launches use probe-only `sessionExists` during current-run event reconciliation. Reconciliation never calls the attach-or-create path.
- Agent messaging reuses the existing project switch, rate limits, delivery queue, receipt watcher, and typed outcomes, with the Server creator gate injected into the same flush-time validation path.
- Server upgrades are documented and tested against a disposable data directory and ephemeral port. Restarting a shared live service is an explicit operator action, never a test or repair step.

## Validation

- `npm test`: 687 files passed, 9,372 tests passed (1 file / 27 tests skipped)
- `npm run typecheck`
- `npm run build`
- `npm run server:build`
- Disposable Server boot test with canvas control enabled: missing persisted backend remained absent, queued launch remained durable, hooks were disabled, and the test-owned server closed normally

No live nodeterm service was restarted, upgraded, or otherwise controlled while validating this change.

---

## Review follow-ups (maintainer-side, on top of the author's commits)

Merged `main` (not rebased, so the reviewed commits stay where the review comments point). Seven files conflicted; the merge commit explains each resolution. Three well-typed disagreements the merge surfaced that no conflict marker could show: `supportsSessionIdFlag`'s new REQUIRED grok probe (the server now passes a real `grokCaps` dep rather than a hard-coded `false`), main's tmux-socket-isolation guard flagging this branch's new boot e2e test (allowlisted with its reason, and the test now refuses to run outside the run-wide `TMUX_TMPDIR` sandbox), and this branch's new plain-terminal assertion reading `NODETERM_AGENT_ID` without neutralising the ambient one — nodeterm is developed inside nodeterm, so that test failed for the developer while passing in CI.

### Why this is not split into five PRs

Staging was proposed in review and the author agreed; landing it as one unit is the maintainer's call, taken deliberately. Three pieces did come out and are separately reviewable: **#652** (codex daemon reconnect, merges clean on `main`), and **#653** filed for the desktop-side `legacy`-caller gap that reviewing this made visible. What remains is one feature whose value is the boundary itself: the flag, the identity gate, the ownership ledger and the verbs are not independently meaningful — stage 1 alone is a surface that refuses everything, and the ledger only earns its review attention next to the verbs it governs. Splitting further would have spread one security argument across five review windows without shrinking it. The commit history is preserved, so it can still be read stage by stage.

### Security posture, re-measured on the merged tree

- **The flag is default-off and inert.** `canvasControl` is parsed from `--canvas-control` / `NODETERM_SERVER_CANVAS_CONTROL` and absent means off; `initServerCanvasControl` runs only under `config.canvasControl === true`, and a failed init keeps it disabled rather than half-on.
- **The gate is the strongest verdict.** `createServerEditionControlHandler` refuses every v1 verb unless `verified` — at the boundary, before `parseControlRequest`'s result reaches a factory method and before anything reads shared state. `legacy` does not pass. Ownership is process-local and never rebuilt from disk.
- **`sourceCanControl` refuses an unknown agent** (`!!agentId && canControlCanvas(agentId)`), so an expired or absent mirror entry denies rather than falls through — the answer promised in review, now pinned by the merged tree.
- **No credential rides argv.** The new server paths add no `execFile`/`spawn` carrying a secret; the hook bearer still travels by 0600 endpoint file, and `NODETERM_HOOK_SOCK` is an address, not a credential.
- **The install decision is now unspellable by omission**: `installAgentIntegrations` is a required plain `boolean` on both `initContextLink` and `initServerCanvasControl`, the desktop states `true` and the server passes its `installHooks` gate.
- **The flag announces what it opens** in `docs/SERVER.md`, in the `--canvas-control` help text, and as a boot line naming the server user: it grants arbitrary command execution on the host, not on the canvas.

### This does NOT close #490

The author's reply suggested the required-field change closes [#490](https://github.com/eneskirca/nodeterm/issues/490); measured against the merged tree, it does not, and the PR should not claim it. #490 is about the **desktop** writing into `~/.codex/AGENTS.md`, `~/.gemini/GEMINI.md` and friends at boot with no consent, setting or uninstall path — and `src/main/index.ts` still calls `initContextLink(..., { installAgentIntegrations: true })` and a bare `initCanvasControl()` on every launch. What this PR genuinely fixes is narrower and worth stating as itself: the write can no longer happen because a caller said nothing, which is what let a unit test rewrite a developer's own agent configuration. The consent/setting/uninstall path #490 asks for is still owed, on the desktop, separately.


howieyoungdahl:
Server Codex identity follow-up landed on the head branch in c105fb99. Root cause: Server Edition already wired the shared app-server and signed thread handlers, but headless open-agent hard-coded sharedIdentity=false. The factory now consumes the boot-populated Codex capability, preserves bare-codex fallback, and has both factory-level and Server composition regressions (including a mutation check). Verified locally: focused Server identity/canvas tests, full 9,290-test suite, typecheck, desktop build, Server build, and a live stripped-env headless Codex probe. The running service was restarted without restarting the shared Codex daemon; all 18 tmux sessions survived.

howieyoungdahl:
Server restart recovery is now included at e9cb50ba. This closes the post-restart dead end where verified agent sessions and saved projects survived, but the process-local project grant did not and Server permanently refused open-project. The enabled Server handler now re-grants only an exact already-saved local project by cwd; it cannot create/add/rename/recolor/focus projects, refuses unknown/unverified/SSH targets, caps grants, and binds the grant to the exact caller. Fail-before reproduced the permanent refusal. Verification: focused 91/91, full Vitest 9,287 passed / 27 skipped (677 files passed, 1 skipped), typecheck green, desktop build green, Server build green. Live deployment preserved all 35 tmux sessions and the shared Codex daemon; open-project restored the saved loop project and launched an explicit Opus director node there.

Abhirup0 (commented):
Reviewed the full diff.

Key areas checked:
- **Ownership & security**: Verified that creator ownership in `HeadlessNodeFactory` is process-local and fails closed across restarts. Unowned mutation checks in `close`, `group`, and `link` correctly reject mixed target sets before applying any destructive pane kill or graph mutation.
- **Concurrency**: `runExclusive` properly serializes `WorkspaceStore` load/save cycles against concurrent `/control/*` requests.
- **Lifecycle arms**: The `awaitWorking` tracking in `pendingLaunch` cleanly handles the startup `done` blip for fresh agent spawns before releasing dependent launches.
- **Delivery settlement**: `sendSettledEnvelope` avoids racing the TUI by validating visible pane footer settlement before emitting `Enter`.

One minor non-blocking note:
- In `openProject` (`src/server/headless-node-factory.ts:680`), `path.resolve(project.cwd) === cwd` assumes exact string matching. On case-insensitive dev filesystems (e.g. macOS host testing), matching via normalized realpath can avoid false `open-project-server-existing-only` errors if case differs between launch arguments and stored project paths.

Overall, this is solid and well-tested. LGTM.

eneskirca:
I reviewed this one as a security change rather than a feature, because `CLAUDE.md` records canvas control's absence on the Server Edition as a deliberate scope decision, and re-opening it turns a read/observe surface into an execution surface on a host that is frequently shared. The short version: **the identity and ownership model is sound, and in one respect stricter than the desktop's.** My objections are about default posture, one documentation omission, and the size of the change relative to what a reviewer can actually certify.

## What I verified, and what holds

- **The gate is the strongest verdict, not the permissive one.** `hook-server.ts:578` passes `verified: verdict === 'verified'`, and `createServerEditionControlHandler` refuses *every* v1 verb when that is false — before `parseControlRequest`'s result reaches any factory method, and before anything reads shared state. `legacy` ("we cannot judge this") does not pass. Given that `NODE_IDENTITY_STRICT_AFTER` is 2026-10-13 and today is not, that is the right side of the line to be on and it does not depend on the cutoff date at all.
- **This edition is stricter than desktop.** `hook-server.ts:161` restricts only the messaging verbs plus `sticky` to a verified caller; here `open-terminal`, `open-agent`, `close`, `link`, `group`, `rename`, `color` and `open-project` are all verified-only too. That is the correct asymmetry for a headless host — but it does raise the obvious question in the other direction: **on desktop, `open-terminal --cmd` is arbitrary command execution and is reachable by a `legacy` caller.** That is pre-existing and out of scope here, but this PR is the moment the contrast becomes visible, and I think it deserves an issue of its own.
- **Ownership is literal and unforgeable from disk.** `ownsSpawn` is `ownerOf(nodeId)?.sourceNodeId === sourceNodeId` over a process-local ledger, `unownedMutation` validates the complete target set before any write or PTY kill, and the docs state plainly that the ledger is never rebuilt from `project.json`, titles, hook records or tmux names. That last sentence is the one that matters: every cheap way to reconstruct ownership after a restart is a way to forge it, and refusing all of them — at the cost of an agent losing control of its own nodes across a restart — is the right trade, made explicitly.
- **`open-project` cannot create or reach anything new.** Absolute path required, `path.resolve` compared against `project.cwd` of an already-saved **non-SSH** project, no creation, no rename, no focus, and a per-caller `GRANT_CAP`. That is about as narrow as that verb can be made while still being the restart-recovery path it is meant to be.
- **Boot is inert**, and the failure path degrades to the named permanent refusal (`hookServer.setControlHandler(serverEditionControlHandler)` stands until the runtime is fully initialised, and a throw leaves it in place). Both are the correct directions.
- The `installAgentIntegrations: config.installHooks !== false` wiring at the call site in `index.ts` correctly reuses the existing gate, and the shim itself goes under `platform().userDataDir`, not a hard-coded home.

## What I would want changed

**1. `initServerCanvasControl`'s own default writes into a human's home directory.** The parameter is `deps.installAgentIntegrations !== false`, so an *omitted* value installs into `~/.claude/skills/...`, `~/.codex/AGENTS.md`, `~/.gemini/GEMINI.md` and every managed account dir. The production call site passes the flag correctly today; the hazard is the next call site, and a test that forgets it. This repo already made exactly this call the other way and wrote down why: `session-memory-service.ts`'s `remote` option pair is deliberately asymmetric, with `isRemoteProject` **required** because "reading-without-knowing is a compile error". Same shape here — make the field required, or default it to `false`. Writing into another user's `$HOME` from a service process is also its own documented mine on this project (root-owned files in `/home/*` become `EACCES` traps for the owning user).

**2. The docs never say the sentence that matters most.** `docs/SERVER.md`'s new section explains identity, ownership and boot inertness well — but nowhere does it say that **`open-terminal --cmd` runs an arbitrary command as the server user**, which is what `args.cmd` becomes (`headless-node-factory.ts`, the `let command = args.cmd` branch). An operator reading "opt-in canvas control" and "creator ownership" could reasonably conclude the blast radius is the canvas. It is the host. That belongs in the *first* paragraph of the flag's documentation, in the `--canvas-control` help text, and ideally in a startup log line when the flag is on, in the same register the repo already uses for `insecureHttp`. Everything else about the design is honest about its own limits; this one omission is the one that could get somebody's server owned by an agent behaving badly.

**3. This is too large to certify as one unit, and the split has already been proven.** 4,764 added lines across 72 files, of which `headless-node-factory.ts` alone is 1,411 — for a change whose entire value proposition is "you can trust the boundary". You have already demonstrated the extraction works: **#540** is the shared-Codex identity repair pulled out of this branch onto current main, and it is a clean, reviewable PR. I would like the same treatment here:

   1. `ServerConfig.canvasControl` + the flag plumbing + the enabled-mode named refusal for every deferred verb (a no-op by itself: the flag turns on a surface that refuses everything).
   2. The **non-executing** verbs — `rename`, `color`, `link`, `group`, `sticky` — plus the ownership ledger and the all-or-nothing target validation. This is where the ledger earns its review attention without an execution surface attached.
   3. `open-project` (existing-only) and `close`.
   4. `open-terminal` / `open-agent` / `--after`, i.e. the actual execution surface, reviewed on its own with the ledger already merged and understood.
   5. Messaging (`send`/`reply`/`notify`), which reuses the existing queue and receipt machinery.

   Each stage is independently useful and independently refusable. As one PR, a reviewer either takes all of it on trust or blocks all of it, and neither is a good outcome for work this careful.

   Two things that should come out regardless of whether you stage the rest: `src/server/workspace-external-watch.ts` (hand-edit reload for server-side projects) and the `noteLink.ts` refactor (`15+/88-`) are not canvas control and would both land faster alone.

**4. Overlap with #540.** `src/core/codex-shim-identity.integration.test.ts`, the `context-link-core.ts` shim-builder change and the `docs/shared-codex-node-identity.md` edits appear in both. Whichever lands first, the other needs a rebase — worth saying on both PRs which you intend to be canonical.

## Questions

- **After enabling the flag, can an already-running session use canvas control, or must it be relaunched?** The gate needs `verdict === 'verified'`, which needs the per-node token; `armServerNodeIdentity` says already-running sessions become verified at their next hook event with no restart, but that is about the *secret*, not about a token file the session's shim can read. If a relaunch is required, that is fine and should be in the docs; if it is not, a test pinning it would be worth having, because "I turned it on and every verb says identity-refused" is the first support question this feature will generate.
- **`sourceCanControl(source.node, this.deps.agentIdOf)` reads the agent id from the status mirror.** What happens for a node whose mirror entry has expired or was never populated — refuse, or fall through? The PR's own "missing `agentId` no longer defaults to Claude" note suggests refuse, which is right; worth confirming it is refuse rather than "unknown agent, allow".
- **Is the flag surfaced in the install script / systemd unit at all?** If not, that is the right call for v1 — an operator who has to add it by hand has read something.

## Overall

I do not think this should be blocked on principle: the model is well chosen, the refusals are named and permanent rather than generic (which the repo learned the hard way is what stops an agent retrying), and the boot-inertness and never-rebuild-the-ledger rules are the two hardest things to get right and are both right here. The validation section's "no live nodeterm service was restarted or controlled while validating this change" is also exactly the right instinct for a change of this kind.

My asks before merge are (1) the install default, (2) the arbitrary-execution sentence in the docs, and (3) staging. @eneskirca — (3) in particular is your call; the security review of stage 4 is the one that genuinely needs your eyes rather than mine.

howieyoungdahl:
Thank you for reviewing this as a security change — that is the right frame, and the three asks are all fair. Taking them in your order, with what is already done and what is not.

## 1. The install default — agreed without reservation

`deps.installAgentIntegrations !== false` means an *omitted* value installs into `~/.claude/skills`, `~/.codex/AGENTS.md`, `~/.gemini/GEMINI.md` and every managed account dir. Your cited precedent is exact: `session-memory-service.ts` made this same call the other way and wrote down why — "reading-without-knowing is a compile error". I will make the field **required** rather than defaulting it to `false`, for that reason: a default of `false` is safe but still lets a call site forget to think about it, whereas a required field forces the decision at every site.

Worth connecting: this is also open issue **#490** ("Global agent instruction files are modified at boot with no consent"). So this is not only a review nit on my PR — the fix closes a filed issue, and I would rather it be framed that way in the changelog.

## 2. The arbitrary-execution sentence — agreed, and I have split out the desktop half

You are right that an operator reading "opt-in canvas control" and "creator ownership" could reasonably conclude the blast radius is the canvas when it is the host. That sentence goes in the first paragraph of the flag's documentation, in the `--canvas-control` help text, and as a startup log line in the same register as `insecureHttp`.

On the contrast you drew in the other direction — desktop `open-terminal --cmd` being reachable by a `legacy` caller — I opened **#653** for it, since you suggested it deserves its own issue. I verified it before filing rather than restating your review: `requiresVerified` (`hook-server.ts:188-193`) is `send`/`reply`/`notify`/`sticky`/`open-project`, and `DESTRUCTIVE_VERBS` (`control-verbs.ts:41`) is `write`/`close`/`open-project` — `open-terminal` is in neither. The repo already documents the gap at `control-verbs.ts:50-51`.

## 3. Staging — agreed, and I have started rather than argued

Your five stages hold. What I can add is that **three pieces were already separable**, and one is out:

- **#652 is open now**: `fcd5ac54` "resume sessions after a shared-daemon reset". It is this branch's *first* commit, has nothing to do with canvas control, is 10 files / +493, and merges clean on current `main`. That is 493 lines off this PR at zero risk.
- The two you named should come out regardless, and I agree: `src/server/workspace-external-watch.ts` (+ its test) and the `noteLink.ts` refactor — the latter is a **net simplification** (+15 / −88), so it is pure subtraction from the reviewable surface.
- **The #540 overlap is real.** I suggest **#540 is canonical** — it is smaller, already reviewed, and currently the only one of mine that merges cleanly — and this branch rebases onto it. Said on both PRs.

After those three subtractions this PR is meaningfully smaller before stage 1 begins. I will then follow your ordering: flag + plumbing with everything refusing; the non-executing verbs plus the ledger; `open-project` and `close`; the execution surface on its own; messaging last.

## Your three questions

**Can an already-running session use canvas control after enabling the flag, or must it be relaunched?**
Traced from code, and the answer is "mostly yes — and the gap is discovery, not identity":

1. The env gate is already satisfied. `buildPtyEnv` sets `NODETERM_CANVAS_CONTROL` on `canControlCanvas(agentId)` alone (`hook-server.ts:1091`) — it is **not** gated on the server flag — so a session started before the flag already carries the variable.
2. The token is already on disk. `armServerNodeIdentity` (`src/server/node-identity-arm.ts:19-31`) calls `initNodeTokens({ canvases })`, which materialises a token file for **every persisted node** at boot, not only newly-spawned ones. So the shim has something to present, and the session becomes `verified` at its next hook event with no restart — which answers your worry that the comment is about the secret rather than about a token file the shim can read.
3. The gap is **discovery**: `initServerCanvasControl` — which writes the shim and the per-agent skill/instruction files — runs only under `if (config.canvasControl === true)` (`src/server/index.ts:595`). Before the flag those files did not exist, so an agent already mid-session has instructions in context that never mentioned the shim. It picks them up on its next turn.

**I have not yet confirmed that on a running instance** — it is a code trace, and you asked precisely the question where a confident wrong answer in the docs is worse than none. I will verify it on a disposable instance before it goes in the docs, and pin leg 2 with a test (a node persisted before the flag gets a token file at boot).

**`sourceCanControl` with an expired or absent mirror entry — refuse or fall through?** Refuse, which matches your reading. I will add the test that pins it rather than leaving it to inspection.

**Is the flag surfaced in the install script / systemd unit?** No — and I think that is right for v1, for the reason you gave: an operator who has to add it by hand has read something.

## On the size point generally

You are right that as one PR a reviewer either takes all of it on trust or blocks all of it, and neither is a good outcome. I would rather ship it in pieces you can actually certify — starting with #652, which needs none of this context to review.

eneskirca:
Measured the reply against the branch rather than the prose: the head is still `a919b0d5` (2026-08-31), so none of the three asks has landed on this PR yet — which is consistent with what you wrote, since every item is phrased as "I will". Recording the state so the next reviewer does not re-derive it:

| Ask | Status on `feat/server-canvas-control` |
|---|---|
| 1. `installAgentIntegrations` required field (closes #490) | not pushed |
| 2. Arbitrary-execution sentence in SERVER.md / `--canvas-control` help / startup log | not pushed |
| 3. Staging | started outside the PR: #652 (`fcd5ac54`, +493/−56, 10 files, merges clean on `main`; note its head is *not* an ancestor of this branch, so this branch needs a rebase onto it once merged) and #653 filed for the desktop `legacy`-caller gap. `workspace-external-watch.ts` and the `noteLink.ts` simplification not yet extracted. |
| Q1 running-session pickup | code trace given, marked unverified — the disposable-instance check and the "persisted node gets a token at boot" test are still owed |
| Q2 `sourceCanControl` refuse | agreed, test owed |
| Q3 flag not in install script | agreed, no change needed |

Agreed on the #540-canonical ordering. Verdict unchanged until the pushes land: still FIX-ISTER — (1) and (2) are small and would be worth pushing here now, independent of how the staging shakes out. @eneskirca for the merge-order decision across #540 → #652 → this.

howieyoungdahl:
Asks (1) and (2) are on the branch now — head is `d4b92adf`. Updated status against your three:

| Ask | State |
|---|---|
| 1. `installAgentIntegrations` default installs into a human's `$HOME` | **Done** — `88a885c6` |
| 2. The arbitrary-execution sentence | **Done** — `d4b92adf` |
| 3. Staging | **In progress** — #652 is out (493 lines); the two you named next |

**(1) — and the hazard was not hypothetical.** I made the field a required `boolean` rather than defaulting it to `false`, for the reason your `session-memory-service.ts` precedent gives: a default still lets a call site forget to think, a required field does not. Omission is now a `TS2554`.

While doing it I found the failure had already happened: **`src/core/context-link.handler.test.ts` never redirects `HOME`**, so on the current head every run of that suite installed the canvas skill into the developer's real `~/.claude` and merged the marker block into their `~/.codex/AGENTS.md`. That is the exact "a test that forgets it" case you predicted, and it was live in the repo rather than waiting for a future call site. It is closed now because the suite cannot compile without deciding.

I extended the same treatment to the two sibling deps — `ServerContextLinkDeps` and core's `initContextLink` options — since they carry the identical shape. **If you would rather keep this PR strictly server-side, drop the `src/main/index.ts` hunk and I will re-push**; the server call sites were already passing the flag correctly.

**On issue #490: this is partial, and the smaller part.** It removes one class of unconsented write and makes the rest unreachable by accident, but it adds **no prompt, no setting, and no uninstall**, and desktop boot still installs unconditionally. I would not mark #490 closed on this. I have said the same on that issue.

**(2)** landed in three places: `docs/SERVER.md:534` (first paragraph of the flag's section, not a footnote), `src/server/config.ts:41`, and a boot line at `src/server/index.ts:625`, pinned by `canvas-control-boot-e2e.test.ts`.

Two notes on that. There is **no `--help` handler anywhere in the server**, so the `config.ts` JSDoc is the flag's only usage string — if you want a real `--help`, that is its own change. And I could not match `insecureHttp`'s register as you suggested, because **`insecureHttp` emits no boot warning at all** — it only swaps a word in the listening line. I used the proxy-trust notice instead (`index.ts:167`, the `⚠️ … ENABLED: …` form), which is the repo's actual idiom for a flag that widens what is permitted. Say the word if you prefer the muted form.

Separately: `--canvas-control` is missing from the SERVER.md flags table entirely. I left that alone rather than grow the diff.

One implementation note you may want to check: I used `serverUserLabel()` rather than a bare `os.userInfo()` in the boot line, because the latter throws for a container uid with no passwd entry — and inside that try block it would have disabled the feature the operator had just enabled.

**Your Q1 and Q2 tests follow in a second push**, not this one: the `sourceCanControl` refuse-on-missing-mirror pin, and the "a node persisted before the flag gets a token file at boot" pin for the already-running-session answer. I want to confirm the latter against a disposable instance before I pin behaviour I have only traced.

Gates: typecheck clean; targeted suites green; `vitest run src/server src/core` 1 failed / 3501 passed against a base of 1 failed / 3500 — the same single `pty-single-user.test.ts` failure, reproduced on `a919b0d5` and untouched by either commit. `server:build` OK.

eneskirca:
Picked this up maintainer-side rather than sending it back again — your (a) `88a885c6` and (b) `d4b92adf` were already in, so what was left was the merge and the two things review had asked for that are not code. Pushed as `ba8dec1b` on your branch; your commits are untouched and still where the review comments point.

**Merged `main`, not rebased.** Seven files conflicted. The three doc ones are the same sentence extended on main (#540's shim-neutrality guard, and the record half being desktop-only until the Server Edition registers the codex-thread handlers), so main's region is taken whole. The four code ones are all the same shape — this branch MOVED a module main kept editing — and both sides are kept: `canvas-control-core.ts` keeps main's dry-run block, rename no-op paragraph and `--after` link correction beside your Server-Edition ownership text and the new `color` verb; `noteLink.ts` keeps the extraction to `@shared/canvas-link`, and main's `contextCapable` comment fix (ec3dcfb8, grok) is ported into the moved file so the move does not silently revert it.

**Three things the merge broke that no conflict marker could show**, each a well-typed disagreement:

1. `supportsSessionIdFlag` gained a REQUIRED third argument on main — grok's own probe, non-defaulted precisely so a caller cannot forget it and silently get "grok never mints". The server factory called it with two. A hard-coded `false` would have been that forgotten probe waiting for the day grok joins `SERVER_AGENTS`, so the factory takes a `grokCaps` dep beside `cliCaps`, defaulted to the real `grokCliCaps`, stated in the fixture.
2. Main's tmux-socket-isolation guard (#629) flagged your new `canvas-control-boot-e2e` test. It legitimately needs the live socket NAME — the server it boots binds `TMUX_SOCKET` itself and the assertion is that boot spawns nothing — so it is allowlisted with that reason, and the test now REFUSES to run unless the run-wide `TMUX_TMPDIR` sandbox is in effect, which is what the allowlist's second assertion asks every entry to prove.
3. Your new "never exposes gateway credentials to a plain terminal" case asserts `NODETERM_AGENT_ID` is absent from the spawned env without neutralising the ambient one. nodeterm is developed inside nodeterm: run from an agent pane the suite inherits `NODETERM_AGENT_ID=claude`, so it failed for the developer and passed in CI. It now saves/clears/restores both vars like the `ANTHROPIC_AUTH_TOKEN` beside it. (Your underlying change is right and is what the assertion was written for.)

`npm run typecheck` clean; full `npx vitest run` **10386 passed / 40 skipped / 0 failed**; `MERGEABLE`.

**On the split (c):** not doing it, and the PR body now says why rather than leaving it implied. #652 and #653 coming out was the valuable part; what remains is one security argument whose stages are not independently meaningful, and spreading it over five review windows would not have made it smaller.

**On #490:** I checked before letting the body claim it, and **this does not close it.** `src/main/index.ts` still calls `initContextLink(..., { installAgentIntegrations: true })` and a bare `initCanvasControl()` at every launch, so the desktop still writes `~/.codex/AGENTS.md` and friends at boot with no consent, setting or uninstall path — which is exactly what #490 is about. What you did fix is narrower and worth claiming as itself: the write can no longer happen by omission, which is what let a unit test rewrite a developer's own agent configuration. I have written it that way in the body; #490 stays open for a desktop-side consent path.

Nothing here changes your design — it is merge repair plus the two non-code asks. Merge decision is @eneskirca's.

Issue #530, opened 2026-08-30
open-worktree: --base cannot express "branch off what that station is about to produce"


Temikus:
## What happens

`open-worktree` resolves its base eagerly, at creation:

```ts
const baseRef = args.base?.trim() || effectiveWorktreeBaseRef(projectDefaults, entries)
```

and creates the worktree in the same tick. So if we're staging things into waves - wave-two station that needs wave-one's refactor cannot say so: `--base rewrite/test-harness` captures that branch as it exists *now*, which is empty, because the upstream station has not committed anything yet.

## What that forces

The only way to express the dependency today is prose. In a recent seven-station run every downstream station's prompt had to begin with `git merge rewrite/test-harness`, plus a hard-stop guard telling the station to abort if the expected files were missing. That is orchestration logic living inside a prompt, where nothing can enforce it.

## Proposal

Resolve `--base` lazily, at the moment an armed node actually starts, rather than at creation. A worktree armed with `--after A --base <A's branch>` would then branch off A's real, finished work, and the dependency would be expressed once, structurally, instead of restated in every downstream prompt.

## Questions

This gives worktree creation the same deferred-execution treatment `pendingLaunch` gives a command, which raises questions that I wanted to run past you first:

- **What does the node look like while it waits?** A group frame with no worktree behind it yet is a state that does not exist today. I propose a simple dotted-outline node. I can provide a couple of mocks for your perusal, but I'm not sure if you already have an idea in mind.

- **What happens if the base still does not resolve when the node fires** - e.g. the upstream errored, or never created the branch? Refuse and leave the node armed, or create off the default branch and say so loudly? I lean towards refuse, but wanted to check.

- **Persistence.** `pendingLaunch` is persisted because agent state is not, and a deferred worktree would need the same treatment plus a manual escape hatch after a restart. Is that acceptable?

- **Interaction with `--on complete`**, if that lands: branching off "idle" is arguably worse than branching off nothing, since a half-finished branch looks legitimate. WDYT?

eneskirca:
#538 (third commit) ships the **expressivity half** of this and deliberately leaves the timing half open — splitting it that way because the two have very different blast radii.

[...]

Issue #521, opened 2026-08-30
Agent status: a station that fails on its first turn is indistinguishable from one that finished


Temikus:
## What happens

A station whose first turn ends in an API error leaves the error in its pane and nothing else. The `open-agent` call has already returned success, the node's status reads as idle, and an orchestrator has no signal that the station never started work. The only way to find out is for a human to look at the pane.

Observed during a seven-station run: two stations were launched with a malformed prompt, both errored immediately, and both looked healthy from every surface the orchestrator can read.

The main problem is - `--after` fires on idle. An errored station reaches idle *immediately*, so every downstream station armed behind it launches against an upstream that produced nothing, meaning it starts an entire dependency chain on bad ground.

## What would fix it

1. A distinct `errored` state, separate from `done`.
2. `--after` not firing on an errored upstream by default. If it must fire, the arming reply should say so.
3. The error text needs to be readable through the control API, so an orchestrator can diagnose without the user looking at the pane.

## Why this is an issue and not a PR

`AgentState` is a closed union of four values (`working | waiting | blocked | done`, `shared/agents/normalize.ts`). A fifth ripples through both raw hook listeners, which the repo requires to stay in parity.

There is also a wider question: **no agent hook reports an API error at all today.** So the state needs a source before it needs a name. 

@eneskirca I noodled on this a bit and here's what I think might help, roughly in preference order:

- The context tail already reads the transcript; an error entry is visible there.
- A `Stop` hook whose turn produced no assistant message is suggestive but not conclusive.
- Pane scraping, which I assume you would rather not.

The first one seems closest to how the rest of the system already learns things, but I want to know what you think before I propose anything further.

eneskirca:
Design response — this issue, #529 and #531 are being designed as one **station lifecycle** model (`spawned → started → turns → declared-complete | failed-to-start/errored | closed(+durable ref)`); see the comment on #529 for the frame. This comment covers the *failed-to-start / errored* leg.

Your analysis splits cleanly into **two distinct failures**, and they behave oppositely today:

1. **Never started** — CLI didn't launch, prompt never delivered. The station emits no hook event, its state stays *unknown*, and unknown is deliberately not-satisfied — so dependents hang forever. Honest, but silent.
2. **Started and errored on turn 1** — hooks fired, `Stop` landed, state is `done`. Dependents fire on bad ground. This is the arming bug, and your seven-station run is the proof.

## Errored (the arming bug)

[...]

Issue #30, opened 2026-07-21
Option: drag the canvas without having to hold down the spacebar.


nebula42x:
For someone with ADHD, having to hold down the spacebar just to drag the canvas is far too tedious.

Since dragging the canvas is such a frequent action in this software, I suggest adding an option to drag without needing to hold down a key. In other words, when the cursor is not focused on special elements (like the terminal or a sticky note), it should change to the drag icon.  This would follow the same interaction logic as Figma.

eneskirca:
Thanks for the suggestion!

First, a note on what's already there: you can pan without holding any key today — **two-finger scroll on a trackpad, or middle-mouse drag** — which is the same interaction set Figma offers (in Figma, left-drag on empty canvas is rubber-band selection too; pan there is Space-drag / middle mouse / the Hand tool).

That said, the pain is real for mouse users who pan constantly, so this is now shipped as an opt-in setting: **Settings → Behavior → Canvas left-drag → "Pan the canvas"**. With it on, left-dragging empty canvas pans the map directly (grab cursor), and box-select moves to **Shift+drag**. The default stays selection so existing muscle memory doesn't break.

Lands in the next release.

PR #538, opened 2026-08-30
canvas-control: multi-line briefs (--prompt-file), --dry-run, and --base <stationId> (#520, #532, #530)

### LINKED ISSUE #520: canvas-control: --prompt cannot carry a multi-line brief (it rides argv on a line typed into the pane)
Documentation for this shipped in #516. This issue is about actually fixing it.

## What happens

`assembleLaunchCommand` collapses every whitespace run in the prompt:

```ts
shellSingleQuote(inputs.initialPrompt.replace(/\s+/g, ' ').trim())
```

The prompt is a positional argument on the agent CLI's launch command line, and that line is *typed into the pane* by `writeWhenShellReady` -> `deliverCommand`, which echo-verifies it before submitting. A newline would submit the half-typed line and cause verification to fail.

## Why it matters

A station prompt is the entire contract with that station. 
Anything structured, e.g.:
- numbered acceptance criteria
- a file list
- "stop if X is missing" guard
, all arrive as one paragraph. In practice, that means orchestration instructions get written as run-on prose, which will degrade quickly through fan-out.

## Suggested direction

Stop putting the prompt in argv. Launch the CLI bare, then deliver the body through the existing `sendText` path, which already frames multi-line text correctly via tmux `paste-buffer -p` (with the ARG_MAX and copy-mode hazards already worked out and documented in CLAUDE.md).

Things that need deciding, hence an issue rather than a PR:

- This changes how **every** agent node starts, including cold-restore resume and hibernation wake, not just canvas-control opens.
- Ordering: the prompt can only be sent once the CLI is up and accepting input, which is a different readiness signal from "a shell owns the pane".
- The one-shot `data.initialCommand` semantics — the prompt currently *is* part of the launch line, and splitting them means two things to make exactly-once rather than one.

Happy to take a swing at it if the shape above is roughly what you would want. @eneskirca let me know what you think.

### LINKED ISSUE #532: canvas-control: no dry-run, so a spawn-team payload can only be validated by opening nodes
There is no way to check that a call is well-formed without performing it. In practice that means:

- A `spawn-team --team '[...]'` payload is validated by running it. So if one makes a typo in the JSON, a role is missing a `prompt`, or a bad `agent` ID is discovered, then all of them have to be cleaned up by hand.

- A `--group` id can only be confirmed real by using it.

  - Same for `--after` ids, which is the flag where being wrong is quite expensive, due to the arming logic.

The cost is asymmetric: verbs are easy to call, but their mistakes are hard to undo, which pushes an orchestrator towards opening one node at a time and inspecting it, which (IMO) is the opposite of what the fan-out verbs are for.

## Proposal

A global `--dry-run` that parses, validates and reports what *would* happen, changing nothing.

## Questions

Before I propose something, I wanted to check with you since from what I'm reading - validation is split across two processes, so how far a dry-run can see depends on where it stops:

- **Main-side only** - the presence rules in `parseControlRequest`, JSON shape, the project/identity gates. Cheap and entirely synchronous, but it cannot tell you whether a `--group` or `--after` id exists, which is most of what you actually want to check.

- **Round-tripping to the renderer** - this would mean we can resolve ids against live canvas state and give a real answer. There's a tradeoff though - it means every verb's handler grows a "do everything except the mutation" path. That can rot quietly: the dry-run path stays correct only as long as someone remembers to update it alongside the real one, so we need some sort of validator/tests around it that would have to be maintained.

A middle option is to make it read-only-verbs-only (resolve ids, report, refuse anything else), which is less useful but cheaper.

I do not have a strong preference, so keen to hear what you think and propose a fix.

eneskirca:
Three small, sharp gaps from @Temikus's orchestration feedback series, all on the same canvas-control verb surface, one commit each. Closes #520, closes #532; ships the expressivity half of #530 (that issue stays open for the deferred-creation design — see my comment there).

## 1. `--prompt-file` — a multi-line brief for a station (#520)

`--prompt` rides argv on a launch line that is typed into the pane and echo-verified, so `assembleLaunchCommand` collapses every whitespace run — a structured brief arrived as one paragraph, and orchestration contracts degraded into run-on prose.

Rather than moving the prompt off the launch line entirely (the issue's sketch — it changes how *every* agent node starts, cold-restore and hibernation wake included, and needs a new "CLI is accepting input" readiness signal), this keeps the delivery path byte-identical and moves only the *content*:

- `open-claude` / `open-agent` take `--prompt-file <abs path>` (and a `spawn-team` role may carry `promptFile`).
- The assembler composes the prompt as `"$(cat '<path>')"`: the **typed line stays one line** — echo verification untouched — and the pane's shell expands the file at execution into a single argv element, newlines and structure intact. Only the path is interpolated (single-quoted); the file content is a substitution *result*, never re-parsed as shell syntax. It also sidesteps our own argv limits — the brief never rides tmux/ssh argv at all (the CLI still receives it as one exec arg, so ~100 KB is the practical ceiling on Linux).
- On an SSH project the path and the `cat` are both on the host, which is exactly where the node runs.
- Validation at the verb: absolute path only, no control characters, `--prompt` XOR `--prompt-file`, and an existence check against the filesystem the node will read from — a missing file would otherwise silently start the agent with an *empty* prompt. Check errors fail open (the pane shows `cat`'s own error).
- Caveat stated in the docs: the file is read at *launch* (later than the call for an `--after`-armed node — keep it in place), and the expansion needs a POSIX-style shell in the new node.

This overlaps #516 (docs-only, still open): both agent-facing bodies here document the one-line `--prompt` fact and the leading-`/` trap in the same breath as the fix, so if this lands first #516 becomes mostly redundant — your call on ordering.

## 2. `--dry-run` — validate a spawn call without performing it (#532)

The fan-out verbs were easy to call and expensive to undo: a `spawn-team` payload could only be validated by opening nodes; a wrong `--after` id was discovered after arming.

- `DRY_RUN_VERBS` (`shared/control-verbs.ts`): `open-terminal`, `open-claude`, `open-agent`, `spawn-team`, `open-worktree` honour `--dry-run`. This is the issue's "round-trip to the renderer" option, but **without a second validation path to rot**: each dispatch case runs its *real* validators (`--group`/`--after` resolved against the live canvas, team JSON parsed, prompt-file existence, worktree path computed) and stops just before the mutation.
- Every other verb **refuses** the flag — a `close --dry-run` that closes would be worse than no flag. The refusal is main's `setControlHandler`'s *first* gate (covers `browser` and `open-project` too) and its sentence derives from the set. `--dry-run` + `--project` is refused (that open lands in another project's store through a different path; a validate-only copy of it is exactly the rot risk).
- A valueless `--dry-run` arrives from the sh shim as an empty string and reads as ON; an unrecognized value fails toward DRY (the safe direction).
- The dry run is deliberately *stricter* in one place: an unknown `--agent` id is caught at dry-run time (the real open still builds the node whose pane then shows the typo — unchanged).
- `spawn-team`'s parser moved to the pure `lib/teamSpec.ts` with **named refusals** — this changes real-path behavior in the refusal direction: a role missing its prompt used to be silently filtered (the team just opened short), a 9th role silently sliced, an unknown agent id opened a broken node. All three now refuse with the role named.

Tested in both parsers per house rules: the shim's valueless-flag translation under real `sh` (`control-shim-parse.test.ts`) and `parseControlBody` (`canvas-control-shim.test.ts`), plus `teamSpec.test.ts` and `control-verbs.test.ts`.

## 3. `open-worktree --base <ref|stationId>` (#530)

`--base` only took a git ref, so "wave two branches off what the wave-one station is producing" lived in prose (`git merge …` at the top of every downstream prompt).

- `resolveWorktreeBase` (`shared/worktree.ts`, pure + tested): a `--base` value naming an existing node or group resolves through its worktree-bound group frame (nested frames climb to the nearest bound ancestor) to that station's **branch** — the dependency is stated once, by identity. Every reply, dry run included, names the branch the id resolved to.
- Explicit refusals, never a guess: an id naming a node *outside* any worktree frame is refused rather than reinterpreted as a ref; self-base (the resolved branch == the branch being created) is refused; a value that is neither a node id nor a valid git ref is refused (`isValidGitRef` also closes leading-dash flag smuggling); a forged `parentId` cycle in the git-shared project.json terminates as a refusal.
- **Deliberately not built here: lazy resolution.** The issue's deferred-creation half (a pending frame with no worktree behind it, refuse-at-fire, persistence + a post-restart escape hatch) is a real design with its own states and is left open for you to rule on — see my reply on #530. Until then the docs state the capture timing outright and give the working pattern: arm the downstream agent with `--after` and create its worktree when the upstream has committed, or have it merge first.

## Cross-cutting

- All agent-facing text updated in the same PR (`buildCanvasSkillBody` + `buildCanvasControlInstructions`), pinned in `canvas-control-core.test.ts` per the #269 stale-doc rule; the dry-run verb list *renders from* `DRY_RUN_VERBS` so it cannot drift.
- New verbs/flags parse identically under the old and new shim loops (every flag has a value, or is valueless-safe) — no dependence on a shim rewrite reaching SSH hosts.
- Surfaces: desktop only in practice (canvas-control is not wired on the Server Edition; mobile N/A). SSH projects covered as noted per feature.
- Full suite: 9348 passed; the two failures are node_modules-freshness pins (`@xterm`/`@xyflow` dists not installed in my worktree), unrelated.

### DIFF
--- a/src/main/canvas-control-core.test.ts
+++ b/src/main/canvas-control-core.test.ts
@@ -316,6 +316,18 @@ describe('parseControlRequest', () => {
     }
   })
 
+  it('both agent-facing texts document --base accepting a station id (issue #530)', () => {
+    for (const body of [buildCanvasSkillBody('/x/shim.sh'), buildCanvasControlInstructions('/tmp/nodeterm.sh')]) {
+      // The flag surface must show the widened grammar…
+      expect(body).toContain('--base <ref|stationId>')
+      // …and both hard truths beside it: what a station id resolves to, and that the base is
+      // captured at CREATION (the deferred-resolution half of #530 is not built — an agent that
+      // reads this text and assumes lazy capture bases a wave on an empty branch).
+      expect(body.toLowerCase()).toContain('station')
+      expect(body).toMatch(/captured when the worktree is CREATED/i)
+    }
+  })
+
   it('both agent-facing texts document --dry-run, derived from DRY_RUN_VERBS (issue #532)', () => {
     for (const body of [buildCanvasSkillBody('/x/shim.sh'), buildCanvasControlInstructions('/tmp/nodeterm.sh')]) {
       expect(body).toContain('--dry-run')
--- a/src/main/canvas-control-core.test.ts
+++ b/src/main/canvas-control-core.test.ts
@@ -14,6 +14,7 @@ import {
 } from '../core/agents/hook-sandbox-hint-sh'
 import { RETRYABLE } from '../core/agents/agent-message-decide'
 import { PROJECT_TARGETABLE_VERBS } from './project-grants'
+import { DRY_RUN_VERBS } from '../shared/control-verbs'
 import { STRICT_CONTROL_VERBS } from '../core/agents/node-identity-policy'
 import { BROWSER_ACTION_KEYS } from '../core/browser-verb'
 import { BROWSER_RETRYABLE, BROWSER_OUTCOME_LABEL } from '../core/browser-outcomes'
@@ -312,6 +313,19 @@ describe('parseControlRequest', () => {
       expect(body.toLowerCase()).toContain('ignore')
       // Per-role model is what lets ONE spawn-team call mix tiers; the JSON example must show it.
       expect(body).toContain('"model"')
+    }
+  })
+
+  it('both agent-facing texts document --dry-run, derived from DRY_RUN_VERBS (issue #532)', () => {
+    for (const body of [buildCanvasSkillBody('/x/shim.sh'), buildCanvasControlInstructions('/tmp/nodeterm.sh')]) {
+      expect(body).toContain('--dry-run')
+      // The verb list is RENDERED from the set (dryRunDocLines) — walk the real set so a verb
+      // added to the gate lands in the text the day it is added, and a removed one reds here.
+      for (const v of DRY_RUN_VERBS) expect(body).toContain(v)
+      // Both hard edges must be stated, or an agent discovers them by losing a call to each:
+      // unsupported verbs refuse, and --project cannot be combined.
+      expect(body.toLowerCase()).toContain('refuses `--dry-run`')
+      expect(body).toContain('cannot be combined with `--project`')
     }
   })
 
--- a/src/main/canvas-control-core.test.ts
+++ b/src/main/canvas-control-core.test.ts
@@ -315,6 +315,21 @@ describe('parseControlRequest', () => {
     }
   })
 
+  it('both agent-facing texts document --prompt-file and the one-line --prompt fact (issue #520)', () => {
+    for (const body of [buildCanvasSkillBody('/x/shim.sh'), buildCanvasControlInstructions('/tmp/nodeterm.sh')]) {
+      // `assembleLaunchCommand` collapses every whitespace run in a --prompt literal (it rides
+      // argv on a line typed into the pane). An agent that does not know this writes a numbered
+      // brief and gets one paragraph — and the fix, --prompt-file, is useless undocumented.
+      expect(body.toUpperCase()).toContain('ONE LINE')
+      expect(body).toContain('--prompt-file')
+      // The per-role escape on spawn-team must be named too, or teams stay prose-only.
+      expect(body).toContain('promptFile')
+      // The failure that costs a whole station: flattened, a leading slash command swallows the
+      // task as its argument, and the node then reads as idle to `--after`.
+      expect(body.toLowerCase()).toMatch(/begin a prompt with `\/`|start a prompt with `\/`/)
+    }
+  })
+
   it('both agent-facing texts document the sticky verb', () => {
     for (const body of [buildCanvasSkillBody('/x/shim.sh'), buildCanvasControlInstructions('/tmp/nodeterm.sh')]) {
       expect(body).toContain('`sticky --node')
--- a/src/main/canvas-control-core.ts
+++ b/src/main/canvas-control-core.ts
@@ -185,6 +185,23 @@ const VERBS: ControlVerb[] = [
  * dispatch stop agreeing.
  */
 export { isDestructiveVerb, DESTRUCTIVE_VERBS } from '../shared/control-verbs'
+// Imported (not only re-exported) because the agent-facing bodies RENDER the dry-run verb list
+// from the set — the same derive-don't-retype rule as `messagingGuidanceLines`, so the docs can
+// never name a verb the gate does not honour.
+import { DRY_RUN_VERBS } from '../shared/control-verbs'
+
+/** The `--dry-run` paragraph both agent-facing bodies share, rendered from `DRY_RUN_VERBS`. */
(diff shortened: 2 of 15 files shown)


Temikus:
Thank you for turning these round so quickly. The approach is right on all three, and I would merge it. I am asking for two changes first, and I owe you a correction to my own evidence that changes how you should sequence #516.

Please read the correction first, because it is the basis for my first request.

## A correction to my evidence

I went back through the seven-station run's transcripts, and I had the failure the wrong way round.

The flatten did not destroy my stations' task text. Both `/model sonnet` stations received their full brief, did the work, and merged as #25 and #30. Their text survived because their prompts did not begin with a slash. The `400` I quoted in the issues came from a probe node that I opened later, to reproduce the mechanism deliberately:

```
API error: 400 {"type":"error","error":{"type":"invalid_request_error",
"message":"model: String should have at most 256 characters"}}
```

So the two failures rank in the opposite order to how I first reported them:

- The flatten costs structure. `--prompt-file` fixes that well, and that is the part of this PR I care about most.
- A leading slash costs the whole station. Prose is still the only thing guarding it.

The run does show one thing that bears directly on prose as a guard. Both `/model sonnet` stations opened with this sentence:

```
First run '/model sonnet' to keep this station cheap, then do the work below. ...
```

Neither station ran it. Both did the rest of the work correctly, both stayed on Opus, and from the orchestrator I could see none of that. An instruction written in prose can quietly not happen, and the reader cannot tell afterwards.

Temikus:
@eneskirca I think 2 additions would be good here:

### 1. Refuse a leading slash instead of documenting it

Both agent-facing bodies now say "Never begin a prompt with `/`", and the same text says there is no supported way to run a slash command at launch. If a leading slash is never legitimate, the verb can refuse it rather than describe it.

The skill text is read by the orchestrator rather than by the station, so the audience differs from my `/model sonnet` case. I think we need a more deterministic guard here.

`--prompt-file` would be another example. Today, a leading slash swallows one flattened line. With a file, the CLI receives the whole brief as a single argv element, so the slash command swallows the entire brief.

The verb already stats the file, so I think refusing when the first non-whitespace character is a slash would be a better approach. You can apply the same rule to `--prompt`. I don't think anything legitimate begins that way to cause issues.

### 2. Refuse to launch when the brief cannot be read

You have documented that the verb checks the file at call time and reads it at launch. For a node armed with `--after`, the gap between those two moments can be quite long, e.g.:

- If the file has gone by then, `cat` writes to stderr and `"$(cat …)"` expands to an empty string. The CLI therefore starts with no prompt. The node then sits idle at its prompt, and anything armed behind it fires on that idle, causing a similar failure to #521.

If you can, please make the launch line refuse to start the CLI when the read fails, rather than starting it with an empty prompt. Something along the lines of `brief=$(cat '<path>') && <launch> "$brief"` keeps the line single and keeps the echo verification. It also turns a silent empty start into a station that visibly never started.

If the leading-slash refusal lands in this PR, we can also close #516 as redundant.

I am glad to test this on a real fan-out once it lands 👍 Thanks again!

PR #540, opened 2026-08-31
Fix shared Codex identity in local canvas/context shims

howieyoungdahl:
## Why

A Codex tool shell spawned by the shared app-server carries `CODEX_THREAD_ID`, but not the pane `NODETERM_*` variables. The managed status hook already recovers the signed thread-to-node binding. The local canvas-control and linked-context shims checked their environment first, so the same real Codex node was incorrectly reported as outside nodeterm.

This is the focused, current-main extraction of the identity repair discovered while exercising #537; it does not pull in the Server Edition feature branch.

## What changed

- Build each local `nodeterm.sh` and `context.sh` with `codexThreadIdentityResolverSh(codexThreadIdentityRoot())` before its environment gate.
- Keep the exported machine-neutral variants unchanged for SSH installation, so no local record path leaks to a remote host.
- Add a real `/bin/sh` integration regression covering both shims with only `CODEX_THREAD_ID` present.
- Document the invariant in `CONTRIBUTING.md`, `CLAUDE.md`, and the shared-Codex identity guide.

## Verification

- `npx vitest run src/main/codex-shim-identity.integration.test.ts src/main/canvas-control-core.test.ts src/core/context-link-core.test.ts src/server/context-link.test.ts` — 87 passed.
- `npm run typecheck` — passed.
- `NODE_OPTIONS=--max-old-space-size=4096 npm run build` — passed.
- Mutation check: removing the canvas resolver made the new real-shell test fail with the original "not a nodeterm agent node" symptom; restoring it returned 2/2 green.
- Full local suite: 9,314 passed, 27 skipped, with three pre-existing host/environment failures in untouched surfaces (`tmux-paste.realtmux`, bogus-SSH timeout, headless listener probe). Each reproduced in isolation; none of their source or test files differ from `origin/main`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1076,6 +1076,14 @@ else, and its context links must keep classifying across restarts).
     layout by construction on all three surfaces) and then the well-known data dirs; it is monotone
     — advertised dir first, keyed by node-id filename in every candidate, and a foreign instance's
     dir yields a foreign `kid` = `legacy` = exactly what presenting nothing already gave.
+  - **Every LOCAL generated sh client recovers shared-Codex identity before its env gate.** A tool
+    shell forked by the account-scoped app-server carries `CODEX_THREAD_ID`, not the pane's
+    `NODETERM_*`. Managed hooks, local `nodeterm.sh`, and local `context.sh` therefore prepend
+    `codexThreadIdentityResolverSh(codexThreadIdentityRoot())` before testing
+    `NODETERM_NODE_ID`/`NODETERM_CANVAS_CONTROL`. Before this was shared, status hooks recovered the
+    node while both user-facing shims declared that same first-class Codex session outside
+    nodeterm. The SSH constants remain machine-neutral: the local record root is not valid on a
+    remote host and must never be baked into its copy.
   - **Every generated sh client walks the SAME endpoint failover** (`nt_candidates`/`nt_adopt`,
     `core/agents/hook-endpoint-failover-sh.ts`) — issue #445, the endpoint-level twin of #384: a
     session is pinned for life to the endpoint PATH it got at tmux creation, so an app
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1083,7 +1083,39 @@ else, and its context links must keep classifying across restarts).
     `NODETERM_NODE_ID`/`NODETERM_CANVAS_CONTROL`. Before this was shared, status hooks recovered the
     node while both user-facing shims declared that same first-class Codex session outside
     nodeterm. The SSH constants remain machine-neutral: the local record root is not valid on a
-    remote host and must never be baked into its copy.
+    remote host and must never be baked into its copy — enforced by
+    `main/remote-ssh/remote-shim-neutrality.guard.test.ts`, two legs (the exported neutral bodies
+    carry no record root or prelude, and `remote-hooks.ts` cannot even NAME a parameterised
+    builder), because the failure is silent and one-sided: a remote shim carrying the prelude keeps
+    working, and the only symptom is this machine's userData layout sitting in a file on someone
+    else's server.
+  - **That prelude EXPORTS WHAT THE RECORD SAYS — it never decides.** `NODETERM_AGENT_ID` and
+    `NODETERM_CANVAS_CONTROL` were once constants there (`codex`, granted); both are
+    `buildPtyEnv`'s answers about the PANE, which labels a node with its OWN agent id
+    (`custom:<uuid>` for a custom agent whose `baseAgent` is codex, not `codex`) and gates the grant
+    on `canControlCanvas`. The constants therefore mislabelled every custom codex-based node and
+    asserted a grant that agrees with the pane only because
+    `SHARED_IDENTITY_CAPABLE ⊆ CANVAS_CONTROL_CAPABLE` — a coincidence that list's own comment
+    invites the next shared-identity agent to break, and breaking it hands a tool shell a capability
+    its pane was denied. So the record carries `agentId` + `canvasControl` INSIDE the 6-tuple HMAC,
+    and the prelude reads them; the grant is exported only when the record grants it and is left
+    UNSET otherwise (absent, never `0` — the shape both shims' `[ -z … ]` gates expect). The **pane
+    echoes its own label** on `/codex-thread/{start,bind}` (a tmux session outlives the app, so
+    after a restart nothing server-side still knows what agent a node runs), but the **grant is
+    never echoed**: the route derives it with `canControlCanvas`, so there is ONE decider and a
+    forged id cannot manufacture a grant the table refuses. The three preimage generations are
+    **selected by the record's shape, never tried in turn** — a record naming an agent must not
+    verify under a preimage that ignores one — and a pre-agent record's implied `codex` + grant is
+    keyed on the LINE being absent, never on the value being empty, so nothing that names an agent
+    falls back to the guess. The env vars were never a security boundary in any case (anyone who can
+    run the shim can `export` them by hand); the per-node token is, and
+    `docs/shared-codex-node-identity.md` states that argument in full.
+  - **A shell that forwards this identity cannot be type-checked into correctness.** A handler that
+    destructures the request without `agent`, and a record write that omits its optional trailing
+    argument, are BOTH well-typed — so the whole dimension can be plumbed through core, the route,
+    the launcher and the prelude, pass `npm run typecheck` and every unit test, and ship INERT.
+    `main/codex-identity-record-wiring.test.ts` pins it at source level, the same remedy
+    `hook-verified-parity.test.ts` uses for the same class of hole.
   - **Every generated sh client walks the SAME endpoint failover** (`nt_candidates`/`nt_adopt`,
     `core/agents/hook-endpoint-failover-sh.ts`) — issue #445, the endpoint-level twin of #384: a
     session is pinned for life to the endpoint PATH it got at tmux creation, so an app
@@ -1395,8 +1427,9 @@ else, and its context links must keep classifying across restarts).
   injects `isRemoteNode`/`readRemoteFile`/`runRemoteCommand`, bounded tail reads), its hook-fed
   path is jailed at ingest (`isSafeRemoteTranscriptPath`), and `resolveLinkTranscript` REFUSES
   the local locators for remote nodes (they'd resolve a stranger's local transcript). Server
-  Edition passes no deps → local-only (context link is NOT wired there at all — `initContextLink`
-  is never called from `src/server`). Discovery is per-agent: claude installs a
+  Edition IS wired (`src/server/context-link.ts` calls `initContextLink(ptyManager, {})`) but
+  passes no remote deps → **local-only**, which is the complete answer there: that shell runs ON the
+  host whose transcripts and tmux it reads, and SSH projects are a desktop-only concept. Discovery is per-agent: claude installs a
   `get-linked-context` skill; codex/gemini get an idempotent marker block
   (`<!-- nodeterm:get-linked-context:start/end -->`) merged into `~/.codex/AGENTS.md` /
   `~/.gemini/GEMINI.md`. On connect an idle-gated one-line note is injected into each endpoint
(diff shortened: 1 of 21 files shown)


eneskirca:
Reviewed this one closely because it moves a credential-adjacent prelude into two user-facing shims. The core judgement is right, the extraction is well-scoped, and I traced the parts that matter rather than taking the description:

- **The seam is correct.** `writeCliFiles` in `src/core/context-link.ts` is what `src/server/context-link.ts` reaches through `initContextLink`, so the Server Edition inherits the fix without a second implementation. That is exactly the "one definition in `src/core` consumed by both shells" rule CLAUDE.md keeps re-learning, and it is satisfied here by construction rather than by two callers agreeing.
- **The SSH neutrality claim holds today.** `remote-hooks.ts` writes `CONTROL_SHIM_SCRIPT` / `CONTEXT_SHIM_SCRIPT` at all four call sites (L617, L663, L689, L722), and both are now `buildXShimScript()` with no root, so no `userData` path can reach a host.
- **The test is the right kind.** Real `/bin/sh`, a fake `curl` that records argv, both shims from one table — the same discipline as `canvas-control-shim.test.ts` and `remote-claude-usage.test.ts`. The stated mutation check (drop the resolver → the original "not a nodeterm agent node" symptom) is what makes it a regression test rather than a smoke test.

Three things I would like addressed.

**1. The neutrality invariant is documented in three files and enforced by none.** Everything above is true of *this* commit. The failure it guards against — someone "simplifying" `remote-hooks.ts` to call `buildControlShimScript(codexThreadIdentityRoot())` because that is what the local installer does — is silent: the remote shim keeps working (the prelude no-ops when the path does not exist), and the only symptom is this machine's `userData` layout sitting in a file on someone else's server. This repo's answer to exactly this shape is a guard test (`fs-atomic.guard.test.ts` is the precedent). Two assertions would close it permanently:

```ts
expect(CONTROL_SHIM_SCRIPT).not.toContain('codex-thread-nodes')
expect(CONTEXT_SHIM_SCRIPT).not.toContain('codex-thread-nodes')
```

...or, better, a source-level scan that `remote-hooks.ts` never imports the builders. Three prose statements of an invariant and zero tests is the configuration CLAUDE.md warns about by name.

**2. The prelude hardcodes `NODETERM_CANVAS_CONTROL=1`, and this PR is what makes that load-bearing.** The signed record carries `nodeId`, `endpoint` and `accountId` — not whether the pane was granted canvas control. `codexThreadIdentityResolverSh` nonetheless exports:

```sh
NODETERM_AGENT_ID=codex
NODETERM_CANVAS_CONTROL=1
```

Meanwhile `hookServer.buildPtyEnv` sets that variable conditionally, `...(canControlCanvas(agentId) ? { NODETERM_CANVAS_CONTROL: '1' } : {})`, and `hook-server.env.test.ts:49` pins that a **custom agent** (`custom:abc`) gets no such variable. A custom agent declaring `baseAgent: 'codex'` runs the codex CLI, so its tool shells carry `CODEX_THREAD_ID`; if the launcher writes an ownership record for those nodes too, the prelude now re-grants canvas control to a session `buildPtyEnv` deliberately withheld it from — and re-labels it `NODETERM_AGENT_ID=codex`, which is also wrong for status attribution.

This is pre-existing in the hook prelude and harmless there (the hook script does not consult `NODETERM_CANVAS_CONTROL`). It stops being harmless the moment the same prelude sits above `nodeterm.sh`'s gate, which is what this PR does. Could you check whether `bindCodexThreadIdentity` is reached for custom codex-based nodes? If it is, the fix is for the record to carry the grant (and the agent id) rather than the prelude assuming it; if it is not, a sentence in `codex-thread-identity-sh.ts` saying *why* the hardcode is safe would stop the next reader having to re-derive it.

**3. Worth stating explicitly that the env gate was never a security boundary.** Anyone who can run `nodeterm.sh` can also `export NODETERM_NODE_ID=… NODETERM_CANVAS_CONTROL=1` by hand; the actual authorisation is the per-node token (`NODE_TOKEN_READ_SH`) and, for context-link, the requester-keyed link doc. So widening what sets those variables does not widen what a local attacker can reach, and I do not think this PR changes the threat model. That reasoning is not written down anywhere in the diff, though, and it is the first question a security reviewer will ask about "a shim now derives its own identity". One paragraph in `docs/shared-codex-node-identity.md` would settle it.

Smaller notes:

- The endpoint charset filter (`tr -cd 'A-Za-z0-9._/ -'`) admits spaces. Fine for a socket path, and pre-existing, but it is the sort of allowance worth a comment since the value is later interpolated into a `curl --unix-socket` invocation.
- `CLAUDE.md` currently states context link "is NOT wired there at all — `initContextLink` is never called from `src/server`", which `src/server/context-link.ts:87` contradicts. Not caused by this PR, but you are already editing that file's neighbourhood and are clearly reading carefully — a one-line correction would be welcome.
- Documenting the invariant in `CLAUDE.md` *and* `CONTRIBUTING.md`, split by audience, is exactly what the Conventions section asks for. Nice to see it done without being asked.

Good, disciplined extraction — my only real ask before merge is (1), and an answer on (2).

howieyoungdahl:
Thanks — (1) is done, and (2) took some digging, so here is what I actually found rather than just a patch. Pushed as `cca3503f`.

**1. The guard test.** `06eb5828`, in the shape of `fs-atomic.guard.test.ts`, with **both** legs: the value assertion you wrote, and the source-level scan that `remote-hooks.ts` never imports the builders. Positive controls throughout, so the test is seen to fail as well as pass.

**2. On the prelude — your premise does not hold for the case you named, and I want to be precise about why, because the conclusion still lands where you pointed.**

`canControlCanvas` resolves through `capabilityAgentId` (`config.ts:279`, `:274`), and the custom-agent base resolver **is** registered in the process that builds the pty env (`pty-manager.ts:1476`). So a custom agent declaring `baseAgent: 'codex'` resolves to `codex`, codex is in `CANVAS_CONTROL_CAPABLE`, and `buildPtyEnv` has **already granted** it. The prelude is not re-granting; it agrees. The `hook-server.env.test.ts:49` pin is a custom agent with *no* base — and a base-less custom agent never runs the codex CLI, so it gets no thread record and cannot reach the prelude at all.

I also checked the two things that would have made it live anyway, and neither does. The hardcoded `NODETERM_AGENT_ID=codex` is **latent** — nothing reads that variable at runtime; the hooks carry the agent id in the POST path, not the env. And there is no window where the env withholds the grant while the prelude asserts it, because `hasSharedIdentity` and `canControlCanvas` resolve through the same predicate in the same `create()` call, so a cold resolver denies the launcher too and no record is written.

**So the bug today is structural rather than reachable — and I think it is still worth fixing on your preferred shape.** The prelude's assumption holds only because `SHARED_IDENTITY_CAPABLE` happens to be a subset of `CANVAS_CONTROL_CAPABLE`, and that list's own comment invites the next person to add an agent to one and not the other.

So the record now carries `agentId` and `canvasControl` rather than the prelude assuming them, inside a 6-tuple HMAC. Details that matter: preimage generations are **selected by shape, never tried in turn** (trying them in turn is a forgery oracle); the pre-agent fallback keys on the *line being absent*, never on its value; and bind never downgrades a declared record nor promotes an implication into a claim. The pane echoes its own label because tmux outlives the app, but the **grant is derived by the route and never echoed**.

**3.** Written — and I verified it before asserting it, rather than restating the reasoning. No verb is gated on the env var alone: the bearer is required on every route, `/control/*` and `/context-link/*` both go through `identityGate` for the per-node token, and context link is requester-keyed (`linkDocs.get(req.nodeId)`). The paragraph says that with the file references.

**On your smaller notes**, one correction and one agreement. The `tr -cd 'A-Za-z0-9._/ -'` space is not an indulgence — it is **required**, because the macOS path contains one; and the value is dot-sourced rather than passed to `--unix-socket`, so the framing in the note does not quite match what the code does. I commented it accordingly rather than tightening the filter. The stale `initContextLink` claim in `CLAUDE.md` is fixed (plus two other docs repeating it). I deliberately did **not** touch the same row in `docs/grok-agent.md` — main has already rewritten it and my edit conflicted, so I reverted it and re-verified the merge.

Gates: typecheck clean; 12 targeted files 244/244; `src/main` 1316 passed; `src/server src/core` 5 pre-existing failures in two untouched files, reproduced identically on `origin/main` in the same worktree; `server:build` OK. Eleven mutations applied and reverted — one of them caught that a `src/main/index.ts` edit had been silently lost while typecheck stayed green, which would have shipped the feature inert, so there is now a source-level wiring guard for it too.

**Two coordination notes.** This PR still merges clean on current `main`. It overlaps #537 as you spotted — I would suggest **#540 is canonical** and #537 rebases onto it, since this is the smaller, already-reviewed change and currently the only one of my three that merges cleanly. Separately, the launcher lines here will conflict with #652 (the codex daemon-reconnect fix, just extracted from #537 as its own PR); I will rebase whichever lands second.

eneskirca:
Re-verified `cca3503f` in a fresh isolated worktree on current `main` (3411efdf): typecheck clean, full `npx vitest run` **9357 passed / 27 skipped / 0 failed** on the branch head, and — because that head is ~360 commits behind — also merged into current `main` locally: typecheck clean, the identity/shim/guard/boundary suites (18 files, 233 tests) green. `git merge-tree` vs `main` and vs #612: 0 conflicts. Approved the fork CI runs.

(1) The guard test is exactly the shape asked for — both legs, positive controls on each, and it reads the real `remote-hooks.ts`. Done.

(2) Your correction stands: `canControlCanvas` goes through `capabilityAgentId`, the resolver is registered in the pty-env process, so a `baseAgent: 'codex'` custom agent was already granted — the prelude agreed rather than re-granted, and the base-less pin at `hook-server.env.test.ts:49` never reaches the prelude. I agree the fix was still worth making on the record rather than the prelude. Traced the new shape:
- the record carries `agentId` / `canvasControl`, signed as a 6-tuple; verification is selected by record shape (`agentDeclared` → 6-tuple; `accountLinePresent` → account-scoped; else legacy and only for the system scope), never tried in turn;
- the pane echoes only `NODETERM_AGENT_ID`; the grant is derived server-side by `canControlCanvas(rawAgentId)` after `isSafeCodexAgentId`, never taken from the pane;
- `bindCodexThreadIdentity` carries a declared agent forward and never downgrades; the prelude exports what the record says and falls back to `codex`+grant only for a pre-agent record (absent line, not empty value), and bails on a malformed id.

(3) The "never a security boundary" section is in `docs/shared-codex-node-identity.md` and matches the code (bearer on every route, `identityGate` on `/control/*` and `/context-link/*`, requester-keyed link doc). The `tr -cd` space note is corrected in place — I had the framing wrong, the value is dot-sourced. The stale `initContextLink` claim in CLAUDE.md is fixed.

One thing to say in the docs rather than fix here: on the **Server Edition** the shim files are written through the shared seam, but `src/server/handlers/index.ts:88-92` deliberately registers no `setCodexThread*Handler`, so no record is ever written there and the prelude always takes the fallback. That is pre-existing and documented in that comment, but "Server Edition inherits the fix" reads as more than it is; one sentence that the record half is desktop-only until the server registers the handlers would keep the next reader honest.

Merge-ready from my side, and I agree #540 should be canonical with #537 rebasing onto it. Merge is @eneskirca's call.

howieyoungdahl:
Added in `9d1de3ee`, and I verified the claim before writing it rather than restating it.

`src/server/handlers/index.ts` registers no `setCodexThreadStartHandler` / `setCodexThreadBindHandler`, and `src/main/index.ts` is the **only** non-test caller of `writeCodexThreadIdentity` and `bindCodexThreadIdentity`. So on the Server Edition no ownership record is ever written and the prelude always takes the fallback — which is exactly what your comment there already says, and what "Server Edition inherits the fix" was overstating.

The sentence is in `docs/shared-codex-node-identity.md` in two places (the resolver section, and the "both shells" clause about the signed record secret, which is where the overstatement actually lived), and in `CLAUDE.md`.

I deliberately left `CONTRIBUTING.md` alone: it instructs what a shim must prepend and never makes a claim about which shell writes records, so adding the caveat there would be noise in the audience-split this repo keeps.

I have kept it precise in the other direction too — the shim/seam half genuinely **is** shared, and the docs still say so. It is only the record half that is desktop-only until the server registers those handlers.

PR #541, opened 2026-08-31
Dismiss the start screen when navigating from the sidebar

ezwep:
## What I hit

I have eight to ten projects open on a normal day. I clicked `+` in the tab bar to start a new project, then changed my mind — I wanted to go back to something already running. So I clicked one of my existing sessions in the sidebar on the left.

Nothing appeared to happen. The session did not open and the start screen stayed where it was. Clicking a project instead of a session behaves the same. The only thing that works is clicking the empty background first to dismiss the screen, and then clicking the session: two clicks for one intention, with nothing on screen saying that is the order it wants.

What makes it read as broken rather than as "that is disabled" is that the sidebar looks entirely normal — fully visible, rows highlight on hover, nothing greyed out. So it does not feel like the app asking me to finish what I started; it feels like the click was lost.

## What is actually happening

`.welcome` is `position:absolute; inset:0; z-index:5` with an opaque `--canvas-bg`, and `.sessions-sidebar` is `z-index:12`. The sidebar stays on top, so the click *is* delivered: `focusNodeById` selects the node, pans the canvas to it, and hands keyboard focus to its terminal — all behind the still-mounted screen. Neither `focusNodeById` nor `switchProject` cleared `welcomeOpen`.

The part I would flag hardest is not the dead click: **keyboard focus goes to a terminal you cannot see**, so the keystrokes after it land somewhere invisible.

Every project create/reopen path already cleared it — `reopenProject`, `createSshProject`, `onRepoCloned`, the canvas-control reopen. Only the two navigation paths were missed.

## Why this shape of fix

`focusNodeById` already special-cases the kanban board for precisely this failure:

> "The board is a full-page overlay: framing the node on the canvas underneath it is invisible, which is why the notch's Go (and every other 'go to node' path) read as broken there."

The start screen is the same kind of overlay and never got the equivalent branch. This adds it, rather than making the sidebar inert — the screen has a close button, Escape and click-outside, and its CSS comment keeps it below the tab bar "so tabs stay reachable", so navigating away while it is open looks intended and the sidebar was simply not part of that set. If you would rather it were fully modal, the honest fix is the opposite one and I am happy to turn it around.

`switchProject` clears before its same-project guard: clicking the already active project is still "take me back to my canvas", and that path returns early without touching any state.

## Testing

Pinned in `canvas-wiring.test.tsx`, which exists for this exact shape — Canvas has no render harness, so a deleted line here is invisible to every other test. Four assertions: both callbacks clear it, and both do so before their early returns. Verified they fail on `main` and pass with the change.

Worth being explicit that this is a source read, not a behavioural test: it cannot prove the overlay disappears for a user, only that the wiring is still there. If you would prefer no test over a source read here, say so and I will drop it.

`npx vitest run src/renderer` — 267 files, 3488 tests green; typecheck clean. v0.3.4, macOS.


Abhirup0 (commented):
Looks good. Verified that:
- Placing `setWelcomeOpen(false)` ahead of the same-project guard in `switchProject` correctly handles clicking the active project to exit the start screen.
- Keyboard shortcuts (`Cmd/Ctrl+1..9`) route through `switchProject` and inherit the dismissal behavior.
- First-run behavior with 0 projects (`!hasProjects`) remains intact.
- The paren-balancing checks in `canvas-wiring.test.tsx` fit the established test pattern in that file.

Ready to merge.

eneskirca:
Reviewed alongside the rest of ezwep's current set, in an isolated worktree off `origin/main`.

**Verified locally:** typecheck clean; `canvas-wiring.test.tsx` 30/30.

I checked the one thing that could plausibly have broken and it doesn't: the render gate is

```tsx
{(!hasProjects || welcomeOpen) && (
  <WelcomeScreen …
```

(`Canvas.tsx:10602`), so `setWelcomeOpen(false)` on a first run with zero projects is a no-op and the
screen stays up. Nothing can dismiss the app into a blank canvas. (Abhirup0 asserts the same above; I
wanted it confirmed against the source rather than agreed twice.)

I also walked every `switchProject` call site rather than sampling: the ⌘1-9 project chord (`:5718`),
the sidebar project row (`:9762`), presence travel (`:9824`), the canvas-control cross-project open
(`:9034`), the node-owner jump (`:6840`), the command palette (`:10077`). All six are "take me
somewhere" — there is no path that calls `switchProject` as a side effect of something the start
screen should survive, so the unconditional dismissal is right rather than merely convenient.

The shape of the fix is the right one. `focusNodeById` already carries the kanban branch for exactly
this class of bug, with a comment saying why; the start screen is the second full-page overlay and
never got the equivalent. Making the sidebar inert instead would have been the wrong direction — the
screen has a close button, Escape and click-outside, so navigating away from it is clearly intended,
and the sidebar was simply not enumerated in that set.

The part of the report I would also have flagged hardest is the same one you did: **keyboard focus
lands on a terminal the user cannot see.** A dead-looking click is annoying; keystrokes going
somewhere invisible is a different category, and it is worth that sentence staying in the commit
message.

### On "would you prefer no test over a source read"

Keep it. `canvas-wiring.test.tsx` exists precisely because Canvas has no render harness, and the only
realistic future failure here is someone deleting the line during an unrelated refactor — which is
exactly what a source read catches. A test that fails on `main` and passes here is real evidence for
that failure mode, and you are explicit in the file about what it cannot prove. That is the honest
version.

### Set context

Independent of the four Windows PRs (#573/#576/#577/#579) and of #568 — no shared files, no shared
behaviour. Mergeable at any point in that sequence.

**Verdict: approve.**

ezwep:
Thanks — particularly for checking the render gate against the source rather than taking the second
opinion, and for walking every `switchProject` call site instead of sampling them.

Noted on keeping the source-read test. Nothing is pending from my side on this one.

PR #542, opened 2026-08-31
RFC: Evaluate Bun 1.4 as alternative package manager / runtime

AlexanderWillner:
# RFC: Evaluate Bun 1.4 as alternative package manager / runtime

> **Status:** Draft / RFC — prototype measurements only, no adoption proposed yet.
> **Branch:** `AlexanderWillner:chore/bun-migration` → `eneskirca:main`
> **Question:** Is this an interesting direction to pursue?

## Context

[Bun 1.4](https://bun.com/blog/bun-v1.4) advertises significantly better install and runtime performance. We created an experimental branch to test what “replace Node with Bun” would actually mean for nodeterm.

**Heads-up:** The Electron desktop cannot be replaced by Bun — it embeds its own Node (v22 in Electron 42). `node-pty` and `smart-whisper` are rebuilt with `electron-rebuild` against Electron’s ABI and must stay on `node`. The scope of the experiment is therefore:

- `npm ci` → `bun install`
- `npm run <script>` → `bun run <script>`
- `node out/server/main.cjs` → `bun out/server/main.cjs` (Server Edition)
- `node out/session-host/host.cjs` → `bun out/session-host/host.cjs` (session host)

Both runtimes remain installed in CI (Node for the native rebuild, Bun for fetch/run).

## What’s on the branch

- `package.json`: `packageManager: bun@1.4.0`, `engines.bun`, `trustedDependencies` (Bun 1.4 gates lifecycle scripts), reduced `npm run` indirection, added `test:bun`/`server:*:bun`/`bench` helpers
- `bunfig.toml` + `bun.lock` (kept alongside `package-lock.json` for the experiment)
- `.github/workflows/ci-bun.yml` — parallel CI job for this branch
- `scripts/bench.mjs` — build/typecheck/install bench via `/usr/bin/time -l`
- `scripts/bench-runtime.mjs` — runtime bench (RSS, startup, HTTP throughput)
- `docs/bun-migration.md` — full write-up

## Prototype measurements

Host: MacBook Air M2, 8c / 24 GB, bun 1.4.0 (34cbb9a), node 26.7.0 / npm 11.19.0, 31 Aug 2026. Not a lab — laptop variance noted.

### Build / install

| Task | npm (node) | bun 1.4 | Δ |
|---|---|---|---|
| `install` clean fetch only (`--ignore-scripts`) | 42.9 s / 916 MB RSS | 5.7 s + 10 s rebuild | **7.5× fetch / 2.7× overall** |
| `install` clean with `electron-rebuild` | ~52 s (est.) | **15.7 s** / 510 MB | **2.7×**, −44% RSS |
| `typecheck` (`tsc` ×2) | 0.58 s / 326 MB | **0.44 s** / 298 MB | **1.3×** |
| `vitest run` (668 files) | **67.1 s** / 400 MB | 72.4 s / 400 MB | **0.93×** (1 extra failure: `localStorage.clear` in `cardModalSize`) |
| `electron-vite build` + relay | 15.57 s / 2.53 GB | **13.26 s** | **1.17×** |

`bun install` dominates. Script-runner wins are small. `vitest` under Bun’s Node compat is currently slightly slower/less compatible.

### Runtime (the only processes Bun can actually replace)

`node|bun out/server/main.cjs --port 0 --data-dir <tmp>` spawned 3×, `ps -o rss` before/after, 500–1000× `GET /` → 401.

| Process | Node | Bun | Δ |
|---|---|---|---|
| Server idle RSS | ~70 MB | ~46 MB | **−34 %** |
| Server load RSS | ~78–80 MB | ~55–61 MB | **−28 %** |
| Session-host idle RSS | ~56 MB | ~31 MB | **−45 %** |
| Server startup to `listening` | ~98–120 ms | ~71–92 ms | **−20 %** (noisy) |
| Server RPS / avg latency | ~3800–5800 rps / 0.17–0.39 ms | ~4200–8300 rps / 0.12–0.23 ms | **no stable winner** — variance 30–50 %, mean ±10 % |

**Takeaway:** Less RAM yes (−30 % server, −45 % host — relevant for a 1 GB VPS: ~25 MB/instance + ~25 MB/host). Less CPU no — throughput is within noise.

## Reproduce

```bash
git fetch origin chore/bun-migration && git checkout chore/bun-migration
bun install                          # or npm ci — both work
npm run server:build
node scripts/bench-runtime.mjs
BENCH_REQUESTS=1000 node scripts/bench-runtime.mjs
node scripts/bench.mjs
```

## Open questions for maintainers

- Is a dual-lockfile (`package-lock.json` + `bun.lock`) acceptable for an experiment, or should we keep one canonical?
- Are the ~30 % RSS savings on Server Edition worth supporting `bun` as a documented alternative for self-hosted deploys, even though desktop stays on Node?
- Should CI try `bun install` for one job (biggest ROI: ~27 s saved/run vs ~5 s for the build) or keep npm-only until the `vitest` compat issue is resolved?
- Any concerns about `trustedDependencies` or keeping `electron-rebuild` on Node?

No breaking change proposed — this draft just makes the prototype visible and asks whether you find the approach worthwhile. Happy to close, narrow to “CI-only”, or iterate based on feedback.

Closes: n/a (RFC)

### DIFF
--- a/.github/workflows/ci-bun.yml
+++ b/.github/workflows/ci-bun.yml
@@ -1,0 +1,51 @@
+name: CI (Bun)
+
+# Bun variant of the quality gate — runs only on the bun-migration branch
+# (and on PRs targeting it) so main is not affected while the experiment
+# lives. Mirrors the npm job step-for-step, swapping npm ci / npm run for
+# bun install / bun run to measure speedup + memory and to catch
+# bun-specific breakage (lifecycle, native rebuild, vitest compat).
+#
+# Bun 1.4 per https://bun.com/blog/bun-v1.4 — install + script-runner wins
+# should be visible here; Electron itself still runs on its embedded Node,
+# so the main-process runtime is not replaced.
+
+on:
+  push:
+    branches: [chore/bun-migration, experiment/bun-1.4]
+  pull_request:
+    branches: [chore/bun-migration, experiment/bun-1.4]
+
+permissions:
+  contents: read
+
+jobs:
+  quality-bun:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: '1.4.0'
+
+      - uses: actions/setup-node@v7
+        with:
+          node-version: '22'
+          # Node is still needed for electron-rebuild (native modules are
+          # compiled against Electron's Node ABI, not Bun's).
+
+      - name: Install dependencies (bun)
+        run: bun install
+
+      - name: Type check (bun)
+        run: bun run typecheck
+
+      - name: Tests (bun)
+        run: bun run test
+
+      - name: Build (bun)
+        run: bun run build
+
+      - name: Server build (bun)
+        run: bun run server:build
--- a/bun.lock
+++ b/bun.lock
@@ -1,0 +1,1349 @@
+{
+  "lockfileVersion": 2,
+  "configVersion": 0,
+  "workspaces": {
+    "": {
+      "name": "node-terminal",
+      "dependencies": {
+        "@xterm/addon-serialize": "^0.14.0",
+        "@xterm/addon-unicode11": "^0.9.0",
+        "@xterm/addon-web-links": "^0.11.0",
+        "@xterm/headless": "^6.0.0",
+        "dompurify": "^3.4.13",
+        "electron-updater": "^6.8.9",
+        "emoji-picker-react": "^4.19.1",
+        "lucide-react": "^0.545.0",
+        "marked": "^12.0.2",
+        "monaco-editor": "^0.52.2",
+        "node-pty": "^1.0.0",
+        "qrcode": "^1.5.4",
+        "sharp": "^0.35.3",
+        "smart-whisper": "^0.8.1",
+        "tweetnacl": "^1.0.3",
+        "ws": "^8.21.0",
+        "zustand": "^4.5.7",
+      },
+      "devDependencies": {
+        "@electron/rebuild": "^4.2.0",
+        "@tailwindcss/vite": "^4.3.3",
+        "@types/node": "^26.2.0",
+        "@types/qrcode": "^1.5.6",
+        "@types/react": "^18.3.31",
+        "@types/react-dom": "^18.3.7",
+        "@types/ws": "^8.18.1",
+        "@vitejs/plugin-react": "^4.3.1",
+        "@xterm/addon-fit": "^0.11.0",
+        "@xterm/addon-search": "^0.16.0",
+        "@xterm/addon-webgl": "0.18.0",
+        "@xterm/xterm": "^5.5.0",
+        "@xyflow/react": "^12.11.3",
+        "electron": "^42.9.1",
+        "electron-builder": "^26.15.3",
+        "electron-vite": "^5.0.0",
+        "esbuild": "^0.28.2",
+        "jsdom": "^30.0.1",
+        "react": "^18.3.1",
+        "react-dom": "^18.3.1",
+        "tailwindcss": "^4.3.1",
+        "typescript": "^7.0.2",
+        "vite": "^7.3.6",
+        "vitest": "^4.1.10",
+      },
+    },
+  },
+  "packages": {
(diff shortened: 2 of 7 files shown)


eneskirca:
This is a good RFC: the scope is bounded honestly (Electron embeds its own Node, so the desktop was never in play), the measurements are labelled as laptop-variance rather than lab numbers, and — the part I appreciate most — it reports the results that do **not** favour the proposal. "Less RAM yes, less CPU no", `vitest` slightly slower with one extra failure, RPS "no stable winner, variance 30–50%". That is the write-up of someone measuring rather than advocating, and it is worth reading even if the answer is no.

Taking your four questions in order.

**Dual lockfile: no, and this is the strongest argument against adopting Bun anywhere in CI.** `package-lock.json` and `bun.lock` will not stay in agreement, and the mechanism is already in this repo: Dependabot opens PRs against `package-lock.json` only (three are open right now). Every one that merges moves the npm tree and leaves the Bun tree where it was, so the parallel `ci-bun.yml` job would be testing a dependency graph nobody reviewed, diverging silently and monotonically. This codebase's recurring lesson is that a duplicated definition drifts and the drift is invisible until it costs an incident — a second lockfile is that pattern with the highest possible blast radius. If Bun is ever adopted it should be *instead of* npm, not beside it.

**~30% RSS on Server Edition: real, but check the proportion.** `docs/session-memory.md` and the session-memory work in CLAUDE.md measured the thing that actually consumes memory on a nodeterm host: a single `claude` process averages **335 MB** and peaked at **1159 MB**, and 95 of them held **31.1 GB** on the production host that prompted the feature. Against that, saving 25 MB on the server process and 25 MB per session host is roughly **0.1%** of what a busy instance holds. On a 1 GB VPS running one or two sessions it is a real fraction; on the deployment shape this project actually sees, it is noise beside the agent CLIs. That does not make the number wrong — it makes it the wrong lever to change runtimes for.

**CI `bun install` only: tempting, and still no — because of the lockfile above, plus one specific hazard.** Bun 1.4 gates lifecycle scripts behind `trustedDependencies`, and this repo's `postinstall` is not optional: `scripts/patch-node-pty.mjs` rewrites node-pty's darwin `pty_posix_spawn` before `electron-rebuild` compiles it, because unpatched node-pty leaks a ptmx device on every spawn and exhausts `kern.tty.ptmx_max` within hours — terminals then simply stop opening. A gate that silently skips that is the difference between a build and a time bomb. The saving grace is that `src/main/node-pty-patch.test.ts` asserts the marker in the compiled-from source, so a skipped patch fails loudly rather than shipping. Worth naming that guard explicitly in `docs/bun-migration.md` — it is the single most important thing a future reader of this experiment needs to know.

**`vitest` under Bun is a step backwards, and it is the one that matters.** `npm run typecheck` and `npm test` are this project's correctness gates. A runtime that runs the suite ~8% slower *and* fails a test the canonical runtime passes (`localStorage.clear` in `cardModalSize`) is trading a real guarantee for an install-time saving. The install win is the only unambiguous result in the whole table, and install time is the cheapest thing in a developer's day.

**My recommendation to @eneskirca: close this as answered, and keep the artefacts.** Specifically:

- `docs/bun-migration.md` is worth keeping *as a measurement record*, ideally moved next to the other probe write-ups so the next person who proposes this finds the numbers instead of re-running them. The repo already treats "write down what you measured so nobody re-derives it" as a rule.
- **`scripts/bench-runtime.mjs` is the piece worth extracting on its own merits**, independent of Bun. A reproducible RSS / startup-to-listening / RPS harness for the Server Edition process is something this project does not have, and it complements the session-memory panel directly — that feature exists because nobody could see what was holding memory. Landing that script alone (running under Node, no Bun anywhere) would be a real contribution out of this branch.

The condition under which this is worth revisiting is narrow and worth writing down: **Server Edition shipping as a container image on memory-billed hosting, with the agent CLIs running elsewhere.** In that shape 25 MB per process is a line item rather than a rounding error. Until then the runtime that matters is Electron's, and Bun cannot touch it.

Thanks for doing the work properly and for asking before proposing adoption — that is the right order.

eneskirca:
Thanks for the thorough RFC — closing as a decision rather than leaving it to stale. The measured trade doesn't favor switching: a second lockfile drifts against Dependabot's npm updates, the memory win is ~0.1% on our heaviest real-world host, and vitest ran slower with one extra failure. The `scripts/bench-runtime.mjs` harness itself is genuinely useful though — a standalone PR contributing just that benchmark would be welcome.

PR #544, opened 2026-08-31
feat: zoom presets in the dock

basmilius:
## Zoom presets in the dock

The dock's percentage readout was inert text between the `−` and `+` buttons. Getting back to
100% meant clicking a step button until the number happened to land there, and it rarely did:
the wheel and both step buttons move zoom *multiplicatively*, so the set of zooms reachable from
an arbitrary starting point contains no round number at all. `Fit view` was the only one-click
camera command, and it lands on whatever the content bounds imply — never on 1.

This turns the readout into a menu button with a preset scale.

### What changed

- **The readout is now a button** (`aria-haspopup="menu"`) opening a preset menu: **25 / 50 / 75 /
  100 / 150 / 200%**, plus a `Zoom to fit` row so the two camera commands live together. The
  preset the camera currently sits on is checked.
- **Every preset is center-anchored**, like the existing ⌘/Ctrl+0 reset: `viewportAtZoom1` is
  generalized to `viewportAtZoom(current, centre, target)` and the old function becomes a wrapper
  at `target = 1`. Zooming about the origin throws the user's work off screen, which makes a jump
  to a chosen zoom feel like a teleport rather than a change of scale.
- **The zoom range now lives in one place** (`canvas/zoom-limits.ts`). `CANVAS_MIN_ZOOM` /
  `CANVAS_MAX_ZOOM` were a private copy in `wheel-zoom.ts` carrying a "keep in sync with
  Canvas.tsx" comment; the preset menu is a third reader, and it is exactly the drift that comment
  predicted. The `<ReactFlow minZoom/maxZoom>` props, the wheel-step clamp and the preset list
  now read the same two constants.

### Details worth a second look

- **The preset list is bounded by the canvas's own range.** `CANVAS_MAX_ZOOM` is 2, so 200% is the
  last entry — a row React Flow would silently clamp is a row that lies about what it does. A test
  asserts every preset survives the clamp, so raising or lowering the range fails loudly here
  instead of shipping a menu with a dead row.
- **The checkmark compares the ROUNDED percentage the dock already displays**, not the raw zoom. At
  0.996 the readout says "100%", and a menu checking nothing there reads as broken. Rounding is the
  display's contract, so the checkmark shares it rather than re-deriving its own tolerance.
- **`viewportAtZoom` returns the SAME object when it has nothing to do** (already at the target, or
  either side not a usable number). Callers rely on that identity to skip the animated
  `setViewport` entirely, so a no-op stays a no-op rather than becoming a 200 ms nudge.
- **`.dock-zoom-menu .dock-menu__check svg` overrides the generic `.dock-menu svg` muting.** The
  checkmark is a state marker, not a leading icon, so it keeps the accent color that says "this is
  where you are".

### Surfaces

- **Desktop** — the feature.
- **Server Edition** — works as-is: this is pure renderer code (`Dock`, `Canvas`, two lib modules,
  CSS) with no new `window.nodeTerminal` member and no main-process leg.
- **Mobile companion** — N/A. *nodeterm mobile* attaches to tmux sessions over the transport
  protocol and has no canvas, so there is no camera to zoom.
- **Kanban board** — N/A. The dock is not rendered over the board.

### Tests

`src/renderer/lib/zoomPresets.test.ts` (new) covers the list invariants and the checkmark;
`zoomReset.test.ts` gains coverage for the generalized function — including **zooming in**, the
direction the `*1` wrapper never exercises, and the refusal of an unusable target rather than
producing a `NaN` viewport. 27 tests pass across the three zoom suites, and `npm run typecheck` is
clean.

### Verified

Ran the app (`npm run dev`) and used the menu: presets land on their exact percentage, the
checkmark follows the readout, the screen center holds across a jump, and ⌘0 / ⇧1 are unchanged.

image


eneskirca:
Good change, and the reasoning for *why* the readout needed to become a control is the strongest part: the wheel and both step buttons move zoom multiplicatively, so from an arbitrary zoom the reachable set contains no round number at all, and `Fit view` lands on whatever the content bounds imply. That is a real dead end, not a nice-to-have.

Consolidating `CANVAS_MIN_ZOOM`/`CANVAS_MAX_ZOOM` into `canvas/zoom-limits.ts` is the right call — the old copy in `wheel-zoom.ts` literally carried a "keep in sync with Canvas.tsx" comment, and a third reader is exactly the drift it predicted. The test asserting every preset survives the clamp is the part that makes the constant sharing pay off rather than just look tidier.

**On the interaction with `lib/zoomShortcut.ts`'s refusal matrix**, since that was the thing I most expected to break: I checked and it is clean. The refusals were never inside `zoomTo100` — they sit at the two call sites (the keydown handler at `Canvas.tsx:5699` and the `app:zoom-actual-size` IPC receiver at ~L5906, both gating on `zoomShortcutAllowed`/`zoomShortcutChord`). Turning `zoomTo100` into `useCallback(() => zoomToScale(1))` moves no gate. And the refusals themselves would not apply to the new surface anyway: they exist because a *chord* can fire while focus is in a text surface (where Shift+1 is just `!`) or while the board is up, whereas a click on a dock button is an unambiguous gesture, and the dock is not rendered over the kanban board. So no contradiction, and no gate needs adding.

One pre-existing thing the refactor makes visible: the ⌘K palette entry `{ id: 'zoom-100', ..., run: zoomTo100 }` also calls it ungated. That is unchanged by this PR and is fine for the same "explicit gesture" reason — flagging only so nobody reads the ungated `zoomToScale` as a regression.

Smaller notes:

- **The `viewportAtZoom` identity contract is load-bearing and only documented in prose.** Callers skip the animated `setViewport` on `next === current`, so a future refactor that returns a fresh object on the no-op path turns every no-op into a 200 ms nudge. A test asserting `toBe` (not `toEqual`) on the already-at-target case would pin it; the description says the suite covers the generalized function, so this may already be there.
- **`zoomMenuOpen` and `menuOpen` are mutually exclusive by hand** — each button closes the other and the shared backdrop closes both. That works, but three places now have to stay in agreement. A single `openMenu: 'add' | 'zoom' | null` would make it structural. Minor.
- **Keyboard access.** The button correctly carries `aria-haspopup="menu"` / `aria-expanded`, but the popup is a plain `div` of buttons — no `role="menu"`, no Escape-to-close, no arrow-key navigation, and focus is not returned to the trigger on close. If the other dock menu (`dock-menu`) has the same shape then this is consistent with what is already there and not something to fix in this PR; if it does not, worth matching.
- `zoomFromPct` / `isOfferableZoomPct` being separate from `ZOOM_PRESETS` is the right split — it means raising `CANVAS_MAX_ZOOM` fails loudly in the test instead of shipping a row that lies.

The centre-anchoring decision (generalizing `viewportAtZoom1` rather than adding a second path) and the "check the rounded percentage the readout already displays" detail are both good calls with the reasoning written down where the next reader will find it. No blockers from me.

PR #546, opened 2026-08-31
fix: snap subagent and loop cards to the grid on layout

basmilius:
## Problem

With **Snap to grid** enabled, ephemeral fan-out cards (subagent cards and the /loop card) do not land on the grid when they appear. They only jump into place after the user nudges them by a pixel.

React Flow's `snapToGrid` is purely a *drag* constraint. The fan-out cards, however, never go through a drag on their way onto the canvas: their positions are computed in the `ephemeralNodes` memo in `Canvas.tsx` via `offsetFrom(parent, storedOffset, fallback)`, which returns the parent agent's position plus a laid-out column/row offset. Nothing in that path consults the grid, so the card renders wherever the arithmetic puts it, and the first drag is what finally snaps it.

## Fix

`offsetFrom` takes an optional `snap` parameter (`0` = off) and rounds its result onto the grid. `Canvas` passes `ephSnap = settings.snapToGrid ? settings.gridSize || GRID : 0` at both call sites — the subagent cards and the loop card.

Two deliberate choices:

- **The snap applies to the resulting position, not to the offset.** The stored offset is relative to the agent node, which is not necessarily grid-aligned itself, so rounding the offset would leave the card off-grid. Rounding the composed (parent-space) position is what actually puts the card on a grid intersection.
- **A user-dragged offset is left alone.** It was already snapped at drag time if the mode was on, so rounding it again is a no-op there; and for a card placed by hand while the mode was off, re-rounding would move it. That matches how regular nodes behave — turning `snapToGrid` on is a constraint on future drags, it does not rearrange what is already on the canvas (`autoAlignGrid` is the setting that arranges once, and it explicitly skips `subagent`/`loop` nodes).

Only the position is snapped, not the card size — same scope as React Flow's own `snapToGrid`.

## Notes

- Ephemeral cards remain outside the persisted node array: nothing here is written to `project.json`, enters undo, or marks the canvas dirty.
- The card inherits the agent's `parentId`, so the snap runs in the same coordinate space as the rest of the fan-out math — group-relative when the agent sits in a frame, absolute otherwise.
- Grouping/ungrouping the agent still cannot teleport a card, since what is stored stays an offset.

## Surfaces

- **Desktop:** fixed.
- **Server Edition:** fixed by the same code (pure renderer logic, no new bridge member).
- **Mobile:** N/A — no canvas, so no fan-out cards and no grid.

## Testing

- `npm run typecheck` passes.
- Ran the dev app to verify by hand; the visual check has not been completed yet, so treat the behavior as reviewed-by-reasoning until confirmed on the canvas.

### DIFF
--- a/src/renderer/canvas/Canvas.tsx
+++ b/src/renderer/canvas/Canvas.tsx
@@ -644,14 +644,25 @@ const NO_EPHEMERAL: { ephemeralNodes: CanvasNode[]; ephemeralEdges: Edge[] } = {
  * inherits the agent's `parentId`), which is the whole point of storing an offset — grouping or
  * ungrouping the agent flips that space between absolute and group-relative, and a stored
  * position would then teleport the card by the group's own x/y.
+ *
+ * `snap` (0 = off) rounds the LAID-OUT default onto the grid: React Flow's `snapToGrid` only
+ * constrains a drag, so a fan-out card landed off-grid and only jumped into place once the user
+ * nudged it. A DRAGGED offset is left alone — it was already snapped at drag time if the mode was
+ * on, and re-rounding it here would move cards the user placed by hand while it was off.
  */
 const offsetFrom = (
   parent: { position: { x: number; y: number } },
   stored: { x: number; y: number } | undefined,
-  fallback: { x: number; y: number }
+  fallback: { x: number; y: number },
+  snap = 0
 ): { x: number; y: number } => {
   const off = stored ?? fallback
-  return { x: parent.position.x + off.x, y: parent.position.y + off.y }
+  const x = parent.position.x + off.x
+  const y = parent.position.y + off.y
+  if (stored || !snap) {
+    return { x, y }
+  }
+  return { x: Math.round(x / snap) * snap, y: Math.round(y / snap) * snap }
 }
 
 // Delivering an armed node's held launch (canvas-control `--after`) can lose the race against
@@ -1606,6 +1617,8 @@ export function Canvas() {
   // Selection state for ephemeral nodes (they live outside React Flow's managed nodes), owned by
   // the agent-nodes store so the cards themselves can set it — see `selectable: false` below.
   const ephSelId = useAgentNodes((s) => s.selectedId)
+  // Grid the laid-out fan-out cards round onto (0 = snap off) — see offsetFrom.
+  const ephSnap = settings.snapToGrid ? settings.gridSize || GRID : 0
   const { ephemeralNodes, ephemeralEdges } = useMemo(() => {
     // Common case: no /loop running and no subagents → return a stable empty result so
     // this memo (which depends on `nodes`, i.e. recomputes every drag frame) stays cheap
@@ -1644,7 +1657,7 @@ export function Canvas() {
         // with the group). Deliberately no extent:'parent' — the fan-out may hang below the
         // frame border without being clamped into it.
         ...(parent.parentId ? { parentId: parent.parentId } : {}),
-        position: offsetFrom(parent, ephemeralPos[lid], { x: -250, y: ph + 60 }),
+        position: offsetFrom(parent, ephemeralPos[lid], { x: -250, y: ph + 60 }, ephSnap),
         draggable: true,
         // NOT selectable: React Flow's rubber band would otherwise sweep a whole fan-out of cards
         // into the selection alongside the real nodes, and every selection action (Group,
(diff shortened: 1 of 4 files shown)


eneskirca:
Reviewed against the vendored `@xyflow/system` source. The premise checks out: `snapToGrid` is consumed only by `XYDrag` (`getPointerPosition` → `snapPosition`) and by `XYResizer`; nothing on the "a node appeared" path consults it, so an ephemeral card really does land wherever the arithmetic puts it. The two deliberate choices in the description are the right ones — snapping the composed position rather than the offset, and leaving a user-dragged offset alone — and the second is what keeps this consistent with "turning snap on is a constraint on future drags, not a re-layout".

One thing worth a second look before this is called done, because it is the case that survives the fix.

**Coordinate space for a card whose agent sits in a group frame.** `offsetFrom` composes against `parent.position`, which for a node inside a frame is *parent-relative*, so the rounding here happens in frame space. React Flow's own drag snap does not: in `XYDrag` the pointer is converted to flow coordinates, `nextPosition` is snapped there (`snapPosition(nextPosition, snapGrid)` at `@xyflow/system` ~L2131), and only then does `calculateNodePosition` subtract the parent origin. So for a frame whose own `position` is not a grid multiple — which is the norm, since `groupSelectedNodes` creates the wrapper at `(minX - 28, minY - 62)` — the card is snapped onto the *frame's* grid, and the very first drag will still move it onto the canvas grid. That is the same "it only jumps into place after you nudge it" symptom the PR is fixing, just narrowed to grouped agents.

If you agree it is worth closing, the shape is small: resolve the parent chain to a root-space offset (`rootPosition` already exists in `state/workspace.ts` for exactly this), snap there, and convert back. If you would rather not, a sentence in the JSDoc saying the snap is frame-relative by design would at least stop the next reader assuming parity with React Flow.

Two smaller notes:

- `settings.gridSize || GRID` quietly treats `gridSize: 0` as "use the default" rather than "off". That matches what the rest of `Canvas.tsx` does with the value, so it is consistent, but a hand-edited `0` with `snapToGrid: true` snaps to 20 rather than doing nothing.
- Ephemeral cards are outside the persisted array, so there is no `-0` hazard here (the sibling PR #587 handles that for real node positions) — worth noting only so a later refactor that starts persisting these positions knows to add it.

Nothing here is merge-blocking as far as I can tell. Good, precisely-scoped change with a description that explains *why* each half of the rule exists rather than just what it does.

basmilius:
Thanks for checking the premise against the vendored source; that matches what I found.

**Coordinate space for a card inside a group frame - agreed, and fixed.** You are right that `offsetFrom` was rounding a frame-relative value while `XYDrag` snaps `nextPosition` in flow coordinates and only then subtracts the parent origin. Since `groupSelectedNodes` creates wrappers at `(minX - 28, minY - 62)`, an off-grid frame origin is the normal case, so this was not an edge case: a card under a grouped agent landed on the frame's grid and the first drag moved it again, which is the same symptom narrowed to grouped agents.

Rather than resolve the chain inline I pulled it into `renderer/lib/gridSnap.ts`, since #587 needs exactly the same round trip: `containerOrigin(parentId, nodes)` (root at top level, `rootPosition` of the frame otherwise, root again for a dangling parentId) plus `snapPointInRootSpace` / `snapRectInRootSpace`. `rootPosition` is now exported from `state/workspace.ts` instead of being re-derived. `offsetFrom` takes `snap?: {grid, origin}`; the origin is resolved once per parent at each call site, and a top-level agent short-circuits to `(0, 0)` without touching the node array.

Tests pin the mechanism rather than the arithmetic: that a card inside a frame at `(-28, -62)` lands on the canvas grid, that feeding the result back through React Flow's own snap-then-subtract sequence is a fixed point (so the first drag does not move it), and that the old frame-relative rounding produces a different answer.

**`settings.gridSize || GRID`** - left as is, with a line saying why: it reads a hand-edited `0` as "use the default" exactly as the rest of `Canvas.tsx` does, and making this one site treat `0` as "off" would be a second meaning for the same value.

**The `-0` note** - worth flagging that the shared helper normalizes it anyway. Ephemeral cards still are not persisted, but the helper is now on the real node path via #587, so a later refactor that starts persisting these positions inherits the normalization instead of needing to remember it.

#587 is rebased on top of this branch, since it consumes the same helper.

eneskirca:
Re-reviewed the follow-up push (1a2fe58) in an isolated worktree: typecheck clean, the `gridSnap` / `workspace` / canvas suites green (35 files, 600 tests), CI green, and it merges clean onto current main.

The root-space fix answers the frame-origin point, and the fixed-point test (our snap → React Flow's snap-then-subtract → the same value) is the right pin for it. I also checked the dirty/undo boundary, since these cards are never persisted: the snapped position exists only inside the ephemeral `useMemo`, never in the managed `nodes` array, and `markDirty` / the undo snapshot are untouched by this change — so nothing here can reach `project.json` or the history stack.

Ready to merge from my side, in the order you proposed (#546 → #587 → #554).

PR #548, opened 2026-08-31
feat(browser): native context menu in webview guests

basmilius:
Right-clicking inside a browser or web node did nothing at all. This adds a native context menu to
`<webview>` guests.

Nothing here was ever deliberately disabled. Electron ships no context menu for web content
(Chromium's own is a Chrome-browser feature), so the menu simply was never built.

## What it does

One `context-menu` listener in the existing `web-contents-created` guest-hardening block, with the
template built by the pure, unit-tested `main/webview-context-menu.ts`.

| Section | Shown when |
| --- | --- |
| Spelling suggestions (max 5) | there is a misspelled word under the cursor |
| Open Link in New Browser Node, Copy Link Address | the link is http(s) |
| Copy Image, Copy Image Address | `mediaType === 'image'` and it has contents |
| Undo/Redo/Cut/Copy/Paste/Delete/Select All | the field is editable, each enabled from `editFlags` |
| Copy | there is a selection outside an editable field |
| Back, Forward, Reload, Inspect Element | always |

The menu is popped with `menu.popup({ frame: params.frame })`. Electron drops the OS-level menu
items (autofill, writing tools, services) unless it is given the invoking frame, so passing it is
what keeps them available on an input field. **Do not drop the `frame` argument to simplify the
call**, or those items silently disappear again.

The editing block uses Electron roles rather than explicit `webContents.cut()` calls, so the items
carry the OS's own localized labels and accelerators. Enablement comes from `editFlags`, because a
role does not disable itself.

Links go through the same `IPC.browserNewWindow` path as `setWindowOpenHandler` (a new browser
node, never a real popup) and are http(s) only for the same reason. Inspect Element uses Electron's
default `inspectElement`; making that window behave properly on macOS is a follow-up.

## Surfaces

- **Desktop:** the feature.
- **Server Edition:** N/A by construction. There is no `<webview>` in a browser, so there is
  nothing to degrade.
- **Mobile companion:** N/A. No browser nodes.

## Testing

11 unit tests plus `npm run typecheck`. They cover role ordering, `editFlags` disabling, separator
tidying, non-http links refused (`javascript:`, `file:`, `data:`), the image rows, and the
suggestion cap.

Verified by hand on macOS: the OS-level items appear on an input field, links open as new browser
nodes, and copy/paste works in a guest page.

## Note on the diff

`src/main/index.ts` gains only additions here; no existing line is touched.


eneskirca:
Nice find on the root cause — "Electron ships no context menu for web content, so this was never built rather than deliberately disabled" is the right framing, and keeping the template in a pure, Electron-free module so it can be unit-tested under vitest is exactly the seam this repo wants.

I checked the placement in `index.ts`: the listener is added *after* the `if (contents.getType() !== 'webview') return` guard at L1114, so it is genuinely webview-only and cannot appear over the app's own renderer where the canvas has its own context menu. Good.

Points worth a look:

**1. Re-validate the link scheme at the handler, not only in the template.** `openLinkInNode` is handed `ctx.linkURL` by a closure the template built, and the only `isHttpUrl` check is at template-build time. The `setWindowOpenHandler` right above it re-checks inline (`/^https?:\/\//i.test(url)`) before calling `sendToMain`, which is the house rule this repo states explicitly for the permission-mode flag: re-validate at the point of use, because the type is compile-time only. Since `params.linkURL` is attacker-controlled page content and the destination is a node whose `<webview src>` is set from it, I would mirror the existing check inside the handler. It costs one line and makes the guarantee local to the dangerous call.

**2. `Inspect Element` is unconditional.** It opens devtools on an arbitrary remote page inside the app's guest session. That is normal browser behaviour and user-initiated, so I do not think it is a defect — but the repo hardens these guests fairly aggressively (`nodeintegration` off, `will-navigate` allowlist, popups denied), so it is worth stating that this is a considered exception rather than an oversight. If you would rather not ship it to every user, gating it on `!app.isPackaged` or on an existing debug setting would keep the rest of the menu intact.

**3. `params.frame` and the AppKit submenus.** The comment says passing `frame` is what turns on AutoFill / Writing Tools / Services, and the fallback when `frame` is null is documented. Both are the kind of thing that silently regresses under a "simplify the call" refactor — the "**Do not drop the `frame` argument**" line in the description is worth moving into the code comment at the `popup(...)` call itself, since that is where someone will be tempted.

**4. `tidy` is a second copy of `tidySeparators`.** The reason given (the renderer may not import `src/main`) is correct, but the direction the repo prefers for a duplicated rule is one definition in `src/core` consumed by both shells — `src/main` and `src/renderer` may both import `@shared`. A four-line helper is a small thing to drift, but it is the exact category CLAUDE.md warns about. Not blocking; a `@shared/menu-separators` would settle it permanently.

**5. Surfaces.** "Server Edition: N/A by construction, no `<webview>` in a browser" is right. Worth noting that the same is true of the *renderer's* browser node in the bridge, so nothing degrades — the description already says this, I am just confirming I checked rather than took it.

Test coverage is good and covers the interesting refusals (non-http schemes, no-image, no-source-URL, suggestion cap, separator tidying). The role-based editing block with `editFlags` enablement is the right call over explicit `webContents.cut()` calls.

Nothing here blocks merge in my view except possibly (1), which I would fix regardless.

PR #554, opened 2026-08-31
UI polish: icons, spacing and tooltips across the canvas chrome

basmilius:
UI polish across the canvas chrome: icons, spacing and tooltips.

## What changed

**Icons.** Text glyphs used as icons are replaced with SVGs from the shared icon set, so every button centers its icon on both axes instead of sitting on a text baseline. Close buttons used two different glyphs across the app and are now one icon everywhere. `IconSessions` contained two zero-length `<line>` elements that only rendered because of round caps; redrawn.

**Spacing.** New `--float-gap` (22px, the dock's distance) and `--tabbar-h` tokens, and every overlay that floats over the canvas is anchored on them: the dock, the pill cluster, the minimap, the controls, both sidebars, the top-right cluster, the banners, the resume card, the pinned explorer and the dictation pill each carried their own 14 to 22px before.

**Tooltips.** Custom tooltips on every dock and controls button. The dock's open upward, since it hugs the bottom edge; the controls' open to the right, since an upward bubble on a vertical stack covers the button above the one being pointed at. This also brings the `Tooltip` refactor (positioning into a `useTooltip` hook, plus viewport clamping) into this branch ahead of its handle consumers, so the new placement extends the shared component rather than colliding with a rewrite of it. `GroupNode` was the last node file still on native `title` rather than the shared `Tooltip`; its five interactive controls now use it, with `aria-label` on the icon-only ones, while the informational spans and the one button whose label is four lines long keep `title` - the tooltip bubble is `nowrap`.

**One recipe for node header actions.** Ten drifted copies of the icon-button recipe collapse into one, with color reserved for the destructive action.

**Fixes with a specific cause.** The refresh spinner rotated the *button*, not the glyph inside it, so it wobbled around a center the glyph did not share; it now rotates the icon. `.dock-btn:hover:not(:disabled)` out-specificed `.dock-add:hover`, so the add button lost its accent on hover. Tailwind preflight put a ring around the primary button.

**Accessibility.** Icon-only dismiss buttons carried their accessible name via `title` only, which is a tooltip first and a name by fallback. They now carry an explicit `aria-label`.

## Behavior changes

This is not a pure no-op pass. Four things behave differently:

- **The bottom-left Controls are restyled**, not removed. An earlier revision of this PR deleted the widget (zoom and fit already live in the dock, so only the lock added value). That is dropped per @eneskirca: with ~5k users on the current layout, removing a visible muscle-memory surface is not a risk worth taking in a polish pass. The widget keeps its position, its four actions and the lock where they were. The plate, blur and border now sit on the column rather than on each button, so the four read as one enclosure instead of four stacked chips; the rule under every button becomes one separator between the camera moves and the lock; and a `--controls-w` token sizes the buttons *and* offsets the pill cluster that shares that corner, so changing the button size cannot silently drop the pills back on top of them. `fit-view.ts` still reserves the controls rect in `CANVAS_CHROME_SELECTOR`, so `fitAll` lands where it always did.
- **The zoom step is animated in both places.** The controls stepped instantly while the dock tweened, which is one action feeling like two depending on which button you reach for. Both now read `ZOOM_STEP_DURATION_MS`.
- **`.dock-zoom-btn`'s 30px override is gone**, so the two dock zoom-step buttons are the same 38px as every other dock button. That is the point of having one recipe.
- **The group frame's two actions become one.** `ungroup` and the `x` beside it called the same handler under different tooltips, which the icon change made obvious; the `x` is dropped and the survivor reads "Ungroup (keeps nodes)". Deleting a frame with its children is still the context menu's Delete, and a bound frame still releases its worktree binding through `unbind` first.

## Split out of this PR

The **WebNode reload button** (`webviewReload.ts` + its tests) is a feature, not polish, and is not in this branch any more. It goes up as its own small PR once this one lands, as agreed on the review.

## Rebase

Rebased onto current `main`. #544's dock zoom presets and #451/#452's wheel-zoom handling landed after this branch and both survive untouched: `zoomPresets.ts`, `zoomShortcut.ts`, `zoom-limits.ts` and `wheel-zoom.ts` are byte-identical to `main`, and the preset menu keeps its logic, its `onZoomTo`, its `activeZoomPreset` and its mutual exclusion with the add menu. Only its presentation moved to `Tooltip` + `aria-label`.

## Verification

`npm run typecheck` is green. The renderer suite passes apart from pre-existing failures that reproduce identically on unmodified `main` in the same checkout (they read files out of `node_modules` or need a `localStorage` the env lacks).

### DIFF
--- a/src/renderer/bridge/dialog-picker.tsx
+++ b/src/renderer/bridge/dialog-picker.tsx
@@ -17,6 +17,7 @@
 // here, so `..`, absolute and empty names are refused exactly as they are in the Explorer.
 
 import { useCallback, useEffect, useState } from 'react'
+import { IconArrowUp } from '../components/icons'
 import { createRoot, type Root } from 'react-dom/client'
 import type { DirEntry } from '../../shared/types'
 import { newEntryPath } from '../lib/explorerCreate'
@@ -214,7 +215,7 @@ function DirectoryPicker({
             disabled={parent === null}
             title="Up one level"
           >
-            ↑
+            <IconArrowUp />
           </button>
           <span className="dir-picker__path" title={dir}>
             {dir}
--- a/src/renderer/canvas/Canvas.tsx
+++ b/src/renderer/canvas/Canvas.tsx
@@ -741,6 +741,11 @@ const offsetFrom = (
 }
 
 
+/** Zoom-step tween. One constant because the same step sits on two surfaces, the dock and the
+ *  bottom-left controls, and a user who reaches for whichever is nearer must not get two
+ *  different gestures. */
+const ZOOM_STEP_DURATION_MS = 150
+
 // A canvas-control request whose source node lives in another project switches that project in
 // first, and the active-project effect hydrates React Flow ASYNCHRONOUSLY — so the handler waits
 // for the node to appear instead of reading an empty canvas one tick too early. Bounded well under
@@ -13378,16 +13383,15 @@ export function Canvas() {
               default modes. */}
           {glyphLayerActive && <SharedGlyphLayer />}
           {/* The library's own zoom/fit glyphs are swapped for the shared icon set so this cluster
-              matches every other floating control; the actions behind them are unchanged, and
-              zoom stays instant here (the dock's buttons are the animated pair). */}
+              matches every other floating control; the actions behind them are unchanged. */}
           <Controls showZoom={false} showFitView={false} showInteractive={false} position="bottom-left">
             <Tooltip label="Zoom in" placement="top">
-              <ControlButton aria-label="Zoom in" onClick={() => zoomIn()}>
+              <ControlButton aria-label="Zoom in" onClick={() => zoomIn({ duration: ZOOM_STEP_DURATION_MS })}>
                 <IconPlus />
               </ControlButton>
             </Tooltip>
             <Tooltip label="Zoom out" placement="top">
-              <ControlButton aria-label="Zoom out" onClick={() => zoomOut()}>
+              <ControlButton aria-label="Zoom out" onClick={() => zoomOut({ duration: ZOOM_STEP_DURATION_MS })}>
                 <IconMinus />
               </ControlButton>
             </Tooltip>
@@ -13988,8 +13992,8 @@ export function Canvas() {
         onAddWorktree={() => openWorktreeDialog(null)}
         onSave={persist}
         onFitView={fitAll}
-        onZoomIn={() => zoomIn({ duration: 150 })}
-        onZoomOut={() => zoomOut({ duration: 150 })}
+        onZoomIn={() => zoomIn({ duration: ZOOM_STEP_DURATION_MS })}
+        onZoomOut={() => zoomOut({ duration: ZOOM_STEP_DURATION_MS })}
         onZoomTo={zoomToPct}
         onDictate={toggleDictation}
         dictateActive={dictationOpen}
--- a/src/renderer/canvas/Canvas.tsx
+++ b/src/renderer/canvas/Canvas.tsx
@@ -91,9 +91,13 @@ import { Dock } from '../components/Dock'
 import { TabBar } from '../components/TabBar'
 import { ContextMenu, type MenuItem } from '../components/ContextMenu'
 import { CommandPalette, type Command } from '../components/CommandPalette'
+import { Tooltip } from '../components/Tooltip'
 import {
-  IconCollapse,
   IconBranch,
+  IconCanvasView,
+  IconClose,
+  IconCollapse,
+  IconDino,
   IconDuplicate,
   IconEditor,
   IconExplorer,
@@ -101,22 +105,23 @@ import {
   IconGear,
   IconGrid,
   IconGroup,
-  IconDino,
   IconJump,
   IconKanban,
-  IconCanvasView,
   IconLock,
   IconMarkdown,
-  IconReload,
-  IconSmiley,
-  IconPower,
+  IconMinus,
   IconNote,
   IconPhone,
+  IconPlus,
+  IconPower,
   IconProject,
+  IconReload,
   IconRemote,
   IconSave,
+  IconSearch,
   IconSelectAll,
   IconSessions,
+  IconSmiley,
   IconSwitch,
   IconTerminal,
   IconTrash,
@@ -946,9 +951,9 @@ export function Canvas() {
     return () => clearTimeout(t)
   }, [notice])
   const [zoomPct, setZoomPct] = useState(100)
-  // Canvas lock (bottom-left Controls): freezes the viewport against GESTURES — pan (drag +
+  // Canvas lock (bottom-left Controls): freezes the viewport against GESTURES: pan (drag +
   // scroll), zoom (pinch / Cmd+wheel / double-click), node dragging and edge connecting.
-  // Deliberate button clicks (Controls +/−/fit, dock zoom, ⌘K fit) still work, matching React
+  // Deliberate button clicks (Controls +/-/fit, dock zoom, ⌘K fit) still work, matching React
   // Flow's own lock convention. Transient by design: a lock that survives restart reads as
   // "the app is frozen" to whoever opens it next.
   const [canvasLocked, setCanvasLocked] = useState(false)
@@ -13011,7 +13016,7 @@ export function Canvas() {
               title="Dismiss"
               onClick={() => setMigrationNote(null)}
             >
-              ✕
+              <IconClose />
             </button>
           </div>
         )}
@@ -13026,7 +13031,7 @@ export function Canvas() {
               title="Dismiss"
               onClick={() => setSyncNote(null)}
             >
-              ✕
+              <IconClose />
             </button>
           </div>
         )}
@@ -13041,7 +13046,7 @@ export function Canvas() {
               title="Dismiss"
               onClick={() => setCopyError(null)}
             >
-              ✕
+              <IconClose />
             </button>
           </div>
         )}
@@ -13060,7 +13065,7 @@ export function Canvas() {
               title="Dismiss"
               onClick={() => setNotice(null)}
             >
-              ✕
+              <IconClose />
             </button>
           </div>
         )}
@@ -13202,7 +13207,7 @@ export function Canvas() {
           title="Command palette"
           onClick={() => setPaletteOpen(true)}
         >
-          <span className="cluster-search__icon">⌕</span>
+          <IconSearch />
           {paletteChip && <span className="kbd">{paletteChip}</span>}
         </button>
         <button title={commandTooltip('Explorer', 'panel.explorer')} onClick={() => showExplorer('toggle')}>
@@ -13368,14 +13373,37 @@ export function Canvas() {
               mounted in the experimental 'shared' renderer mode; nothing about it exists for the
               default modes. */}
           {glyphLayerActive && <SharedGlyphLayer />}
-          <Controls showInteractive={false} position="bottom-left" onFitView={fitAll}>
-            <ControlButton
-              className={`canvas-lock-btn${canvasLocked ? ' locked' : ''}`}
-              title={canvasLocked ? 'Unlock view (pan/zoom)' : 'Lock view (pan/zoom) — nodes stay movable'}
-              onClick={() => setCanvasLocked((v) => !v)}
+          {/* The library's own zoom/fit glyphs are swapped for the shared icon set so this cluster
+              matches every other floating control; the actions behind them are unchanged, and
+              zoom stays instant here (the dock's buttons are the animated pair). */}
+          <Controls showZoom={false} showFitView={false} showInteractive={false} position="bottom-left">
+            <Tooltip label="Zoom in" placement="top">
+              <ControlButton aria-label="Zoom in" onClick={() => zoomIn()}>
+                <IconPlus />
+              </ControlButton>
+            </Tooltip>
+            <Tooltip label="Zoom out" placement="top">
+              <ControlButton aria-label="Zoom out" onClick={() => zoomOut()}>
+                <IconMinus />
+              </ControlButton>
+            </Tooltip>
+            <Tooltip label="Fit view" placement="top">
+              <ControlButton aria-label="Fit view" onClick={fitAll}>
+                <IconFit />
+              </ControlButton>
+            </Tooltip>
+            <Tooltip
+              label={canvasLocked ? 'Unlock view (pan/zoom)' : 'Lock view (pan/zoom); nodes stay movable'}
+              placement="top"
             >
-              {canvasLocked ? <IconLock /> : <IconUnlock />}
-            </ControlButton>
+              <ControlButton
+                className={`canvas-lock-btn${canvasLocked ? ' locked' : ''}`}
+                aria-label={canvasLocked ? 'Unlock view' : 'Lock view'}
+                onClick={() => setCanvasLocked((v) => !v)}
+              >
+                {canvasLocked ? <IconLock /> : <IconUnlock />}
+              </ControlButton>
+            </Tooltip>
           </Controls>
           {/* Peer cursors live INSIDE <ReactFlow>: PresenceLayer uses ViewportPortal +
               useReactFlow, which throw outside the provider — and cursors are flow coordinates. */}
--- a/src/renderer/canvas/Canvas.tsx
+++ b/src/renderer/canvas/Canvas.tsx
@@ -13383,26 +13383,28 @@ export function Canvas() {
               default modes. */}
           {glyphLayerActive && <SharedGlyphLayer />}
           {/* The library's own zoom/fit glyphs are swapped for the shared icon set so this cluster
-              matches every other floating control; the actions behind them are unchanged. */}
+              matches every other floating control; the actions behind them are unchanged. The
+              tooltips open to the RIGHT rather than upward like the dock's: this is a vertical
+              stack, so an upward bubble covers the button above the one being pointed at. */}
           <Controls showZoom={false} showFitView={false} showInteractive={false} position="bottom-left">
-            <Tooltip label="Zoom in" placement="top">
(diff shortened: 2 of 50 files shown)


eneskirca:
A lot of this is straightforwardly good, and the diagnoses are specific in the way this repo likes: the spin animation sitting on the *button* while the button held a `⟳` text glyph (so it rotated around a centre the glyph did not share) is a genuinely nice catch, as is `.dock-btn:hover:not(:disabled)` out-specificing `.dock-add:hover` and the Tailwind-preflight explanation for the ring around the primary button. Ten drifted copies of one icon-button recipe collapsing into one, with colour reserved for the destructive action, is the right rule and it is stated as a rule rather than as a diff.

**The one thing I would change before this merges is the opening claim.** "No behaviour changes. Everything here is something that was already wrong on screen, not a redesign." is not accurate, and this repo cares about that distinction more than most — a claim written ahead of the mechanism is the failure mode it keeps re-learning. At least three things here change behaviour:

1. **React Flow's `Controls` widget is removed** (`Canvas.tsx`, plus `.react-flow__controls` dropped from `CANVAS_CHROME_SELECTOR` in `fit-view.ts`). Zoom in / zoom out / fit-view lose a surface, and the canvas lock moves from bottom-left to the dock. That is a UI decision worth its own line — CLAUDE.md's "bottom-left canvas lock" sentence is being edited *because* the location changed. Defensible (the dock does duplicate zoom and fit), but it is a redesign, not a centring fix, and it deserves to be the headline rather than a bullet under "Redundant canvas controls".

2. **A new Reload button on web nodes**, with new `webviewReload.ts` + two new test files. That is a feature. The design is good — I like that a forced reload never silently downgrades to a cached one, and the `live` constant existing so the button and the body cannot disagree about whether there is a guest is the right instinct. It just is not polish.

3. `fit-view.ts` no longer reserves the controls rect, which changes where `fitAll` lands.

None of these are wrong. I would just retitle/re-describe so a reviewer skimming for risk is not told there is none, and ideally split (2) out — a 53-file cosmetic sweep and a new node action have very different review costs, and right now the feature is the least-reviewed thing in the PR.

**Sequencing.** This collides with three other open PRs from you, all touching `Canvas.tsx` and/or `Dock.tsx`:

- **#544** (zoom presets) adds `onZoomTo` to `DockProps` and rewrites the zoom cluster in `Dock.tsx`; this adds `canvasLocked`/`onToggleLock` and restyles the same cluster. Guaranteed conflict, and they also disagree about the same question — #544 assumes the dock's zoom readout is the place for camera commands, which is the same argument this PR makes for deleting `Controls`. They probably want to land adjacent, in a chosen order.
- **#587** and **#546** both touch `Canvas.tsx` too, though in different regions.

A note on the PR saying which order you would like them applied would save whoever merges from guessing.

Smaller things:

- `ShortcutCaptureBanner.test.tsx` moving from selecting on the literal `✕` to `title="Dismiss"` is the right direction, but several other Canvas dismiss buttons now render `<IconClose />` while still carrying `title="Dismiss"` and no `aria-label`. Since the sweep is adding `aria-label` elsewhere (WebNode's buttons get one), it would be worth being uniform: an icon-only button whose accessible name comes from `title` is a weaker guarantee than one with `aria-label`.
- `IconSessions` containing two zero-length `<line>` elements that only rendered because of round caps is a good catch and exactly the kind of thing worth the redraw.
- The `--float-gap` / `--tabbar-h` tokens are the right fix for four different edge distances, and I would keep that even if the rest is split.

Verdict: the CSS/icon work is welcome and I would take it. Please correct the "no behaviour changes" framing and consider lifting the WebNode reload out into its own PR; the `Controls` removal is a product call for @eneskirca rather than something a reviewer should wave through.

eneskirca:
Decision on the Controls question: **we're keeping the bottom-left Controls for now.** With ~5k users on the current layout, removing a visible, muscle-memory surface isn't a risk we want to take in a polish pass — that call may be revisited once the dock has carried the same functions for a while.

The rest of this PR is genuinely wanted though. Could you split it?
1. **Icons / spacing / tooltips polish** — welcome as-is (rebase onto current main first; #544's dock zoom presets and #451/#452's wheel-zoom handling landed after this branch, and the review will check those survive the rebase untouched).
2. **WebNode reload button** — separate small PR, happy to take it.
3. **Controls removal** — dropped for now, per above.

Sorry to slice up a careful PR — the polish half deserves to land without riding a UX gamble.

basmilius:
Thanks, that is a fair call on the Controls and I have dropped that part entirely. Rebased onto current main and split as you asked.

**1. Controls stay.** The widget is back in the bottom-left with its position, its four actions and the lock exactly where they were, and `.react-flow__controls` is back in `CANVAS_CHROME_SELECTOR` so `fitAll` reserves its rect again. The CLAUDE.md sentence about the bottom-left canvas lock is reverted too. What I did keep is the presentation, and testing turned up three things worth naming.

The column was painting its plate on each button, so the four read as four stacked chips and the border around them enclosed nothing. Plate, blur and border now sit on the column, matching `.dock`, with transparent buttons and a hover wash. I kept the 10px radius rather than copying the dock's 16px: the dock's is 31% of its height, and on a 30px-wide column 10px is already 33%.

The rule under every button became one separator above the lock. Zoom in, zoom out and fit view are all camera moves; the lock is the only one whose state changes what the other three may do. It rides the existing `.canvas-lock-btn` class, so nothing has to count the Tooltip spans.

The pill cluster shared the bottom-left corner with the column and sat on top of it. It now clears it through a `--controls-w` token that also sizes the buttons, plus the 12px the rest of the chrome already uses between two adjacent clusters (`.resume-card` sets the precedent). The other left-anchored chrome did not need it: the sidebar and its icon cluster are top-left, and the dictation pill is centered.

One behavior change I did make deliberately: the controls stepped zoom instantly while the dock tweened. Same action, two surfaces, two different gestures depending on which button is nearer. Both now read one `ZOOM_STEP_DURATION_MS`.

**2. WebNode reload is out of this branch.** It goes up as its own PR once this lands, as you suggested. You were right that it was the least-reviewed thing in a 53-file cosmetic sweep.

**3. The description is rewritten.** The "No behaviour changes" claim is gone, replaced by a Behavior changes section naming all four.

**4. `aria-label` on the dismiss buttons.** Done as its own commit across all of them, not just the one the test touched. Agreed that an accessible name coming from `title` is the weaker guarantee.

**Two things the icon work uncovered.**

The frame's ungroup button and the `x` beside it called **the same handler**, under two different tooltips; the `x` never deleted anything. It only became obvious once both were icons rather than a word and a glyph. @eneskirca's call was to keep Ungroup and drop the `x`, so the frame now offers one action reading "Ungroup (keeps nodes)". No capability is lost: deleting a frame together with its children is the context menu's Delete and never was this button. I checked the worktree path deliberately, since it is now the only unbind reachable from the frame itself - `ungroup` still routes a bound frame through `unbind` before `ungroupNodes`.

And `GroupNode.tsx` turned out to be the last node file not importing the shared `Tooltip` at all. Its five interactive controls now use it. The informational parts keep `title` on purpose - the branch path, the stale-worktree warning, the dirty/ahead/behind counters and the setup status are text, not controls, matching what the sweep did elsewhere. The setup-failed *button* keeps `title` too, which is the one place the rule bends and why it carries a comment: `.tooltip` is `white-space: nowrap`, and that label is four explanatory lines that would render as one bar off the edge of the window. One CSS rule came with it, because the tooltip wrapper becomes the flex item and does not inherit the `flex: none` those buttons declare. I verified the two structural hazards: no wrapper sits on or above `.group-node__label`, so the frame's `dragHandle` still resolves, and the wrappers inherit `pointer-events: auto` from the containers that re-enable it under the frame's `pointer-events: none`.

**On the rebase.** #544 and #451/#452 both survive untouched, and I checked rather than assumed: `zoomPresets.ts`, `zoomShortcut.ts`, `zoom-limits.ts` and `wheel-zoom.ts` are byte-identical to main. In `Dock.tsx` the zoom cluster keeps #544's preset menu, `onZoomTo`, `activeZoomPreset`, `zoomMenuOpen` and the mutual exclusion with the add menu.

**Merge order**, since you asked for one:

1. **#546** (subagent/loop card snap) - smallest, and it introduces the shared root-space snap helper
2. **#587** (snap-to-grid) - consumes that same helper, which answers your point 2 on both PRs at once
3. **this PR** - the polish, rebased
4. **WebNode reload** - new PR, after this one merges
5. **#566** (terminal focus on window activation) - independent of all of the above, so it can go anywhere in the sequence that suits you; it is last only because nothing else waits on it

basmilius:
Two more commits after another pass over the restored controls.

**The controls' tooltips open to the right now** rather than upward like the dock's. On a vertical stack an upward bubble covers the button above the one being pointed at, which is the opposite of what a tooltip is for. The dock is horizontal and keeps `top`.

That pulled the tooltip refactor forward into this branch. A separate branch had already rewritten `Tooltip.tsx` into a `useTooltip` hook plus a thin wrapper, and adding a third placement to the old component here would have guaranteed a conflict with it for no gain. So the refactor lands first, unchanged and on its own commit, and the placement builds on it. Only `Tooltip.tsx` came across: the handle consumers and the 30 lines of `data-tip` CSS they retire stay on that branch, which rebases onto this one leaving a pure consumer diff (`StickyNode`, `TerminalNode`, that CSS) with `Tooltip.tsx` dropping out of the commit entirely. I verified that rather than assuming it. The public props are unchanged and the defaults moved intact into the hook, so no existing caller needed touching.

The clamp was the part that actually needed thought. It slides a bubble back inside the window on the assumption that it is centered on `left`, which is true for `top`/`bottom` and false for `right`, where `left` is the bubble's own left edge; clamping it as if centered would push it half its width off the trigger near an edge. It now reads how much of the bubble sits either side of the anchor, so the guarantee the refactor introduced covers the new placement instead of having an exception carved into it. The anchor and class-modifier logic are pure and tested, because the anchor and its `.tooltip--*` transform are two halves of one position: change one alone and the bubble drifts by its own width, which renders fine and merely looks slightly off.

`useTooltip` is exported here without a caller until the handle branch lands. I left it rather than hiding it and re-exporting it a commit later.

The description is updated to match.

eneskirca:
Re-reviewed 53ad554 in an isolated worktree: typecheck clean, the renderer component / canvas / node / lib suites green (174 files, 2029 tests), CI green. The split is as asked: the Controls are back bottom-left with the same four actions, `fit-view.ts` and CLAUDE.md are untouched in the diff, and the WebNode reload is out.

**Scroll-wheel zoom: not affected, checked in code rather than assumed.** `wheel-gesture.ts`, `wheel-zoom.ts`, `zoomPresets.ts`, `zoomShortcut.ts`, `zoom-limits.ts` and `zoomReset.ts` are byte-identical to main; no `zoomOnScroll` / `panOnScroll` / `zoomOnPinch` prop on `<ReactFlow>` changes; nothing under `src/shared` or `state/settings` is touched. The `Dock.tsx` diff is Tooltip wrapping plus icons — `onZoomTo`, `activeZoomPreset`, `zoomMenuOpen` and the add/zoom menu exclusion from #544 are intact. The one deliberate behaviour change (Controls step zoom tweened at the same `ZOOM_STEP_DURATION_MS` as the dock) is named in the description.

**One thing before merge:** the branch now conflicts with main in `ShortcutsPanel.tsx` — #622 (4d7e5b6a) rewrote the panel from the keybinding registry after your rebase. Your change there is only ✕ → `<IconClose />` + `aria-label`, so it is a one-line reapply after another rebase. I resolved it locally by taking main's file to check the rest of the merge: typecheck and the suites stay green (178 files, 2030 tests).

Two small notes, neither blocking:
- The terminal `×` keeps the tooltip "Close (ends the session)" but its `aria-label` is now just "Close". The warning is the part that matters, so the accessible name should carry it too.
- The Controls buttons are now wrapped in Tooltip spans, so they are no longer direct children of `.react-flow__controls` — worth one look on a device that the column plate, the lock separator and the right-opening bubbles render as described. I confirmed the removed frame `×` really called `ungroup` (both `onClick={ungroup}` in the removed lines), so dropping it loses nothing.

basmilius:
Rebased onto current main, which merged #546 and #587 in the meantime. One conflict, in `ShortcutsPanel.tsx`: `4d7e5b6a` rewrote that panel to derive its rows from the keybinding registry, and this branch's only change there was swapping the close button's `x` glyph for the shared icon. I kept the rewrite whole and re-applied just that swap on top, with the import moved into the file's relative-import group to match its own ordering. The new `ShortcutsPanel.test.tsx` watchdog passes.

I checked that nothing of the two merged PRs got partially reverted rather than assuming it: `zoomPresets.ts`, `zoomShortcut.ts`, `zoom-limits.ts`, `wheel-zoom.ts`, `gridSnap.ts`, `resizeSnap.ts`, `nodeSizing.ts` and `state/workspace.ts` are all byte-identical to main, and the only zoom-related lines this branch still changes in `Canvas.tsx` are the two literal `150`s becoming one shared constant.

The rebase also turned up an accessibility regression this branch had been carrying, so there is one more commit. On main the close buttons of the shortcuts, log and source-control panels have `x` as their text content, which is their accessible name; replacing that text with `<IconClose />` - a bare `<svg>` with no `<title>` - left them nameless. The earlier `aria-label` commit missed them because it upgraded buttons that already carried a `title`, and these three carried nothing. I scanned the whole branch for icon-only buttons with neither an `aria-label` nor a `title` rather than fixing the three I knew about: it found a fourth, the terminal header's AI-name button, which was the only one of its row without a label. All four now have one. A `Tooltip` wrapper renders a bubble, it does not supply an accessible name.

basmilius:
@eneskirca Rebased onto current main (`d76b4996` -> the 25 commits that landed since, including #663 and #665's canvas-focus work and the session-memory pause). Two conflicts, both in `SessionMemoryPanel.tsx`, both the same import line: main added `useState` for the new pause control while this branch added `IconClose`/`IconReload` for the sweep. Kept both, nothing else in that file changed - the diff there is now exactly the two glyph swaps (`x` -> `<IconClose />`, the reload arrow -> `<IconReload />`).

I checked rather than assumed that main's new work survived: `SessionMemoryPanel.tsx` is +3/-2 against main, and the pause slot, `sessionPauseOffer`, the in-flight guard and the `Pause {title}` label are all untouched. `TerminalNode.tsx`, `CardModal.tsx` and `Canvas.tsx` auto-merged; the unannounced-death chip and its one-click resume are intact.

`npm run typecheck` is clean and the renderer suite is green: 322 files, 4260 tests. The two failures in `state/cardModalSize.test.ts` are a local Node/localStorage quirk, not this branch - I ran that file on a detached `origin/main` and it fails identically there.

One thing I deliberately left alone: the new pause button renders a bare glyph, so it is the one icon-only control in that panel the sweep does not cover. It arrived after this branch and swapping it here would widen a rebase into a change to your feature, so I would rather do it in a follow-up if you want it.

Could you take another look when you have a moment? The branch has now been rebased three times waiting on the review, and each pass costs a bit more as main moves under it. No rush on my side - I just would rather not keep rebasing indefinitely if the answer is that it needs another split or is not wanted after all. Happy either way.

basmilius:
Change of mind on the one thing I said I would leave out: the pause control is now in this branch after all, in `f79fffe8`.

I had reasoned that touching it would widen a rebase into a change to your feature, but that is not what it is. It is a control the sweep would have caught if it had existed when the branch was written — the panel drew a bare `⏻` while the node menu's "Pause session", running the exact same `pauseAgentNode`, drew `IconPower`. Leaving it out would have meant asking you to review the same rule twice, once here and once in a follow-up, so a separate PR was the wrong shape and I have closed it.

`IconPower` is the right icon rather than merely an available one: its doc-block reserves it for acting on a **process** rather than a view, and a pause quits the agent CLI while leaving the tmux session, its scrollback and the conversation alone. The in-flight `…` becomes the shared `.ui-spinner`. No new icon, no new CSS.

While in there, the pause and kill rules are now identical: both get the centering an SVG child needs and a text child did not, in the order the kill button already used, and neither keeps the `font-size` that stopped meaning anything the moment its child stopped being text. The kill button's dead `14px` was mine, left behind by the `x` -> `<IconClose />` swap earlier on this branch — dropping one and not the other is the drift the branch is against.

New test pins the swap in both directions: icon at rest, spinner while the promise is pending, icon back once it settles. It settles inside `act`, because a promise still pending at teardown resolves into an unmounted tree and vitest reports that as an unhandled error.

Typecheck clean; renderer suite 322 files, 4261 tests. The two `state/cardModalSize.test.ts` failures are the local Node/localStorage thing from my last comment, identical on a detached `origin/main`.

eneskirca:
Re-verified `f79fffe8` in an isolated worktree, merged onto current `origin/main` — **the rebase is clean**, no conflicts. `npm run typecheck` clean; the full renderer suite green: 325 files, 4298 tests. (The two `state/cardModalSize.test.ts` failures you saw locally do not appear here — that file is not on `main`, so it never runs.)

**Scroll-wheel zoom: re-confirmed on the rebased head, by file set rather than by reading.** The branch touches 50 files and **none** of them sit under `src/renderer/lib/`, `src/shared/` or `src/renderer/state/`, so `wheel-gesture.ts`, `wheel-zoom.ts`, `zoomPresets.ts`, `zoomShortcut.ts`, `zoom-limits.ts`, `zoomReset.ts`, `gridSnap.ts`, `resizeSnap.ts` and `nodeSizing.ts` are byte-identical to main by construction rather than by inspection. The `Canvas.tsx` diff is 186 lines and carries no `zoomOnScroll` / `panOnScroll` / `zoomOnPinch` / `minZoom` / `maxZoom` change; its only zoom-related lines are the two literal `150`s becoming one `ZOOM_STEP_DURATION_MS` — the Controls step-zoom tween, which is the single behaviour change the description names.

The late `SessionMemoryPanel` commit is +3/-2 against main (three glyph swaps plus the `IconPower`/spinner pair) and leaves the pause slot, `sessionPauseOffer`, the in-flight guard and the `Pause {title}` label untouched — I checked rather than assumed, since that feature landed under the branch.

Merge-ready from my side, and thanks for absorbing three rebases. Merge is @eneskirca's.

Issue #447, opened 2026-08-26
Add devin cli support


Asherslab:
Loving the different agents and LLM support. Work required us to use Devin (windsurf) for our dev work, would love to see first class support added for it

eneskirca:
Thanks — glad the multi-agent side is earning its keep. Two answers here: what you can do **today**, and what first-class (builtin) support would take.

**Today: add it as a custom agent, no code needed.**

Settings → Agents → Custom agents lets you define any CLI as an agent: a label, a launch command (whatever starts the Devin CLI in a terminal), optional extra args and env vars (with `${env:VAR}` expansion at spawn time). You get a real canvas node that spawns it in a persistent tmux session, with process/terminal-title status, grouping, kanban cards, the works.

One important caveat: custom agents can declare a `baseAgent` to inherit a builtin harness's capabilities (hooks, resume, status badges…) — but that exists for CLIs that are *literally* harness-compatible, e.g. a Claude-compatible proxy. Devin is its own harness, so **leave `baseAgent` unset**: inheriting, say, the Claude harness would light up features (busy/idle detection, session resume, context meter) that would silently do the wrong thing against a different CLI. Without a base you get spawn + terminal-title + process status only — honest, if minimal.

**What builtin, first-class support requires.**

Every capability in nodeterm is a membership list gated on measured facts about the real CLI, not assumptions — that rule exists because assuming another agent behaves like Claude has shipped us real bugs (wrong context percentages, badges that strobe, flags silently swallowed). For Devin the checklist would be:

1. **Registry entry** — id, label, spawn command, prompt convention.
2. **Status detection** — does the Devin CLI have a hooks/event mechanism (like Claude/Codex/Gemini/Grok do) we can subscribe to for working / waiting / blocked / done? Without one, no RUNNING/NEEDS-YOU badges.
3. **Session resume** — a documented `resume <id>`-style command, so sessions survive a machine reboot (cold restore) and support in-place restart.
4. **Context/usage reporting** — where and in what format it reports token usage, if at all. Each CLI's formula is different and we don't guess denominators; no trustworthy number means no meter, rather than a wrong one.
5. **Approval/permission flags** — whether it has a permission-mode axis and what each mode actually means.

The practical blocker: none of us has a Devin seat, and every one of those points has to be measured against the real CLI (we keep a per-agent doc of measurements for each builtin). So the realistic path is a collaboration — if you (or anyone with work access) can answer a few concrete questions, I can scope the integration properly:

[...]

Issue #409, opened 2026-08-24
Codex approval prompts break after nodeterm installs PermissionRequest hook


blastart:
### What happened

Codex approval prompts stopped appearing for me after launching nodeterm. This affected both Codex nodes inside nodeterm and Codex started from a regular terminal.

I traced it to the `PermissionRequest` hook that nodeterm adds to the global `~/.codex/hooks.json`. Removing only that hook and restarting Codex fixes approvals. Launching nodeterm again restores the hook and the problem comes back.

### Environment

- nodeterm 0.3.2 desktop
- Codex CLI 0.149.1
- macOS 26.6.1 (25G76), Apple Silicon
- `approval_policy = "on-request"`
- `approvals_reviewer = "user"`
- nodeterm `hookReplyApprovals`: `false`

### Steps to reproduce

1. Launch nodeterm.
2. Check `~/.codex/hooks.json`. A nodeterm handler is registered under `PermissionRequest`.
3. Start Codex in nodeterm and run a harmless command that needs approval.
4. The approval prompt does not appear or remains stuck.
5. Start Codex from a regular terminal. It has the same problem because the hook is in the global Codex config.
6. Remove only `.hooks.PermissionRequest` from `~/.codex/hooks.json` and restart Codex.
7. Approval prompts work again.
8. Relaunch nodeterm. It writes the hook back and the problem returns.

I repeated this in A/B/A order. With the hook removed, two real approval requests appeared and could be accepted normally. After restarting nodeterm, `hooks.json` was modified and the `PermissionRequest` entry was restored, even though `hookReplyApprovals` was still disabled.

### Workaround

After starting nodeterm, remove `.hooks.PermissionRequest` from `~/.codex/hooks.json`, then start or restart Codex. The other nodeterm hooks can remain enabled.

### Possible fix

`PermissionRequest` is currently included unconditionally in `CODEX_EVENTS`, while `docs/hook-reply-approvals.md` says hook-reply approvals are Claude-only and the Codex decision contract has not been verified. `NODETERM_PERM_WAIT_SECS` is also only enabled for Claude.

It may be safer to stop installing `PermissionRequest` for Codex until that hook is supported and tested, or add a setting to disable individual managed hooks. At the moment nodeterm reinstalls it on every launch.

Possibly related: #51, although that report is about hook trust indices rather than approval prompts.

eneskirca:
[...]

PR #555, opened 2026-08-31
feat: add Devin CLI support

Asherslab:
Adds **Devin** (`devin` on PATH, measured at 3000.4.25) as a builtin agent id alongside claude, codex, gemini, copilot, opencode and grok, following the "Adding a new agent" rules in CLAUDE.md.

Relates to #447.

## What Devin claims

| Capability | Devin | Why |
|---|---|---|
| `AGENT_HOOK_TARGETS` | joined | `normalizeDevin` + `DEVIN_HOOK_EVENTS` (7 events) + installer writing the shared managed hook into `~/.config/devin/config.json` |
| `RESUMABLE_AGENTS` | joined | `devin --resume <SESSION_ID>` |
| `PERMISSION_MODE_CAPABLE` | joined | own dialect: `--permission-mode auto\|accept-edits\|smart\|dangerous` |
| `CANVAS_CONTROL_CAPABLE` | joined | claude-shaped skill layout, `~/.config/devin/skills/<name>/SKILL.md` (confirmed with `devin skills paths`) |
| `CONTEXT_LINK_CAPABLE` | joined | `locateDevin` + `linesFromDevin` over `~/.local/share/devin/cli/transcripts/<sessionId>.json` |
| `MODEL_SWITCH_CAPABLE` | **not** joined | `--model` takes Devin-native slugs (`swe-1-7`) and the CLI has no gateway base-url/api-key env |
| subagent / usage / chat / title / shared-identity / minted session id | not joined | no measured surface |

Launch grammar: `promptInjectionMode: 'argv'` with `argvPromptSeparator: '--'`, so the permission flag goes **before** the separator (grok's branch) — the prompt is a positional and `list`/`auth`/`models`/`update` would otherwise be read as subcommands.

Hook dialect: snake_case envelope (`hook_event_name`, `session_id`, `prompt_id`), **no `transcript_path`** (hence the locator), and `Stop` carries only `stop_hook_active` — no `last_assistant_message`, so `lastMessage` stays unset.

## Second commit: review findings

The first commit's capability claims were reviewed and six were wider than what had been measured. Each is closed by dropping the claim, not by guessing:

- **perm-wait (would have re-opened #409)** — the hook-reply-approval gate had been widened from `agentId === 'claude'` to `hasPermissionMode`, but the wait branch is claude-only at *build* time in `managed-script.ts`. Devin gained nothing, while codex/gemini/grok/devin sessions started carrying `NODETERM_PERM_WAIT_SECS` — exactly the inherited-wait exposure #409 fixed. New `PERM_WAIT_CAPABLE = ['claude']` + `hasPermWait`; devin joins the "wait branch is absent from every non-claude script" test.
- **model switch** — devin removed from `MODEL_SWITCH_CAPABLE`. `modelGatewayEnv` had no devin case, so it returned `{}` while `withAgentModel` still appended `--model`: a switcher that rewrote the launch line while routing stayed on Devin's servers.
- **hook config shape** — `devin.test.ts` asserts the exact JSON written: per-event entries, `.*` matchers on the tool events only, idempotent re-install, the user's other settings preserved, stale managed entries swept, clean removal.
- **`manual` mislabelled** — Devin's own default is `auto`, which auto-approves read-only tools, so emitting no flag would have delivered `auto` under an "Ask each time" label (codex's documented trap). `manual` and `plan` are now unsupported for devin and the derived settings copy says so. `modeSupported` no longer early-returns for an agent that *has* a dialect; gemini keeps `manual` via an explicit empty-value key, and that convention (present+empty = supported, no flag; absent = unsupported) is now written down.
- **claude cross-fire** — measured: devin does **not** read `~/.claude/settings.json` (a `SessionStart` hook there did not fire), so the claude hook cannot double-fire in a devin session. The comment claiming the opposite is corrected.
- **discovery installed** — both capability lists set `NODETERM_CANVAS_CONTROL` / context-link env while nothing told devin the verbs existed. The canvas-control and get-linked-context SKILL.md are now written into `~/.config/devin/skills/` locally and on SSH hosts.
- `dialectFor` resolves through `capabilityAgentId`, so a custom agent declaring `baseAgent: 'devin' | 'gemini' | 'codex'` inherits that flag *and* value table instead of falling back to claude's spelling. Pinned by test.

## Surfaces

- **Desktop:** full.
- **Server Edition:** hooks, normalize, resume, permission modes and the capability lists are all in `src/core` / `src/shared`, so they work there. Canvas control is not wired on that shell at all, so the per-account skill install is desktop-only, as it already is for claude.
- **Mobile:** N/A — the companion attaches to tmux sessions and carries no per-agent capability concept.

## Verification

`npm run typecheck` clean. Touched suites green. `src/core/agents/hooks/opencode.test.ts` has 4 failures on this branch **and on `main`** — pre-existing, unrelated.

`docs/devin-agent.md` carries the full measurement write-up and a six-item device checklist for what could not be verified here: the `PermissionRequest` decision contract, the remote skill path on a real host, remote context-link reads (no `transcript_path` to jail), title sync, the usage meter, and the Windows skill/instruction paths.


eneskirca:
Audited against the seventeen "Adding a new agent" rules in CLAUDE.md, in an isolated worktree. **`npm run typecheck` clean, full suite green: 683 files, 9344 passed / 27 skipped, 0 failed** — including `src/core/agents/hooks/opencode.test.ts`, which your notes list as failing; it is green here, so whatever caused that is local to your checkout, not the branch.

This is a strong contribution, and the second commit is the reason. Walking back six capability claims because they were wider than what you had measured — rather than defending them — is exactly the discipline the rules ask for, and the perm-wait one would have re-opened #409 in production. A few of the calls are worth naming because they are the ones most new-agent branches get wrong:

- **`manual` and `plan` unsupported for devin** (rule 8). Devin's default is `auto`, so emitting no flag under an "Ask each time" label would have been codex's trap repeated. You found it and closed it by narrowing the promise, not by substituting a nearest match.
- **Not in `USAGE_CAPABLE`** (rule 2). That list gates three features, not one, and joining it for a meter would have routed a devin node through claude's `resolveTranscript` cwd fallback. Staying out was the right read.
- **Not in `MODEL_SWITCH_CAPABLE`** (rules 17/18), with the reason stated: `--model` takes Devin-native slugs and there is no gateway env, so a switcher would have rewritten the launch line while routing stayed on Devin's servers.
- **The permission flag is asserted at the composed layer**, `assembleLaunchCommand` producing `devin --permission-mode accept-edits -- 'version'` — not only through `withPermissionMode`. Rule 5 exists because the unit test passes while the composed line is wrong; you tested the thing the rule points at.
- **`PermissionRequest` matched with `===`** (rule 7), not a substring — grok's `includes('permission')` strobed NEEDS YOU on every tool call, and this does not repeat it.
- **Both raw listeners are untouched and stay symmetric** (rule 11). Devin needs neither grok's session-directory branch nor a context tail, so "no change in either shell" is the correct answer here rather than a gap — worth one sentence in the doc saying that, since the rule's default assumption is the opposite.
- Rule 12 has nothing to jail by construction: no `transcript_path` in the envelope, so `locateDevin` derives the path and the remote consequence is on the checklist rather than papered over. `resolveLinkTranscript` keeps refusing local locators for remote nodes, so devin inherits that refusal for free.
- Rule 16 is satisfied — six honest checklist items in `docs/devin-agent.md` §9, each naming what was not measured.

Two things I want closed before merge, one of them small.

**1. `hasPermWait` silently widens the gate for custom agents, relative to `main`.**

`main` reads `(options.agentId ?? 'claude') === 'claude'`. This branch reads `hasPermWait(options.agentId ?? 'claude')`, and `includes()` resolves through `capabilityAgentId` — so a **custom agent declaring `baseAgent: 'claude'`** now gets `NODETERM_PERM_WAIT_SECS` in its session env, where before it got nothing.

I think that is probably the behaviour you want: a claude-based custom agent runs claude's hook script, so it can honour the decision reply, and every other capability in this system resolves through the base. But it is an undeclared change to the exact environment variable #409 was about, and the PR body describes `PERM_WAIT_CAPABLE` as a pure narrowing. Please either state it as intentional and pin it with a test (`hasPermWait('custom:x')` with a claude base is `true`, with a devin base is `false`), or compare against the raw id if it was not intended.

**2. The `/exit` claim for devin is asserted in a test comment and nowhere else.**

`EXIT_SEQUENCES.devin = '/exit'`, and the only record of why is a comment inside `agent-restart.test.ts` saying `/exit` and `/quit` both exit. `docs/devin-agent.md` does not mention the exit sequence at all, and neither does the device checklist — so the one fact behind it has no measurement note and no item admitting it is unverified.

This matters more than the line count suggests, because joining `RESUMABLE_AGENTS` plus this table makes devin nodes eligible for the bulk "restart idle agents" action and for Eco hibernation, both of which type that string into a live session. Rule 15 asks for the CLI's documented primary, **bare**, and the reason it is a rule is gemini's `/quit --delete`, which exits *and permanently deletes the session history* the restart exists to resume. So: is `/exit` Devin's documented primary quit, and does it take any argument that destroys session state? If you measured it, a sentence in the doc and a `not.toContain` pin like gemini's closes it. If you did not, checklist item 7.

**Nits**

- `withPermissionMode`'s doc comment in `approval-mode.ts` still reads "with one (grok's `--`)" — devin now takes that branch too, and the comment above `dialectFor` already knows it. Rule 10's whole subject is a rule written in two places drifting.
- The new convention you introduced — *present key with an empty value = supported, emits no flag; absent key = unsupported* — is a genuinely good disambiguation and it is now load-bearing for gemini's `manual` and devin's. It is documented at `GEMINI_MODES`, which is the right home, but CLAUDE.md's Conventions section asks that a rule other developers must know before touching the code also land in `CONTRIBUTING.md`. One line there would do it.
- `installCanvasSkillInto(devinConfigDir())` and `devinSkillPath()` create `~/.config/devin/` on machines that have never had Devin. That is the house pattern already (codex/gemini/copilot/opencode instruction files are written unconditionally too), so no change requested — noting it only so the choice is visible as a choice.

The `dialectFor` fix resolving through `capabilityAgentId` is a real improvement that outlives this PR: a custom agent with `baseAgent: 'devin' | 'gemini' | 'codex'` was previously falling back to claude's flag spelling. Good catch, and it is pinned.

Asherslab:
Both blockers closed in c7e356bb, plus the two nits that asked for a change.

**1. `hasPermWait` — intentional, now declared and pinned.**

Your read is the one I want: a custom agent declaring `baseAgent: 'claude'` runs claude's binary and therefore claude's hook script, so it can honour the decision reply, and resolving through `capabilityAgentId` is what every other capability here already does. The PR body describing `PERM_WAIT_CAPABLE` as a pure narrowing was wrong about this one edge, and the widening was undeclared — that part is fair.

It is now stated at the list itself, in `docs/devin-agent.md` checklist item 1, and pinned:

```ts
it('is inherited by a claude-based custom agent, and by no other base', () => {
  setCustomAgentBaseResolver((id) =>
    id === 'custom:proxy' ? 'claude' : id === 'custom:d' ? 'devin' : undefined
  )
  expect(hasPermWait('custom:proxy')).toBe(true)
  expect(hasPermWait('custom:d')).toBe(false)
  expect(hasPermWait('custom:plain')).toBe(false)
})
```

**2. `/exit` — measured, not deferred to the checklist.**

The CLI is installed here (devin-cli 3000.4.25, homebrew cask), and it ships its own docs. `share/devin/docs/reference/commands.mdx:439`:

> `/exit` — Exit the application (alias: `/quit`). You can also type `exit` or `quit` without the `/` prefix.

So `/exit` is devin's documented primary and `/quit` the alias — the opposite orientation to grok and gemini, which is the detail I should have written down the first time. It takes **no arguments**. Devin's destructive verb is a separate command, `/rm-session <session-id>` — *"Irreversibly delete a session and all its data"* (`commands.mdx:437`) — so the gemini `--delete` trap has no analogue on the exit line here, but the separation is exactly what makes it worth pinning rather than asserting:

```ts
expect(exitSequence('devin')).toBe('/exit')
expect(exitSequence('devin')).not.toContain('rm-session')
expect(exitSequence('devin')).not.toContain(' ')
```

Written up in `docs/devin-agent.md` §5.1 with the citation and with why the line count understates it — devin being in `RESUMABLE_AGENTS` plus this table is what makes its nodes eligible for the bulk restart and for Eco, both of which type the string into a live session. (`restartEligibility` still refuses `working` and `blocked`, so it never lands in an open permission prompt.)

**Nits**

- `withPermissionMode`'s doc comment now reads "grok's and devin's `--`", with the end-of-options consequence spelled out — rule 10 catching itself.
- The present-key-empty-value vs absent-key convention is now a house rule in `CONTRIBUTING.md`, naming both sides (gemini's `manual` vs gemini's `auto` / devin's `manual` and `plan`) and the reason it must be a key and not a value: same emitted command, opposite promise, and the Settings copy derives from the distinction.
- Took your suggestion on rule 11 as well — `docs/devin-agent.md` §8 now says in one paragraph that both raw listeners are unchanged *by construction* (no context tail, no session-directory derivation, no subagent correlation), so "no change in either shell" reads as the answer rather than as a gap.
- The `~/.config/devin/` creation nit: left as is, noted as a deliberate choice matching the codex/gemini/copilot/opencode pattern.

**Verification.** `npm run typecheck` clean; the three suites this touches are green (137/137). My full run is not comparable to yours — 384 failures, 390 of them `act(...) is not supported in production builds of React`, plus the `opencode.test.ts` / realtmux ones you already identified as local to my checkout. All in files this branch does not touch. Consistent with your green run, so I am treating it as my `node_modules`, not the branch, but flagging it rather than claiming a green suite I did not get.

eneskirca:
Re-verified `c7e356bb` in an isolated worktree: `npm run typecheck` clean, and the 66 suites it touches or gates (`src/shared/agents`, `src/core/agents`, `agent-restart`, the `no-electron` / `fs-atomic.guard` / `nodeterm-events` / `permissionMode` guards) are green — 1143/1143. Both blockers are closed the way the review asked: `hasPermWait` is declared at the list and pinned for a claude-based custom agent vs a devin-based / baseless one, and `/exit` is now measured against the CLI's bundled docs with the `rm-session` / whitespace pins beside gemini's `--delete` one. The three nits landed too. Nothing further on the content.

One mechanical thing before it can go in: **the branch needs a rebase.** `main` has moved 262 commits since your base, and a trial-merge now conflicts in three files — `agent-restart.ts`, `agent-restart.test.ts` and `CONTRIBUTING.md` — all adjacent-line additions. The one that is semantic rather than textual: main has since given **opencode** its own `/exit` in `EXIT_SEQUENCES`, so the `expect(exitSequence('opencode')).toBeNull()` line (and its comment) in your test must take main's side (`toBe('/exit')`). I also checked the agent tables main gained since your base (`account-binding.ts`, `account-color.ts`, `launch.ts`): they are claude/codex-scoped by design, so devin needs no entry there.

Also worth knowing: GitHub created **no CI run at all** for `c7e356bb` (the `56eeece2` runs are the last ones on the branch), so the rebase push is what will produce one — we'll approve it as soon as it appears.

howieyoungdahl (commented):
Read this one against the "Adding a new agent" checklist in `CLAUDE.md`, rule by rule, and traced
the parts that decide behaviour rather than taking the description. `git merge-tree` against current
`main` (`db39636e`) reports conflicts in 10 files — `CLAUDE.md`, `CONTRIBUTING.md`,
`context-link-render.{ts,test.ts}`, `context-link.ts`, `agent-restart.{ts,test.ts}`,
`approval-mode.ts`, `config.ts`, `launch.test.ts` — so a rebase is owed regardless of the notes
below. I did not run the suite.

## What I verified and agree with

- **The capability lists are joined with a named reason each, and the refusals are pinned as tests,
  not prose.** `devin capabilities` in `config.capabilities.test.ts` asserts the negative side —
  `canSubagent`, `canRecur`, `canBranch`, `hasUsage`, `canChat`, `canTransferFrom`, `canRename`,
  `canReadTitle`, `hasSharedIdentity` all false — which is the half that usually gets left out. Rule
  6's harder version ("if you cannot establish a trustworthy denominator, ship no meter") is honoured
  by keeping devin out of `USAGE_CAPABLE` rather than inventing a window.
- **Rule 2 was applied to the list this PR itself widened.** Joining `PERMISSION_MODE_CAPABLE` had
  pulled the hook-reply wait along with it; `PERM_WAIT_CAPABLE = ['claude']` + `hasPermWait` puts it
  back, `pty-manager.ts` reads the helper, and `managed-script.test.ts` gains devin to the
  "wait branch is ABSENT from a non-claude script at BUILD time" loop. That is the right shape:
  the narrowing is a list, not a special case at the call site.
- **The launch grammar is right at the composed layer, which is where rule 5 says the mistake
  lives.** `argvPromptSeparator: '--'` with the flag before the separator, and
  `config.capabilities.test.ts`'s "is the ONLY agent that asks for a separator" was rewritten to
  "one of the agents" with the remaining builtins still asserted `undefined`, rather than deleted.
- **Rule 15 is done better than the rule asks.** `EXIT_SEQUENCES.devin = '/exit'`, bare, cited to the
  CLI's own bundled `share/devin/docs/reference/commands.mdx:439`, and pinned three ways —
  `toBe('/exit')`, `not.toContain('rm-session')`, `not.toContain(' ')`. Generalising gemini's
  `--delete` trap to devin's separate `/rm-session <id>` verb is exactly the transfer that rule was
  written for.
- **Rule 7 holds.** `normalizeDevin` matches `ev === 'PermissionRequest'` — a closed compare, not
  grok's `includes('permission')` — and returns `null` for anything outside the seven subscribed
  events.
- **Rule 8's honesty test is passed in the harder direction.** `manual` and `plan` are declared
  UNSUPPORTED for devin rather than mapped to a nearest match, because devin's own default is
  `auto`; and the `modeSupported` refactor that makes that expressible (present-key-with-empty-value
  = supported-and-unflagged, absent = unsupported) does not change gemini or codex. I checked that
  rather than assuming it: `approvalFlags` returns `[]` for a falsy value
  (`approval-mode.ts:120-121`), `GEMINI_MODES` gains `manual: ''`, `CODEX_MODES` already carries
  `manual: 'untrusted'`, and the guards are tests **already on `main` that this PR leaves
  untouched** — `approval-mode.test.ts:33` (`approvalFlags('gemini','manual')` is `[]`) and `:66`
  (`modeSupported('gemini', m)` is `m !== 'auto'` for every mode). A refactor of that function that
  moved either answer would go red without anyone having to notice.
- **`dialectFor` resolving through `capabilityAgentId` is a real pre-existing bug fix.** Before it, a
  custom agent with `baseAgent:'gemini'` found no dialect and fell through to claude's spelling —
  `--permission-mode auto_edit`-shaped nonsense on a CLI that wanted `--approval-mode`. The new test
  pins all three bases.
- **Rule 11 is answered rather than skipped.** Neither raw listener changed, and
  `docs/devin-agent.md` §8 says why. I checked independently: no context tail (out of
  `USAGE_CAPABLE`), no subagent correlation (out of `SUBAGENT_CAPABLE`), no session-name leg (out of
  `TITLE_READ_CAPABLE`), and no hook-reported transcript path to ingest. "No change in both shells"
  is the correct parity here.
- **Rule 12 needs nothing, for the right reason.** Devin's envelope carries no `transcript_path`
  (`hook-events.ts`), so there is no path to jail; the locator derives one instead
  (`locateDevin`, `handoff/locate.ts`), and `resolveLinkTranscript` already refuses local locators
  for remote nodes.
- **Rule 16 and the two-docs convention are both satisfied**: `docs/devin-agent.md` with a six-item
  device checklist, plus CLAUDE.md and CONTRIBUTING.md updated split by audience — the
  empty-key-vs-absent-key paragraph in CONTRIBUTING is the right half to put there.

## What I would change

**1. `$XDG_CONFIG_HOME` — the local half is not in the checklist, and the failure is silent and
total.** `devinConfigDir()` (`src/core/agents/hooks/devin.ts`) resolves `%APPDATA%\devin` on Windows
but hardcodes `~/.config/devin` on POSIX, with no `$XDG_CONFIG_HOME` check. Directly next door,
`opencodeConfigDir()` (`src/core/agents/hooks/opencode.ts:14-21`) does the check and its comment
names the affected population by exactly this description: *"$XDG_CONFIG_HOME/opencode when the env
var is set (Linux/Server Edition users do this)"*. The same three hardcodes appear on the remote
side — `AGENT_TARGETS`' `.config/devin/config.json` and both `writeRemoteSkill` calls at
`${remoteHome}/.config/devin` — sitting a few lines from the helper this repo built for precisely
this problem, which expands `"${XDG_CONFIG_HOME:-$NT_H/.config}/opencode/AGENTS.md"` on the HOST and
is verified against real `/bin/sh`.

Why I would not defer it to the checklist: the failure mode is hooks written to a path the CLI never
reads, i.e. **a devin node that reports no agent status at all** — which is the bug the comment
immediately above `AGENT_TARGETS` exists to commemorate for remote gemini, and it is invisible
without a devin install to test on. The asymmetry also argues for it independently of measurement:
honouring `$XDG_CONFIG_HOME` when devin ignores it costs nothing (the fallback is the same path),
while not honouring it when devin respects it costs the whole integration. A tool that follows
`%APPDATA%` on Windows is following a config-convention library on POSIX too, more often than not.
Checklist item 2 names the remote half as an assumption; I would rather see both halves use the
mechanism than see the assumption recorded.

**2. `normalizeDevin`'s `PermissionRequest` branch reads a field nothing can set for devin, and it
drives a real button.** It spreads `...(p.nodeterm_pending_id ? { pendingId: … } : {})`, and
`pendingId` is what renders Approve/Deny on the node (`TerminalNode.tsx:4601`) and what the phone
mirror turns into an approval action (`agent-status-mirror.test.ts:404`). Today it cannot fire —
`PERM_WAIT_CAPABLE` is claude-only and the wait branch is absent from a devin script at build time,
which this PR pins — so this is not a bug. It is the one place in the diff where the code claims
more than `docs/devin-agent.md` §9 item 1 admits, and it is what the next person will cite when they
add `devin` to `PERM_WAIT_CAPABLE` without measuring the decision contract. One line in that branch
saying it is inert until item 1 is measured, or dropping the spread until then, would close it.

**3. The `devin capabilities` docblock asserts the opposite of the test underneath it.** It reads
*"Canvas control, context links and model switching are now enabled because the CLI surface is
there"*, immediately above `expect(canSwitchModel('devin')).toBe(false)`. Left over from the first
commit, and trivially fixed — but this repo's own rule is that a doc line no test backs is a plan,
and here a test backs the opposite. Same sentence would be worth a scan of `docs/devin-agent.md` for
the same leftover.

## Smaller notes (non-blocking)

- `installSkill()` (`src/core/context-link.ts`) now writes the get-linked-context SKILL.md into
  `devinConfigDir()/skills/` at every app launch, and `initCanvasControl` does the same for the
  canvas skill, on machines with no Devin installed. The claude path has always behaved that way so
  this is consistent, not new — just noting that the loop now creates a config tree for a product
  the user may not have.
- `devinConfigDir()`'s Windows branch is the only one of the six installers to have one, while the
  script it points at (`devin.sh`) is the shared POSIX managed script. Correct as groundwork, but
  worth one clause saying it is forward-looking rather than a supported path today.
- `controlRouting.test.ts` gains `copilot` alongside `devin` — a real drive-by fix, since that loop
  had been missing an agent that is in `CANVAS_CONTROL_CAPABLE`. Worth a line in the description so
  it is not mistaken for noise.
- The `Stop` branch destructures `p.last_assistant_message` into `lastMessage` while the comment
  three lines up says devin does not send it. Harmless and forward-looking; just reads oddly
  against a comment that says the opposite.

## Overall

The parts that are hardest to get right — the launch separator at the composed layer, the bare exit
sequence, the closed permission-event compare, the refusal to claim a context meter, and the
narrowing of the wait capability the PR's own first commit had widened — are all right, and the
negative assertions are what make that checkable rather than claimed. My ask before merge is (1);
(2) and (3) are one line each. The rebase against `main` is the other prerequisite.

**What I could not verify:** I have no `devin` install, so every measured claim in the PR body and
`docs/devin-agent.md` (the 3000.4.25 flag vocabulary, the hook payload shape, `devin skills paths`,
the transcript layout, and whether devin reads `$XDG_CONFIG_HOME`) is taken on the author's word. I
did not run the test suite or `npm run typecheck`. The claim about the pre-existing
`opencode.test.ts` failures on `main` is untested by me.

Asherslab:
Rebased onto `main` (`3411efdf`) and force-pushed — `c7e356bb` → `1c8d689c`. Two behaviour fixes ride along that you have not seen, both found by running the branch rather than reading it, so I want them called out rather than buried in a rebase.

## The rebase

Five conflicts, all adjacent-line except the one you flagged. Resolutions:

- **`agent-restart.ts` / `.test.ts`** — took main's side on opencode, as you said: `opencode: '/exit'` kept, and my `expect(exitSequence('opencode')).toBeNull()` line and its now-false comment are gone.
- **`config.ts`** — `CONTEXT_LINK_CAPABLE` is the union of main's grok (with its measured-on-1.0.13 note) and devin.
- **`context-link-render.ts` / `.test.ts`** — main gained a grok parser since my base; `linesFromDevin` sits beside it and both routing arms are present. The test file was add/add: main's real-fixture file kept whole, devin's describes appended.
- **`context-link.ts`** — `LINK_LOCATORS` gains grok and devin.
- **`approval-mode.ts`** — main's issue-#601 duplicate-flag paragraph kept; my devin edit folded into the separator sentence above it.
- **`CONTRIBUTING.md`** — main's four new house rules kept, mine appended.

I also checked the agent tables main gained since my base (`account-binding.ts`, `account-color.ts`, `launch.ts`) — claude/codex-scoped by design, devin needs no entry, matching your read.

## Fix 1 — `auto` mapped to Devin's `smart`, not its `auto` (`853b8fd5`)

Using the branch surfaced this immediately: every devin node started in the default mode however the setting was set. **Devin's `auto` is an alias for `normal`, its own default**, so the flag changed the command line and could not change the session.

The two CLIs mean different things by the word. Claude Code's own mode help:

```
default      — ask before every edit
accept edits — edit freely, ask for commands
plan         — research and propose, never touch files
auto         — Claude decides what is safe        ("use auto for long unattended tasks")
```

A model judges each action. Devin's `normal` judges nothing (reads auto-approved; every edit and shell command prompts). The devin mode that delegates the decision to a model is **`smart`** — workspace edits auto-approved, everything else judged by a fast model, high-risk categories (installs, `rm`, `sudo`, mutating git, dotenv and key material) always prompting.

This is not rule 8's substituted-nearest-match trap, it is the inverse: that trap is claiming a mode the CLI cannot express, whereas here the CLI expresses it exactly under a different name and the matching *name* was the wrong answer.

Measured on 3000.6.12 the flag starts the session in smart with no warning. On 3000.4.25 — which predates the launch-flag route, added in devin 3000.5.20 — the flag is accepted and devin prints its own fallback, exit 0, while `/smart` in that same session switched fine (so the account rollout was on and only the startup path lagged). No version gate added: the degrade is devin's own, announced in the pane, and lands on `normal`, which is where the session would have started anyway.

`plan` stays unsupported, and its reason is now measured instead of asserted — devin has an in-session `/plan`, but `--permission-mode plan` is **rejected by the parser** on both versions, and a rejected value kills the launch.

The commit also corrects a pre-existing comment in `config.ts` describing claude's `auto` as "approve most things but not edits". That description is what produced my wrong mapping. The gemini conclusion resting on it is unaffected — `auto_edit` is a blanket auto-approval of edits, not a judgement.

## Fix 2 — a delayed Enter for composers that debounce input (`1c8d689c`)

Text written into a devin session sat in its composer unsubmitted: the canvas-control `write` verb, the agent-message envelope, note pushes and the restart's `/exit`.

Devin's behaviour, not our delivery — measured on 3000.6.12, idle session:

| written by | Enter in the same burst | Enter ≥ ~80 ms later |
|---|---|---|
| bracketed paste (`paste-buffer -p`, our path) | stays in the composer | submitted |
| unframed paste | stays in the composer | submitted |
| `send-keys -l` literal | stays in the composer | submitted |

Threshold between 50 and 80 ms, and the identical bracketed-paste delivery into claude submits immediately (control test). Devin keys off timing, not the paste markers — the one case the `localFramedDelivery` tombstone's "a paste-aware composer cannot re-chunk into the paste" does not cover.

It mattered most for the envelope, which answers **`sent`** to an orchestrating agent: a swallowed submit was reported as a handoff that had started. The restart failed more kindly (`exit-timeout`, session left running).

**opencode had this first.** `agent-restart.ts` carried a hardcoded `agentId === 'opencode'` branch with the same 150 ms, measured on 1.18.18-1.18.25 — so only the restart path knew, and every other write into an opencode composer had the bug devin has. Devin arriving with identical behaviour is what showed it is a property of the agent, so the number moved to `AGENT_CONFIG.submitEnterDelayMs` (resolved through `capabilityAgentId`, so custom agents inherit) and every write path reads it. Both agents now get it in `sendText` and `sendEnvelope` across all three transports and in the restart's exit phase; every other agent answers `0` and keeps the single-invocation delivery byte for byte. `CONTRIBUTING.md` gains the general rule.

The split gives up something real, so it is re-established explicitly: one tmux command list makes "the Enter can never fire after a failed text send" structural, because tmux abandons the rest of a failed list. `runDelayedEnterDelivery` sends the Enter only when the paste answered true, and the Enter's own failure is the caller's answer — a submit firing into a composer that never got its text would send whatever the *user* had typed there. Three tests pin the order (injected clock), the refusal plus buffer sweep, and the late failure. `opts.agentId` is plumbed through the wire for the case core cannot answer: a node whose tmux session outlived the app has no live `Session` to read the agent id from.

Verified end to end against live devin with the exact delivery this produces — paste, 150 ms, separate Enter — and the turn runs.

## Verification

`npm run typecheck` clean. `src/shared/agents` + `tmux-naming` + `agent-restart` green. For the rest I diffed failing-test *names* against the same tree with my changes stashed, since this checkout has a broken React test env: **zero failures unique to these changes** — `src/core`+`src/shared`+`src/main` 17 vs 18 (the difference is a baseline-only `fs.watch` flake), `src/renderer` 64 vs 64 identical. The one `tmux-paste.realtmux` CONTROL failure is pre-existing here and asserts that the *old* broken behaviour still reproduces, which it does not on this host's tmux.

Both fixes are behaviour changes you signed off before seeing, so treat them as new review surface rather than rebase noise. CI should now produce a run for this head.

Asherslab:
PR should be up to date and ready for review/merge if desired

Issue #414, opened 2026-08-24
Reflect active tab/session name in the native window title (for OS-level time tracking)


kalus-msic:
Hey, first of all thanks for NodeTerm - switched to it last week for terminals + Claude Code/Codex sessions and it's been a great experience so far.

One thing I ran into: I use ActivityWatch to automatically track time per application/window (feeds into a Redmine time-logging workflow). ActivityWatch works by reading whatever title the OS reports for the currently focused window - that's how it distinguishes, say, "which file is open in VS Code" or "which tab is active in iTerm2/Windows Terminal", since those apps update the native window title to match the focused tab.

NodeTerm keeps one static native window title ("NodeTerm") regardless of which tab/node is focused - a terminal, a sticky note, an agent session, etc. Since everything lives inside a single OS-level window, tools like ActivityWatch (or any other window-title-based time tracker) only ever see "NodeTerm" and can't tell which terminal or session was actually active.

Would it be possible to add this as an optional setting? Something like "Show active tab name in window title", which when enabled would make NodeTerm update the native window title (BrowserWindow.setTitle()) to include the name/label of the currently focused node -  e.g. NodeTerm - <terminal/agent name>. That's basically what iTerm2, Windows Terminal, and kitty already do when you switch tabs, though of course not everyone would want that behavior by default, so keeping it opt-in would avoid changing the experience for people who don't need it.

This would make NodeTerm usable with automatic, window-title-based time trackers without any extra integration work - it'd just work the same way it does for regular terminal tabs elsewhere.

Happy to help test this if it's something you're open to adding.

eneskirca:
Thanks for the detailed writeup — and glad NodeTerm is working out for you!

This was a great request: the use case is real and the fix turned out clean, so it's implemented in #415. What you get:

- A new **opt-in** toggle in **Settings → Appearance → "Show active session in window title"** (off by default, exactly as you suggested — nothing changes for people who don't need it).
- When enabled, the native window title becomes **`<node title> — <project name> — node-terminal`** for whichever node was last active — driven by keyboard focus landing in a node (terminal, sticky, editor) or by selecting a single node. When focus moves to app chrome (tab bar, settings), the last session's title deliberately sticks, which should be the right behavior for a time tracker.
- If no node is known it falls back to `<project> — node-terminal`, and it also works in the browser-based Server Edition (tab title) for free.

[...]

PR #566, opened 2026-08-31
fix: restore terminal keyboard focus when the window is activated

### LINKED ISSUE #557: Restore terminal focus after switching back to nodeterm
On macOS, when I switch away from nodeterm with Cmd+Tab and then switch back, the terminal I was using often does not get keyboard focus back.

This is especially confusing with Claude Code and Codex sessions because their cursor keeps blinking. I understand that the cursor is rendered by the program running inside the terminal and does not necessarily mean that xterm has keyboard focus, but visually it looks like the terminal is ready for input.

A common sequence is:

1. Type something in a terminal.
2. Switch to another app with Cmd+Tab.
3. Switch back to nodeterm.
4. Press Backspace to remove the last character.
5. nodeterm opens the confirmation dialog for deleting the selected terminal node instead.

This makes it very easy to trigger a destructive canvas action when the user reasonably thinks they are typing into the terminal.

It would be helpful if nodeterm remembered the last focused terminal and restored keyboard focus when its window becomes active again. This should probably only happen when focus was previously inside a terminal and no dialog, settings field, editor, or other input currently needs focus.

I use nodeterm in a multi-monitor setup. When I switch back to the app, the mouse pointer is often still on another display, so the hover-based terminal focus cannot help. Restoring focus should therefore be based on the previously focused terminal and window activation, not on pointer position or a new hover event.

The default Backspace binding for “Delete selection” makes the problem more noticeable. It can be removed in the shortcut settings, but Backspace is such a common editing key that using it for deleting terminal nodes feels risky whenever terminal focus is uncertain.

Keeping Delete as the default, with Fn+Backspace as the equivalent shortcut on compact Mac keyboards, while leaving plain Backspace unbound might be a safer default.

Clicking the terminal restores focus, and disabling the Backspace binding avoids the deletion prompt, but both are workarounds for something that happens repeatedly during normal app switching.

basmilius:
Fixes #557.

## The bug

Terminal focus in this app is pointer-driven. `TerminalNode`'s hover guard focuses xterm on a dwell or a click (`onBodyEnter` / `onGuardUp`) and **blurs it again on `mouseleave`** (`onBodyLeave`). That is deliberate, because the guard exists so a pointer merely passing over a terminal does not steal the keyboard, but it means focus has no owner once the pointer is elsewhere.

On a multi-monitor desk the pointer routinely leaves the node on its way to another display, so by the time the user Cmd+Tabs away **the terminal has already lost focus**. Coming back, there is no pointer event left to hand it over: `document.activeElement` is `<body>`, the guard is armed again, and the next keystroke reaches the global dispatcher instead. There `canvas.deleteSelection` has a bare `Backspace` default (`shared/keybindings.ts`), so pressing Backspace to delete a character opens the **delete-node confirm** over a terminal the user believed they were typing into.

That is exactly the sequence the reporter describes, and it explains why hovering cannot rescue it: the pointer is on another screen.

## The fix

`useTerminalFocus` gains a `lastNodeId` memory, recorded on the three paths where a terminal actually takes the keyboard (`enterNow`, the dwell `enter`, `insertFiles`). It **deliberately survives a blur**, since the blur is the case being repaired and clearing it there would forget precisely the terminal we want back. Nothing clears it explicitly: a node that is gone or unmounted simply misses the request, the same fail-open the existing one-shot `request` already has.

On window activation the canvas asks for that node through the request `TerminalNode` already consumes, so the restore reuses the click path (`enterNow`) rather than adding a second way to focus a terminal.

## Refusals

The decision is the pure `nodeToRefocus` (`renderer/lib/focusRestore.ts`), so the interesting cases are testable without a window that can be activated. It hands back `null`, leaving focus exactly where the browser put it, when:

- **a modal is open** (`openDialogCount() > 0`). Every modal registers in the dialog stack, the kanban `CardModal` included, whose own `ModalTerminal` focuses itself. Stealing focus out from under one is worse than the bug.
- **a typing surface has focus** (input, textarea, contentEditable, Monaco). Chromium restores the previously focused element on activation, so this is the ordinary case of the user having left the app from a text field.
- **a terminal already has focus**, meaning the pointer never left and nothing was lost.

The DOM read is deferred by a macrotask rather than taken inline: Chromium restores its own focused element around window activation, and deciding before that has settled would read `<body>` and steal focus from the settings field the user actually left the app from.

## What this deliberately does not change

The `Backspace` default on `canvas.deleteSelection` stays. It already goes through a confirm dialog, and dropping it would change behavior for every user who never hit this; the reporter's own workaround (unbinding it in Settings > Keyboard Shortcuts) remains available for anyone who prefers `Delete` / `Fn+Backspace` only.

## Surfaces

- **Desktop**: the target.
- **Server Edition**: works as-is. Pure renderer code, no new `window.nodeTerminal` member, no bridge change.
- **Mobile companion**: N/A, no canvas and no pointer-driven focus guard.

## Testing

- New `renderer/lib/focusRestore.test.ts` (7 cases: restore, and each refusal).
- `renderer/state/terminalFocus.test.ts` gains a case pinning that consuming the one-shot request does **not** clear the memory the restore reads.
- `npm run typecheck` clean; full suite 9299 passed. The 18 remaining failures are pre-existing and environment-bound (verified against a stashed tree): BSD `chmod --`, an over-long unix socket path in `$TMPDIR`, real-tmux/real-sh suites, `localStorage` in the node env, and one flaky `fs.watch` timing test.

Manually verified on macOS with the reported sequence: type in a terminal, move the pointer off the node, Cmd+Tab out and back, press Backspace, and it now lands in the terminal. Also checked that an open dialog, a focused text field and an open card modal all keep their focus.

### DIFF
--- a/src/renderer/canvas/Canvas.tsx
+++ b/src/renderer/canvas/Canvas.tsx
@@ -7282,12 +7282,20 @@ export function Canvas() {
     const onWindowFocus = () => {
       if (timer) clearTimeout(timer)
       timer = setTimeout(() => {
+        // Only the nodes we can prove belong to the canvas on screen: `nodesRef` and
+        // `activeProjectId` do not turn over in one commit, and an id we cannot place must not
+        // become a request that waits for its project to come back.
+        const activeProjectId = useProjects.getState().activeProjectId
+        const liveIds = new Set(
+          nodesProjectIdRef.current === activeProjectId ? nodesRef.current.map((n) => n.id) : []
+        )
         const target = nodeToRefocus({
           lastNodeId: useTerminalFocus.getState().lastNodeId,
           activeElement: document.activeElement as unknown as ContextElement | null,
           openDialogs: openDialogCount(),
-          boardOpen: isKanbanOpen(useProjects.getState().activeProjectId),
-          settingsOpen: settingsOpenRef.current
+          boardOpen: isKanbanOpen(activeProjectId),
+          settingsOpen: settingsOpenRef.current,
+          liveIds
         })
         if (target) {
           useTerminalFocus.getState().request(target, { ack: false })
--- a/src/renderer/canvas/Canvas.tsx
+++ b/src/renderer/canvas/Canvas.tsx
@@ -7259,16 +7259,24 @@ export function Canvas() {
     []
   )
 
+  // The settings page is React state, and the listener below is registered once — so it reads the
+  // flag through a ref rather than taking a dep and re-registering on every open/close.
+  const settingsOpenRef = useRef(settingsOpen)
+  settingsOpenRef.current = settingsOpen
+
   // Cmd+Tab back into the app: hand the keyboard to the terminal that last had it (issue #557).
   // Terminal focus is pointer-driven (the hover guard blurs xterm on `mouseleave`), so a pointer
   // parked on a second display leaves the app with NO terminal focused, and returning gives the
   // canvas dispatcher every keystroke, where a bare Backspace is `canvas.deleteSelection`. The
-  // refusals (a modal is open, a text surface or a terminal already holds focus) are the pure
-  // `nodeToRefocus`; this effect is the DOM read plus the request.
+  // refusals (a modal, the board or the settings page is up, a text surface or a terminal already
+  // holds focus) are the pure `nodeToRefocus`; this effect is the DOM read plus the request.
   //
   // Deferred by a macrotask rather than read inline: Chromium restores its own previously focused
   // element around window activation, and deciding before it has settled would read `<body>` and
   // steal focus from the settings field the user actually left the app from.
+  //
+  // `{ ack: false }`: coming back to the window is not evidence that the user has SEEN the node,
+  // and the ACK reaches past this machine (the notch capsule and the paired phone's Live Activity).
   useEffect(() => {
     let timer: ReturnType<typeof setTimeout> | undefined
     const onWindowFocus = () => {
@@ -7277,9 +7285,13 @@ export function Canvas() {
         const target = nodeToRefocus({
           lastNodeId: useTerminalFocus.getState().lastNodeId,
           activeElement: document.activeElement as unknown as ContextElement | null,
-          openDialogs: openDialogCount()
+          openDialogs: openDialogCount(),
+          boardOpen: isKanbanOpen(useProjects.getState().activeProjectId),
+          settingsOpen: settingsOpenRef.current
         })
-        if (target) useTerminalFocus.getState().request(target)
+        if (target) {
+          useTerminalFocus.getState().request(target, { ack: false })
+        }
       }, 0)
     }
     window.addEventListener('focus', onWindowFocus)
--- a/src/renderer/canvas/Canvas.tsx
+++ b/src/renderer/canvas/Canvas.tsx
@@ -207,6 +207,8 @@ import {
 } from '../lib/globalKeybindings'
 import { isTerminalTarget, type ContextElement } from '../lib/keyContext'
 import { installTerminalFocusMirror } from '../lib/terminalFocusMirror'
+import { nodeToRefocus } from '../lib/focusRestore'
+import { openDialogCount } from '../components/dialog-stack'
 import {
   applyWindowTitle,
   composeWindowTitle,
@@ -7256,6 +7258,36 @@ export function Canvas() {
       }),
     []
   )
+
+  // Cmd+Tab back into the app: hand the keyboard to the terminal that last had it (issue #557).
+  // Terminal focus is pointer-driven (the hover guard blurs xterm on `mouseleave`), so a pointer
+  // parked on a second display leaves the app with NO terminal focused, and returning gives the
+  // canvas dispatcher every keystroke, where a bare Backspace is `canvas.deleteSelection`. The
+  // refusals (a modal is open, a text surface or a terminal already holds focus) are the pure
+  // `nodeToRefocus`; this effect is the DOM read plus the request.
+  //
+  // Deferred by a macrotask rather than read inline: Chromium restores its own previously focused
+  // element around window activation, and deciding before it has settled would read `<body>` and
+  // steal focus from the settings field the user actually left the app from.
+  useEffect(() => {
+    let timer: ReturnType<typeof setTimeout> | undefined
+    const onWindowFocus = () => {
+      if (timer) clearTimeout(timer)
+      timer = setTimeout(() => {
+        const target = nodeToRefocus({
+          lastNodeId: useTerminalFocus.getState().lastNodeId,
+          activeElement: document.activeElement as unknown as ContextElement | null,
+          openDialogs: openDialogCount()
+        })
+        if (target) useTerminalFocus.getState().request(target)
+      }, 0)
+    }
+    window.addEventListener('focus', onWindowFocus)
+    return () => {
+      if (timer) clearTimeout(timer)
+      window.removeEventListener('focus', onWindowFocus)
+    }
+  }, [])
 
   // Active session → native window title (issue #414, opt-in `settings.windowTitleActiveSession`):
   // lets window-title-based time trackers (ActivityWatch) tell sessions apart. Two latest-wins
--- a/src/renderer/lib/focusRestore.test.ts
+++ b/src/renderer/lib/focusRestore.test.ts
@@ -1,0 +1,49 @@
+import { describe, expect, it } from 'vitest'
+import { nodeToRefocus, type FocusRestoreState } from './focusRestore'
+import { XTERM_INPUT_CLASS } from './keyContext'
+
+const el = (tagName: string, cls: string[] = [], contentEditable = false) => ({
+  tagName,
+  isContentEditable: contentEditable,
+  classList: { contains: (name: string) => cls.includes(name) }
+})
+
+const state = (over: Partial<FocusRestoreState> = {}): FocusRestoreState => ({
+  lastNodeId: 'term-1',
+  activeElement: el('BODY'),
+  openDialogs: 0,
+  ...over
+})
+
+describe('nodeToRefocus', () => {
+  it('restores the last focused terminal when nothing else owns the keyboard', () => {
+    expect(nodeToRefocus(state())).toBe('term-1')
+  })
+
+  it('restores when focus sits on a non-typing element (a button, the canvas pane)', () => {
+    expect(nodeToRefocus(state({ activeElement: el('BUTTON') }))).toBe('term-1')
+  })
+
+  it('refuses without a remembered terminal', () => {
+    expect(nodeToRefocus(state({ lastNodeId: null }))).toBeNull()
+  })
+
+  it('refuses while a modal is open', () => {
+    expect(nodeToRefocus(state({ openDialogs: 1 }))).toBeNull()
+  })
+
+  it('refuses when a terminal already has focus', () => {
+    const term = el('TEXTAREA', [XTERM_INPUT_CLASS])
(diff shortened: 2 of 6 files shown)


eneskirca:
The diagnosis is right and the shape is the one I would have wanted: a pure decision (`nodeToRefocus`) plus a DOM read, reusing the one-shot `request` channel `TerminalNode` already consumes rather than adding a second way to focus a terminal. Deferring by a macrotask so Chromium's own activation-time focus restore settles first is a real detail, not decoration — reading inline would have read `<body>` and produced exactly the theft the refusals exist to prevent.

Two gaps in the refusal set, both of the same kind: `openDialogCount()` answers for things registered in the dialog stack, and the app has full-page surfaces that are not.

**1. The kanban board.** `KanbanView.tsx` does not import `dialog-stack` (only `CardModal` does). So with the board open — an *opaque* overlay over a still-mounted canvas — Cmd+Tab back hands the keyboard to a canvas terminal the user cannot see, and their next keystrokes go into a hidden pane. This is the same hazard `lib/zoomShortcut.ts` already refuses for ("refuses while the kanban board is up"), so there is precedent for the refusal and a store to read it from (`state/viewMode.ts`). The `CardModal` case is covered, but the board without a card open is not.

**2. The settings page** (and, less severely, any other non-dialog full-page overlay). `SettingsPage.tsx` is not in the dialog stack either. The typing-surface refusal saves the common case, but only once a field actually has focus — a freshly opened Settings page with nothing focused would hand the keyboard to a terminal underneath it.

Both are cheap to add as extra `FocusRestoreState` fields so they stay in the pure function and testable.

**3. A quieter one, worth a decision rather than a fix.** The restore goes through `enterNow`, which also does `setActive(id, true)` and `clearUnread(id)`. Per CLAUDE.md, clearing unread "ACKs the finish across phone/notch surfaces". So after this, merely Cmd+Tabbing back into the app — with the pointer on another display and no user gesture at all — acknowledges the last-focused agent's completion on the phone and clears its dot. That may well be what you want (the user *is* back at the machine), but it is a behavioural change to a cross-surface acknowledgement that the PR description does not mention, and it is not something the reporter asked for. If it is deliberate, saying so in the description is enough; if not, the restore could focus without the ACK.

Smaller notes:

- `lastNodeId` deliberately surviving a blur is correct and well argued. It also survives a **project switch**, so the remembered node can belong to another project. Today that fails open (the node is unmounted, the one-shot request goes unconsumed), which matches the existing contract — worth one line in the JSDoc so a future change that starts routing requests across projects knows this is load-bearing.
- `window.addEventListener('focus')` does not receive element focus (those do not bubble), so this only fires on real window activation. I checked because a bubbling variant would have made this fire constantly.
- Leaving the `Backspace` default on `canvas.deleteSelection` alone is the right call, and the reasoning for it belongs in the description exactly where you put it.

Good, well-argued fix. I would want (1) at minimum before merge; (3) is a question rather than an objection.

basmilius:
Thanks - both gaps are real, and I have pushed a commit for them.

**1. The kanban board.** Fixed. `KanbanView` registers nothing in the dialog stack, so `openDialogCount()` could never see it, and with the board up the restore handed the keyboard to a terminal behind an opaque overlay. `nodeToRefocus` now takes a `boardOpen` field, read at the call site through `isKanbanOpen(activeProjectId)` - the same store and the same helper `lib/zoomShortcut.ts` uses, for exactly the reason you point at.

**2. The settings page.** Fixed the same way, as a `settingsOpen` field. Worth noting it is the more clear-cut of the two: `SettingsPage` renders as `fixed inset-0` with an opaque background, so it is a genuine full-page surface, and the typing refusal only covers it once a field actually holds focus. Canvas reads its own `settingsOpen` state through a ref rather than adding a dep, so the listener still registers once.

Both live on `FocusRestoreState` rather than being folded into one `overlayOpen` boolean, so each refusal is named and separately tested. I added a line to the interface doc saying that a new full-page surface outside the dialog stack owes a field here, since the next one will not be found by the dialog count either.

On the other non-dialog surfaces I checked while I was in there: `ShortcutsPanel` and `LogPanel` sit on a `.sc-overlay` scrim and `SourceControlPanel` on `.drawer-overlay`. They are modal-shaped but unregistered, which is a pre-existing gap in the dialog stack rather than something this restore introduces; I left them alone rather than growing the field list past the two surfaces that are actually opaque full-page views. Registering those panels in the dialog stack looks like the right fix, but it is a separate change with its own Enter/Escape implications.

**3. The ACK.** You are right, and it was not deliberate - I reused `enterNow` for the focus and inherited the acknowledgement with it. It is now decoupled: the restore focuses without acking.

The reasoning: `clearUnread` calls `ackDone`, which resolves the notch capsule's blob and the paired phone's DONE Live Activity and Inbox card. Activating a window is not evidence that the user has seen anything - the pointer can be on another display and nothing was aimed at that node - and an acknowledgement that reaches past this machine needs better evidence than "the app is frontmost again". Every other caller of the request channel (sidebar click, notification tap, Cmd+K) *is* a human pointing at a node, and those keep acking.

Mechanically I kept the single request channel you liked and put the flag on the request: `request(nodeId, { ack?: boolean })`, defaulting to true, so every existing call site is byte-identical and only the activation restore passes false. The flag belongs to the caller because only the caller knows whether a human aimed at the node.

I deliberately kept the other two things `enterNow` does. `setActive(id, true)` sets one local `activeId` meaning "the keyboard is in this node", which is simply true after the restore - leaving it stale would make the window title (#414) name a different session than the one receiving keystrokes. `presence.reportFocus(id)` is a live "where I am working" cast, not a read-acknowledgement of a finished turn, and it is accurate for the same reason. Both are released again by the existing `onBodyLeave` path, so neither can get stuck. `clearUnread` was the only one of the three that resolves state on another device.

**Smaller notes.** Added the JSDoc on `lastNodeId` about surviving a project switch: it fails open today because the request is consumed by a mounted `TerminalNode` and an inactive project's nodes are unmounted, and the note says a change that starts routing focus requests across projects has to clear or qualify it - otherwise a window activation would pull the user out of the project they are looking at. Agreed on the `window.addEventListener('focus')` check and on leaving the `Backspace` default alone.

eneskirca:
Re-reviewed c2bb1cf in an isolated worktree, including a merge onto current main (the branch is ~250 commits behind but merges clean): typecheck clean, the `focusRestore` / `terminalFocus` / canvas / nodes suites green after the merge. The three points are answered, and carrying the ACK on the request (`{ ack: false }` only at the activation site) is the right shape.

**One thing to fix before merge, and it is the `lastNodeId` JSDoc's own claim.** A request aimed at an unmounted node does not fail open — it stays **latent**. `useTerminalFocus.nodeId` is only cleared by the consumer, and `TerminalNode`'s `lastFocusReqRef` starts at 0, so a node whose id matches consumes the request the moment it mounts. That is exactly how the cross-project `pendingFocusRef` path works today: `Canvas.tsx` ~L2195–2209 requests right after the project loads, before the node has mounted, and relies on the mount consuming it. So: the user leaves project A with terminal T focused (T is `lastNodeId`), switches to project B, Cmd+Tabs out and back with no terminal focused → `request(T, { ack: false })` sits in the store; the next time they switch back to A, T takes the keyboard, drops its hover guard and reports itself active — a focus nobody aimed at, minutes later. Smallest fix: refuse when the id is not among the active project's live nodes — a `liveIds` (or `mounted`) field on `FocusRestoreState` fed from `nodesRef.current`, so the refusal stays pure and tested — and reword the JSDoc. Clearing `lastNodeId` on project switch would also work but forgets the terminal for a plain switch-and-back.

Non-blocking: the restore goes through `enterNow`, which also runs `setArmed(false)` on a node the pointer is not over. The guard is a pointer-event overlay and keyboard focus does not need it down, so the next time the pointer enters that node the dwell is skipped and a wheel scrolls tmux instead of panning the canvas, until one mouseleave re-arms it. Focusing xterm while leaving the guard armed on the `ack: false` path would keep the hover contract intact; your call.

CI: the `quality-windows` failure is `agent-launch.test.ts › Git Bash executes the POSIX encoder` timing out at 5 s — a Windows-runner flake unrelated to this renderer-only diff (main had three other timeouts of the same class this morning). I re-ran the job.

basmilius:
Thanks, and you are right about `lastNodeId` - the JSDoc claim I added was simply wrong. I verified the mechanism rather than taking it on trust: `lastFocusReqRef` is `useRef(0)`, `focusReq` reads `s.nodeId === id ? s.nonce : 0`, and the store's `nodeId` is cleared in exactly one place, the consumer effect. So a mount with a matching stored request fires immediately, and `Canvas.tsx`'s cross-project `pendingFocusRef` path depends on precisely that: it calls `request(pending)` inside a `setTimeout(..., 0)` after the project load, before the node exists. Latent, not fail-open, and load-bearing for that path.

That is why I took your smallest fix rather than clearing `lastNodeId` on a project switch: the clearing version also forgets the terminal on a plain switch-and-back, and it would break the pending-focus path's reliance on latency if it ever grew into the store. `nodeToRefocus` now takes a `liveIds` set and refuses an id that is not on the canvas the activation returned to, with a test for your exact scenario.

One addition to the shape you suggested. `nodesRef` is only trustworthy while its epoch tag matches the active project - `nodesProjectIdRef` exists because the node array and `activeProjectId` do not turn over in one commit - so `liveIds` is empty unless they agree, and an empty set refuses. "We cannot tell which project's nodes these are" is not evidence that the node is live. The JSDoc now describes the latent behaviour and points at where the staleness is actually refused.

**The hover guard: taken, and measured first.** You are right that it is a pointer contract and keyboard focus does not need it down. I checked that focus still lands with the guard up rather than assuming it: `.term-hover-guard` is `position:absolute; inset:0; z-index:2` with no `pointer-events` or visibility handling, and a jsdom probe (textarea under a stacked overlay, programmatic `.focus()`) puts `document.activeElement` on the textarea. Hit-testing does not gate `.focus()`. I also confirmed nothing is stranded by leaving it armed: `onMouseEnter`/`onMouseLeave` are on `.term-node__body` with the guard as a child, so the dwell and re-arm still fire, and a click still routes through `onGuardUp` with the ack.

Rather than add a second option I gated it on the existing `ack`. Its doc already said the flag means "only the caller knows whether a human aimed at this node", and both consequences follow from that one fact: aimed drops the guard and marks the finish read, unaimed does neither. A `keepGuard` that is always `!ack` would be redundant state. There is a local `aimed` so the two reads say why.

Agreed on the `quality-windows` flake; thanks for re-running it.

eneskirca:
Re-verified `72157c6d` in an isolated worktree, merged onto current `origin/main` (clean). `npm run typecheck` clean on both projects; the renderer suites green after the merge — 203 files, 2626 tests, `focusRestore` / `terminalFocus` / canvas / nodes included.

**The `lastNodeId` finding is closed, and the epoch guard is the right addition.** Canvas builds `liveIds` as `nodesProjectIdRef.current === activeProjectId ? nodesRef.current.map(n => n.id) : []`, so "we cannot tell which project's nodes these are" refuses rather than guesses, and `nodeToRefocus` returns `null` for an id that is not on the canvas the activation returned to. `focusRestore.test.ts` pins both directions — my exact scenario ("refuses a remembered node that belongs to another project") and the unresolvable-nodes case. The JSDoc now describes the latent behaviour and points at where the staleness is actually refused, so the claim I objected to is gone rather than merely softened.

**The hover guard is taken too**, gated on the existing `ack` through the single `aimed` read in `enterNow`: aimed drops the guard and marks the finish read, unaimed does neither. Putting both consequences on the one flag rather than adding a `keepGuard` is the better shape, and the measurement you did first (a programmatic `.focus()` is not hit-tested, so focus lands with the guard up) is what makes it safe.

Merge-ready from my side. Merge is @eneskirca's.

PR #568, opened 2026-08-31
fix(terminal): verify both ends of the launch line, so a swallowed first character cannot submit

### LINKED ISSUE #556: Startup command loses its first character when the shell's rc file reads stdin (e.g. oh-my-zsh update prompt)
## What happens

When a node opens, nodeterm types the startup command into the fresh PTY and hits Enter
immediately, without waiting for the shell to actually reach its prompt. If anything in
the user's rc file reads from stdin during startup, it consumes the first character of
that command and the rest is executed as a broken command.

Reproducible with a stock oh-my-zsh install, which prompts for an update every 13 days
by default (`zstyle ':omz:update' mode prompt`):

```
claude --permission-mode auto --session-id d65bbada-...
[oh-my-zsh] Would you like to update? [Y/n]
[oh-my-zsh] You can update manually by running `omz update`
  ofeo-app  laude --permission-mode auto --session-id d65bbada-...
zsh: command not found: laude
```

The `[Y/n]` prompt reads exactly one character, so it swallows the `c`, answers itself
with an invalid response, and hands the remaining `laude --permission-mode ...` to zsh.

Exactly one character disappears, always the first one.

## Why it is not just an oh-my-zsh problem

oh-my-zsh is the most common trigger, but any interactive prompt during shell startup
does the same thing: nvm/asdf version prompts, direnv authorization questions, dotfile
managers asking to reload, corporate rc files asking for a VPN/2FA choice. All of them
land on a shell that is not yet at a prompt while nodeterm is already writing to it.

## Steps to reproduce

1. Install oh-my-zsh with default settings (no `zstyle ':omz:update'` in `~/.zshrc`).
2. Wait until the update check is due, or force it:
   `rm -f "${ZSH_CACHE_DIR:-$HOME/.oh-my-zsh/cache}/.zsh-update"`
3. Open a new agent node in nodeterm so it sends a startup command.
4. The first character of the command is eaten; the shell reports `command not found`.

A minimal rc-level repro without waiting for the update window - add to `~/.zshrc`:

```zsh
read -k 1 '?press a key: '
```

Then open a node: the startup command loses its first character the same way.

## Suggested fixes

- Wait for the shell to be ready before writing the startup command: send a bare newline
  first and wait for the prompt to be echoed back, or match a known prompt pattern /
  inject a sentinel (`PS1` marker, `printf '\e]133;A\a'` OSC 133 prompt marks) and only
  then send the command.
- As a cheaper mitigation, send a leading newline before the command so any pending
  single-char read gets consumed by that newline instead of by the command's first letter.

## Environment

- macOS (Darwin 27.0.0), zsh + oh-my-zsh + powerlevel10k
- Startup command: `claude --permission-mode auto --session-id <uuid>`

Closest existing issue I found is #47 (Enter swallowed as paste under a multiplexer),
but that is a different mechanism - here the shell itself consumes the character.

ezwep:
Closes #556.

### The reported diagnosis is not what happens

The report says nodeterm "types the startup command into the fresh PTY and hits Enter
immediately, without waiting for the shell to actually reach its prompt". That machinery does
exist — `command-delivery.ts` writes the line *without* Enter, waits for the shell to echo it,
and only then submits, with a Ctrl-U rewrite and three attempts. So the wait is there; the
verification is what lets this through.

### What actually happens

`echoedIntact` compared only the **tail**:

    return cleanedSoFar.includes(cmd.slice(-ECHO_TAIL_CHARS))   // last 24 chars

That was deliberate — the docstring says *"the head is polluted by the prompt"* — and it is
exactly blind to a lost **head**. An rc file that reads the same tty during startup swallows
the first character (oh-my-zsh's update prompt does this every 13 days by default). The shell
then echoes a line whose tail is fully intact, verification passes, Enter goes out, and the
shell runs the mangled remainder — matching the report's `laude --permission-mode auto`.

Verified in the shipped 0.3.4 bundle, not only on `main`.

### It is in two places

`src/session-host/session.ts` carries its own independent copy of the same check. Fixing only
the renderer would leave the Server Edition / SSH leg broken, so both are changed.

That leg also behaves differently in a way that shaped the fix: on retry exhaustion the
renderer submits unverified (fail-open), while the session-host clears the line and **refuses
the launch**.

### The fix

Both ends, each as its own substring match:

    return (
      cleanedSoFar.includes(cmd.slice(0, ECHO_EDGE_CHARS)) &&
      cleanedSoFar.includes(cmd.slice(-ECHO_EDGE_CHARS))
    )

A truncated head or tail fails; junk in the middle — or before the line — is tolerated.

I first tried requiring the whole command contiguously (`includes(cmd)`). That is too strict: a
ZLE redraw can splice printable prompt text into the echoed line, so the cleaned stream need
not contain the command in one piece. On the renderer that would only stall and fail open, but
on the session-host leg it would refuse the launch outright — a cosmetic false negative turned
into a node that never starts. Two edge windows keep the detection without that.

### Tests

Three added, and each was counter-proved against the variant it guards:

| change under test | result |
|---|---|
| revert to tail-only | the two #556 tests fail |
| require the whole command contiguously | the interleaving-tolerance test fails |
| head + tail (this PR) | 16/16 pass |

All 13 pre-existing delivery tests pass unchanged in every variant — the tail-only window was
not protecting a documented case.

`npm run typecheck` clean. Full suite: 9304 passed, 8 failed in 4 files — identical to the
baseline on `main` in my environment (I stashed the change and re-ran the same files to
confirm). Those 8 spawn a real `sh`/`tmux` or a socket and fail under my agent shell's sandbox,
not from this change.

### Open question for you

This doubles the number of windows that must match, so every echo-fidelity risk that already
applied to the tail now also applies to the head: a redraw splicing text inside a window, tab
or control-char expansion, or NFC/NFD differences in an edge. One mitigation is that a ZLE
re-wrap splices at the wrap column — deep into a long line — so the head window is probably
less exposed than the tail already was.

`ECHO_EDGE_CHARS` is left at the existing 24 for both ends, but that number is now doing two
jobs. A smaller head window would shrink the interleave exposure while still catching a lost
leading character; it would also raise the chance of a coincidental match against prompt text.
I did not want to pick that trade-off silently — happy to change it to whatever you measure.

### Not covered

The session-host leg has no unit coverage for launch delivery (the check is inline in a class
method driven by a real pty), so its change is verified by reading, not by a test. And the
line-based variant of the rc-file race — an rc `read` that consumes until a newline rather
than a single key — is untested; the reported shape swallows exactly one character.

### DIFF
--- a/src/renderer/terminal/command-delivery.test.ts
+++ b/src/renderer/terminal/command-delivery.test.ts
@@ -1,6 +1,7 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import {
   DELIVERY_ATTEMPTS,
+  KILL_LINE,
   VERIFY_TIMEOUT_MS,
   cleanEcho,
   deliverCommand,
@@ -64,6 +65,15 @@ describe('echoedIntact', () => {
   it('does not match a truncated echo (flush ate the tail)', () => {
     expect(echoedIntact(CMD.slice(0, -6), CMD)).toBe(false)
   })
+  it('does not match an echo that lost its head (rc file read ate the first chars)', () => {
+    expect(echoedIntact(`% ${CMD.slice(1)}`, CMD)).toBe(false)
+  })
+  it('tolerates redraw text interleaved mid-line — the command need not be contiguous', () => {
+    // A ZLE re-wrap can splice printable prompt text into the echoed line. Both ENDS are still
+    // there, so this must verify: on the session-host leg a false negative refuses the launch.
+    const mid = Math.floor(CMD.length / 2)
+    expect(echoedIntact(`% ${CMD.slice(0, mid)}% ${CMD.slice(mid)}`, CMD)).toBe(true)
+  })
 })
 
 describe('deliverCommand', () => {
@@ -78,6 +88,19 @@ describe('deliverCommand', () => {
     f.emit('\x1b[32m% \x1b[0m' + CMD.slice(0, 40) + '\r\n')
     f.emit(CMD.slice(40))
     expect(f.writes).toEqual([CMD, '\r'])
+  })
+
+  it('does not submit a line whose head the shell never echoed (#556)', () => {
+    const f = fakeIo()
+    deliverCommand(f.io, CMD)
+    // An rc file reading the same tty during startup (oh-my-zsh's update prompt) swallows the
+    // first character, so the echo is missing its head while the TAIL is fully intact.
+    f.emit(`% ${CMD.slice(1)}`)
+    expect(f.writes).toEqual([CMD]) // no '\r': submitting here runs a mangled command
+    // The rewrite lands after the rc prompt consumed its answer, and then verifies.
+    vi.advanceTimersByTime(VERIFY_TIMEOUT_MS)
+    f.emit(`% ${CMD}`)
+    expect(f.writes).toEqual([CMD, KILL_LINE, CMD, '\r'])
   })
 
   it('kills the line and rewrites when the echo never completes, then succeeds', () => {
--- a/src/renderer/terminal/command-delivery.ts
+++ b/src/renderer/terminal/command-delivery.ts
@@ -3,16 +3,13 @@
 // rc/ZLE setup resets the tty with a FLUSH that can eat part of the queued line, and a
 // mangled line submitted anyway strands the shell at `quote>` (field report: 3 spawned
 // team agents, none started, each needed a manual `'` + Enter). So: write WITHOUT Enter,
-// wait until the shell has echoed the tail of the line back, THEN submit. A verify timeout
+// wait until the shell has echoed both ENDS of the line back, THEN submit. A verify timeout
 // kills the pending line (Ctrl-U) and rewrites; the LAST attempt submits unverified —
 // fail-open, a terminal whose echo we can't recognize must never block the launch (that
 // worst case is exactly the pre-fix behavior).
 
 export const VERIFY_TIMEOUT_MS = 2000
 export const DELIVERY_ATTEMPTS = 3
-/** Long enough to be unambiguous in the echo stream, short enough that a ZLE wrap/redraw
- *  sequence interleaved mid-line rarely lands inside the matched window. */
-export const ECHO_TAIL_CHARS = 24
 /** Ctrl-U — clear the pending input line before a rewrite. Exported because the in-place restart
  *  choreography clears the line the same way before typing its exit command (agent-restart.ts). */
 export const KILL_LINE = '\x15'
@@ -28,9 +25,33 @@ export function cleanEcho(chunk: string): string {
   return chunk.replace(ESC_SEQ, '').replace(/[\r\n]/g, '')
 }
 
-/** Has the shell echoed the full line? Tail-match: the head is polluted by the prompt. */
+/** Long enough to be unambiguous in the echo stream, short enough that a ZLE wrap/redraw
+ *  sequence interleaved mid-line rarely lands inside either matched window. */
+export const ECHO_EDGE_CHARS = 24
+
+/**
+ * Has the shell echoed the line? BOTH ENDS must appear, each as its own substring match.
+ *
+ * Tail-only (the original) was blind to a lost HEAD: an rc file that reads the same tty during
+ * startup swallows leading characters — oh-my-zsh's update prompt does exactly this — and the
+ * shell then echoes a line whose tail is fully intact, so verification passed and Enter
+ * submitted a mangled command the shell ran as garbage (#556).
+ *
+ * Requiring the WHOLE command contiguously would catch that, but it is too strict: a ZLE
+ * redraw can interleave printable prompt text inside the echoed line, so the cleaned stream
+ * need not contain the command in one piece. That matters asymmetrically — the renderer would
+ * merely stall and fail open, but the session-host leg clears the line and REFUSES the launch,
+ * turning a cosmetic false negative into a node that never starts.
+ *
+ * Two edge windows keep both properties: a truncated head or tail fails, while junk in the
+ * middle (or before the line) is tolerated. Commands shorter than 2×ECHO_EDGE_CHARS simply
+ * compare as a whole, which is the strictest form and safe at that length.
+ */
 export function echoedIntact(cleanedSoFar: string, cmd: string): boolean {
-  return cleanedSoFar.includes(cmd.slice(-ECHO_TAIL_CHARS))
+  return (
+    cleanedSoFar.includes(cmd.slice(0, ECHO_EDGE_CHARS)) &&
+    cleanedSoFar.includes(cmd.slice(-ECHO_EDGE_CHARS))
+  )
 }
 
 export interface DeliveryIo {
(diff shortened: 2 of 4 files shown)


eneskirca:
Reviewed alongside ezwep's Windows set (#573, #576, #577, #579), in an isolated worktree off
`origin/main`.

**Verified locally:** typecheck clean; `command-delivery.test.ts` 16/16.

**I re-ran your counter-proof rather than taking it.** Replacing the head condition with `true` in
`echoedIntact` and re-running the file:

```
× does not match an echo that lost its head (rc file read ate the first chars)
× does not submit a line whose head the shell never echoed (#556)
Tests  2 failed | 14 passed (16)
```

Exactly the two new tests, nothing else — so the head window is load-bearing and the 13 pre-existing
delivery tests really were not protecting the tail-only shape. That is the mutation evidence this
kind of change needs, and it holds.

The diagnosis correction is right and worth having on the record: the report blames "types and hits
Enter immediately", the machinery to wait already existed, and the defect was in what "verified"
meant. Rejecting `includes(cmd)` for the ZLE-splice reason — and noting the asymmetry that makes a
false negative cheap on one leg and fatal on the other — is exactly the right way round.

---

### The finding: the session-host leg's echo buffer is bounded, and that bound was sized for a tail

`src/session-host/session.ts:583`:

```ts
echoed = (echoed + cleanEcho(chunk)).slice(-Math.max(command.length + 256, 512))
```

A **tail** match is safe under that truncation by construction — the tail is what the window keeps. A
**head** match is not. Once more than the window's worth of bytes has arrived after the head, the head
is evicted and can never be seen again, so the two substrings need never coexist in the buffer even
though both were genuinely echoed. The check then fails, all three attempts fail, and this leg does
not fail open — `session.ts:568-572` writes `KILL_LINE` and `finish(new Error('session-host could not
verify launch command delivery'))`, i.e. **the node never starts**. Which is precisely the outcome the
PR body says it rejected `includes(cmd)` to avoid; the bounded buffer reintroduces a narrower version
of it through the back door.

How narrow? For a typical launch line (~75 chars) the window is 512. So it takes ~512 visible
characters between the head and the tail. A single ZLE full-line redraw with a long prompt
(powerlevel10k with a right prompt, a couple of repaints) is in that range once ESC sequences are
stripped and `cleanEcho` has removed the newlines that would otherwise have bounded it. Not the common
case — but the common case is not what this leg's refusal is for.

The renderer half is immune: `echoed += cleanEcho(chunk)` is unbounded there, which is why the tests
cannot see this.

Cheapest fix that keeps the bound — make the head **sticky**, so eviction cannot unsee it:

```ts
sawHead ||= (echoed + cleaned).includes(command.slice(0, ECHO_EDGE_CHARS))
echoed = (echoed + cleaned).slice(-Math.max(command.length + 256, 512))
if (sawHead && echoed.includes(command.slice(-ECHO_EDGE_CHARS))) submit()
```

with `sawHead = false` reset in `tryOnce()` beside `echoed = ''`. Same memory bound, same semantics,
no eviction hazard. (Widening the window instead would work too but leaves the failure mode, just
further away.)

### Your open question on `ECHO_EDGE_CHARS`

I would leave it at 24 for both ends and not add a second constant. The asymmetry you are worried
about is real but small: the head window is adjacent to the prompt, so a coincidental match is more
plausible there than at the tail. But the strings being matched are launch command lines
(`claude --permission-mode auto --sess…`), and I cannot construct a prompt that contains 24 contiguous
characters of one. A smaller head window would trade a hazard I can describe for one I cannot, and
that is the wrong direction. If you ever do split them, the head is the one that should stay large.

### On "not covered"

The session-host leg having no unit coverage is a real gap but not one to fix here — the check is
inline in a class method driven by a real pty, and extracting it would be a bigger change than the fix.
If you take the sticky-head suggestion, that *is* extractable as a tiny pure helper shared with the
renderer, which would close both this and the CLAUDE.md "a duplicated rule drifts" concern that this
second copy raises. Worth considering, not worth blocking on.

---

### Merge-order note (why this is in the Windows set even though it isn't a Windows PR)

**This should land before #579.** Per #575, the session-host bundle has never been built into or
packaged with any release — so `src/session-host/session.ts` is **dead code in every shipped build
today**, and its half of this fix currently protects nobody. #579 is what makes it live. If #579 goes
first, the first Windows build that actually has a session host ships with the tail-only check this PR
replaces.

**Verdict: approve the approach and the renderer half as-is; I would take the sticky-head change on
the session-host leg before merging, given that leg refuses rather than fails open. Merge before
#579.**

ezwep:
Applied in 565a67e8: the head is latched before the window is trimmed, and reset per attempt beside
`echoed`. Your version, unchanged.

**I revisited the coverage question, and the existing setup can cover this without an extraction.**
`session.test.ts` already drives `HostSession` through a fake `IPty` rather than a real pty, so the
delivery path is reachable as it stands. Two tests:

- head → 700 visible characters → tail **submits**.
- a tail-only echo **does not** submit. That covers the case where the latch must not start true,
  and would catch a future change that initialises it that way.

For the first one I reverted the fix and raised the test timeout, so that a slow failure would not
be cut short by vitest before the code had finished. It fails after **6010 ms** — three attempts at
2000 ms — and the failure is `Error: session-host could not verify launch command delivery`, i.e.
the refusal itself rather than a timed-out test. So the outcome you predicted is observed, not only
reasoned.

`tsc -p tsconfig.node.json` passes cleanly, and all 20 tests in `session.test.ts` pass.

**`ECHO_EDGE_CHARS`.** Agreed — 24 at both ends, no second constant. Your point that a smaller head
window would trade a hazard you can describe for one you could not construct is the argument I was
missing when I raised the question.

**On extracting the shared helper.** I checked the mechanics rather than guessing: esbuild does
resolve `@shared/*` through `tsconfig.node.json` in a session-host bundle (probe file, bundled in
7 ms). So there is no bundler obstacle, and my only argument is scope.

`src/session-host/` currently imports nothing outside its own directory — only `./`-relative
siblings, node builtins and npm packages. The extraction would make the renderer's command-delivery
its first cross-boundary import, landing in the bundle that by your account has never been packaged
into a release until #579. It is also not needed for this fix.

My preference is to keep this PR scoped to the fix and send the extraction as a follow-up once #579
has landed. If you would rather have it here, say so and I will add it.

**Merge order.** Agreed — before #579.

eneskirca:
**Verdict: merge-ready** (before #579, as noted).

Re-verified on Linux in an isolated worktree off `origin/main`:

- The sticky-head change (`565a67e8`) is applied exactly as suggested — `sawHead` latched on `merged` before the window is trimmed, reset per attempt beside `echoed = ''`. Same memory bound, no eviction hazard, and the session-host leg no longer refuses a launch whose genuinely-echoed head scrolled out of the bounded buffer.
- Both ends are fixed (renderer `command-delivery.ts` + `session-host/session.ts`), so the two-shell parity holds.
- The two new `session.test.ts` cases drive `HostSession` through the fake `IPty`: head → 700 chars → tail submits, tail-only does not — which also pins that the latch doesn't start true.

Locally: `npm run typecheck` clean; `command-delivery.test.ts` + the `session-host` suites = 106 passed / 4 skipped. Agreed on leaving `ECHO_EDGE_CHARS` at 24 for both ends, and on keeping the shared-helper extraction as a follow-up after #579 lands rather than making the session-host bundle take its first cross-boundary import here.

**Merge order confirmed: #568 before #579** — `session.ts`'s launch-verify is dead code in every packaged build today, and #579 is what revives it, so this must land first or the first real Windows session host ships with the tail-only check this replaces.

PR #570, opened 2026-08-31
fix(terminal): re-derive DOM renderer row spacing on reattach

basmilius:
### What

Switching between projects left every terminal's characters a full cell too far apart for roughly 150 ms, after which the text snapped back to normal.

### Why

xterm's DOM renderer does not lay glyphs out on a grid: it pins the advance with CSS `letter-spacing`, derived once per renderer instance as `cell.width - widthCache('W')` (`DomRenderer._setDefaultSpacing`). That width cache measures with `offsetWidth`, which is **0** for an element that is not in the rendered tree.

`WebglAddon.dispose()` is also the put-xterm-back-on-a-DOM-renderer path, and it runs from the lifecycle effect's cleanup, i.e. after React has detached the node's DOM. Parking a terminal therefore builds its DOM renderer against a 0-wide `W` and bakes a whole extra cell into the spacing; the adopting mount paints that until the WebGL grant lands (`WEBGL_ACQUIRE_DEBOUNCE_MS`, 150 ms), which is exactly the split second users see. Focus mode's `display:none` wrapper reclaims hidden grants the same way, so it has the same shape.

Nothing in the reattach path healed it: xterm re-derives the spacing on a char-size, dpr, or options change, and **not** on a resize.

### How

New `src/renderer/terminal/dom-renderer-spacing.ts` exporting `resyncDomRendererSpacing(term)`, called from `applyFit` in `TerminalNode.tsx`.

- **Change-gated.** It runs on every fit, so the comparison costs one cached width-cache read, and a terminal whose spacing already agrees is left completely alone.
- **It refuses to guess.** A measurement of 0 means the element still is not rendered, so the helper bails instead of re-baking the same wrong number.
- **DOM renderer only.** Only that renderer carries a width cache, so the WebGL renderer and the shared glyph addon bail on the first gate.
- **Guarded internals, fail-open**, in the same style as `quantizeCharSize` and `restoreDomRenderer`: a future xterm that renames a field silently keeps its stock behavior.

`applyFit` is deliberately the single call site: it covers adopt (park), leaving focus mode, closing the kanban board, and resize in one place.

### Surfaces

- **Desktop:** the fix.
- **Server Edition:** identical. This is pure renderer code with no new bridge member.
- **Kanban card modal:** N/A. `ModalTerminal` creates a fresh xterm per open against an attached element and deliberately has no WebGL budget, so its renderer is never rebuilt while detached.
- **Mobile:** N/A, no xterm.

### Testing

- `src/renderer/terminal/dom-renderer-spacing.test.ts`, 6 cases: recovery from the detached measurement, the no-op when spacing already agrees, the bail while the measurement is still 0, the bail on a renderer without a width cache, fail-open on missing internals, and the bail when the cell width is not known yet.
- `npm run typecheck` green, `src/renderer/terminal` 481 tests green.
- Manual: switch back and forth between two projects with open terminals, and enter/leave focus mode with terminals around the focused node.


eneskirca:
I traced this through the vendored xterm 5.5.0 source, and the diagnosis is correct in every detail that matters.

- `DomRenderer._setDefaultSpacing` really is `this.dimensions.css.cell.width - this._widthCache.get('W', false, false)`, applied both to `_rowContainer.style.letterSpacing` and to `_rowFactory.defaultSpacing` (`browser/renderer/dom/DomRenderer.ts:286-290`).
- `WidthCache._measure` is `el.offsetWidth / REPEAT`, which is 0 for an element not in the rendered tree.
- The recompute triggers are exactly `handleCharSizeChanged` (L318), `handleDevicePixelRatioChange` (L293) and the options-change path (L421). `handleResize` (L312) does **not** call `_setDefaultSpacing`, so the claim that nothing in the reattach path heals it holds.

One thing I went looking for specifically, because it would have made the fix a no-op, and it is fine: `WidthCache.get` only writes the flat cache `if (width > 0)`, so a measurement taken while detached is **not** cached. The guard read therefore re-measures on each attempt rather than returning a poisoned 0 forever, which is what lets the helper recover once the element is back. That is load-bearing and invisible from the diff — it may be worth one line in the JSDoc, since an xterm release that started caching 0 would silently turn this into a permanent bail.

`_core._renderService._renderer.value` is right (`RenderService._renderer` is a `MutableDisposable`), and it matches the access style already in `glyphgrid-attach.ts` and `TerminalNode`'s `restoreDomRenderer`, so no new precedent is being set.

Two small notes:

- **The bail on `measured === 0` costs a forced layout each time it fires.** Once the value is non-zero it is cached and the check is a flat-array read, as the doc says. But in the detached/`display:none` window — which is precisely when `applyFit` may be running from a `ResizeObserver` — every call sets `textContent` to `'W'.repeat(32)` and reads `offsetWidth`, i.e. a synchronous reflow per fit. Probably immaterial at the frequency `applyFit` actually runs, but it is the opposite of the "one cached read" the comment promises, and worth a sentence so nobody later moves the call somewhere hotter.
- `SPACING_EPS` at `0.01` against a value that is `cellWidth - measured`: at typical cell widths the letter-spacing deltas are ~0.4px, so the epsilon is comfortably below the signal and above float noise. No concern, just noting I checked.

The `'shared'` glyphgrid renderer and the WebGL addon both bail on the `_widthCache` gate, which is the correct first gate — worth keeping that ordering if the guards are ever rearranged.

Good change: narrow, correctly gated, fails open, and the test suite pins the *mechanism* (recovery, the still-0 bail, the missing-cache bail) rather than a screenshot. The CLAUDE.md paragraph is in the right place and in the file's own voice.

Issue #565, opened 2026-08-31
tmux-hint splits PATH on ':', which shreds Windows paths


ezwep:
## What happens

`src/core/tmux-hint.ts:158`:

```ts
const dirs = [...(env.PATH ? env.PATH.split(':') : []), ...COMMON_BIN_DIRS]
```

On Windows the `PATH` separator is `;`, and the entries themselves contain a `:` (`C:\...`). Splitting
on `:` therefore yields fragments like `C` and `\Program Files\...` rather than directories.
`COMMON_BIN_DIRS` (`tmux-hint.ts:55`) is POSIX-only as well.

Notably, `src/core/exec-path.ts:72` in the same codebase gets this right with `path.delimiter`, so
this looks like an oversight rather than a deliberate choice.

**Expected** — `path.delimiter`, so the PATH is split correctly on every platform.
**Actual** — meaningless fragments on Windows.

## Suggested fix

`env.PATH.split(path.delimiter)`, and an empty `COMMON_BIN_DIRS` on win32.

---

Windows 11 Pro 10.0.26200 · nodeterm 0.3.4 (unsigned Windows beta) · code references against
`main` @ 4b6cfd9

ezwep:
Correcting my own severity here, having gone through the call graph while preparing a fix.

`findCommand` has exactly one caller: `pty-manager.ts:1748`, and only as the `hasCommand` callback
handed to `tmuxInstall(process.platform, ...)`. `tmuxInstall` returns `null` for anything that is
not `darwin` or `linux` before the callback is ever invoked, and the callback is lazy. So on Windows
**`findCommand` is never called** — the `:` split cannot misfire there in practice.

That makes this a latent correctness issue rather than an observable Windows bug: the code is wrong
about `path.delimiter`, but nothing on Windows reaches it. I also found that "fixing" it is not free —
`tmux-hint.test.ts:49` asserts against `PATH: '/usr/bin:/bin'`, which stops splitting the moment the
separator becomes platform-dependent, so a correct fix has to take the test with it.

I have therefore left it out of the PR I am opening for the executable-resolution bugs, and I would
understand entirely if you want to close this. Flagging it rather than quietly dropping it, since my
original report implied a live failure.

eneskirca:
Confirmed on current `main` — `src/core/tmux-hint.ts:158` is still `env.PATH.split(':')`, and `COMMON_BIN_DIRS` (`:55`) is seven POSIX paths.

[...]

Issue #559, opened 2026-08-31
Windows: executable lookup never appends PATHEXT, so every CLI probe fails and agent nodes come back without --resume


ezwep:
## What happens

On Windows, closing nodeterm and starting it again brings every agent node back — but with an empty
conversation. The nodes relaunch their CLI *fresh* instead of `claude --resume <session-id>`.

The cold-restore code that should do this is present and correct. It just never receives a session
id. The reason is one line of executable lookup, and the same line silently disables several other
features.

## Root cause

`src/core/exec-path.ts:71-83` — `findInPathString()` resolves a binary with `path.join(dir, bin)`
plus `fs.accessSync(..., X_OK)`. It never tries a `PATHEXT` extension. On Windows no executable is
named `gh`, `ssh` or `claude`; they are `gh.exe`, `ssh.exe`, `claude.exe`. (`X_OK` is equivalent to
`F_OK` on Windows, so it is no safety net either.)

Measured on a Windows 11 machine:

```
claude -> ...\claude.exe    extensionless path exists: False
gh     -> ...\gh.exe        extensionless path exists: False
ssh    -> ...\ssh.exe       extensionless path exists: False
```

`findExecutableSync()` then falls back to the caller's well-known locations, and those are POSIX-only
everywhere (`/usr/bin/...`, `/opt/homebrew/...`). None exist on Windows. The result is `null`, always.

## The chain to the missing `--resume`

1. `src/core/claude-cli.ts:51` — the capability probe calls `findInLoginPath('claude')`, gets `null`,
   and returns `UNKNOWN_CLAUDE_CLI_CAPS` **without ever running the CLI**. So `sessionIdFlag: false`
   and `version: null`.
2. `src/renderer/state/workspace.ts:611-612` —
   `const mintedSessionId = sessionIdFlagSupported ? uuid() : undefined`. No capability, no id.
3. `workspace.ts:680` — `...(mintedSessionId ? { agentSessionId: mintedSessionId } : {})`, so nothing
   is persisted on the node.
4. `src/renderer/nodes/TerminalNode.tsx:3073` — on cold restore
   `priorId = st?.sessionId || data.agentSessionId`. Both empty, so `assembleResumeCommand` builds a
   command with no `--resume`.

The comment at `workspace.ts:600` says the minted id exists *"so this node always has something to
resume with"*. That is exactly the safety net that disappears on Windows.

## Evidence from disk

[...]

PR #573, opened 2026-08-31
fix(core): resolve executables through PATHEXT on Windows, and stop hardcoding gh's path

### LINKED ISSUE #560: GitHub features are dead on Windows: gh is looked up only at three hardcoded POSIX paths
## What happens

`src/core/git-service.ts:24-35`:

```ts
function findBin(names: string[]): string | null {
  for (const c of names) { try { if (fs.existsSync(c)) return c } catch {} }
  return null
}
const GH_PATH = findBin(['/opt/homebrew/bin/gh', '/usr/local/bin/gh', '/usr/bin/gh'])
```

This lookup **never consults `PATH` at all** — only three absolute Unix paths. On Windows `GH_PATH`
is therefore guaranteed `null`, independently of the `PATHEXT` problem filed separately. It is also
a module-level `const`, so the answer is fixed when the module loads.

Consequences in the same file:

- `:64`, `:91` — early `return false`
- `:360`, `:461` — `ghAvailable: false` to the UI
- `:798` — `{ ok: false, message: 'GitHub CLI (gh) not found.' }`

**Expected** — `gh` is found on `PATH`. On this machine it is installed at
`C:\Program Files\GitHub CLI\gh.exe` and authenticated (`gh auth status` reports a logged-in
account).
**Actual** — the app reports that the GitHub CLI cannot be found, and every GitHub action is
unavailable.

**Frequency** — always, on every Windows install.

## Why it hurts

The message points the user the wrong way: they go install something that is already there and
working.

## Suggested fix

Use `findExecutableSync('gh', [...])` the way the ssh lookups do (with the `PATHEXT` fix), and make
the lookup lazy instead of a module-level const so a `gh` installed after launch is picked up.

---

Windows 11 Pro 10.0.26200 · nodeterm 0.3.4 (unsigned Windows beta) · code references against
`main` @ 4b6cfd9

ezwep:
Fixes #560. Addresses #559 — deliberately NOT a closing keyword: the maintainer is keeping
#559 open until the npm-installed (`.cmd`) case is settled, which this PR does not close.

## The bug

`findInPathString` (`src/core/exec-path.ts`) resolved a binary as `path.join(dir, bin)` and never
tried a PATHEXT extension. On Windows nothing answers to a bare name — `gh` on disk is `gh.exe`,
`claude` is `claude.exe`, an npm shim is `<name>.cmd` — so every lookup missed, fell through to its
POSIX fallback list (`/usr/bin/…`, `/opt/homebrew/…`) which does not exist there either, and
returned `null`.

Because every caller reads `null` as "not installed", the failure is silent and lands somewhere
else entirely:

- **Agent nodes cannot resume after an app restart.** `claude-cli.ts` probes capabilities through
  `findInLoginPath('claude')`; on `null` it returns `UNKNOWN_CLAUDE_CLI_CAPS` without running the
  CLI, so `sessionIdFlag` is false → `workspace.ts` mints no session id → no `agentSessionId` is
  persisted → cold restore has nothing to pass to `--resume`. The conversation is gone while its
  transcript sits on disk. The comment at `workspace.ts` says the minted id exists *"so this node
  always has something to resume with"* — that safety net was missing on Windows.
- `ssh` / `scp` are never resolved (`pty-manager`, `remote-ssh/*`, `ssh-setup-runner`).
- Permission mode `auto` reads as unsupported — same probe, same `version`.

`git-service.ts` carried a **second, independent copy**: `GH_PATH` was a module-level const over
three absolute POSIX paths that never consulted `PATH` at all, so it was `null` on every Windows
install regardless of the above. The UI reported "GitHub CLI (gh) not found." on machines where gh
was installed and authenticated.

## The change

`executableCandidates(bin, platform, pathext)` yields the names a bare command can have on disk.
Bare name first, so an extensionless binary still wins; a name that already carries a PATHEXT suffix
comes back untouched (no `gh.exe.EXE`). Platform and PATHEXT are parameters rather than ambient
reads, so the mapping is unit-testable from any OS — the same shape as `tmuxInstall(platform,
hasCommand)` next door. PATH entries are unquoted before use, as `where.exe` does.

The access check drops to `F_OK` on Windows: Node documents `X_OK` as degrading to `F_OK` there, so
the extension list — not the executable bit — is what makes a hit real.

`ghPath()` replaces the const, goes through the shared resolver, and memoizes **on hit only** so a
miss is re-probed instead of frozen at import time (which also stops it racing the async
login-shell PATH probe).

**Off Windows both paths are byte-identical to before**: `executableCandidates` returns `[bin]` and
the access mode stays `X_OK`.

## One behaviour worth knowing

A win32 hit carries PATHEXT's own casing, so `gh.exe` comes back as `…\gh.EXE`. Same file — the
platform is case-insensitive — and every consumer here spawns the path rather than comparing it.
Recovering the on-disk spelling would need a readdir per directory, on a function whose whole point
is staying cheap on the main thread. Documented at the call site; the tests compare the way the
platform does.

## Tests

`src/core/exec-path.test.ts` is new. The name mapping is covered on every platform (PATHEXT
fallback, whitespace/empty entries, the no-double-extension guard, case-insensitivity), and a second
block walks a real temp directory end to end. The win32-only half asserts the regression directly: a
bare name reaching `.exe` and `.cmd`, and an extensionless file still winning when both exist.

## Three surfaces

- **Desktop** — where the bug bites; this is the fix.
- **Server Edition** — shares `src/core`; unaffected on Linux by construction (the win32 branches
  are inert).
- **Mobile** — not applicable. The companion talks to the host, and the host is the side being
  fixed. No `nodeterm-ios` change is implied.

## Verification

- `npm run typecheck` — **clean** (both `tsconfig.node.json` and `tsconfig.web.json`).
- `npx vitest run src/core/exec-path.test.ts` — **12 passed**, including the win32-only PATHEXT
  block, which by definition never runs on the Linux/macOS CI.
- `npx vitest run src/core` — **no new failures**. Diffing the failing-file set against a clean
  `upstream/main` checkout on the same machine: 50 files fail on main, 49 on this branch, and the
  set difference is one-directional.

**And one file goes from red to green:** `src/core/commit-message.test.ts`.

```
upstream/main          Tests  2 failed | 3 passed (5)
  × spawns in the project folder for a local project
  × does NOT spawn in an SSH project's remote path — that is the ENOENT
  AssertionError: expected undefined to be '/tmp'

this branch            Tests  5 passed (5)
```

Those are the repo's own pre-existing tests, not ones I wrote. `expected undefined` is the spawn
never happening, because `resolveBinary` could not find `claude` — the exact failure this PR
describes, already encoded in the suite and red on every Windows checkout.

The remaining 49 failing files fail identically on `upstream/main` and are environmental
(`listen EACCES` on unix-domain sockets in the hook tests, plus the missing native
`node-pty`/`smart-whisper` builds — the `smart-whisper` postinstall does not compile here, so the
suite was installed with `--ignore-scripts`). One of them, `brandPulse.test.tsx`, turns out to be a
line-ending assumption rather than an environment limit; filed separately as #578.

Reviewed with `codex exec review` before opening, which returned no actionable findings and asked
for the end-to-end filesystem test that is now included.
## Not in this PR

#565 (`tmux-hint` splitting PATH on `:`) is left out on purpose — `findCommand` is only reachable
from the darwin/linux branches of `tmuxInstall`, so it never runs on Windows. I have commented to
that effect on the issue.


eneskirca:
Reviewed as part of the six-PR Windows set (#573, #576, #577, #579, #568, #541), each in its own
isolated worktree off `origin/main`.

**Verified locally:** `tsc -p tsconfig.node.json` clean; `exec-path.test.ts` + `commit-message.test.ts`
green (15 passed, 2 skipped — the two skipped are the `describe.skipIf(process.platform !== 'win32')`
PATHEXT blocks, so on Linux the regression itself is covered only by the pure `executableCandidates`
mapping, which is the honest position and matches what the body says). Off-win32 the resolver is
byte-identical: `executableCandidates` returns `[bin]` and `ACCESS_MODE` stays `X_OK`.

The shape is right — a parameterised mapping beside `tmuxInstall(platform, hasCommand)`, quoting
handled the way `where.exe` does, `ghPath()` memoised on hit only. Two things below.

---

### 1. Bare-name-first may leave #559's headline unfixed for an npm-installed CLI

This is the one I would measure before calling #559 closed.

`executableCandidates` puts the bare name first, and the doc comment explains why: *"so an
extensionless executable still wins where one exists."* On POSIX that is obviously right. On Windows
an extensionless file sitting in a PATH directory is, in the overwhelmingly common case, **not an
executable** — it is an npm shim. `npm i -g <pkg>` writes three files into the global bin directory:

```
%APPDATA%\npm\claude        <- POSIX sh script, for Git Bash
%APPDATA%\npm\claude.cmd
%APPDATA%\npm\claude.ps1
```

So for an npm-installed Claude Code, `findInPathString('claude', …)` now returns the **sh script**,
and `claude-cli.ts:53` does `execFileP(bin, ['--version'])` with no shell — which cannot run it on
Windows. The probe throws, the `catch` returns `UNKNOWN_CLAUDE_CLI_CAPS`, and the chain you traced in
#559 (no `sessionIdFlag` → no minted id → no `--resume`) is exactly where it was.

And the `.CMD` candidate would not rescue it either: since the CVE-2024-27980 fix, Node refuses to
spawn `.bat`/`.cmd` through `child_process` without `shell: true` (this repo pins Node 22 in
`.nvmrc`). The codebase already knows this — `src/core/agent-launch.ts:432` classifies a `.cmd`/`.bat`
hit as `cmd-script` precisely so it is *not* invoked as a native executable, with a comment saying so.
`findInPathString`'s callers (`claude-cli.ts`, `commit-message.ts`, `pty-manager`) have no such
classification.

To be clear about severity: **this is not a regression.** Today the lookup returns `null` for
everything on Windows, so every outcome here is an improvement. It means the fix may be incomplete
for a very common install shape, and the issue it says it fixes may not actually close.

Two directions, both cheap:

- On win32, try the PATHEXT candidates **before** the bare name. That is also what Windows itself
  does: for a bare command with no extension, `cmd.exe`/`CreateProcess` resolve through PATHEXT and
  never execute an extensionless file. The current order inverts the platform's own rule. (The test
  *"prefers an extensionless file when one exists"* would flip with it.)
- And/or route a `.cmd`/`.bat` hit through the existing `agent-launch.ts` classification at the
  callers that spawn directly.

Could you check on your box with `npm i -g @anthropic-ai/claude-code` present? Your #559 measurements
show `claude -> ...\claude.exe`, so your install is the native one and would not have surfaced this.

### 2. `exec-path.test.ts:54,59` — the backslashes are not in the string

```ts
expect(unquotePathEntry('"C:\Program Files\GitHub CLI"')).toBe('C:\Program Files\GitHub CLI')
```

`\P` and `\G` are not escape sequences, so JS drops the backslashes: the value under test is
`"C:Program FilesGitHub CLI"`. The assertion passes and the quote-stripping is genuinely exercised —
but the *Windows path shape* the test names is not. `\\` on both, same for line 59.

---

### Not findings, recorded so they are not re-derived

- `ghPath()` re-probes on every miss, so `status()` walks PATH on each call where gh is absent. Fine:
  `status()` already spawns eleven git subprocesses (see the comment at
  `renderer/state/worktrees.ts:33`), so a handful of `accessSync` calls is noise. I would not cache
  the miss — the "gh installed while the app runs" property is worth more.
- `ACCESS_MODE` reads `os.platform()` at module load, so only `executableCandidates` is cross-OS
  testable and the filesystem half correctly skips. Consistent with what the body claims.
- `unquotePathEntry` runs on every platform. A POSIX directory literally named `"…"` would be
  mangled; not worth guarding.
- **#583** (GIT_ENV prepending POSIX paths with `:` on Windows) lives in `git-service.ts`, which this
  PR already edits, and is untouched here. Not a request to widen the scope — just noting it is the
  natural next patch in this file, and that this PR does not close it.

**Verdict: good change, merge it first in this set (#576 needs it).** The npm-shim question above does
not block the merge; it decides whether #559 can actually be closed.

ezwep:
Pushed a correction — a second review pass found a real defect in the candidate **order** of the
original commit, and it is worth calling out rather than burying.

It returned the bare name FIRST and the PATHEXT variants after it. That is backwards from how
Windows resolves, and the layout where it bites is not exotic: **npm installs a global CLI as both
`<name>` (a POSIX shell shim, for Git Bash) and `<name>.cmd`, in the same directory.**
Bare-name-first hands back the shim, which `CreateProcess` cannot run — so the caller gets a spawn
failure on a file that plainly exists. That is a worse outcome than the `null` this PR set out to
fix, because the path looks correct.

Reproduced with both files in one directory:

```
before:  mytool       <- the POSIX shim, unspawnable
after:   mytool.CMD   <- the one that runs
```

The bare name is now the LAST candidate rather than dropped, since an extensionless PE is executable
on Windows; it simply must not beat a real PATHEXT match.

One of the original tests asserted the wrong rule ("prefers an extensionless file when one exists"),
so it is replaced by three that pin the corrected behaviour: the shim-vs-`.cmd` case, an
extensionless-only file still resolving, and a name given in full.

Re-verified on Windows 11: `npm run typecheck` clean, `exec-path.test.ts` 14 passed,
`commit-message.test.ts` 5 passed (still the 2-red-to-green improvement over `upstream/main`).

ezwep:
**End-to-end proof on Windows, across a full OS reboot.** This was the last unverified link.

Setup: a node created on a build carrying #573, given one short conversation so a transcript existed,
then the machine was rebooted (which kills the detached session host, so the cold-restore path is the
one that must fire).

```
system boot          21:31:53

pid 16196  21:33:42  claude.exe … --resume 0dfeb960-18fb-4d42-a9bb-9637902b9c5a --permission-mode auto
pid 16660  21:33:42  claude.exe … --permission-mode auto                        <- node predating the fix
pid  2712  21:34:09  claude.exe … --permission-mode auto --session-id 44acde9d-…  <- a newly created node
```

`0dfeb960-…` is exactly the id that had been minted and persisted as `agentSessionId` in
`project.json` before the reboot. The full chain therefore holds:

1. the PATHEXT fix lets the capability probe find `claude.exe`, so `sessionIdFlag` is detected;
2. `workspace.ts` mints an id and persists it on the node;
3. after a reboot, cold restore relaunches with `--resume <that id>` and the conversation comes back.

The three processes above show all three states side by side, including why an **existing** node
still comes back empty: it never got an id, so there is nothing to resume. That is not fixable
retroactively, and it is worth knowing when verifying this issue against a build — a tester whose
projects all predate the fix will see no change at all and reasonably conclude it did not work.
(#577 has the same shape for project names, and now heals those on read.)

`--permission-mode auto` on every line is the second half of the same fix: on the released build it
was silently dropped, because `autoSupported` is derived from the same failed probe.

Windows 11 Pro 10.0.26200, Claude Code CLI 2.1.251, build from `windows-fixes-combined`.

ezwep:
A side effect of this PR that I did not anticipate, and that I think needs your judgement rather
than mine. It surfaced from a user question — "what changes if I install tmux on Windows?" — not
from testing, so it is **reasoned from the code, not observed**: this machine has no tmux, so I
cannot demonstrate it.

## The change

`findTmux` has a dedicated win32 branch (`pty-manager.ts:344`) whose comment says what it is for:

> *MSYS2/Cygwin tmux, if the user put it there, is [on the Windows PATH]*

It resolves via `findInPathString('tmux', …)`. **Before this PR that could never match**, because the
walk only tried the literal name `tmux` and an MSYS2 build is `tmux.exe`. The branch was aspirational
— it read as if it worked, and could not. **After this PR it matches.**

That flips a backend decision. In `create()`:

```
tmuxPath && tmuxEnabled && persistKey   -> tmux          (:2979)
… else … sessionHostSupported()          -> session host  (:3068)
```

tmux wins. So a Windows user with MSYS2 on PATH moves off the session host — the backend this
branch of work exists to make available — and onto an MSYS tmux, which I would expect to work
badly there: it manages MSYS ptys rather than ConPTY, and nodeterm hands it `-f <userDataDir>\tmux.conf`,
a Windows path an MSYS tmux is unlikely to read as written.

Net effect: on the machines where it triggers, this PR could *replace* working persistence with
broken persistence. That is the opposite of its purpose, which is why I am flagging it rather than
quietly shipping it.

## Something that suggests the intent is already "host wins on Windows"

The two paths disagree with each other today:

- **spawn** (`:2979`): `this.tmuxPath && …` — a found tmux wins, on every platform.
- **recycle** (`:4463`): `(this.runtimePlatform === 'win32' || !this.tmuxPath) && …` — win32 takes
  the host **regardless** of tmux, with the comment *"on Windows whenever persistence is on and the
  host bundle exists, elsewhere only when no local tmux is installed"*.

So the recycle path already encodes "Windows uses the host even if a tmux is present", and the spawn
path does not. Before this PR that inconsistency was unreachable, because `tmuxPath` could never be
set on Windows. This PR makes it reachable. (`hostBackendEligible()`, added on #579, follows the
recycle form.)

## What I have not done

I have not changed either path. Which way it should go is a product decision — accept an MSYS tmux
and make the conf path work, or make win32 prefer the host the way the recycle path already assumes
— and it is yours, not something to fold into a PATHEXT fix.

If you want the conservative option, the smallest version is a win32 gate in `findTmux` so the
lookup keeps returning null there until an MSYS tmux is known to work; that restores today's
behaviour exactly while keeping everything else this PR fixes. Say the word and I will push it.

ezwep:
Both findings addressed, and I measured the npm case you asked for. Answering the second one first
because it is the one that decides #559.

## 1. Bare-name-first — already changed, and your prediction is confirmed

The ordering flip landed before your review reached me (commit `4718d5d9`, comment above): PATHEXT
first, bare name last, and the *"prefers an extensionless file"* test replaced by three that pin the
corrected rule. A second review pass on my side had reached the same conclusion independently, which
I take as a good sign about the finding rather than about me.

**Measured with an exact npm-shim layout** — `claude`, `claude.cmd`, `claude.ps1` in one directory:

```
old order  ->  claude       execFile -> ENOENT
this PR    ->  claude.CMD   execFile -> EINVAL
```

So the fix resolves the *right* file now, and **you are still right that #559 does not close**. It
moves from "picks a file Windows cannot execute" (`ENOENT`) to "picks the correct file and the
caller refuses to spawn it" (`EINVAL`, the CVE-2024-27980 behaviour you named). Better, and not
enough.

Which makes the second of your two directions the load-bearing one: the callers that spawn a
resolved path directly — `claude-cli.ts:53`, `commit-message.ts` — need the `cmd-script`
classification `agent-launch.ts:432` already has, rather than treating every hit as a native
executable. I have not done that here: it means either a `shell: true` carve-out on a path that
comes from a PATH walk (so quoting becomes load-bearing on a string the user's `PATH` supplied), or
resolving past the shim to the real binary. That is your design call, and I would rather implement
the one you pick than pick one in a PATHEXT fix. Happy to send it as its own PR either way.

For the record, my own install is the native `claude.exe`, which is why my #559 reboot run passed —
I have said so on the issue too, since reporting it as "Windows" without naming the install shape
was too broad.

## 2. `exec-path.test.ts:54,59` — fixed

Correct, and slightly embarrassing: the value under test was `"C:Program FilesGitHub CLI"`. The
assertions passed and the quote-stripping was genuinely exercised, but the Windows path *shape* the
test names was not — which is the only reason those cases exist. Doubled on both, and on the
lone-quote case beside them. `exec-path.test.ts` 14 passed.

## On the recorded non-findings

Agreed on all four, and thank you for writing down the `ghPath()` reasoning — "I would not cache the
miss, the gh-installed-while-running property is worth more" is exactly why it is memoised on hit
only, and it is better recorded on the PR than rediscovered later.

On **#583**: agreed it is the natural next patch in this file and that this PR does not close it. I
have left it alone deliberately — same reason I left #565 out, which is its pair.

PR #576, opened 2026-08-31
fix(core): commit-agent binary lookup — absolute Windows paths, POSIX-only fallbacks, gate ordering

### LINKED ISSUE #574: Windows: Name with AI and Generate commit message always report "claude CLI not found" — a third private executable lookup, POSIX-only in three ways
## What happens

"Name with AI (from terminal output)" does nothing useful on Windows — it reports

> claude CLI not found. Install Claude Code or pick another agent in Settings.

on a machine where Claude Code **is** installed and on `PATH`. The same path backs Generate commit
message and the canvas group naming, so all three are affected.

## Root cause: a third, independent copy of the executable lookup

`src/core/commit-message.ts` does not use the shared resolver first. `resolveBinary(name)`:

1. `if (name.startsWith('/')) return fs.existsSync(name) ? name : null` — a POSIX-only test for
   "is this an absolute path". A Windows absolute path (`C:\...`, or a UNC `\server\share`) does
   not start with `/`, so it is never recognised as one.
2. Five hardcoded candidates, all built POSIX-style and all extensionless:
   `~/.local/bin/<name>`, `~/.claude/local/<name>`, `/opt/homebrew/bin/<name>`,
   `/usr/local/bin/<name>`, `/usr/bin/<name>`. On Windows the real binary is `<name>.exe` (or a
   `.cmd` shim), so every one of them misses.
3. `if (!/^[A-Za-z0-9._-]+$/.test(name)) return null` — then
   `findInPathString(name, shellPathNow() ?? process.env.PATH)`, which until now appended no
   PATHEXT extension and so missed too.

Three lookups in a row, none of which can succeed on Windows, and the resulting `null` is rendered
as "not found — go install it".

Measured on Windows 11 (the five candidates, exactly as the code builds them, for `claude`): all
five miss.

## What #573 fixes, and what it does not

PR #573 (for #559/#560) teaches `findInPathString` about PATHEXT, so step 3 now resolves
`claude` → `claude.exe` and this feature should start working. I have verified that lookup
resolves on Windows with that change in place.

Two things in `resolveBinary` are **not** covered by it and are worth fixing here:

- **A custom agent command cannot be an absolute Windows path.** `C:\Tools\myagent.exe` fails the
  `startsWith('/')` test at step 1, then fails the `^[A-Za-z0-9._-]+$` gate at step 3 (it contains
  `:` and `\`), so it returns `null` no matter what. Settings → Commit messages → custom command is
  effectively bare-name-only on Windows. `path.isAbsolute()` is the platform-correct test.
- **The candidate list can produce an unspawnable hit.** Because the candidates are extensionless,
  on Windows they match any extensionless file of that name — including a POSIX shell script, which
  `spawn()` cannot execute there (no shebang handling). The list is dead weight on Windows and
  would be better skipped, or filtered through the same PATHEXT logic.

## Steps to reproduce

1. Windows, with the Claude Code CLI installed and on `PATH`.
2. Any terminal node with some output → the ✦ button, or the sidebar's "Name with AI".

**Expected** — a short generated title.
**Actual** — "claude CLI not found. Install Claude Code or pick another agent in Settings."

**Frequency** — always.

## Why it hurts

The message is confidently wrong, and it sends the user to reinstall a CLI that is already there.
Same for Generate commit message, which fails identically.

## Suggested fix

Fold `resolveBinary` onto `findExecutableSync` from `src/core/exec-path.ts` — one resolver instead
of a third private copy — and use `path.isAbsolute()` for the absolute-path branch. Relates to
\#559, and step 3 is fixed by #573.

---

Windows 11 Pro 10.0.26200 · nodeterm 0.3.4 (unsigned Windows beta) · Claude Code CLI on PATH ·
code references against `main` @ 4b6cfd9

ezwep:
> **Depends on #573.** On its own this does not fix the headline symptom — see *What this does not
> fix* below. Together the two close #574.
>
> Fixes #574. (Carried here rather than on #573 because this is the
> one that lands second; flagged in review as missing from both.)

## The bug

"Name with AI (from terminal output)", "AI commit message" and the canvas group naming all resolve
their CLI through `resolveBinary` (`src/core/commit-message.ts`), and on Windows it cannot succeed.
The user sees

> claude CLI not found. Install Claude Code or pick another agent in Settings.

on a machine where Claude Code is installed and on `PATH` — a confidently wrong message that sends
people to reinstall a CLI they already have.

Three things are POSIX-only, in a row. This PR fixes the first two plus an ordering bug beside them;
#573 fixes the third.

## 1. `startsWith('/')` as the absolute-path test

A Windows absolute path (`C:\Tools\agent.exe`, or a UNC `\server\share\agent.exe`) does not start
with `/`, so it missed that branch, fell through to the bare-name path, and was then rejected by
that branch's charset gate for containing `:` and `\`. **A custom commit agent could not be given as
a full path on Windows at all.** Now `path.isAbsolute`.

## 2. Five extensionless POSIX fallback locations

`~/.local/bin/<name>`, `~/.claude/local/<name>`, `/opt/homebrew/bin/<name>`, `/usr/local/bin/<name>`,
`/usr/bin/<name>`. On Windows the real binary is `<name>.exe` or a `<name>.cmd` shim, so every one
misses — and being extensionless they can only ever match an extensionless file, which on Windows
may well be something `spawn()` cannot execute (a POSIX shell script, say). Skipped on win32 rather
than left to misfire.

**Order is otherwise untouched.** They still run *before* the PATH walk on macOS and Linux, which is
deliberate: a GUI process inherits a minimal PATH and these are where user installs live. I
considered folding the whole thing onto `findExecutableSync` and decided against it precisely
because that resolver checks PATH first, which would have silently changed which binary macOS users
get.

## 3. A behaviour change worth reviewing on its own

The charset gate `^[A-Za-z0-9._-]+$` now runs **before** the fallback loop instead of after it. It
was only ever applied on the way to the PATH walk, so a name containing `../` normalized straight
out of the directories that loop exists to confine it to — `~/.local/bin/../../evil` is `~/evil`.
Same rule, same set of rejections, applied where it actually confines. The value comes from a user
setting rather than from a remote, so this is hardening rather than a live exploit, but it seemed
wrong to port the function and leave it.

## What this does not fix

The third POSIX-only step is the PATH walk itself: `findInPathString` appends no PATHEXT extension,
so a bare `claude` never reaches `claude.exe`. That is #573's change and it is not in this branch.
**Until #573 lands, the default configuration on Windows — a bare agent name — still reports "claude
CLI not found."** What this PR buys on its own is an absolute path in the custom-command setting
starting to work, plus the two correctness fixes above.

A positive bare-name PATH test on Windows is deliberately absent for the same reason: it could not
pass without #573. This was raised as a P1 in review and is the reason the title no longer claims
the lookup "works on Windows" — happy to rebase this onto #573 instead if you would rather review
them as one.

## Tests

`resolveBinary` is exported for the tests; every caller is still in this module (the repo already
exports `findCommand`/`tmuxInstall` for the same reason). Covered: the absolute-path branch both
ways, the relative and injection-shaped rejections, and a bare name that is nowhere.

## Three surfaces

- **Desktop** — where the bug bites.
- **Server Edition** — shares `src/core`. The win32 branch is inert there, but the gate move is a
  **behaviour change on POSIX too**, and worth stating rather than glossing: in the old order the
  fallback loop `return`ed *before* the charset gate ran, so a name failing that test was never
  rejected downstream — it could resolve out of `~/.local/bin/...` and be spawned. That is the
  `../../evil` hole this closes. The cost is that a custom commit-agent value containing a space,
  a slash, or anything else outside `[A-Za-z0-9._-]` now returns `null` on macOS and Linux as
  well, where those five directories previously resolved it. I think that is right — confining
  is what those directories are for, and a value that escaped them was never meant to work — but
  it is a change, not a no-op.
- **Mobile** — not applicable.

## Verification

- `npm run typecheck` — **clean**.
- `npx vitest run src/core/commit-message.test.ts` — **2 failed | 6 passed (8)**.

The 6 passing include all 3 tests added here. **The 2 failures are the dependency, demonstrated
rather than argued:**

```
× spawns in the project folder for a local project
× does NOT spawn in an SSH project's remote path — that is the ENOENT
  AssertionError: expected undefined to be '/tmp'
```

Those two are pre-existing tests, red on `upstream/main` on Windows and still red here, because the
bare-name PATH walk has no PATHEXT. On the #573 branch the same file is **5 passed (5)**. So this PR
fixes what it claims to fix and no more — exactly the point of the section above.

Reviewed with `codex exec review` before opening. Its single P1 was that the original title claimed
the lookup "works on Windows" when it needs #573 to do so; the title and framing were rewritten, and
the numbers above are the evidence for that correction.
---

Windows 11 Pro 10.0.26200 · nodeterm 0.3.4 (unsigned Windows beta)


eneskirca:

**Verified locally:** `tsc -p tsconfig.node.json` clean; `commit-message.test.ts` 8/8 green (the 3
added plus the 5 pre-existing — on Linux the two you report red on Windows pass, which is consistent
with your diagnosis that they are the missing-PATHEXT symptom).

The dependency framing is right and I want to record that I checked it rather than took it: there is
**no file overlap** with #573 (`exec-path.ts`/`git-service.ts` vs `commit-message.ts`), so the two
merge in either order with no conflict — but on win32 this PR sets `candidates = []`, which means the
whole answer now rides on `findInPathString`, and that is #573's change. So the ordering is semantic,
not textual. **No need to rebase** — merge #573 first and this second, and the branch is fine as it
stands. Leaving them separate is also the better review artefact: the "what this does not fix"
section is the kind of thing that gets lost in a squash.

All three fixes are correct:

- `path.isAbsolute` — and note it stays correct in the cross direction too: on POSIX a `C:\…` string
  is still not absolute, still falls to the charset gate, still returns `null`. Unchanged behaviour
  where the platform is unchanged.
- Skipping the extensionless POSIX candidates on win32, with the ordering on POSIX explicitly left
  alone. The reasoning for *not* folding onto `findExecutableSync` (it checks PATH first, which would
  change which binary macOS users get) is the right call and worth having written down.
- The gate move. Agreed this is hardening rather than a live exploit, and agreed it was wrong to port
  the function and leave it.

---

### One correction to the body, not the code

> **Server Edition** — shares `src/core`; unchanged on Linux (the win32 branch is inert, and the gate
> move only narrows what was already rejected downstream).

The parenthetical isn't quite true, and I think the true version is a better argument for the change.
In the old order a fallback hit **returned before the gate ran**:

```ts
if (name.startsWith('/')) return …
const candidates = [ `${homedir}/.local/bin/${name}`, … ]
for (const c of candidates) if (fs.existsSync(c)) return c   // <- returns here
if (!/^[A-Za-z0-9._-]+$/.test(name)) return null             // <- gate never reached
```

So a name failing the charset test was *not* rejected downstream — it could resolve out of
`~/.local/bin/…` and be spawned. That is precisely the `../../evil` hole you close, and it means the
gate move **is** a POSIX behaviour change: a custom commit-agent setting containing a space, a slash
or anything else outside `[A-Za-z0-9._-]` that previously resolved through those five directories now
returns `null` on macOS and Linux too.

I think that is the right change — the whole point of those directories is confinement, and a value
that escapes them was never intended to work. But this repo's house rule is that a behaviour change
is stated rather than described as a no-op, so I would just say it plainly in the body: *"on POSIX
this also stops a non-bare-name resolving out of the fallback directories, which is the hole being
closed."*

### #574 will not auto-close

The body says *"Together the two close #574"*, but there is no `Fixes #574` / `Closes #574` keyword in
either PR, so GitHub will close neither. Worth adding the keyword here (this is the one that lands
second), or closing #574 by hand once both are in.

**Verdict: approve. Merge immediately after #573.**

ezwep:
Both corrected — thank you, the first one is the more interesting of the two and you are right that
the true version argues for the change better than my wrong version did.

## The Server Edition claim

I wrote that the gate move *"only narrows what was already rejected downstream"*, and that is simply
false. Your trace is the reason: the fallback loop `return`s before the gate runs, so a name failing
the charset test was never rejected downstream at all — it could resolve out of `~/.local/bin/...`
and be spawned. That IS the `../../evil` hole, and it means the move is a behaviour change on POSIX
and not only on Windows.

The body now says so plainly, including the cost: a custom commit-agent value containing a space, a
slash, or anything outside `[A-Za-z0-9._-]` that previously resolved through those five directories
now returns `null` on macOS and Linux too. I agree it is the right change for the reason you give —
confinement is what those directories are for — but calling it a no-op was me describing what I
wished were true rather than what the code did.

## `Fixes #574`

Already added, and GitHub now lists `#574` as this PR's linked issue — I put it here rather than on
#573 for the same reason you name: this is the one that lands second. Your comment and my edit
crossed.

PR #577, opened 2026-08-31
fix(renderer): name a project after its folder on Windows, not its whole path

### LINKED ISSUE #561: Sidebar: project rows show the full Windows path instead of the folder name, and that name leaks into the git-shared project.json
## What happens

In the Sessions panel and the project tab bar, every project row shows the full path
(`C:\Users\<user>\code\my-app`) instead of `my-app`. The name wraps onto two lines, and the only part
that identifies the project — the folder name — sits at the very end. Two projects under the same
parent are indistinguishable until the second line.

## Root cause

The project name is derived with a basename that only splits on forward slashes:

- `src/renderer/state/projects.ts:301` — `folder.split('/').filter(Boolean).pop() || 'Project'`
- `src/renderer/state/projects.ts:636` — same, for `fallbackName`
- `src/renderer/lib/projectOpen.ts:138` — same
- `src/renderer/lib/projectOpen.ts:20` — `normalizeProjectCwd` only strips `/+$`, never touches `\`
- sibling: `src/renderer/canvas/Canvas.tsx:10031` — same pattern for a folder name

On a Windows path with backslashes `split('/')` yields a single element: the whole path.

## The derived name is also written to disk

Every `.nodeterm/project.json` created on Windows contains:

```json
"name": "C:\Users\<user>\code\my-app"
```

Per the comment at `projects.ts:65`, `name` is git-shared through `project.json`. A Windows user who
commits that file therefore shares their absolute local path as the project name with the whole
team, including teammates on macOS.

## Steps to reproduce

1. On Windows, open a folder as a project (Open folder), e.g. `C:\Users\<user>\code\my-app`.
2. Open the Sessions panel, Project tab.
3. Look at the project row, then open `.nodeterm/project.json` in that folder.

**Expected** — `my-app`, as on macOS/Linux.
**Actual** — the full path, wrapped over two lines, and the same string in `project.json`.

**Frequency** — always, for every project created on Windows.
**Workaround** — Rename project overwrites the name by hand.

## Why it hurts

The sidebar is where you switch projects, and it becomes unreadable as soon as two projects share a
parent folder. The leak into `project.json` turns a cosmetic issue into a shared one.

## Suggested fix

One platform-independent basename helper
(`p.replace(/\/g, '/').split('/').filter(Boolean).pop()`), as `session-host/host.ts:426` and
`session-host/session.ts:198` already do, used at all five sites.

---

Windows 11 Pro 10.0.26200 · nodeterm 0.3.4 (unsigned Windows beta) · code references against
`main` @ 4b6cfd9

ezwep:
Fixes #561. Independent of #573/#576 — no shared files.

## The bug

Every project created on Windows is named after its absolute path. The sidebar shows
`C:\Users\<user>\code\my-app` wrapped over two lines instead of `my-app`, with the only
distinguishing part — the folder — last. Two projects under one parent are indistinguishable until
the second line, and the tab bar truncates them to the same prefix.

The cause is an inlined basename that only splits on forward slashes:

```ts
folder.split('/').filter(Boolean).pop() || 'Project'
```

On a backslash path that returns **one** segment: the whole thing. It appears at five sites
(`projects.ts` twice, `projectOpen.ts`, `Canvas.tsx`, and `normalizeProjectCwd` stripping only a
trailing `/`), so it could not be fixed once.

## It is not only cosmetic

`name` is git-shared through `project.json` — see the comment at `projects.ts`. A Windows user who
commits that file publishes their absolute local path as the project name to everyone who pulls it,
teammates on macOS included.

## The change

One `folderName()` in `lib/projectOpen.ts` answers for all five sites. Backslashes are folded to `/`
before splitting, the way `session-host/host.ts:426` and `session-host/session.ts:198` already do
it, so a single implementation serves both platforms.

`normalizeProjectCwd` now strips a trailing separator of either kind — leaving a `\` in place
produced an empty final segment downstream — **except where removing it changes what the path
means**. `C:\` is the root of drive C while `C:` is the current directory *on* drive C, so a
stripped drive root would have deduped an opened drive root against the wrong project through
`findProjectByCwd`'s exact-string match. A path that is nothing but separators is kept for the same
reason. This was a P1 in review; it is now guarded and asserted.

The helper returns `''` rather than a default for a path with no folder in it, and each caller keeps
its own fallback: `'Project'` is right for a project name and wrong for Canvas's notification label,
which prefers the node title.

## Deliberately not converted

`ExplorerPanel.tsx`, `SshPassphrasePrompt.tsx` and `SshProjectDialog.tsx` inline the same split, but
those paths are remote (SSH) and genuinely POSIX. Folding backslashes there would corrupt a
legitimate filename — `\` is a valid character in a POSIX path. Happy to be told otherwise if any of
them can receive a local path.

## Tests

POSIX and Windows shapes, a UNC share, a mixed-separator path (Node hands back both), redundant
separators, and the empty answers; for `normalizeProjectCwd`, the trailing separator on both sides
plus the preserved drive root.

## Three surfaces

- **Desktop** — where the bug bites.
- **Server Edition** — serves Linux; unchanged (no backslashes to fold, and the drive-root guard
  cannot match).
- **Mobile** — not applicable; it reads project names from the host, so it inherits the fix.

## Verification

- `npm run typecheck` — **clean** (which also confirms the missing-comma build breaker review caught
  is gone).
- `npx vitest run src/renderer/lib/projectOpen.test.ts` — **26 passed**. That is the 6 added here
  plus all 20 pre-existing `planOpenProject` tests, which matters because `normalizeProjectCwd`
  feeds `findProjectByCwd`'s dedupe.
- `npx vitest run src/renderer/state src/renderer/lib` — **1565 passed, 1 failed**. The one failure
  is `brandPulse.test.tsx`, which fails identically on a clean `upstream/main` checkout on this
  machine: it slices CSS with a hardcoded `'}\n}'` while a Windows checkout has CRLF. Unrelated to
  this PR; filed separately as #578.

Reviewed with `codex exec review` before opening, which caught two things now fixed: the missing
comma in the `Canvas.tsx` import, and the drive-root semantics (`C:\` must not normalize to `C:`,
which is a different location and would have deduped against the wrong project).
---

Windows 11 Pro 10.0.26200 · nodeterm 0.3.4 (unsigned Windows beta)


eneskirca:

**Verified locally:** typecheck clean; `projectOpen.test.ts` 26/26; the wider
`src/renderer/state src/renderer/lib` sweep 613/613 across 48 files (I ran the state half deliberately
— `normalizeProjectCwd` feeds `findProjectByCwd`'s exact-string dedupe, so that is where a mistake
here would land, not in the naming). Independence from #573/#576/#579 confirmed: no shared files, and
`projectOpen.ts` imports only `@shared/one-line`, so pulling `folderName` into `state/projects.ts`
introduces no cycle.

This one I have nothing to push back on. Specifically:

- **The drive-root guard is the right catch and the tests pin it both ways.** `C:\` → kept, `C:` →
  never produced. That is genuinely the thing that would have silently deduped an opened drive root
  against the wrong project through `findProjectByCwd`, and it is not a mistake I would expect a
  reviewer to spot after the fact.
- **`folderName` returning `''` rather than `'Project'`** so each caller keeps its own fallback — the
  Canvas notification label prefers the node title, and baking `'Project'` in would have quietly
  changed that string. Right call.
- **The "deliberately not converted" list is correct.** `ExplorerPanel` / `SshPassphrasePrompt` /
  `SshProjectDialog` are remote POSIX paths and `\` is a legal character in a POSIX filename, so
  folding it there would corrupt a real name. To answer your "happy to be told otherwise": I traced
  them and they are fed from SSH browse/remote-cwd state, never from a local `dialog:select-folder`
  result. Leave them.

---

### One scope question, not a blocker

This fixes **new** projects. Every project already created on Windows keeps its bad `name` on disk —
`fallbackName` at `projects.ts:636` only applies when `name` is absent, so an existing
`.nodeterm/project.json` carrying `"name": "C:\\Users\\<user>\\code\\my-app"` stays that way, including
one already committed and pulled by teammates. Which is the leak half of #561.

That is defensible (the user has "Rename project", and silently rewriting a git-shared field on load
is its own kind of surprise), but #561 asks for both halves, so it is worth deciding out loud:

- accept it, and say in the body that existing projects need a manual rename — so #561 closes with a
  known remainder rather than looking fully fixed; or
- add a one-time repair: when `name === cwd` (i.e. the name is verbatim the path, which is only ever
  the bug's fingerprint and never something a user typed), re-derive it through `folderName`. Narrow
  enough to be safe, and it fixes the already-committed files on next open.

I would take the first for this PR and file the second if you want it — I mention it only so the issue
gets closed accurately.

**Verdict: approve. Independent of the rest of the set — merge whenever.**

ezwep:
Pushed a second commit, because a tester on another Windows machine reported this as **still
broken** on a build that contains the fix — and they were right.

The original change only names a project correctly when it is **created**. Nothing re-derives a
stored name: `probeFolder` → `fileToProject` takes `f.name` from `project.json` verbatim. So every
project a user already had keeps its pre-fix name forever, and only a manual rename clears it. For
anyone upgrading, that means nothing visibly changes — which is exactly how it was reported.

`healPathAsName` now replaces a stored name that **is** the project's own cwd, on read. Narrow by
design: only a name EQUAL to the cwd is touched, since such a name carries no information the cwd
does not already carry, so re-deriving it cannot lose something the user chose. A deliberate rename
that merely looks path-ish, or names a *different* path, is left exactly as typed.

One thing worth flagging, because the first version of this commit had it wrong: the comparison is
**per-platform**. It originally lowercased and folded backslashes on every path — both are Windows
rules, and both are wrong on POSIX, where `\` is a legal character in a filename (`/repo/a\b` is one
folder) and paths are case-sensitive (`/repo/Foo` is not `/repo/foo`). A name differing only in case
from its cwd is a name the user chose, and the naive version would have destroyed it. `looksWindows()`
now gates those rules on a drive letter or UNC prefix, and `folderName` folds separators on the same
condition.

`folderName` moves to `shared/project-name.ts` because the core-side loader needs the same rule and
`src/core` may not import from `src/renderer`. It is still exported from `projectOpen.ts`, so no
call site changes.

Verified on Windows 11: typecheck clean, 38 tests pass across `project-name`, `projectOpen` and
`no-electron`. `workspace-files.test.ts` shows two failures about cwd relativization — they fail
identically on `upstream/main` here and are the same class as #578 (a test written for POSIX
separators).

Issue #575, opened 2026-08-31
Windows has no session-persistence backend at all: the session-host bundle is never built into or packaged with the app, and the bundled tmux is mac-only


ezwep:
This is, I think, the answer to my own #454 — and the reason sessions do not persist on Windows at
all.

## What I observe

On Windows, agent sessions do not survive. Switching to another node shows a fresh `claude` rather
than the existing conversation, and quitting and reopening the app loses every running session.

## What ships in the Windows build

The packaged Windows app has **no session-persistence backend of any kind**:

- `resources/session-host/host.cjs` — **absent**. `resolveSessionHostScript()`
  (`src/core/session-host-launcher.ts:26`) looks for exactly that path, so
  `sessionHostSupported()` returns false. Searched the whole install: no `host.cjs` anywhere. The
  string `session-host` does appear inside `app.asar` (the compiled code referencing it), but the
  bundle itself is not there.
- `resources/bin/tmux` — **absent**, and there is no `resources/bin` directory at all. The bundled
  tmux is a **mac-only** `extraResources` entry (`package.json`, `build.mac.extraResources`).

So `PtyManager`'s backend choice at `pty-manager.ts:4463` —
`(runtimePlatform === 'win32' || !tmuxPath) && settings.tmuxEnabled && hostAvailable` — can never
select a host-backed session on Windows, and there is no tmux to fall back to either. Every session
is a plain direct PTY owned by the Electron main process.

Confirmed live on this machine: the running processes are exactly the five standard Electron
processes plus `powershell.exe` children of the **main** process, each with `claude.exe` beneath it.
No session-host process exists.

## Why it is absent: two independent gaps in packaging

1. **`host:build` is never run by any build or dist script.**
   - `build` = `electron-vite build && npm run build:codex-relay`
   - `dist:win` = `make-icon && electron-vite build && electron-builder --win …`
   - `host:build` (`esbuild src/session-host/host.ts … --outfile=out/session-host/host.cjs`) is
     only ever invoked by hand.

   Worth noting because the comment at `src/core/session-host-backend.ts:32-36` states the bundle is
   missing "only in a dev checkout that never ran `npm run host:build` (or `npm run build`, which
   now runs it too)". `npm run build` does not run it.

2. **Nothing copies `out/session-host/` into `resources/`.** `extraResources` is
   `build/icon.png` at the top level, `resources/bin/tmux` + licenses under `mac`, and **nothing
[...]

Issue #454, opened 2026-08-27
Windows: what is left between the merged session host and a downloadable build?


ezwep:
## The question

What is the intended path for Windows, and what is still between the merged session-host work and a downloadable build? Asking as a user with Windows machines in the mix, so I can plan around it rather than guess.

I did the reading first, so this is a "what is left / what should I use meanwhile" question rather than a feature request. I searched open and closed issues and found nothing covering it — but this is a roadmap question rather than a bug, so if Discussions is the better home for it, say the word and I will move it there.

## What I found already in place

- **The session host is merged** (#305, 20 Aug) — a tmux-equivalent persistence backend for platforms with no tmux, with `docs/windows-session-host.md` behind it and `pty-manager.ts` selecting it as the third backend when no tmux is found.
- **Packaging exists** — `dist:win`, an NSIS + zip x64 target, `build/icon.ico`, and `win-package-smoke.yml` proving `electron-builder --win` produces an installer from a clean checkout.
- **CI already runs on Windows** — the `quality-windows` job on `windows-latest`. (Which proves the suite passes there, not that packaging, signing and distribution are ready — that distinction is exactly what I am asking about.)
- The landing page says Windows support is coming (`d8ff2780`).

And what is deliberately not there yet, from your own comment in `win-package-smoke.yml`:

> BUILD ONLY — this workflow must NEVER publish. […] Windows is not a shippable target until the session-host phase merges; release wiring (signing, `latest.yml` on nodeterm.dev/updates, a `release.yml` job) is a deliberate follow-up.

Since the session-host phase *has* now merged, that comment reads as partly stale — which is what prompted the question.

## What I am actually asking

1. **Is the remaining gap only release wiring** — signing, the update feed, a `release.yml` job — or is there functional work still open that the merged session host does not cover?
2. **Is an unsigned self-build a reasonable interim** for someone who accepts no auto-update, or would you rather people did not run `dist:win` output yet? A line either way would settle it; right now the honest answer to "can I use nodeterm on Windows" is buried in a CI comment.
3. **Is the Server Edition the intended Windows answer for now?** From the README it runs headless on a Linux/macOS host and is used from any browser, which makes Windows a fully supported *client* today with nothing to install. If that is the recommendation, the download page saying so next to "Windows support is coming" would save the next person this dig.

## Why it matters to me

[...]

PR #579, opened 2026-08-31
fix(build): ship the session host, and resolve it from inside the asar

ezwep:
Addresses #575 — deliberately NOT a closing keyword. #575 asks for two things and this covers
one: the packaging. Its request for a visible degraded-mode notice on Windows is a separate
change and should not be auto-closed away by this merge. Independent of #573/#576/#577 — no shared files.

## The bug

A packaged build contains **no session-host bundle at all**, so `sessionHostSupported()` is false
everywhere. On macOS and most Linux boxes that goes unnoticed: `PtyManager` falls back to tmux. On
Windows there is no tmux and none is bundled (`resources/bin/tmux` is a mac-only `extraResources`
entry), so the backend predicate

```ts
(runtimePlatform === 'win32' || !tmuxPath) && settings.tmuxEnabled && hostAvailable
```

can never be satisfied. Every session is a plain direct PTY owned by the Electron main process, and
sessions do not survive — which is the app's central promise.

Inspected on an installed 0.3.4 Windows build: no `host.cjs` anywhere under the install, `out/`
inside `app.asar` holds only `main`, `preload`, `renderer`, and the running process tree is the five
standard Electron processes with `powershell.exe` children of **main**. No host process exists.

## Gap 1 — `host:build` was never invoked

`build` was `electron-vite build && npm run build:codex-relay`, and every `dist:*` script calls
`electron-vite build` **directly** rather than going through `build`, so fixing it in one place would
not have been enough. `host:build` is now in `build` and in every script that invokes
electron-builder: `dist`, `dist:linux`, `dist:win` and `release`. That last one is what the release
workflow actually calls — I missed it first time round and review caught it, and without it
production macOS artifacts would have kept shipping without the host.

## Gap 2 — where the bundle is looked for

`resolveSessionHostScript` checked `<resourcesPath>/session-host/host.cjs`, and its comment
attributed that to an `extraResources` entry — *"see package.json's `build.extraResources`"* — which
does not exist and never did.

**Adding that entry would not have worked either**, which is why this takes the other route.
Electron patches `Module._nodeModulePaths` so a script under `resourcesPath` may only resolve from
paths under `resourcesPath`. Measured on the packaged Windows build, the search list for
`<resourcesPath>/session-host` is exactly:

```
<resourcesPath>/session-host/node_modules
<resourcesPath>/node_modules
```

`node-pty` is in neither — and the bundle needs it, since `host:build` marks it `--external`. From
**inside the asar** the list instead reaches `<resourcesPath>/app.asar/node_modules`, where
electron-builder's unpacked-native redirect makes `node-pty` resolve:

```
<resourcesPath>/app.asar/out/session-host/node_modules
<resourcesPath>/app.asar/out/node_modules
<resourcesPath>/app.asar/node_modules      <-- node-pty resolves here
<resourcesPath>/node_modules
```

So the bundle is resolved at `<appPath>/out/session-host/host.cjs`, and **nothing has to copy it**:
`build.files` is already `["out/**/*", "package.json"]`, so it rides into the asar the moment gap 1
builds it. `app.getAppPath()` is the asar when packaged and the repo root in dev, so one candidate
answers both. `repoRoot` stays as a fallback for a shell that supplies no app path, and the
`resourcesPath` candidate is kept **first** so an installation that does ship such a copy is
unaffected.

`appPath` is added to `CorePlatform` beside `resourcesPath`, optional for the same reason — an
Electron notion the Server Edition omits.

## Tests

Seven, covering the resolution order, the dev/packaged split, a clean miss when the bundle was never
built, absent opts, and a candidate that throws. `no-electron.test.ts` still passes, which matters
since this touches `src/core`.

## Three surfaces

- **Windows** gains a persistence backend it never had.
- **Linux** gains one wherever no system tmux is installed (same predicate).
- **macOS** is unchanged in practice — bundled tmux keeps winning — but the bundle now ships, which
  makes that fallback real rather than theoretical.
- **Mobile** — not applicable directly, though a phone attaching to a host that survives is the
  whole point.

## Verification

Built and verified end to end on Windows 11 (Node 24.19.0, vitest 4.1.11).

- `npm run typecheck` — clean. `session-host-launcher.test.ts` + `no-electron.test.ts` — **9 passed**.
- `npm run dist:win` — **exit 0**, producing `nodeterm-Setup-0.3.4.exe`.
- The packaged `app.asar` now carries `out/{main, preload, renderer, session-host}` with
  `host.cjs` at 376387 bytes. The 0.3.4 release build has only the first three.
- Spawned the host the way `spawnSessionHost` does
  (`ELECTRON_RUN_AS_NODE=1 nodeterm.exe <resources>/app.asar/out/session-host/host.cjs <userDataDir>`):
  it stays up, no crash, empty stderr.
- **`node-pty` resolves from inside the asar** — the point the design turns on:
  `RESOLVE OK -> ...^Gpp.asar
ode_modules
ode-pty\lib\index.js`, and requiring it yields
  `spawn, fork, createTerminal, open, native`. The last search path is `<resources>
ode_modules`,
  which is where an `extraResources` copy would have had to find it — and does not. That contrast is
  now measured on a real build rather than inferred.

Reviewed with `codex exec review` before opening; its one finding was the missing `release` script,
now fixed.


ezwep:
Update: **I have now built and verified this end to end on Windows.** The "not verified" caveat in
the description can be considered closed — I have edited it accordingly.

`npm run dist:win` on this branch, Windows 11 / Node 24.19.0, exit 0, producing
`nodeterm-Setup-0.3.4.exe`.

**1. The bundle ships.** Reading the packaged `app.asar` header:

```
out/ in asar    : main, preload, renderer, session-host
session-host    : host.cjs (376387 bytes)
node-pty in asar: yes (unpacked)
```

The 0.3.4 release build has only `main, preload, renderer` there — so this is the gap closing.

**2. The host starts from inside the asar.** Spawned exactly the way `spawnSessionHost` does it
(`ELECTRON_RUN_AS_NODE=1 nodeterm.exe <resources>/app.asar/out/session-host/host.cjs <userDataDir>`):
it stays up, no crash, empty stderr.

**3. `node-pty` resolves — the point the whole design turns on.** From
`<resources>/app.asar/out/session-host`:

```
search paths:
   ...\app.asar\out\session-host\node_modules
   ...\app.asar\out\node_modules
   ...\app.asar\node_modules          <-- hit
   ...\resources\node_modules

RESOLVE OK  -> ...\app.asar\node_modules\node-pty\lib\index.js
REQUIRE OK  -> exports: spawn, fork, createTerminal, open, native
```

Note the last search path: `<resources>\node_modules`. That is where an `extraResources` copy would
have had to find `node-pty`, and it is not there — which is the measured reason this PR resolves
from inside the asar instead. The contrast is now demonstrated on a real build rather than inferred
from the module-path patch.

Happy to run anything else against this build if it would help.

ezwep:
Update: **the part I could not verify is now verified**, on a real `dist:win` build produced with
this branch.

## The bundle ships

Reading the packaged `app.asar` header of the build this branch produces:

```
out/ in asar : main, preload, renderer, session-host
session-host : host.cjs (376387 bytes)
```

On the released 0.3.4 build the same read gives only `main, preload, renderer`. So gap 1 is closed:
`host:build` running inside `dist:win` puts the bundle where `build.files` already carries it, with
no `extraResources` entry.

## And it can load `node-pty` from there

This was the open question — whether a host resolved from inside the asar can reach its external
native dependency. Resolving exactly as the host itself would, via
`createRequire('<resourcesPath>/app.asar/out/session-host/host.cjs')`, run under the packaged
binary with `ELECTRON_RUN_AS_NODE=1`:

```
node-pty RESOLVED -> <resourcesPath>\app.asar\node_modules\node-pty\lib\index.js
node-pty GELADEN  -> exports: spawn, fork, createTerminal, open, native
```

It resolves *and* loads, `native` included — so electron-builder's unpacked-native redirect carries
the binding through the asar path. That is the evidence behind choosing this over an
`extraResources` copy, which cannot reach `node-pty` at all (search list in the PR body).

## What is still not covered

I have not yet run the installed app long enough to confirm a session actually survives an app
restart end to end — only that the host bundle ships and its dependency loads, which were the two
mechanical unknowns. Happy to report back once I have run on it for a while.

Built on Windows 11, Node 24.19.0, from `windows-fixes-combined` (this branch plus #573, #576, #577).

ezwep:
Final update: **running on it now, and the backend is live.** This closes the last gap I flagged.

I installed the `dist:win` build from this branch on Windows 11 and inspected the process tree of the
running app:

```
session-host pid 16888
  spawned as: nodeterm.exe <install>\resources\app.asar\out\session-host\host.cjs
  parent: nodeterm main (33192)

  powershell 18140   <- session
  powershell 24320   <- session
  powershell  6592   <- session
     all three: ParentProcessId = 16888
```

Every session is a child of the **session host**, not of the Electron main process. On the released
0.3.4 build the same inspection shows the shells parented directly to main, with no host process in
existence — which is the failure this PR fixes.

So the full chain is now verified on a real build rather than reasoned about:

| | |
|---|---|
| `host.cjs` ships inside the asar | 376,387 bytes at `out/session-host/` |
| host is spawned from that path | pid 16888, from `app.asar\out\session-host\host.cjs` |
| `node-pty` resolves and loads from there | including the `native` export |
| sessions are host-backed | 3/3 parented to the host |

The `NOT verified` caveat in the PR description no longer applies — I have left the wording in place
rather than rewriting history, but please read it as superseded by this and the previous comment.

Windows 11 Pro 10.0.26200, Node 24.19.0, built from `windows-fixes-combined` (this branch plus #573,
#576, #577).

eneskirca:

**Verified locally:** typecheck clean; `session-host-launcher.test.ts` + `no-electron.test.ts` +
the `session-host` suites green (72 passed across 7 files).

**Verified on real Windows CI, independently of your machine:** I pushed this branch and dispatched
`win-package-smoke.yml` against it —
[run 33536406642](https://github.com/eneskirca/nodeterm/actions/runs/33536406642), **success** on
`windows-latest`. The log shows the gap-1 fix doing its job inside `dist:win`:

```
> npm run make-icon && electron-vite build && npm run host:build && electron-builder --win --x64 …
> esbuild src/session-host/host.ts --bundle --platform=node --format=cjs --outfile=out/session-host/host.cjs …
  out\session-host\host.cjs  367.6kb
  • building  target=nsis file=dist\nodeterm-Setup-0.3.4.exe
```

So the packaging half is now confirmed on a clean runner, not only on your box. The asar-vs-
`extraResources` reasoning is well argued and your measured `node-pty` search list settles it — an
`extraResources` copy genuinely cannot reach the native module, and that is the kind of thing that
would otherwise have been "fixed" the wrong way twice.

**Relationship to #575: confirmed.** #575 names exactly two gaps (`host:build` never invoked by any
build/dist script; nothing places the output where the resolver looks) and this closes both, by a
better route than #575's own suggestion. One remainder: #575 also asks for *"a visible degraded-mode
notice on Windows"* for the case where there is no backend at all. Not in this PR, which is fine —
but `Fixes #575` will auto-close it, so either drop that from #575 first or note it as a follow-up.

---

### The one thing I want answered before this merges

> **macOS** is unchanged in practice — bundled tmux keeps winning

That is true of the **backend-selection** predicate you quote (`pty-manager.ts:4463`), which is gated
on `(win32 || !this.tmuxPath)`. It is not true of `sessionHostSupported()` as a whole. Two call sites
are gated on **that flag alone**, with no `!this.tmuxPath`:

- `pty-manager.ts:2367`, inside `sessionExists()`:
  `if (this.getSettings().tmuxEnabled && sessionHostSupported()) probes.push(sessionHostHasSession(…).catch(() => true))`
- `pty-manager.ts:4272`, inside `listNodetermSessions()`:
  `this.getSettings().tmuxEnabled && sessionHostSupported() ? sessionHostListSessions() : …`

Every *other* host call site in that file carries the guard — 3765, 3882, 3960, 4008, 4694 are all
`live?.sessionHost || !this.tmuxPath`. These two look like they were written under the assumption that
`sessionHostSupported()` is a proxy for "Windows", which it has been in practice, because the bundle
has never shipped. **This PR is what makes it stop being true.**

Both reach `getClient().request()` → `ensureConnected()` → `doConnect()` → `spawnSessionHost(…)`
(`session-host-client.ts:360`). So after this lands, on a fully tmux-backed Mac:

1. The first relay `projects.list`, or the first node registration
   (`main/remote/host-service.ts:301` calls `pty.sessionExists(nodeId)`), **spawns a session-host
   process** that has nothing to do and no session to own — with a 30 × 150 ms connect loop behind it
   if it does not come up.
2. Worse: `sessionExists`'s host probe is `.catch(() => true)`. That fail-safe direction is correct
   *for tmux*, where unprobeable means "assume warm, don't type into it" — the function's own doc
   comment says so. Applied to a host that isn't there, a genuinely **cold** node answers
   `exists: true` to the relay host, and the caller "treats it as a warm join and types nothing into
   it."

The blast radius is the phone-relay path only (`sessionExists` and `listNodetermSessions` have no
other callers — I checked), so this is not a canvas-wide cold-restore break. But it is a real
behaviour change on macOS, on the surface where it is least visible, and the PR body currently says
there isn't one.

Simplest resolution: add `!this.tmuxPath` (or `live?.sessionHost || !this.tmuxPath`) at those two
sites, matching the other five. If you think the mac/tmux path *should* now consult a host, that is a
defensible position too — but then say so in the body, because it is a new fact about macOS.

### Two smaller things

**Linux.** `dist:linux` and the release Linux job now ship the bundle, so a Linux box with **no tmux**
silently moves from a plain direct PTY to session-host-backed sessions. You state this ("Linux gains
one wherever no system tmux is installed"), and it is the outcome you want — but it is the first time
that backend runs in a shipped non-Windows build, and `server:dev` (`npm run build && …`) flips the
Server Edition the same way. The Windows smoke above proves packaging; it proves nothing about that.
A tmux-less Linux run before the next release would be worth more than another Windows one.

**`ci.yml:58`.** The standalone `npm run host:build` step is now redundant with `npm run build` at
`:34`. Harmless, but it no longer proves what it was there to prove. Leave it or drop it — just don't
read it as independent coverage.

---

### Merge ordering note

**#568 should land before this one.** Its `src/session-host/session.ts` half — the launch-line echo
verification — is currently dead code in every packaged build, for exactly the reason this PR fixes.
This is what makes it live. Merging #579 first would ship the first real Windows session host with the
tail-only check #568 replaces.

**Verdict: the packaging change is right and now CI-verified. Holding on the
`sessionExists`/`listNodetermSessions` question above — a one-line guard, or a corrected claim in the
body.**

ezwep:
Thanks for running the smoke workflow — and for both review findings. The first one is now fixed on
this branch; the second I have deliberately left alone. Reasoning for each.

## `sessionHostSupported()` without the backend guard — fixed

You are right, and it is worse than an inconsistency: `hasSession` and `listSessions` go through
`request()`, and `request()` is what **establishes the connection on a cold client**
(`session-host-client.ts:1319` says so in a comment) — which spawns a host when none is published.
So a read-only existence probe has the side effect of starting a daemon.

That was invisible before this PR, because no packaged build had a bundle to spawn. Shipping it is
exactly what makes those two sites reachable, so this belongs here rather than in a follow-up.

`hostBackendEligible()` lifts the spawn path's own predicate out and both sites now use it.

**One trade-off, stated rather than hidden:** on a Mac that *has* tmux, those two queries no longer
consult the host at all — so a session that is host-backed on such a machine (possible only if it was
created while tmux was absent) becomes invisible to them. The alternative that avoids the trade-off
is a **non-spawning** client path — connect only when a host is already published, never launch —
which would let the probe stay unconditional. That is a change to your connect logic rather than to
the packaging, so I did not make it unilaterally; say the word and I will.

## `sessionExists`'s `.catch(() => true)` — not changed

Left as is on purpose. It is the same fail-safe direction `tmuxSessionExists` takes, and the comment
above it argues for that direction explicitly. It predates this PR, and flipping it is a behaviour
decision about what the relay should believe when a probe fails — not something to fold quietly into
a packaging change. Happy to do it as its own PR if you want it flipped; with the guard above, its
blast radius on macOS is now nil anyway.

## Merge order

Noted, and agreed on the reasoning: #568 before #579, because `src/session-host/session.ts` is dead
code in every packaged build today and this is what revives it. I have no way to enforce that from
my side — flagging it here so it is recorded on the PR itself and not only in the issue.

## The degraded-mode notice

Also agreed it is not in here, and I think it should not be: it is a different change (a UI notice)
with a different test story, and it protects a case this PR is trying to make impossible. Happy to
open it separately if you want it tracked — say which surface it belongs on and I will.

Re-verified after the change on Windows 11: typecheck clean, `session-host-launcher` +
`no-electron` 9 passed, `pty-manager` suites 18 passed.

ezwep:
Answering your open question, and closing the loop on the two housekeeping points.

## The two unguarded `sessionHostSupported()` call sites — fixed

You are right, and it is worse than an inconsistency: `hasSession` / `listSessions` go through
`request()`, and `request()` is what *establishes* the connection on a cold client — so a read-only
existence probe **spawns a host**. Shipping the bundle is exactly what makes those two sites
reachable, so it belongs in this PR rather than a follow-up. `hostBackendEligible()` lifts the spawn
path's own predicate out and both sites use it (commit `47caf9e0`, pushed before your review
reached me — comment above has the detail).

One trade-off I stated there rather than hid: on a Mac that *has* tmux, those two queries no longer
consult the host at all, so a session that is host-backed on such a machine — possible only if it
was created while tmux was absent — becomes invisible to them. The alternative that avoids the
trade-off is a **non-spawning** client path (connect only when a host is already published, never
launch), which would let the probe stay unconditional. That is a change to your connect logic rather
than to packaging, so I did not make it unilaterally. Say which you prefer and I will send it.

`sessionExists`'s `.catch(() => true)` is untouched on purpose: same fail-safe direction as
`tmuxSessionExists`, predates this PR, and flipping it is a behaviour decision about what the relay
should believe when a probe fails. With the guard above, its blast radius on macOS is nil.

## `Fixes #575` — removed

Good catch, and the right call: #575 asks for two things and this covers one. The body now says
**Addresses #575** with a non-closing keyword, and names the degraded-mode notice as the remainder —
GitHub's linked-issue list for this PR is now empty, so the merge cannot auto-close it.

Happy to open the notice as its own issue if you would rather track it separately than leave it
living inside #575; say which surface it belongs on and I will write it up.

## Merge order

Understood: **#568 before #579**, because `src/session-host/session.ts` is dead code in every
packaged build today and this is what revives it. Recorded here as well as on the issue so it is on
the PR that has to wait, not only on the one that explains why.

Thank you for dispatching `win-package-smoke.yml` against the branch — an independent Windows runner
confirming `host:build` inside `dist:win` is worth more than my own box saying the same thing.

ezwep:
A consequence of this PR that showed up the first time someone installed a build carrying it, and I
think you need it before merging: **the session host makes the installer hang.**

## What happened

A tester quit nodeterm, ran the installer, and it did not complete. Task Manager still showed
`nodeterm.exe`. Only after killing that process did the install go through.

## Why, and the part I can prove

The host is spawned as `process.execPath` — the Electron binary — so **it is itself
`nodeterm.exe`**. Verified on a running install carrying this branch:

```
6768   MAIN            11:37:36
8232   SESSION-HOST    11:37:37     <- nodeterm.exe … app.asar\out\session-host\host.cjs
6736 / 16488 / 17780   electron-child
```

And it is *supposed* to outlive the app: `spawnSessionHost` detaches and unrefs precisely so
`app.quit()` cannot take it down, and `host.ts:287` only exits after a grace window **once its last
session is gone**. In normal use there are sessions, so quitting the app leaves a live
`nodeterm.exe` behind. By design — that is the whole feature.

The build is `nsis` with `oneClick: true`, whose model is "close the running app, replace the files,
relaunch it". To that installer a detached host is indistinguishable from the app itself: same
executable name, same install directory.

**Stated honestly:** the process tree and the deliberate survival are measured. That the installer
specifically waits on or fails to kill that process is inferred from the symptom — I could not read
the NSIS logic out of the compressed installer to confirm which of the two it does.

## Why it is worth solving before this merges rather than after

Before this PR no packaged build ever left a `nodeterm.exe` running after a quit, because no host
shipped. This is what makes it start happening — and it will happen on **every** install and every
update on Windows, not just the first. `electron-updater` drives the same NSIS flow, so the built-in
updater is likely to inherit it. It also lands on top of #454, which is already about the Windows
build being awkward to update.

## Directions, none of which I want to pick for you

- **Give the host a distinct executable name** — spawn it from a renamed copy of the Electron binary
  (`nodeterm-session-host.exe`) so the installer's process match no longer catches it. Cleanest from
  the installer's point of view; costs a copied binary in the package.
- **Have the installer shut the host down first**, gracefully, and let the next launch respawn it.
  Sessions would not survive that particular install, which is a fair trade for an install.
- **Custom NSIS include** that excludes the host from the kill. Fiddly: NSIS matches on image name,
  and telling the two apart needs the command line.

Happy to implement whichever you prefer. I have not touched it here because it is a packaging and
process-identity decision, not a bug in the resolution logic this PR is about — and because getting
it wrong means an installer that either kills live sessions or refuses to run.

eneskirca:
Re-reviewed on Linux in an isolated worktree off `origin/main`. The packaging change and the follow-up commit are both right.

**The `sessionExists`/`listNodetermSessions` question is resolved.** `47caf9e0` lifts the spawn predicate into `hostBackendEligible()` and both read-only sites now use it, so a probe can no longer start a session-host on a tmux-backed Mac. Verified locally: typecheck clean; `session-host-launcher` + `no-electron` (core + server) + `pty-manager` + `session-host` suites = 92 passed. The stated trade-off (a host-backed session on a tmux Mac becomes invisible to those two queries — reachable only if created while tmux was absent) is acceptable and correctly documented; the non-spawning-client alternative is fairly deferred to whoever owns the connect logic. Leaving `sessionExists`'s `.catch(() => true)` as-is is also fine given the guard makes its macOS blast radius nil.

**The one thing I'd want settled before this merges is the installer-hang you flagged 2026-09-02** — and I think you're right that it belongs here, since this PR is what first ships a `nodeterm.exe` that outlives a quit. I confirmed the mechanism against the electron-builder 26.15.3 NSIS template on this checkout:

- `nsis` is `oneClick: true`, `perMachine: false`. `_CHECK_APP_RUNNING`/`FIND_PROCESS` match by image name / by path under `$INSTDIR`, and the detached host is spawned from `process.execPath` = `<INSTDIR>\nodeterm.exe`, so it matches on **both** axes.
- On an update, `KILL_PROCESS ... 0` (graceful) runs first — the host has no window and won't close on that — then force-kill, with retry/`MessageBox` loops in between. That is the hang/"still running in Task Manager" symptom, and `electron-updater` drives the same NSIS flow, so the built-in updater inherits it.

Of your three directions, **the distinct executable name is the cleanest** and the one I'd recommend: spawn the host from a renamed copy of the Electron binary (`nodeterm-session-host.exe`) so neither the image-name nor (if you also want) the `$INSTDIR` path match catches it, and sessions survive the install. The "shut the host down first" option loses sessions on every update — acceptable for a manual install, worse for the auto-updater path this feeds. The custom-NSIS-include route can't tell the two `nodeterm.exe` apart by image name, so it's the most fragile.

**Merge order: #568 before #579** — agreed and recorded on both PRs.

Verdict: **the code is merge-ready; holding only on the installer-identity decision above, which is a maintainer call.** Note the still-open verification debt in your own comments: a real tmux-less Linux `dist:linux` run (first shipped non-Windows use of this backend) and end-to-end session-survival across an app restart. Neither is reproducible on this Linux box beyond the packaging half your CI run already proved.

eneskirca:
Decision on the installer hang you reported: let's give the session host its **own executable name** (`nodeterm-session-host.exe`) rather than special-casing the NSIS script. Reasoning: the one-click installer identifies the running app by image name and `$INSTDIR` path (checked against electron-builder 26.15.3's template), so a detached host that is also `nodeterm.exe` will always be caught by that wait/kill loop; a distinct name takes it out of the loop structurally instead of by exception. The code itself is merge-ready as reviewed (the `hostBackendEligible()` guard resolved the earlier point) — once the rename lands we can merge. Tmux-less Linux and restart-survival E2E remain a device-verification item on our side.

ezwep:
Rename pushed — `nodeterm-session-host.exe`, via a hard link, per your decision.

## How, and why a link rather than a copy

`hostLauncherPath()` spawns the host through a hard link beside the binary. Same file, so no second
copy of a 224 MB executable, no admin, and Windows reports the process under the **link's** name.
Measured on a packaged install, first with a trivial script and then with the real bundled
`host.cjs`:

```
PROCESNAAM: nodeterm-session-host.exe
```

It has to sit in the install directory — Electron needs its sibling DLLs, `locales/` and
`resources/`, and a link in `%TEMP%` exits immediately (`0x80000003`, measured before moving it).

## Three failure modes, two of which review found rather than I did

- **Stale link after an update.** An installer replaces `nodeterm.exe`; a link to the old inode would
  keep launching the previous version forever. Re-created whenever it does not point at the current
  `execPath`.
- **Identity compared with BIGINT stats.** An NTFS file id is 64-bit and routinely exceeds
  `Number.MAX_SAFE_INTEGER`, so the default numeric `ino` reports two *different* files as the same
  one — here that means launching a stale or unrelated binary and believing it is current. A zero id
  proves nothing and re-links rather than trusting the comparison.
- **Async spawn failure.** `spawn()` reports `ENOENT` on the `'error'` event, which the surrounding
  try/catch cannot see — and an unhandled `'error'` on a ChildProcess takes the main process down.
  Reachable: two app processes can race on the alias, and the loser can unlink it between the winner
  creating it and spawning through it. There is now a listener, plus one retry through `execPath`
  when the alias is what failed.

Everything else — read-only install dir, a filesystem without hard links, a directory squatting the
name, a link that cannot be removed — returns `execPath`, which is today's behaviour. Off Windows the
function returns before touching the filesystem.

`hostLauncherPath(execPath, platform, fs)` takes its filesystem injected, in the shape
`tmuxInstall(platform, hasCommand)` already uses here, so all eight fallback paths are unit-tested
without needing a 224 MB binary and a writable install directory. 19 tests pass across
`session-host-launcher` + `no-electron`; typecheck clean.

## What I have NOT proven

**A packaged upgrade with the host running.** The NSIS uninstaller template moves old files with
`Rename` and aborts on failure (`app-builder-lib/templates/nsis/uninstaller.nsh:70`, `:171`), so the
question is not only "is the image name matched" but "can the file be renamed while a process holds
it under another name". A running image has shared-file locking constraints that a second name does
not bypass. I would not claim the installer issue is closed until someone runs an actual upgrade
with a live host — I can do that on the next build and report, but it is worth stating rather than
assuming.

## One thing this does not fix, flagged rather than widened

`main/codex-relay-daemon.ts:557` also spawns a detached `process.execPath`, so it carries the same
image-name conflict for the same reason. Out of scope here; say the word if you want it in the same
change or as a separate one.

## Still open on this PR

`sessionExists`'s `.catch(() => true)` remains untouched, for the reason in my earlier comment — same
fail-safe direction as `tmuxSessionExists`, predates this PR, and flipping it is a behaviour decision
about what the relay should believe when a probe fails. Happy to send it separately if you want it
flipped.

eneskirca:
Re-reviewed, and first the good news: **the finding from my last round is properly closed.** `hostBackendEligible()` is now the one predicate both the spawn path and the probes share, and its comment names the exact failure — a request on a cold client establishes the connection, so an existence probe *spawns a host*. That was the thing that made "macOS is unchanged in practice" untrue, and it is no longer true of this branch. The ordering constraint is also satisfied: #568 merged on 2026-09-02, so the first packaged build that revives `src/session-host/session.ts` already carries its echo-verification fix.

**Your red `quality` run is not stale, and I measured it rather than guessing.** I merged this head into current `main` in a clean worktree (no conflicts), ran `tsc -p tsconfig.node.json` (clean) and the full suite:

```
Test Files  1 failed | 852 passed | 5 skipped (858)
     Tests  4 failed | 11664 passed | 51 skipped (11719)
```

All four are `src/core/session-host-launcher.test.ts`, and all four come from one line:

```ts
// src/core/session-host-launcher.ts:112
const link = path.join(path.dirname(execPath), WINDOWS_HOST_EXE)
```

The function takes an explicit `platform` argument, but builds the path with the **running** platform's `path`. On a Linux runner with `platform: 'win32'` injected, `path.dirname('C:\\Program Files\\node-terminal\\electron.exe')` is `'.'` — POSIX sees no separator in a backslash path — so the join returns the bare `nodeterm-session-host.exe` the test reports. On real Windows `path` *is* `path.win32`, which is why `quality-windows` is green and only `quality` is red. Production behaviour on Windows is correct; what is wrong is that the function's platform parameter is a lie everywhere else.

The fix is one line:

```ts
const link = path.win32.join(path.win32.dirname(execPath), WINDOWS_HOST_EXE)
```

This is also the house rule rather than my preference — CLAUDE.md: *"Resolve path separators / absolute-path checks / file-link dialects against the filesystem-owning core's platform, not the viewer's."* We merged the mirror image of this a few hours ago (#776, a test that assumed vitest never runs on Windows), so the repo has just paid for both directions of the same assumption.

Please push that plus a rebase onto `main`, and I will take it — everything else in the branch stood up to the merged run.

eneskirca:
@ezwep — we carried this the last step so it can land. **The credit stays yours.** It is still your branch and your commits; nothing was rewritten or force-pushed.

## What we pushed

Two commits on top of `61070790`, fast-forward only:

1. **`cca75f1e` — `fix(core): build the host link path with path.win32`.** This is the one-liner from the last review:
   ```ts
   const link = path.win32.join(path.win32.dirname(execPath), WINDOWS_HOST_EXE)
   ```
   The platform is a parameter, so the separators now follow it rather than the OS running the code. The only test change is a comment saying the `hostLauncherPath` suite is **deliberately not** platform-gated. The existing literal `C:\Program Files\…` expectations are now the proof on both the Linux `quality` job and `windows-latest`.
   - Mutation-checked on Linux: reverting just that line turns exactly those four tests red again (`4 failed | 13 passed`); with the fix, all 17 pass.
2. **`130f9bf7` — merge of current `main` (`6a332e0d`, v0.3.7).** It is a merge, not a rebase, so your history is untouched. There were no conflicts.

Local run on the merged head (Linux):
- `npm run typecheck`: clean.
- `npm run host:build`: `out/session-host/host.cjs` 370.8 kB.
- `vitest run`: `858 passed | 5 skipped` files, `11735 passed | 51 skipped` tests, 0 failed.

CI is re-running on the push. Merging stays with @eneskirca.

## What this PR still does not prove — Windows device checklist

Nothing below has been observed on a Windows device **at this head**. Your 0.3.4-based verification covered packaging, the in-asar `node-pty` resolution, and the host starting. It predates the hard-link rename and the three weeks of `main` merged here. Anyone on Windows 11 can take these — @ezwep, @FabricioCasali:

```
Packaging
 1. `npm run dist:win` on this head: exit 0, and app.asar carries out/session-host/host.cjs.
 2. Fresh install, open a terminal node: Task Manager shows a `nodeterm-session-host.exe` process
    (the hard link), NOT a second `nodeterm.exe`, with the node's shell beneath it.

Persistence — the point of #575
 3. Start `claude` in a node, quit nodeterm fully, relaunch: does the node re-join the SAME live
    session (conversation still there), with no fresh prompt?
 4. Switch to another project and back after > 5 min (past the park window): same session?
 5. Delete the node (×): the host drops the session, and it does not come back on relaunch.

Install location / upgrade — the hang this rename exists for
 6. Per-user install (%LOCALAPPDATA%\Programs): upgrade with a host process running. Does the
    installer finish without the "nodeterm is running" hang?
 7. Per-MACHINE install (Program Files, not writable without elevation): the link cannot be
    created, so hostLauncherPath falls back to nodeterm.exe. Does the host still start? Does the
    upgrade hang come back? (Expected: yes, it does. Say so if you see it.)
 8. After an upgrade: is the stale link replaced — does Task Manager show the NEW version's binary
    under the link name?

Other platforms (must be unchanged)
 9. Packaged macOS build with tmux: NO session-host process is spawned at launch, on project
    switch, or when a phone lists sessions over the relay (hostBackendEligible).

Phone over the relay (the reason this blocks Windows phone attach)
10. With phone access on, attach to a running agent node from the iPhone over the relay. Does it
    join the live session (the agent's screen) rather than a blank PowerShell prompt?
```

Thank you for the packaging work — the measured in-asar `node-pty` resolution is what made this safe to carry.

PR #584, opened 2026-09-01
feat(kanban): pull request cards, read-only

Temikus:
Phase 2 of #462, on top of #513's registry: pull requests as a second board lane, read-only.

**Review decision:** is harvesting pull requests from the existing issue poll the right fetch, and is the issue lane unchanged?

## The endpoint call

`/repos/{repo}/issues` returns pull requests and `listIssues` used to skip them. Keeping them costs no extra requests and inherits the `since` watermark, ETags, poll, and cache the issue lane already has. I measured `/repos/{repo}/pulls` first: it ignores `since` (a day-old value returned the same 100 items as none), its items are ~25 KB against ~7 KB, and it adds only `head`/`base` for a second ETag/paging/cache lineage.

Harvested fields are `draft` and `pull_request.merged_at`, the latter being the only thing separating merged from closed. `head` is deliberately absent: phase 3 asks about one branch, which `/pulls?head=owner:branch` answers in one request.

## Two things worth your eye

**`kind` protects the issue lane.** One snapshot now holds both kinds, so `GitHubIssueQuery.kind` (absent = issue) keeps issue items and counts as they were. `moveIssue` refuses a pull request by number, since the two share a number space.

**An overflow evicts pull requests first.** `MAX_ISSUES` and the 64 MB bound are shared, so without this a large repository could tip the whole board into `incomplete`. `evictPullsToFit` drops pull requests oldest-updated first and marks `pullsTruncated`; the existing path fires only if issues alone still miss the bound, and the board says the lane is a subset.

The `pulls` source is `readOnly` in the registry: no drag, no move control. The filter's `GitHub` button becomes `Issues`, beside `Pull requests`.

## Verification

<details>
<summary>Tests, typecheck, surfaces</summary>

- Tests cover eviction order and refusal, `kind` splitting, `moveIssue` refusing a pull request, client decode of draft/merged/malformed, cache validation of `pull`, and the read-only card. Three existing tests changed on purpose (this phase changes behaviour): the discard pin, the filter's buttons, and the store's query count, 3 to 6, both kinds served from the one cached snapshot.
- A live `/issues?state=closed` page through the real client: 20 items, 18 pull requests, merged detected, nothing rejected.
- `npm run typecheck` clean. Suite: 30 failed / 9183 passed, against 31 / 9164 on clean `main` here (unpatched `node-pty`, real-tmux/ssh suites).
- Desktop and Server Edition share this code and `query` rides the bridge opaquely, so no new bridge member. Mobile: N/A.
- **Not run on device** (native modules unbuilt here), so the lane is unexercised by hand.

</details>

Follow-ups unchanged: 3. explicit attachment plus the canvas chip, where `head` gets fetched. 4. `create-from-nodeterm`.


eneskirca:
Reviewed against the phase plan agreed in #462, on top of #513's registry. Fetched the branch into an isolated worktree: **`npm run typecheck` clean; suite 9332 passed / 1 failed**, and that one is `context-tail.test.ts` — a timing flake unrelated to this PR (passes on its own re-run here, and the `pr555` worktree off the same base runs it green).

The endpoint decision is right and the measurement behind it is the part that makes it right: `/pulls` ignoring `since` is what would have cost a second ETag/paging/cache lineage, and you checked instead of assuming. `kind` keeping the issue lane's items and counts byte-identical, `moveIssue` refusing a pull by number (shared number space), the eviction order, and the `pullsTruncated` stickiness surviving until a full reconciliation can honestly clear it — all four are the load-bearing ones and all four are handled. `evictPullsToFit`'s binary search is correct at the ends (`high = length-1` can't return the full set, which is exactly the set already known not to fit), and `byNumber` keying is safe because issues and PRs share one number space per repository.

Two things to change before merge, then this is good to go.

**1. `readOnly` is declared in the registry and read by nobody.**

`kanbanSources.ts` gains `readOnly: true` on `pulls`, `kanbanSources.test.ts` asserts it, and no production code branches on it. What actually prevents a write today is the `Drag` union having no `pulls` variant, plus `GitHubPullCard` simply not having drag props — and `isProviderDrag` branches on `placement`, never on `readOnly`. So the fact "the board never writes this source" lives in two places, and the registry copy is the decorative one.

That is the acceptance criterion for this phase inverted: the point of one registry entry is that the next source declares its behaviour and the consumers follow. Right now a source could declare `readOnly: false` and change nothing, or declare `readOnly: true` and still be draggable the moment someone widens the union. Either consume it — gate the card's drag/move affordance on `kanbanSource(id).readOnly` so the declaration is what does the work — or, if the type-level guard is deliberately the enforcement, say so in the field's doc comment and in the test, so the next reader does not trust a flag that decides nothing.

**2. The `issue-limit` degraded path drops issues for pull requests — the one invariant this PR states.**

```ts
if (!trimmed) {
  await this.incomplete(
    captured, state, 'issue-limit', [...byNumber.values()].slice(0, MAX_ISSUES), …
```

`byNumber` now holds both kinds, and `incomplete()` stores what it is handed verbatim as `partialIssues`. So in the degraded state the slice is a mixed set and pull requests occupy slots that used to hold issues: the issue lane shows *fewer* issues than it does on `main` today. `evictPullsToFit`'s own docstring is the rule being broken here — "a repository big enough to overflow must degrade in the new half, not turn the board it already had read only".

Filtering to issues before the slice is enough (`[...byNumber.values()].filter((item) => !item.pull).slice(0, MAX_ISSUES)`), and the `byte-limit` call below hands `harvest` over with the same dilution. Reached only above 10 000 items / 64 MB, so it is rare rather than urgent — but it is the exact case the eviction was written for, and it is the one path that skips it.

**Nit — `pullsTruncated` in the view is two facts wearing one name.**

`Object.values(github?.pullPages ?? {}).some((page) => page.partial)` is also true in the `partialIssues` state, where nothing was evicted at all. The sentence stays defensible (the lane really is a subset there), but the service already distinguishes the two and the view flattens them back together. If the banner is meant to be "we evicted PRs", `partial` is not the field that says that.

**Smaller notes, no action needed unless you disagree**

- `connect`/`reload` now issue `2 × (columns + 1)` queries. Both passes read one cached snapshot and the whole effect is gated on `board.github`, so this is renderer↔core round trips rather than network — worth knowing, not worth changing.
- `pullFrom` returns `null` for `pull_request: null`, which rejects the entire page via `malformed-response`. Consistent with how the client already treats every other malformed field, so I would leave it; flagging only because GitHub omits the key rather than nulling it, so the branch is unreachable in practice.
- `KanbanSourceFilter` renders from `KANBAN_SOURCES` and the `Issues` / `Pull requests` split falls out of the registry with no new string at a call site — that half of the criterion is met cleanly.

Ungrouped, `validKanban` and the assignment path are untouched, and this phase introduces no session↔PR relationship, so the "never derived from a transcript" rule has nothing to violate yet — the right shape for phase 3 to build on.

Temikus:
@eneskirca thanks for merging, just to check - do you want me to address the 2 tail items above in a separate PR?

Issue #563, opened 2026-08-31
Windows: the UI calls the local machine "This Mac" — including in the license and billing copy


ezwep:
## What happens

Settings → Accounts groups the local Codex accounts under the heading **"This Mac"** — on a Windows
machine.

It is not a one-off slip: `This Mac` / `this Mac` / `your Mac` / `other Macs` appears in **34
user-visible strings across 15 files**, and one of them is the name of the local session itself.

## Where

**Accounts**
- `src/renderer/components/settings/sections/AccountsSection.tsx:662` — the group label `'This Mac'`
- same file `:599`, `:601` ("Add account — On this Mac"), `:635`, `:674` ("Setting up on this Mac…")
- `src/renderer/lib/accountPresentation.ts:58` — tooltip `originDetail = 'This Mac'`
- `src/renderer/session/localSession.ts:11` — the local session is literally named `'This Mac'`

**License and billing — where it does the most damage**
- `src/renderer/lib/licenseCopy.ts:62` — *"**This Mac is not authorized on this license**, so the key
  and device count are unavailable"*
- `licenseCopy.ts:57`, `:78`, `:97`, `:157`, `:182` ("On a Mac (or phone) where Pro is active…")
- `LicenseSection.tsx:154`, `:229`, `:242`, `:243` ("Deactivate Pro on this Mac?", "Your other Macs…")

**Team access / upgrade**
- `TeamAccessSection.tsx:84`, `:119`, `:188`, `:193` ("Share this Mac with your team")
- `UpgradeDialog.tsx:20` ("3 team seats to share this Mac")

**Remote and phone**
- `PhonePairPopover.tsx:100`, `:141`, `:147` ("Reach this Mac from anywhere")
- `RemoteAccessDialog.tsx:107`, `:140`; `RemoteSection.tsx:96`, `:137` ("can run commands on this Mac")
- `PhoneSection.tsx:155`, `:333`

**Other**
- `PresenceIdentitySection.tsx:47` — "other people see when they connect to this Mac"
- `OnboardingFlow.tsx:458` — "makes the capsule sit flush on your Mac"

**Genuinely macOS-specific — leave alone**
- `PtyPressureBanner.tsx:120` — really is about `kern.tty.ptmx_max` and the macOS password prompt.

## Steps to reproduce

1. Start nodeterm on Windows.
2. Settings → Accounts.
3. Look at the heading above the Codex accounts.

[...]

PR #585, opened 2026-09-01
fix(grok): read the session title from the keys summary.json actually has

xaviguardia:
## Summary
- replace the constructed summary fixture with a redacted Grok 1.0.13 capture
- prefer `generated_title` and fall back to `session_summary`
- remove the unsupported `title` guess and close the documentation checklist item

## Why
A real `summary.json` has no `title` field. Accepting that guessed key could let foreign or hand-edited data override Grok's generated session name. The captured fixture preserves the real key set and value types, with paths and identifiers redacted.

## Testing
- `npx vitest run src/core/grok-session.test.ts`: 13/13 passed
- required mutation without `generated_title`: 4 tests failed
- `npm run typecheck`: passed
- `npm test`: exactly the baseline — 681 files, 9,342 tests, 23 pre-existing failures
- `git diff --check`: passed

## Surfaces
Desktop and Server consume the shared core metadata reader. Mobile has no protocol or UI change in this PR; tagging @eneskirca for visibility.

## Not verified
- the reverse rename path—whether renaming a node updates Grok's own title—was not verified and remains outside this read-path change

### DIFF
--- a/docs/grok-agent.md
+++ b/docs/grok-agent.md
@@ -266,22 +266,31 @@ resolve **nothing** there (`GROK_ENCODED_CWD_MAX_BYTES`); a session id must matc
 
 The session **name** (what `/resume` shows and what a node title with `titleAuto` adopts) is read by
 `core/grok-session.ts` → `pickGrokSessionMeta` over `summary.json`, in preference order
-`TITLE_KEYS = ['generated_title', 'session_summary']`, plus `current_model_id` as the model. Reads are capped
+`TITLE_KEYS = ['generated_title']`, plus `current_model_id` as the model. Reads are capped
 at 256 KB and answer `null` — never a throw — for an absent, oversized or unparseable file.
 Resolution is a **direct open** of the directory a hook told us about: `rememberGrokSessionDir` /
 `grokSessionDirFor` / `forgetGrokSession` keep a bounded (512-entry, least-recently-seen-evicted)
 `sessionId → dir` map, populated by the shells' raw listeners.
 
-**The fixture is a real capture.** `src/core/__fixtures__/grok/summary.json` was captured from grok
-1.0.13 at `$GROK_HOME/sessions/<encoded-cwd>/<session-id>/summary.json` on 2026-09-01. Personal
-paths and ids are redacted, while every key and value type is preserved. `generated_title` and
-`session_summary` deliberately use different redacted text so the fixture proves which key wins.
+**The fixture is a real capture, with one authored value — and that value caused a wrong call.**
+`src/core/__fixtures__/grok/summary.json` was captured from grok 1.0.13 on 2026-09-01; personal
+paths and ids are redacted and every key and value type is preserved. But its `session_summary` was
+REWRITTEN as a sentence so the fixture could prove which key wins, and a reader later took that
+sentence for grok's own output and reasoned from it. Measured across 49 local `summary.json` files:
+`session_summary` is never a sentence — 0 to 49 characters, mean 23, and identical to
+`generated_title` in 28 of the 29 files that carry both. It is a title, not prose. The fixture now
+carries a short title-shaped value, because a fixture that does not look like the real thing is a
+loaded trap for whoever reads it next.
 
-The capture has no `title` field. `generated_title` is grok's session name and wins;
-`session_summary` is the human-readable fallback before that title exists. Keeping the old guessed
-key would let a hand-edited or foreign `title` override the name grok actually generated. The
-fixture test's required mutation — removing `generated_title` from `TITLE_KEYS` — fails on the
-captured title, so the precedence is behaviorally pinned.
+The capture has no `title` field. `generated_title` is grok's session name and is now the ONLY title
+key: `session_summary` was dropped in favour of resolving to nothing. The reason is the rule this
+repo states for itself — an uncertain value must degrade to NOTHING, never to something else — and
+`pickGrokSessionMeta` feeds the node title with no length cap anywhere on the path, so a sentence in
+that field would reach the header, the sidebar, the kanban card, the window title and
+`project.json`. **The cost, stated rather than discovered:** of those 49 files, 20 carry
+`session_summary` and no `generated_title` — young sessions, before grok names them. Those now
+resolve to no name until grok generates one. No name is a state the UI already handles; a wrong name
+is not.
 
 **Not captured at all:** nothing, for the context meter. `signals.json` was the last entry here and
 it has been captured (22 sessions): it states the used count, the window total AND the percentage.
@@ -578,9 +587,11 @@ Session identity + restore
 11. Does the session chip fill in? The chip has exactly ONE source: the terminal-title OSC
     (`term.onTitleChange`, path/prompt-looking titles ignored). The summary.json poll feeds the node
     TITLE instead — that is item 12 — so a blank chip with a correct title is not a bug.
-12. **Verified 2026-09-01:** a real grok 1.0.13 `summary.json` stores the session name in
-    `generated_title`, with `session_summary` as a separate human-readable fallback and no `title`
-    field. The redacted capture in `src/core/__fixtures__/grok/summary.json` pins that shape.
+12. **Verified 2026-09-01, refined 2026-09-03:** a real grok 1.0.13 `summary.json` stores the
+    session name in `generated_title` and has no `title` field. `session_summary` exists alongside it
+    but is NOT a separate kind of value: across 49 local files it is identical to `generated_title`
+    in 28 of the 29 that carry both, and never longer than 49 characters. It is not read as a title
+    any more — see §5 for why resolving to nothing beats resolving to something unverified.
 13. Rename the NODE by hand: does grok's own title change (the `/rename` write leg)?
 14. Reboot (or `tmux kill-server`) and reopen the project: does the node cold-restore with
     `grok --resume <id>` and land in the SAME conversation, in the right cwd? Note that after
--- a/docs/grok-agent.md
+++ b/docs/grok-agent.md
@@ -282,6 +282,14 @@ sentence for grok's own output and reasoned from it. Measured across 49 local `s
 carries a short title-shaped value, because a fixture that does not look like the real thing is a
 loaded trap for whoever reads it next.
 
+**`parent_session_id` is absent because 1.0.13 does not write it, not because the capture missed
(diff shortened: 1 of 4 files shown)


eneskirca:
Thank you for actually capturing this — checklist item **14** in `docs/grok-agent.md` has been open since the grok branch landed precisely because nobody had a logged-in grok session to look at, and a redacted real `summary.json` is worth more than the rest of this diff put together. Dropping `'title'` is clearly right: it was documented in-tree as an explicit guess, the capture shows no such key, and the doc + fixture-provenance comment + checklist item are all updated together rather than leaving one of them lying about the state of the world. The differing redacted strings for `generated_title` and `session_summary`, so the precedence cannot pass by accident, is exactly the right way to build that fixture.

**My one substantive objection is the other half of the change: adding `session_summary` to `TITLE_KEYS`.**

It is a separate decision from removing `'title'`, and I do not think the capture supports it.

1. **It is not a title.** In your own fixture it is `"Prepare the project for release with a verified checklist."` — a sentence, against `generated_title`'s `"Plan the release"`. `pickGrokSessionMeta` has no length cap (claude's reader clamps names to 200 chars; this one clamps nothing), and its result is adopted as the node `title` while `data.titleAuto` holds, so it lands in the node header, the sessions sidebar row, the kanban card, the window title and — once the user edits anything — `project.json`. Before this PR a session with no generated title returned `null` and the node kept its own name (`Grok 1`). After it, the node header shows a full sentence.

2. **The ordering claim is an inference, not a capture.** The doc now says `session_summary` is "the human-readable fallback before that title exists", but your capture has *both* keys populated simultaneously, so it cannot establish that one precedes the other. It may well be true; it is just not what this fixture shows, and `docs/grok-agent.md` is otherwise careful to mark composed facts as composed.

3. **It goes against the rule the grok write-up sets for itself.** From CLAUDE.md's "Adding a new agent" section: *"A guess must degrade to nothing, never to something wrong. A title reader that cannot resolve returns `null` (the node keeps its own name)."* Falling back to a summary sentence is degrading to something else rather than to nothing — the same shape as the `'title'` guess you are removing, just in the opposite direction.

Concretely, I would suggest `TITLE_KEYS = ['generated_title'] as const`, keep the `null` fallback, and if the summary really is useful before a title exists, surface it somewhere that is not the title (the session chip, or a tooltip). If you would rather keep it, it needs a captured session that has `session_summary` and **no** `generated_title` to back the ordering claim, plus a length clamp so a two-line summary cannot become a node title.

Two smaller things:

- The security framing ("a hand-edited or foreign `title` could override the generated name") is a bit strong for a file that lives in the user's own `$GROK_HOME` and is written by grok. The correctness argument alone — the key does not exist — is enough and is the one that will still be true next release.
- `parent_session_id` is present in the old constructed fixture and absent from the capture. Nothing reads it today, but the doc's field list still names it as documented in 1.0.0; worth a word on whether 1.0.13 dropped it or it is simply absent on a root session, so the next reader does not treat its absence as a capture error.
- The "Not verified" note about the reverse rename leg is the right thing to say, and it matches reality: grok is in `RENAME_CAPABLE`, so item **15** stays open. Good that you did not quietly close it.

With `session_summary` dropped (or backed by a capture), this is a clean, well-evidenced fix and I would take it.

eneskirca:
Second-pass review (verified in an isolated worktree at a578e41: `npm run typecheck` clean, `grok-session.test.ts` 13/13, CI approved and green).

The branch has not moved since the review above, so the one blocking point still stands: **`session_summary` in `TITLE_KEYS`**. Concretely, this is what would make it mergeable:

```ts
const TITLE_KEYS = ['generated_title'] as const
```

plus the matching sentence in `docs/grok-agent.md` (the fixture keeps `session_summary`, it just stops being a title candidate). Everything else in the PR — dropping the guessed `title`, the captured fixture, closing checklist item 14 — is good as is.

One cross-PR note: none of #614/#617/#618/#619 touch `TITLE_KEYS`, so this rebases cleanly in either order.

xaviguardia:
Done — `TITLE_KEYS = ['generated_title']`, null fallback untouched.

Two things worth putting in front of you before you merge, because one of them is our fault and the other is a cost you may not have priced.

**The sentence you quoted is ours, not grok's.** `"Prepare the project for release with a verified checklist."` is the value in `src/core/__fixtures__/grok/summary.json` — written by us so the test could tell the two keys apart. Its provenance note said "captured from grok 1.0.13" on one line and admitted the strings "intentionally differ" on the next, with nothing marking which was which. That is a trap we set, and you walked into it; sorry.

Measured across 49 real `summary.json` files on a logged-in machine:

| | files | length min/max/mean |
|---|---|---|
| `session_summary` | 49 | 0 / 49 / 23 |
| `generated_title` | 29 | 22 / 49 / 38 |

Where both exist they are **identical in 28 of 29**. The longest value in the whole corpus is `Spanish count 1-40 one-per-line long explanations` — 49 characters. There is no sentence anywhere: in 1.0.13 `session_summary` is the same title, not prose.

**The cost:** 20 of those 49 files have `session_summary` and no `generated_title` — young sessions, before grok has generated one. Those 41% now show no name until it does.

**We made the change anyway**, because your rule holds independently of the premise: nothing on the path from this value to the node title, sidebar, kanban card, window title or `project.json` caps its length, so a guess that degrades to *something else* rather than to *nothing* is what this repo's own `CLAUDE.md` forbids. If a future grok does put prose there, the strict list is the safe place to be.

If you would rather keep the key with a length cap instead, that is a one-line change and we are happy to make it — this is your call, we just did not want it made on a fixture value we invented.

The fixture is corrected in this push: the discriminating value is now a 24-character title, inside the measured range and still distinct from `generated_title`, and the provenance note now names which values are fabricated and why.

xaviguardia:
Follow-up so the record matches the branch: this has been **rebased onto `main` since my last comment**, and I should have said so then rather than leaving it to be discovered.

Head is now `69db79d9` (was `461ef91b`), MERGEABLE. The rebase happened after #592 landed, which is what put it in conflict.

Two conflicts, both in `docs/grok-agent.md`, both in the device checklist and neither touching code:

- The `TITLE_KEYS` item. Your side still listed `TITLE_KEYS[0] = 'title'` as a guess to be verified; mine replaced it with the measured finding. Kept the measured text, but on **your** numbering (12/13/14) rather than the renumbering my branch carried — my numbers would have produced two item 15s once the following block resumed at 15. Also fixed the cross-reference in item 11, which pointed at "item 14" for the summary.json poll and now points at 12.
- The refined wording of that same item, from the follow-up commit: `session_summary` is no longer described as "a separate human-readable fallback", since across those 49 files it is identical to `generated_title` in 28 of the 29 that carry both.

Verified after the rebase: file set identical before and after (4 and 4, lists diffed), `TITLE_KEYS = ['generated_title']` intact, typecheck clean on both configs, `grok-session` + `agent-session-name` 22/22.

Separately — thank you for #666. That was the honest fix for the three copy suites, and pinning `navigator.platform` in the test rather than loosening the assertions keeps #563's point intact. Worth noting for anyone reading this branch's CI: it is rebased on a `main` from *before* #666, so those three still fail on a Mac checkout here. That is a stale branch, not a live failure — they go green the moment either lands.

xaviguardia:
@eneskirca `TITLE_KEYS = ['generated_title']` as you asked, head `69db79d9`, rebased, CI green.

No decision pending from you; the "your call" in my earlier comment is withdrawn. Ready to land.

eneskirca:
Decision from @eneskirca: **keep `session_summary` as a title candidate, with a length cap** — option (b) in your message.

Your measurement is what settles it: in 1.0.13 that key is a title, not prose (49 files, longest value 49 characters, identical to `generated_title` in 28 of 29 where both exist), and dropping it leaves **41%** of young sessions unnamed until grok generates one. That is a visible regression for the common case, and it buys protection only against a dialect change that has not happened.

So: `TITLE_KEYS = ['generated_title', 'session_summary']` with a cap, and the cap is the part that carries the rule this repo actually cares about — a value longer than the cap is **refused, not truncated** (a truncated sentence in a node title, a kanban card and `project.json` is exactly the "degrades to something wrong" failure; no value at all is the correct degrade, the node keeps its own name). Pick the cap from your corpus rather than a round number — your longest observed title is 49 characters, so something in the 60-80 range documents itself as "a title, generously" while a sentence never passes.

Please also keep the corrected fixture and its provenance note, and say in `docs/grok-agent.md` which key is preferred, that the second is accepted only under the cap, and why the cap exists — the next reader will otherwise re-litigate this from first principles. A test that pins a long `session_summary` yielding `null` (not a prefix) would be the one to add.

Thank you for measuring the corpus instead of arguing the principle; the 41% is what changed the answer.

eneskirca:
Rebase needed — and part of this has landed meanwhile, so the rebase should also shrink the PR.

On current `main`:

- `src/core/__fixtures__/grok/summary.json` is already the real capture (it carries `session_summary` and `generated_title`, no `title`).
- `docs/grok-agent.md` already records that no `title` key appears in a real `summary.json`.

What is still open here, and worth keeping:

1. **`session_summary` as the fallback.** `TITLE_KEYS` on main is `['title', 'generated_title']` — there is no `session_summary` leg, so a session whose generated title has not been written yet still reads as nameless. That is the real addition.
2. **Dropping the `'title'` guess.** This one is a judgement call rather than a bug, and main argues the other way in a comment right above the constant: a key that is absent from every file costs nothing, and keeping it first means a manual title would win the day someone confirms grok grew one. If you still want it removed, make that argument explicitly in the PR body (e.g. a foreign or hand-edited `summary.json` reaching the read path), because whoever reads this next will find main's comment first.

Could you rebase onto `main` and reduce the diff to what is left? Happy to take it as soon as it is green.

xaviguardia:
@eneskirca Rebased and shrunk — but this PR's head was our fork `main`, which cannot rebase onto yours without mixing unrelated history. The remaining change is in **#784**, cut down against current `main`.

What landed there, and only that:

- `session_summary` as the fallback (`TITLE_KEYS = ['title', 'generated_title', 'session_summary']`), which is the real addition you named
- an 80-character cap on that fallback: longer values are **refused, not truncated** (the cap from your 9 Sep note — longest observed title in the corpus is 49)

What this does **not** do: drop `'title'`. Main's comment already argues to keep it; we followed that rather than re-arguing it in the body.

Closing this one as superseded by #784.

xaviguardia:
Superseded by #784.

PR #587, opened 2026-09-01
fix(canvas): make snap-to-grid actually snap, on resize and on placement

basmilius:
With snap-to-grid on, a node could never be brought onto the grid: not by resizing it, and not by creating it there. Two causes, one rule.

## Resizing adds to the grid instead of snapping to it

React Flow snaps the resize DELTA, never the resulting edges. From `getDimensionsAfterResize` (`@xyflow/system`):

```js
let distX = Math.floor(isHorizontal ? xSnapped - startValues.pointerX : 0)
const newWidth = startWidth + (affectsX ? -distX : distX)
```

Both pointer reads are snapped, so `distX` is always a grid multiple, and it is **added** to whatever width the node started at. A node that starts at 613 with a grid of 20 can only ever reach 613 + n·20. The size the user is aiming for is unreachable, however carefully they drag.

`lib/resizeSnap.ts` re-snaps the resizer's changes *before* they are applied, so the node tracks the grid during the drag rather than correcting itself on release.

Two details worth reviewing:

- **`resizing` is the marker.** Only the resizer sets it; the `ResizeObserver`'s measurement changes carry no such field and pass through untouched. Forcing those onto the grid would fight the DOM about what every node actually measures.
- **A right/bottom-handle drag sends no position change**, so one is added when snapping moves the anchored edge. Without it the node keeps an off-grid `x` while its width is measured from a snapped one, and the right edge lands between grid lines again.

## New nodes were born off-grid

`placeAt` centres a node on the cursor, so its position is half a width off whatever the pointer hit, and a terminal's size comes from `settings.defaultNodeWidth`/`Height`, which are hand-editable and clamped but never rounded. That is where the resize problem came from in the first place.

`placeNode` snaps the whole box through the same `snapNodeToGrid` that the resizer and align-to-grid use, so all three land a node in the same place rather than each having its own rule. It replaces the `position`/`width`/`height`/`style` block that all ten node factories repeated verbatim, which is what kept the rule from having a single home.

## Scope

Everything is gated on `settings.snapToGrid`. With it off, nothing in either path runs: resizing and placement are byte-identical to before. `alignToGrid` is untouched.

**Surfaces:** desktop and Server Edition share the renderer, so both get it; mobile has no canvas.

## Tests

- `resizeSnap.test.ts` (9): in-flight and final changes snap, a measurement change is left alone, the appended position change appears only when the anchored edge actually moves, a left-handle drag is rewritten in place rather than gaining a second position, collapsed nodes keep their bar height.
- `workspace.placement.test.ts` (6): position and size both land on multiples, with and without a cursor, `style` stays in step with `width`/`height`, minimums are respected at a large grid, and snapping off reproduces the old box exactly.
- `nodeSizing.test.ts` (+2): one snap lands any off-grid box on multiples, and is idempotent.

Two of these came from a wrong assumption of mine that the tests caught: at x 13 with a proposed width of 633 the right edge is at 646, so it snaps to 640 and the **width becomes 620**, not 640. Rounding a small negative coordinate also produces `-0`, which would have ridden into node positions on every resize near the origin; both paths normalize it.

### DIFF
--- a/src/renderer/canvas/Canvas.tsx
+++ b/src/renderer/canvas/Canvas.tsx
@@ -644,14 +644,25 @@ const NO_EPHEMERAL: { ephemeralNodes: CanvasNode[]; ephemeralEdges: Edge[] } = {
  * inherits the agent's `parentId`), which is the whole point of storing an offset — grouping or
  * ungrouping the agent flips that space between absolute and group-relative, and a stored
  * position would then teleport the card by the group's own x/y.
+ *
+ * `snap` (0 = off) rounds the LAID-OUT default onto the grid: React Flow's `snapToGrid` only
+ * constrains a drag, so a fan-out card landed off-grid and only jumped into place once the user
+ * nudged it. A DRAGGED offset is left alone — it was already snapped at drag time if the mode was
+ * on, and re-rounding it here would move cards the user placed by hand while it was off.
  */
 const offsetFrom = (
   parent: { position: { x: number; y: number } },
   stored: { x: number; y: number } | undefined,
-  fallback: { x: number; y: number }
+  fallback: { x: number; y: number },
+  snap = 0
 ): { x: number; y: number } => {
   const off = stored ?? fallback
-  return { x: parent.position.x + off.x, y: parent.position.y + off.y }
+  const x = parent.position.x + off.x
+  const y = parent.position.y + off.y
+  if (stored || !snap) {
+    return { x, y }
+  }
+  return { x: Math.round(x / snap) * snap, y: Math.round(y / snap) * snap }
 }
 
 // Delivering an armed node's held launch (canvas-control `--after`) can lose the race against
@@ -1606,6 +1617,8 @@ export function Canvas() {
   // Selection state for ephemeral nodes (they live outside React Flow's managed nodes), owned by
   // the agent-nodes store so the cards themselves can set it — see `selectable: false` below.
   const ephSelId = useAgentNodes((s) => s.selectedId)
+  // Grid the laid-out fan-out cards round onto (0 = snap off) — see offsetFrom.
+  const ephSnap = settings.snapToGrid ? settings.gridSize || GRID : 0
   const { ephemeralNodes, ephemeralEdges } = useMemo(() => {
     // Common case: no /loop running and no subagents → return a stable empty result so
     // this memo (which depends on `nodes`, i.e. recomputes every drag frame) stays cheap
@@ -1644,7 +1657,7 @@ export function Canvas() {
         // with the group). Deliberately no extent:'parent' — the fan-out may hang below the
         // frame border without being clamped into it.
         ...(parent.parentId ? { parentId: parent.parentId } : {}),
-        position: offsetFrom(parent, ephemeralPos[lid], { x: -250, y: ph + 60 }),
+        position: offsetFrom(parent, ephemeralPos[lid], { x: -250, y: ph + 60 }, ephSnap),
         draggable: true,
         // NOT selectable: React Flow's rubber band would otherwise sweep a whole fan-out of cards
         // into the selection alongside the real nodes, and every selection action (Group,
@@ -1694,10 +1707,15 @@ export function Canvas() {
           type: 'subagent',
           // Same coordinate-space rule as the loop card above: inherit the agent's group.
           ...(parent.parentId ? { parentId: parent.parentId } : {}),
-          position: offsetFrom(parent, ephemeralPos[cid], {
-            x: (i % COLS) * COL_W,
-            y: ph + 60 + Math.floor(i / COLS) * ROW_H
-          }),
+          position: offsetFrom(
+            parent,
+            ephemeralPos[cid],
+            {
+              x: (i % COLS) * COL_W,
+              y: ph + 60 + Math.floor(i / COLS) * ROW_H
+            },
+            ephSnap
+          ),
           draggable: true,
           selectable: false, // see the loop card above
           selected: ephSelId === cid,
@@ -1728,7 +1746,7 @@ export function Canvas() {
     }
     return { ephemeralNodes: eNodes, ephemeralEdges: eEdges }
     // eslint-disable-next-line react-hooks/exhaustive-deps -- loopSig stands in for the byId read
-  }, [agentById, loopSig, ephemeralPos, ephSizes, ephExpanded, ephSelId, nodes])
+  }, [agentById, loopSig, ephemeralPos, ephSizes, ephExpanded, ephSelId, ephSnap, nodes])
 
   // Merge the persisted nodes with the ephemeral ones and the webview keep-alive pool once per
   // change (not per render), so React Flow's array-identity short-circuit holds while
--- a/src/renderer/canvas/Canvas.tsx
+++ b/src/renderer/canvas/Canvas.tsx
@@ -155,6 +155,7 @@ import { reopenVariants } from '../lib/reopenVariants'
 import { modelsForAgent } from '@shared/agents/model-gateway'
 import { useModelGateway } from '../state/modelGateway'
 import { viewportAtZoom } from '../lib/zoomReset'
+import { containerOrigin, snapPointInRootSpace } from '../lib/gridSnap'
 import { zoomFromPct } from '../lib/zoomPresets'
 import { CANVAS_MAX_ZOOM, CANVAS_MIN_ZOOM } from './zoom-limits'
 import { isSpaceRelease, spacePanKeydown } from '../lib/spacePan'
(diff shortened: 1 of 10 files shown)


eneskirca:
I checked the two mechanism claims against the vendored sources before reviewing the fix, and both are exactly right.

**The resize claim is verbatim accurate.** `getDimensionsAfterResize` (`@xyflow/system`, ~L3202):

```js
let distX = Math.floor(isHorizontal ? xSnapped - startValues.pointerX : 0)
const newWidth = startWidth + (affectsX ? -distX : distX)
```

Both pointer reads come from `getPointerPosition(..., { snapToGrid, snapGrid })`, so `distX` is a grid multiple added to an unsnapped `startWidth`. A node that starts at 613 with grid 20 is confined to `613 + 20n` for its whole life. That is a real, unreachable-target bug and the diagnosis is not a guess.

**The `resizing` marker holds.** Only `ResizeControl`'s `onChange`/`onEnd` emit `{ type: 'dimensions', resizing: true|false }` (`@xyflow/react` ~L4879 and ~L4903); the measurement path in `updateNodeDimensions` (`@xyflow/system` ~L1882) emits `{ id, type: 'dimensions', dimensions }` with no such field. So `typeof c.resizing === 'boolean'` really does separate "the user is dragging a handle" from "the DOM reported a size", and the test that pins it is the right test to have.

Three things I would like your read on.

**1. A group frame resized from the top or left may drag its children off by up to half a grid.** When a parent is resized from an anchored edge, `ResizeControl.onChange` receives `childChanges` and emits them as `type: 'position'` changes for the *children*, computed from the delta React Flow itself chose. `snapResizeChanges` then rewrites the parent's own position/dimensions to a different value, but the child position changes carry ids that are not in `boxes`, so they pass through untouched. The compensation and the frame no longer agree, and the children shift relative to the frame by up to `grid/2` per top/left drag — cumulative over several drags. Worth a test with a `group` node plus one child and a top-left handle; if it does drift, the fix is probably to shift the child changes by the same correction applied to the parent.

**2. Snapping is done in parent-relative space, and React Flow's drag snap is not.** `snapNodeToGrid` is fed `node.position`, which for a child is relative to its frame. `XYDrag` snaps `nextPosition` in flow (absolute) coordinates and only then converts to parent-relative. Since `groupSelectedNodes` creates frames at `(minX - 28, minY - 62)`, an off-grid frame origin is the normal case, so a node inside a frame will snap onto one grid while resizing and a different one while dragging. Same note as on #546; it may be worth solving once, in a shared helper, for both.

**3. Grabbing the right handle now moves the left edge.** Deliberate and documented, and I follow the reasoning — but it is the one part a user will *feel*. The alternative is to hold the anchored edge and snap only the edge under the pointer, which keeps the gesture honest at the cost of an off-grid width. I do not think this blocks anything; I would just like to know it was weighed against that, rather than falling out of `snapNodeToGrid` snapping the whole box.

Smaller notes:

- Inside `onNodesChangeWrapped` the new `const settings = useSettings.getState().settings` shadows the component-scope `settings`. Reading through `getState()` is right (no stale closure, no extra dep), but the shadowing makes the next reader check twice — `snapSettings` or similar would cost nothing.
- `nodes.find((n) => n.id === id)` runs twice per resizing id and again in the trailing loop. Irrelevant for one node, but this is on the per-frame resize path with a canvas that can hold a few hundred; a `Map` built once would make it O(n) instead of O(n·k).
- The `+ 0` for `-0` is a good catch, and the placement test pinning it against `project.json` is exactly the kind of test this repo wants.

Overall: strong change with a real, measured root cause and tests that pin the mechanism rather than the symptom. Points 1 and 2 are what I would want answered before merge.

basmilius:
Rebased onto #546, which is where the shared root-space helper now lives.

**1. Group frame resized from the top or left - reproduced, and it was worse than the report.** I wrote the test first. Frame at `(13, 7)` sized 613x371, grid 20, one child at frame-relative `(100, 100)`, i.e. root `(113, 107)`. On a top-left drag React Flow proposes the frame at `(-7, -13)` and compensates the child to `(120, 120)`, which restores root `(113, 107)` exactly. The snap then moved the frame to `(0, -20)` and left the child alone, landing it at root `(120, 100)` - off by `+7, -7`, within your "up to half a grid".

It does not accumulate the way you expected, though: once the frame is on the grid the correction is zero, so it is a one-time shift per off-grid origin rather than a per-drag creep. Still wrong, just bounded.

The test also surfaced a case that is not in your list. On a **right/bottom** drag React Flow emits no frame position change and no child changes at all, because it never moved the origin - but the snap does, pulling `x` from 13 to 20, and every child rides along for the same `+7`. Identical defect with no compensation to desynchronize.

So the fix is one rule rather than a patch on the top/left path: record how far the snap moved each resized node's origin away from what React Flow proposed, and shift its direct children by the opposite. Where React Flow already emitted a child change, that change is adjusted; where it emitted none, one is pushed. A zero correction pushes nothing, so an on-grid frame is byte-identical to before. Confirmed against `XYResizer`'s `childChanges` computation (`@xyflow/system` ~L3499) and `ResizeControl`'s emission of them as `type: 'position'` (`@xyflow/react` ~L4892) - the ids really are the children's, which is why they missed `boxes` entirely.

**2. Parent-relative vs. root-space snapping - fixed, and shared with #546.** `snapResizeChanges` now snaps through `snapRectInRootSpace` against `containerOrigin(node.parentId, nodes)`, so a node inside a frame snaps onto the same grid while resizing as `XYDrag` gives it while dragging. Measured before the change: an inner frame at `(41, 69)` inside an outer at `(-28, -62)` landed 12 px off the canvas grid. There is a test for the nested case.

**3. Grabbing the right handle moves the left edge - weighed, and kept.** The alternative you describe (hold the anchored edge, snap only the edge under the pointer) buys a gesture that feels more honest at the cost of an off-grid width, and an off-grid width is the bug this PR exists to fix: `getDimensionsAfterResize` adds a grid-multiple delta to an off-grid `startWidth`, so a node that starts at 613 stays at `613 + 20n` forever and the size the user is aiming for is simply unreachable. Snapping only the moving edge reproduces that for one of the two axes, permanently, for any node that was ever off-grid. Snapping the whole box makes every target reachable, and the anchored edge settles after the first drag rather than moving on every one. So it is a deliberate trade, not a side effect - I have said as much in the module doc so the next reader does not have to re-derive it.

**4. The frame now lands correctly instead of being corrected afterwards.** From testing: `groupSelectedNodes` created the wrapper at a fixed `(minX - 28, minY - 62)`, which knows nothing about the grid, so with snapping on *every* grouping minted a frame with an off-grid origin - exactly the case points 1 and 2 then have to compensate for on each later resize, and that `offsetFrom` has to compensate for on each fan-out card placed inside it.

With a grid, the padding is now raised to at least one full cell and the box is grown OUTWARD to the surrounding grid lines, so all four edges sit on the grid and the clearance around the members can only increase, never shrink. That last part needed a new primitive rather than reusing `snapNodeToGrid`: nearest-rounding pulls an edge inward by up to half a grid, so at grid 24 a left edge at x=20 rounds *right* to 24 and eats 4px of the padding. `expandRectToGrid` floors the left/top and ceils the right/bottom instead, and keeps the same min-size clamp.

The snap runs in root space through the same helper as the rest of this change, so a nested frame lands on the canvas grid rather than its parent's. `fitGroupToChildren` shares the box computation, because it carries the identical padding arithmetic and runs on the ancestor chain from inside `groupSelectedNodes` - left alone it would have pulled a just-snapped frame straight back off the grid. Members keep their absolute canvas positions throughout.

With snapping off, nothing changes: every path reproduces the original fixed-padding box exactly, and there is a test pinning the pre-change numbers as well as the no-argument call agreeing with them.

**Smaller notes** - `settings` inside `onNodesChangeWrapped` is now `snapSettings`; reading through `getState()` stays. The per-id `nodes.find` and the trailing loop's second scan are gone in favour of one `Map` built up front, so the per-frame cost is O(n) rather than O(n*k). Glad the `+ 0` earned its place; that normalization now lives in the shared helper, so both the resize and the fan-out path get it from one source.

One structural note, since it moves a function you may look for: `containerOrigin` now lives in `state/workspace.ts` beside `rootPosition` (it is `rootPosition` with one branch in front), and `lib/gridSnap.ts` re-exports it. `gridSnap` already imports from `workspace`, so having `workspace` import `containerOrigin` back would have made the module graph cyclic. Same shape as `readMemInfo` being owned by `session-memory.ts` and re-exported by `session-budget.ts`. Every existing import site is unchanged.

eneskirca:
Re-reviewed 977f475 (on top of #546) in an isolated worktree: typecheck clean, the `gridSnap` / `resizeSnap` / `nodeSizing` / `workspace` / canvas suites green (37 files, 639 tests), CI green, merges clean onto current main and stays green after the merge.

Points 1–3 are answered by the code and pinned by tests: the top-left child compensation, the right/bottom origin move that React Flow never emitted a change for, the nested frame snapping against the canvas grid, and `expandRectToGrid` keeping the min-size clamp. Boundary check for the ephemeral cards: `snapResizeChanges` only ever adds changes for ids it finds in `nodes` (a card's id misses `byId`, and the `managed` filter has already diverted it to the agent-nodes store), so snapping cannot produce a dirty mark or an undo entry the resize itself would not have produced.

One gap, non-blocking: `moveSessionToGroup` (sidebar drag-to-group, `Canvas.tsx` ~L10565) calls `reparentNode(ns, nodeId, groupId)` without the grid, so the re-fit inside `fitAncestorChain` runs with `grid = 0` there — a session dropped onto a frame smaller than itself pulls that frame back off the grid, which is the case `groupBox` was written to prevent. `snapGridNow()` as the fourth argument closes it (the canvas-control `move` path already does the equivalent through `fitGroupToChildren(..., snapGridNow())`). Fine as a follow-up if you would rather keep this PR closed.

Ready to merge otherwise.

PR #588, opened 2026-09-01
feat(kanban): Omni swimlane overview (global kanban)

AlexanderWillner:
Instead of one tab per project, the Kanban view now shows all open projects as swimlanes in a single overview.

- **Visually stronger swimlane separation**: margin, rounded corners, colored left border per project, shadow, alternating background, sticky header with gradient, highlight animation
- **Click navigation**: top bar with clickable project pills (1. Project A, 2. Project B …) jumps to the lane via smooth scroll — analogous to Cmd+1..9. Clicking the swimlane header also jumps/highlights.
- **Cmd+1..9**: in the global Kanban it scrolls to the Nth swimlane (via `nodeterm:swimlane-jump`), otherwise switches tabs as before. Also works from a focused terminal/modal (bubble).
- **Omni Kanban optional**: Settings → Behavior → Omni Kanban (global swimlanes) (`settings.omniKanbanEnabled`, default ON, migrated for existing installs). When OFF, per-project Kanban remains and Cmd+1..9 switches tabs; when ON, Cmd+Shift+B shows the global overview.

Included fix: terminal project-jump (Cmd+1..9) did not work from a focused terminal/modal (swallow with `preventDefault` blocked the dispatcher) — now `bubble`.

Branch: `feat/kanban-swimlanes` (fork AlexanderWillner/nodeterm)
Test: `npm run typecheck`, `npm run build` ok, manually verified in `dist/mac-arm64/nodeterm.app`.


eneskirca:
Thanks for building this — a cross-project overview is a genuinely reasonable thing to want from a tool whose whole premise is spatial organisation, and the swimlane + project-pill + ⌘1..9 navigation is a coherent design rather than a bolt-on. This is a **technical pre-assessment only**; whether nodeterm should ship a global board, and whether it should be on by default, is @eneskirca's product call.

## What I would want changed before this is mergeable

**1. Two derivations of "canvas node → kanban card".** `KanbanView` derives the active project's cards **live from React Flow nodes**, which CLAUDE.md calls out as the single live source of truth. The new `toKanbanSessionState.ts` derives every *other* project's cards from the **serialized `CanvasNodeState`** in the projects store. Those two will disagree in exactly the ordinary case: a title rename, a colour change or a new node in the active project is live in React Flow and not yet committed to the store (`commitActiveToStore()` runs on project switch / add / delete). So the active lane and the other lanes are drawn by two different functions from two different snapshots. This repo's stated rule is one definition consumed by both callers — I would derive *all* lanes from serialized state and commit the active project before opening the board, or route both through one function.

The new module also opts out of the type system to do it: `n as unknown as Record<string, unknown>` to reach `initialCommand`/`pendingLaunch`, and `spawn: { ... } as never`. `as never` in particular means the object is not checked against `KanbanSession['spawn']` at all, so a future field added to spawn will silently be missing here.

**2. Two flags for one feature, with opposite defaults.** `settings.omniKanbanEnabled` (in `shared/types.ts`, default `true`) and `useViewMode.globalKanban` (localStorage `nodeterm.globalKanban`, default `false`). Reading the diff they mean different things — "the feature exists" vs "the global board is currently open" — but the names do not say so, and `omniKanbanEnabled !== false` is spelled out in five separate places rather than behind one helper. Two of those are `useSettings.getState()` reads inside handlers, which is fine, but the fifth copy is where this drifts.

Related: `globalKanban` is **persisted**, so a user who opens the global board once boots into it forever. The codebase already reasoned about this shape for the canvas lock — *"Transient by design: a lock that survives restart reads as 'the app is frozen' to whoever opens it next"* — and a global board that outlives a restart has the same property for someone who does not remember enabling it.

**3. Two changes here are unrelated to the feature and should be separate PRs.**

- **`monaco-editor: "^0.56.0"` → `"0.56.0"`.** Pinning an exact version is a dependency-policy decision, it silently freezes patch releases, and 0.56 landed in main deliberately through its own PR. Please drop this hunk (and the `package-lock.json` line with it).
- **The `terminalKeyAction` `swallow` → `bubble` fix.** This one is *valuable* and I think it is right: `'bubble'` returns `false` from `attachCustomKeyEventHandler` (so xterm skips its keymap and no `^@`/`^[` reaches the pty) **without** `preventDefault`, which is exactly what the window dispatcher needs. It is a real regression fix that should land fast, independent of any product decision about swimlanes — and burying it inside an 819-line feature PR guarantees it waits on that decision. Two things it needs on the way out:
  - The **JSDoc above `terminalKeyAction` still says the opposite**: *"When it is true the key must be swallowed HERE ... by the time Canvas's bubble-phase listener calls `preventDefault()` the control byte has already gone to the pty."* That paragraph now contradicts the code directly below it. In this codebase a stale invariant comment is treated as a defect, not a nit — please rewrite it in the same voice, explaining why `bubble` achieves what `swallow` was reaching for.
  - A test pinning that `ownsProjectJump` yields `'bubble'` **and** that the consumer does not `preventDefault` on it. `terminal-config.test.ts` moves 7+/4-, so this may already be covered; worth confirming, because the whole bug was that the dispatcher bails on `defaultPrevented`.

**4. Test coverage is thin for the size.** 378 new lines of `GlobalKanbanView`, a 63-line `Canvas.tsx` change and a new derivation module are covered by 14 lines of `KanbanColumn.test.tsx`. `toKanbanSessionState` is pure and trivially testable (each kind, the sticky title derivation and its 80-char clamp, the `initialCommand`/`pendingLaunch` precedence, the `null` for a non-session kind); the swimlane jump and the omni gate are the other obvious units.

**5. The three-surfaces call is not made.** CLAUDE.md asks every feature to state Desktop / Server Edition / Mobile explicitly, even when the answer is N/A. This looks like pure renderer + settings, so SE should work as-is and mobile is a separate repo — but please say so rather than leave it inferred.

## Questions

- **What happens when a card modal is opened for a session in a non-active project?** `ModalTerminal` co-attaches to the tmux session, and for an SSH project that is not currently connected there is no ControlMaster. Core should refuse (`PtyCreateResult.unavailable`, the `requireRemote` rule), so I expect a failed attach rather than anything dangerous — but "a local shell opened in place of a remote one" is the exact incident that rule exists for, and it deserves a deliberate answer rather than a hope.
- **Does a lane for a closed project appear?** `closeProject` keeps the project and its nodes on disk with tmux running. The description says "all open projects", which sounds right; worth confirming the filter is on `closed`, not on something else.
- Board data lives in `project.kanban` inside each project's `.nodeterm/project.json`. A global view that can move a card writes to *another project's* git-shared file — worth stating whether that is intended, since a cross-project drag is a commit in someone else's repo.

## Small things

- The PR description is in German; the repo's convention is English for everything user- and contributor-facing, and PR bodies are read by contributors who do not speak it.
- `readGlobalKanban` accepting both `'1'` and `'true'` is a nice tolerant read.

Summary: interesting feature, coherent UX, but as one PR it mixes a product decision, an unrelated dependency pin and a genuine keyboard regression fix. Splitting out the `terminalKeyAction` fix (with its comment corrected) would let the useful part land this week regardless of what @eneskirca decides about swimlanes.

AlexanderWillner:
Addressed review https://github.com/eneskirca/nodeterm/pull/588#issuecomment-5498602893 — pushed to `feat/kanban-swimlanes` (d0917c53):

**1. Two derivations → one discipline:** `toKanbanSessionState` now type-safe (no `as unknown`/`as never`), derives sticky title via shared helper and spawn from `pendingLaunch?.command` (typed), with comment stating it is the persisted-state counterpart to `toKanbanSession` and they must stay in sync. Added `GlobalKanbanView` effect + synchronous commit in `view.kanbanToggle` so opening the overview commits live React Flow nodes to `p.nodes` before lanes render — active lane no longer stale vs other lanes.

**2. Two flags → one helper:** Added `isOmniKanbanEnabled(settings)` in `state/viewMode.ts` (feature flag vs view flag distinguished in doc) and replaced 5 inline `!== false` sites (`Canvas`, `TabBar`, `TerminalNode`). `isGlobalKanbanOpen()` now delegates to it and documents why `globalKanban` (localStorage) is persisted intentionally ("you left it open", unlike canvas lock).

**3. Unrelated hunks split:**
- `monaco-editor` pin reverted to `^0.56.0` (package.json + lock).
- `terminalKeyAction` JSDoc rewritten to explain `bubble` (no `preventDefault`) vs old swallow — code already returned `bubble`, now comment matches.

**4. Tests:** Added `toKanbanSessionState.test.ts` (terminal/browser/sticky, markdown heading, 80-char clamp, `pendingLaunch`, null for non-cards). Existing `terminal-config.test.ts` already pins `ownsProjectJump → bubble`.

**5. Three surfaces:** `GlobalKanbanView.tsx` header now states Desktop full, Server Edition works as-is (pure renderer+workspace.save), Mobile N/A.

**Questions:**
- *Modal for SSH project not connected:* `ModalTerminal` co-attaches via transport with `requireRemote`; core refuses (`PtyCreateResult.unavailable`) → no local shell fallback (the incident the rule exists for).
- *Closed projects:* Filtered `!p.closed` — deliberate, parked sessions stay hidden.
- *Cross-project drag:* Write goes to owning project's `.nodeterm/project.json` via `setProjectKanban`→`markWorkspaceDirty`→`workspace.save` — intentional, board travels with repo.

`npm run typecheck` + `npm run build` green, new tests pass.

eneskirca:
Product decision: **we want this, with guardrails.**

1. **Default OFF.** ⌘⇧B keeps today's per-project board for existing users — no silent behavior change.
2. **A dedicated, remappable shortcut** for the global board (a registry command in `shared/keybindings.ts`, so it shows in Settings → Keyboard Shortcuts and ShortcutsPanel like everything else).
3. **A setting to make Omni the default** — users who live in the global view can flip ⌘⇧B to open Omni instead. Opt-in, per person.
4. As already asked in review: please extract the embedded `terminalKeyAction` keyboard fix into its own PR (it's real and mergeable fast), and address the two-derivation//monaco-pin notes.

With those, this becomes a headline feature rather than a surprise. Thanks — the swimlane overview itself is genuinely good.

AlexanderWillner:
Follow-up for product decision https://github.com/eneskirca/nodeterm/pull/588#issuecomment-5500480388 — pushed to `feat/kanban-swimlanes` (b02b40d2):

**1. Default OFF:** `DEFAULT_SETTINGS.omniKanbanEnabled` flipped to `false`; `isOmniKanbanEnabled` now strict `=== true` (no migration of absent key to ON). Existing users see no change — `Cmd+Shift+B` keeps per-project board.

**2. Dedicated remappable shortcut:** Added `view.globalKanbanToggle` (`General`/`app`, unbound) in `shared/keybindings.ts` — shows in Settings → Keyboard Shortcuts and ShortcutsPanel like all registry commands. User can bind e.g. `Cmd+Shift+G` (or any chord).

**3. Opt-in flip for Cmd+Shift+B:** New setting `omniKanbanAsDefault` (`shared/types.ts`, default `false`) in Settings → Behavior → *Make Omni the default for Cmd+Shift+B*. `view.kanbanToggle` now: if global overlay open → close it; else if `omniEnabled && asDefault` → open Omni (with commit), else per-project. `view.globalKanbanToggle` always opens Omni (when feature enabled) regardless of the flip.

**4. Terminal fix extracted:** The `terminalKeyAction` `swallow→bubble` regression fix (point 3 of earlier review) is now standalone PR #611 — can land fast independent of swimlanes. This branch keeps the same fix (so no rebase drift) but the canonical merge should be #611 first.

Typecheck + targeted tests green; feature still pure renderer + `workspace.save` (Server Edition works as-is, Mobile N/A as before).

eneskirca:
Thanks for turning the product decision around so quickly. The three conditions are in the diff as asked: `omniKanbanEnabled` defaults to `false` with a strict `=== true` gate, `view.globalKanbanToggle` is a registry command (unbound, remappable, shows in Settings → Keyboard Shortcuts and, after #622, in ShortcutsPanel automatically), and `omniKanbanAsDefault` is a per-user opt-in. State placement is also right: both flags live in `settings.json` and the open/closed view flag in `localStorage`, i.e. machine-local; board writes go to the owning project's `.nodeterm/project.json`, same as the per-project board today.

I pulled `b02b40d2` into an isolated worktree and ran it. A few things block merge — the first three are mechanical, four to seven are behavioral:

**1. Conflicts with `main` (`src/renderer/canvas/Canvas.tsx`, the `nodeterm:switch-system-account` handler), and because of that GitHub created no CI run for `b02b40d2`** — `pull_request` workflows are skipped when the merge commit cannot be built. Please merge `main` (#622 landed today) so CI runs again.

**2. `src/shared/keybindings.test.ts` fails.** The registry table test pins every command; the new `view.globalKanbanToggle` row is missing from the expected table (`Tests 1 failed | 70 passed`). Add the row there.

**3. ⌘⇧B on desktop does not go through `view.kanbanToggle`.** Under app-first the menu accelerator (`MENU_ITEM_ID_KANBAN`) takes ⌘⇧B above the page and forwards it as IPC to the `onToggleKanban` receiver in `Canvas.tsx` (~line 4835), which still does `useViewMode.getState().toggle(id)`. So `omniKanbanAsDefault` has no effect for ⌘⇧B on desktop, and while Omni is open ⌘⇧B toggles the per-project board *underneath* the overlay. The registry handler only sees ⌘⇧B under terminal-first with a focused terminal, or on Server Edition. Put the decision in one function and call it from both the IPC receiver and the registry handler (the receiver's `noteTerminalCapture` stays).

**4. `isGlobalKanbanOpen()` fails open in the packaged renderer.** It reads settings via `require('./settings')`. Vite emits that call verbatim (checked with esbuild on the file), and the renderer runs with `contextIsolation: true` / `nodeIntegration: false`, so `require` is not defined: the call throws, the `catch` says "treat as enabled", and the feature gate is bypassed at every call site — the dispatcher's `kanbanOpen`, `projectJumpGesture` (⌘1..9 dispatches `swimlane-jump` instead of switching), `focusNodeById` (requests a card instead of framing), the dictation / tidy / focus-mode / breadcrumb guards, and `TerminalNode`'s `registryOwns`. Repro: enable Omni → open it → disable Omni in Settings: the canvas comes back, but ⌘1..9 no longer switches projects and canvas-scope commands are refused. Vitest cannot catch it because vite-node provides `require`. `settings.ts` does not import `viewMode`, so a plain static import of `useSettings` works; and if you keep a try/catch, fail *closed* — this is a default-OFF feature.

**5. Edits to the ACTIVE project's lane are lost.** `onRenameNode`, `onEditSticky`, `onBrowserNav` and `onDeleteNode` in `GlobalKanbanView` write `useProjects.setState` directly. For the active project React Flow is the live source of truth (CLAUDE.md) and nothing re-syncs it from the store — the load effect is keyed on `activeProjectId`/`reloadNonce` only — so the next `commitActiveToStore()` (the very effect that runs when Omni opens, or any project switch) overwrites the edit; a deleted node comes back, with its tmux session already killed. Route active-project mutations through Canvas the way `nodeterm:create-node` already does, or reload the active project after writing.

**6. The delete path bypasses the existing one.** Native `confirm()` instead of `ConfirmDialog`; none of `deleteNodeFromKanban`'s teardown (agent status, dispose, board log); `pty.destroy` is called twice (`api` and `window.nodeTerminal`); and for a node in a non-active **SSH** project `destroy(nodeId)` only touches local sockets, so the remote session keeps running while the card disappears — exactly the orphan the session-memory panel had to fix (`lib/sessionKill.ts` / `sshProject.killSessions` is the routing that handles it).

**7. `pendingLaunch.command` must not become `initialCommand`.** `toKanbanSessionState` maps it to `spawn.initialCommand`, and `ModalTerminal` writes it 300 ms after a fresh attach. `pendingLaunch` is the `--after` DAG launch that fires only when its dependencies report `done`; opening a card modal fires it early. Since `create-node` now switches to the project first (so `TerminalNode` mounts and delivers `initialCommand` through `writeWhenShellReady`), the modal delivery has no remaining legitimate case — please drop both the mapping and the fixed-delay write (a timer into a shell-owned pane is the race `writeWhenShellReady` exists to avoid).

Smaller: the TabBar toggle's `if (viewFor(...) === 'kanban') vm.toggle(p.id) else vm.toggle(p.id)` has identical branches; the header hint says "Cmd+1..9" on Linux/Windows and on Server Edition, where the browser owns the chord; the `switchProject` → `setTimeout(…, 120)` in the create handler is a timing guess; the `GlobalKanbanView` / `TerminalNode.global-kanban` tests assert source text, not behavior; and CLAUDE.md's Kanban bullet + the two settings need a paragraph (Conventions: a change other developers must know goes in CLAUDE.md, the short version in CONTRIBUTING.md).

Once 1–7 are in I'm happy to re-run it. #611 is the right way to land the keyboard fix first.

AlexanderWillner:
Addressed https://github.com/eneskirca/nodeterm/pull/588#issuecomment-5508896704 — pushed `af6c0179` (on top of `b02b40d2`):

**1. Merge main** — merged `upstream/main` (`#622`) to fix `Canvas.tsx` conflict and restore CI (was skipped due to merge conflict).

**2. `keybindings.test`** — added `view.globalKanbanToggle` row to the pinned whole-table expectation.

**3. `Cmd+Shift+B` desktop** — extracted single `performKanbanToggle` (closes global if open, else respects `omniEnabled && asDefault` with commit) and called from both the registry handler (`view.kanbanToggle`) and the desktop menu IPC receiver (`onToggleKanban`; kept `noteTerminalCapture`). Added `view.globalKanbanToggle` handler. Previously the IPC path always did `toggle(id)` and toggled the per-project board underneath the overlay.

**4. `isGlobalKanbanOpen` fail-open** — replaced dynamic `require('./settings')` (Vite emits verbatim, undefined in renderer with `contextIsolation`) with static `import { useSettings }`; catch now fail-closed (`return false`) — default OFF, so an unreadable store no longer enables Omni. This fixes the dispatcher (`kanbanOpen`, `projectJumpGesture`, `focusNodeById`) and `TerminalNode.registryOwns` bypass when Omni is disabled.

**5. Active lane edits lost** — `GlobalKanbanView` now dispatches `nodeterm:global-*` events; Canvas handles them: `renameSession` (already branches), `setNodes+markDirty` for active vs store+`writeDisk` for non-active for sticky/browserNav. Previously direct store writes for the active project were clobbered by the next `commitActiveToStore`.

**6. Delete path** — replaced native `confirm()` with `setConfirm`/`ConfirmDialog`, proper teardown (`agentStatus`/`agentNodes`/`webviewKeepAlive`/`clearAttachConsent`) and SSH-aware kill (`transport.destroy` locally vs `sshProject.killSessions(..., {everySocket:true})` for remote owner) via a single Canvas handler for both active (`deleteNodeFromKanban`) and non-active (store removal + kill).

**7. `pendingLaunch`** — dropped `pendingLaunch.command → spawn.initialCommand` mapping in `toKanbanSessionState` and the 300 ms `setTimeout` write in `ModalTerminal` (DAG `--after` launch must not fire on card open; `create-node` now switches project so `TerminalNode.writeWhenShellReady` delivers the legitimate `initialCommand`).

**Smaller:**
- TabBar identical branches collapsed.
- Header hint now `${mod}+1..` (`Cmd` on Mac, `Ctrl` elsewhere) and close title likewise.
- `create-node` `setTimeout 120→50` with comment (next tick for React Flow remount, previous was a guess).
- `GlobalKanbanView` now routes via Canvas events and handles `NodeIcon` via `global-set-icon` (added missing `onSetIcon` prop for new `CardModal` requirement after merge).
- Docs: `CLAUDE.md` Omni paragraph + `CONTRIBUTING.md` short note.

Typecheck + targeted tests (`keybindings.test`, `toKanbanSessionState.test`, `terminal-config.test`) green; `TerminalNode.global-kanban` still passes. #611 remains the standalone keyboard fix.

eneskirca:
Re-verified `af6c0179` on a fresh isolated worktree: `npm run typecheck` clean; 68 test files / 987 tests green (kanban, keybindings, ShortcutsPanel, keydown-intercept, nodeterm-events, terminal-config, globalKeybindings, state); merges clean with `main`; CI green on this head.

Per blocker:
1. **Conflict / CI** — closed. Merges clean, `quality` + `quality-windows` + CodeQL + dependency review pass on `af6c0179`.
2. **`keybindings.test` table** — closed, row added, suite green.
3. **⌘⇧B via menu IPC** — closed. `performKanbanToggle` is called from both the `onToggleKanban` receiver (with `noteTerminalCapture` kept) and the `view.kanbanToggle` handler; `view.globalKanbanToggle` has its own handler.
4. **Fail-open gate** — closed. `viewMode.ts` now statically imports `useSettings` (no cycle: `settings.ts` does not import `viewMode`), the catch returns `false`, and running the file through esbuild (`--format=esm --platform=browser`) emits zero `require(` calls. No other `require(` remains under `src/renderer`.
5. **Active-lane edits** — closed. Rename goes through `renameSession` (active → `setNodes`+`markDirty`, else store + `writeDisk`, `/rename` push preserved); sticky / browser-nav / icon branch on the active project the same way, and the sticky branch mirrors `editStickyText` (agent-sync stamp cleared). `nodeterm-events.test` confirms every new `nodeterm:global-*` event has a listener.
6. **Delete path** — closed. Active → `deleteNodeFromKanban` (ConfirmDialog + full teardown); non-active → ConfirmDialog, store removal, `sshProject.killSessions(projectId, [nodeId], {everySocket:true})` for an SSH owner / `transport.destroy` locally, plus agentStatus / agentNodes / webviewKeepAlive / attach-consent cleanup.
7. **`pendingLaunch` → `initialCommand`** — closed. Mapping removed from `toKanbanSessionState`, the 300 ms write removed from `ModalTerminal`.

Non-blocking leftovers, fine as a follow-up or a last small commit:
- `TabBar`'s board-toggle button still re-implements the omni/asDefault decision inline rather than calling `performKanbanToggle` (the new CLAUDE.md paragraph says TabBar shares it — today it does not; the `globalKanbanOpen` effect in Canvas does the commit, so behavior is right, only the doc line overstates it).
- `ModalSpawn.initialCommand` is now an unused field.
- The `window as unknown as {…}` cast around `killSessions` is unnecessary — `window.nodeTerminal.sshProject.killSessions(projectId, nodeIds, opts)` is already typed with exactly that signature.
- Non-active sticky edits keep `textUpdatedAt`/`textUpdatedBy` in the store while the active path clears them.
- The `create-node` handler still relies on a 50 ms timer after `switchProject`; the `GlobalKanbanView` / `TerminalNode.global-kanban` tests still assert source text rather than behavior.

From my side this is mergeable. Merge decision is @eneskirca's; no device run on macOS was done here.

PR #592, opened 2026-09-01
feat(grok): subscribe to the whole published hook set, and stop the badge sticking on Esc

xaviguardia:
Grok's integration subscribed **9 of the 15 hook events** Grok 1.0.13 publishes, and classified two of them by substring. Both cost the user a wrong badge, in opposite directions — one too loud, one silent.

This is largely **a correction of `docs/grok-agent.md` against Grok's own behaviour**, not a pile of new features. That file states, in §8, that the whole Grok integration was written with no Grok binary and no account on the machine. This PR is what happens when someone finally runs it.

## What the live capture found

Before opening this, the change was tested against hook payloads **we had written ourselves from Grok's shipped docs**. Capturing real ones from a logged-in Grok 1.0.13 session turned up a bug that had already survived TDD, an independent review and a QA sign-off:

**`PreCompact` / `PostCompact` were dropped on 100% of real payloads.** The guard read `trigger` — Claude's spelling. Grok sends `source`. With the key absent the canonicaliser returned `''`, the closed-vocabulary guard rejected every live event, and the feature was **dead, not degraded**.

It stayed green because a payload invented from the docs carries the key the docs imply. The test proved our code agreed with our reading of the documentation; it could never have proved agreement with the agent. The fixture added here is a live capture under a real account, so it cannot agree with us out of politeness.

Three documented claims were also wrong, and are corrected with the measurement rather than quietly dropped:

- **§8.2** said an interrupted turn *"fires no hook at all, by grok's design"*, and proposed measuring how bad it felt to decide whether a watchdog was worth building. `StopCancelled` exists (`10-hooks.md:98`); the badge stuck for want of a subscription, not because of a limit in Grok.
- **§8.1** treated the `Notification` vocabulary as unknown and inferred from another integration's reader. It is published and closed (`10-hooks.md:99`, `:162`), and the capture confirms it.
- **`transcriptPath` IS sent**, against what the doc claimed — 14 of 15 captured events carry it (`SessionStart` is the measured exception, arriving before the session directory exists). We still derive the directory from `(cwd, sessionId)`, and the capture is what upgrades that from an assumption to a validated choice: **the derived path and the transmitted one are byte-identical in every captured event**. Three reasons the derivation stays are named in the doc — `SessionStart` has no path to read, the field names a *file* while we need the *directory*, and an SSH node must resolve under the host's own `$GROK_HOME`.

One of the false claims was **introduced by this branch itself** (a subscription-list count and a symbol our own change had moved). It is fixed here.

## Why

**The NEEDS YOU badge strobed.** A `permission_prompt` Notification fires before *every* tool call. The shipped mapping matched the permission family by substring, so a node that was simply working produced an unread dot, a chime, an OS notification and a phone inbox card **per tool call**. The published closed set replaces the substring, and an unknown type is a deliberate no-op: a badge stuck on a finished node has no later hook to clear it, so widening the match "to be safe" is the unsafe direction.

**The RUNNING badge stuck on Esc.** `StopCancelled` was not subscribed, so an interrupted turn fired nothing we listened to and the node sat on RUNNING until the next `UserPromptSubmit` re-synced it. Its `reason` distinguishes `user_interrupt` from `permission_rejected`, `permission_cancelled`, `max_turns` and `no_progress`. No watchdog was built, and none is needed.

## The trade-offs, and what was rejected

- A session-level `user_interrupt` clears RUNNING as an **interrupted** `done` (no ordinary completion alert or unread). The other four reasons end the turn **without** the `interrupted` flag, so a non-successful terminal outcome stays visible rather than having its completion edge suppressed.
- A payload carrying `subagentType` returns early and changes nothing. A child's stop is not the session's stop (`10-hooks.md:354`).
- **Cancel-and-send emits no `StopCancelled`** (`:336`) because the turn is being replaced, not stopped, and the agent stays busy. Treating it as a turn end would blank the badge of a node that is working.
- An unknown `reason` stays a no-op rather than defaulting to `done`, so a future cancel-and-send dialect cannot become a false completion.
- **The `idle_prompt` rescue is kept, not retired.** It was built believing no event existed, and the obvious move once `StopCancelled` landed was to delete it. `10-hooks.md:384` documents a turn where no stop hook completes at all — that case still needs a floor. It is narrowed instead: it can only clear `working`, so it can never erase a live approval or question.
- `task_complete` is mapped to nothing. The spec does not define it as a turn end — `Stop` does — so guessing `done` here would risk a false completion.
- The compaction fix reads `source ?? trigger`, keeping the Claude-shaped key. Reading a key that never arrives costs nothing; the SDK path may still present it.

`PermissionDenied`, `SubagentStart`/`SubagentStop` and `PreCompact`/`PostCompact` are now subscribed too, completing the published set.

## About the size of this diff

**19 files, 1,241 insertions, 497 deletions.** That is large for this repo, and worth breaking down before you judge it — only about a third is source code:

| | lines touched |
|---|---|
| Captured JSON fixture (pure data) | 290 |
| Tests | 743 |
| `docs/grok-agent.md` | 186 |
| **Source code** | **519** |

And the source volume is two structural moves, not new behaviour:

1. **The event list moved to a single frontier in `src/core`** (`grok-hook-session.ts`), consumed by both the desktop and the Server Edition. That is why `src/main/index.ts`, `src/server/agent-status.ts` and `src/shared/agents/hook-events.ts` all shrink while one new core module grows. This repo has shipped a one-shell hook-server change three times, and on this branch the remote installer had already subscribed Gemini to *Claude's* event names — a duplicated rule drifts.
2. **A substring was replaced by the published closed set**, which rewrites the mapping table and its tests.

I am aware this is larger than the repo usually takes from an outside contributor. If you would rather review it in pieces, say so and I will split it — the branch history makes that cheap.

## Verification

- `npm run typecheck` — 0 errors.
- Focal suites: 468 passing across 18 files.
- **Mutation-checked, not just green.** Removing the `user_interrupt` branch turns the mirror suite red; removing the `subagentType` early exit turns it red; and reverting the compaction fix to `trigger` alone turns the capture suite red **against the real captured payload**. All restored byte-identically (verified by checksum).
- Full suite: 9,349 passing, with failures confined to **6 files** — `pane-cwd.realtmux`, `pty-session-host`, `control-master.realsh`, `session-host-client`, `tmux-paste.realtmux`, `remote-atomic-write`. Every one is a **strict subset** of the files that already fail on a clean checkout of this exact base with no change applied: the environment-dependent tmux / ssh / session-host suites. **No new failure is introduced.**
- The failure *count* inside those files drifts between runs on the same tree, so the claim is made per FILE — a count would not reproduce. Measured by applying the change to a throwaway worktree checked out at this base, not inferred from a branch carrying other work.
- Reproduction note: `src/core/agents/hooks/opencode.test.ts` fails 4/22 if the suite inherits a `NODETERM_HOOK_SOCK` environment variable, and passes 22/22 without it. That is a pre-existing test-isolation issue unrelated to this change.

## Not verified

The capture covered 15 payloads across 11 event types from one live session, so much of this is now measured rather than read. What it did **not** cover, and why:

- **`TITLE_KEYS[0] = 'title'` is still a guess.** `/rename` was never run in the captured session, so the key Grok writes a manual title to remains unconfirmed. Promoting it on the strength of "we captured something" would be exactly the error this PR corrects.
- **The bare `type` fallback for `Notification` remains inferred** from another integration's reader — every captured Notification used `notificationType`.
- **Canvas-control discovery is unverified**: `grok inspect --json` was never run, so the premise that Grok scans `~/.claude/skills` rests on its shipped docs.
- **The remote (SSH) install path** for the new events is unit-tested, not exercised against a real host.

The device checklist in `docs/grok-agent.md` §9 now carries a table of what the capture could not measure, with the reason for each.

## Surfaces

- **Desktop** — full support; `src/main/index.ts` consumes the shared core frontier.
- **Server Edition** — full parity, and that is the point of the refactor: both raw hook listeners decode Grok events through one definition in `src/core`, so the Server Edition cannot silently miss a field the desktop gained. `src/server/agent-status.test.ts` pins it.
- **Mobile** — the agent-status mirror this feeds is what *nodeterm mobile* reads, so an interrupted Grok turn will now clear the phone's RUNNING card instead of leaving it stuck; that rides the existing mirror and needs no iOS change. What is **not** covered is Grok-specific: the per-node activity line still does not work for Grok (`recordRawToolEvent` gates on a `hook_event_name` Grok's file hooks never send). @eneskirca — flagging it here since `nodeterm-ios` is private and I cannot open that follow-up myself.

---

## Beyond the title: `atomicConfig` touches every agent

`writeManagedHookFileAtomic` / the `atomicConfig` option make the managed **script**
write atomic for *every* agent, not only grok — claude, codex, gemini and opencode
included. That is deliberate (a half-written hook script is a broken hook for whoever
owns it), but it is more than "subscribe to the whole published hook set" promises, so
it is stated here rather than left for the next reader to discover.


eneskirca:
Reviewed at e5e128e in an isolated worktree: typecheck clean, the 12 focal suites (350 tests, incl. `no-electron`, `fs-atomic.guard`, `hook-verified-parity`) green, and the full suite passes here (709 files / 9708 tests, 0 failures — this box has tmux, so none of the environment-dependent suites you listed fail), CI approved and green. I also materialised grok 1.0.13's own `10-hooks.md` (fresh `GROK_HOME` on this box) and it confirms what the PR relies on: `StopCancelled` and its six reasons, the closed `StopFailure` `error` set, and `permission_prompt` firing only while a permission UI is actually waiting. The captured fixture carries `source` on both compaction events and `transcriptPath` on 14/15. So the measured claims check out.

Two things to fix before merge, both small:

1. **A claim the fixture contradicts.** `src/core/agents/hooks/grok.ts` says *"Grok mints a new session id during compaction, and both shells must observe it"*, and `planGrokHookSession` has a `postcompact` branch that retires the previous id. In the capture, `pre_compact` and `post_compact` carry the **same** `sessionId`, and your own doc table (`docs/grok-agent.md`, the compaction row) says *"there is no id to re-key a session association with"*. Please drop the comment's claim (and either drop the `postcompact` branch or label it as defensive-only). Leaving it reads as a measurement it is not.

2. **`StopFailure` now fails toward a stuck badge.** With `if (!GROK_STOP_FAILURE_TYPES.has(...)) return null`, a `StopFailure` whose `error` is absent or a value outside the six leaves the node on RUNNING until the `idle_prompt` backstop. The event itself is the turn end, whatever the class — grok's docs make `unknown` the catch-all for anything the runtime cannot classify, so an out-of-set value can only be a future dialect, and the turn has still ended. Suggest: unmatched class ⇒ still `done` (the class is informational here, nothing branches on it).

Nits, non-blocking: `grok.ts` switched to double-quoted imports (the repo is single-quote throughout); and the `atomicConfig` / `writeManagedHookFileAtomic` change also makes every agent's managed *script* write atomic, which is fine but is more than the title says — a line in the description would help the next reader.

Ordering note for @eneskirca: this conflicts with #618/#619 in `src/main/index.ts`, `src/server/agent-status.ts` and `docs/grok-agent.md` (the stack extends the inline grok branch that this PR moves into `applyGrokHookSession`). Whichever lands second owes the rebase; the natural home for #618's tail-track/untrack and `recordRawToolEvent` calls is inside `applyGrokHookSession`, not beside it in both shells.

xaviguardia:
Rebased onto current `main` (`3411efdf`) — this branch is green and ready for review.

One note that may save time on both sides: it touches the raw hook listener in `src/main/index.ts`, `src/server/agent-status.ts` and `src/shared/agents/normalize.ts`, which puts it in the path of most merges. It has needed rebasing three times in two days, and each rebase is a chance to silently change what the PR does — so reviewing it earlier rather than later would avoid that churn.

Happy to split it into smaller pieces if that would be easier to review.

xaviguardia:
Correction to my comment above: "ready for review" was wrong, and I apologise for the noise. Your review of 2026-09-02 asked for two changes before merge and I had not read it when I wrote that — the rebase was mechanical and I never opened the thread. That is on me.

Both points are accepted, and neither is arguable:

1. The comment in `src/core/agents/hooks/grok.ts` states a measurement the fixture contradicts — `pre_compact` and `post_compact` carry the same `sessionId`. The claim goes, and the `postcompact` branch will be dropped or labelled defensive-only.
2. `StopFailure` with an absent or out-of-set `error` should still be `done`. You are right that the event *is* the turn end whatever the class, and that failing to `null` there leaves the node on RUNNING until the backstop — a silent stuck badge, which is the worse of the two failure directions.

The two nits are taken as well: the double-quoted imports go back to single quotes, and the description will state that the `atomicConfig` change makes every agent's managed script write atomic, not just grok's.

Your ordering note is noted too — this PR moves the inline grok branch into `applyGrokHookSession`, and #618/#619 have since landed extending the inline form, so this branch owes that rebase. It is already rebased onto `3411efdf`; folding #618's tail-track/untrack and `recordRawToolEvent` into `applyGrokHookSession` rather than leaving them beside it in both shells is the right shape and will be in the next push.

I will comment again when the changes are in, not before.

xaviguardia:
Both blocking points and both nits are in — `aaefbbcc`, on top of the rebase onto `3411efdf`.

**1. The compaction claim is gone, and so is the branch.** You were right on the fixture: `pre_compact` and `post_compact` carry the same `sessionId`. We dropped the claim rather than softening it, and removed the `postcompact` branch rather than labelling it defensive — a branch guarding against something that does not happen is a line nobody can prove or disprove, and in six months it reads as a measurement again. It already did: that branch had been cited as evidence in one of our own rebase decisions. The tests that pinned the removed behaviour now pin the distinct-id case instead, which is the input the branch acted on, so restoring it goes red.

**2. `StopFailure` now yields `done` whatever the class.** Your reasoning is the one we would want applied to us: the event *is* the turn end, `unknown` is grok's documented catch-all, so an out-of-set value can only be a future dialect. The test asserts `done` explicitly rather than "not null" — "not null" would also pass if someone returned `working`, which is the state this exists to prevent. Reverting the gate turns nine tests red.

While there, `GROK_STOP_FAILURE_TYPES` went with it: the gate was its only reader. The six documented classes stay in the comment — that part is measured and worth keeping; a `Set` nobody reads is the same dead weight as the branch above.

**Nits:** imports in `hooks/grok.ts` are back to single quotes, and the description now states that the `atomicConfig` change makes the managed script write atomic for *every* agent, not just grok.

**On your ordering note.** Agreed, and the branch is already rebased onto current `main`. On folding #618's `trackSignals` / `untrack` / `recordRawToolEvent` into `applyGrokHookSession` rather than leaving them beside it in both shells — we agree with the principle, and would rather do it as its own change than here. It is ~42 lines plus three injected dependencies (`grokContextTail` is a per-shell instance), it refactors code that is already merged, and doing it in a PR titled after the hook set would be exactly the "more than the title says" you flagged on `atomicConfig`. Happy to open that as a follow-up immediately, or to fold it in here if you would rather see it together — your call.

xaviguardia:
Correction to my comment above, before you spend time on it: the commit I described as done (`aaefbbcc`) went out with two tests red. `b4ea5754` fixes both. I should have caught them before announcing, not after.

**One was a real regression.** Rewriting the `StopFailure` branch dropped `errored: true`. `erroredTurn.ts` is the single definition of "this turn died", shared across agents and both shells — without that field grok became the only agent whose failed turn looks clean. `renderer/lib/erroredTurn.test.ts` was untouched from upstream and pinned it; it is green again, and removing the field turns it red.

The cause is the same one this PR keeps correcting: a conflict resolved by taking the side already in hand without diffing what the other side carried. Your point 2 asked us to stop failing toward silence, and the first attempt at it introduced a different silence.

**The other was a leftover.** `server/agent-status.test.ts` still asserted "replaces the old session association when PostCompact mints a new id" — the behaviour removed in the previous commit. It now pins the opposite, on the distinct-id input the old branch acted on.

Neither file was in the focal set we ran: one because nothing had edited it, the other because its change came from an earlier commit rather than the working tree. Deriving a focal set from edited files does not cover *removing* a behaviour; grepping the behaviour's name across the tests does, and would have named both in a second.

State now: `b4ea5754`, typecheck clean on both configs, and the five suites covering the changed behaviour green (96 tests). Nothing else in the review is outstanding.

xaviguardia:
Full suite on `b4ea5754`, for the record — this time the whole suite, not the focal set that let the last two announcements through.

```
Test Files   10 failed | 741 passed | 2 skipped  (753)
Tests        24 failed | 10255 passed | 37 skipped
```

None of the ten is touched by this branch: `git diff --name-only upstream/main -- <file>` is empty for all ten, checked one at a time. Seven are the environment-dependent suites this box cannot pass (tmux/ssh) plus `node-pty-patch`, whose ConPTY patch is not applied in this `node_modules`.

Three are new since yesterday and worth a moment of your time, because they are yours rather than ours and they fail on any Mac:

- `renderer/lib/licenseCopy.test.ts`
- `renderer/lib/accountPresentation.test.ts`
- `renderer/components/settings/sections/LicenseSection.test.tsx`

They fail in isolation too, on platform wording — `expected '2 of 3 devices in use — this Mac and …' to be '… this computer …'`. The test files, the modules under test (`licenseCopy.ts`, `accountPresentation.ts`, `LicenseSection.tsx`) and the `@shared/types` both read are all byte-identical to `main` here, and this branch touches nothing under `renderer/lib` or `settings/`, so there is no path from our change to them. CI presumably runs on Linux and stays green.

What we did **not** check: whether they already failed on a clean `main` checkout. We are confident they are not ours; we are not claiming to have seen them fail without us.

PR #600, opened 2026-09-01
fix(hooks): normalize path separators when matching managed hook commands on Windows

### LINKED ISSUE #558: Windows: managed hooks are appended, not replaced — settings.json accumulates one duplicate set per install
On Windows, every hook install appends a duplicate instead of replacing the previous one, so `~/.claude/settings.json` grows one extra copy of all 9 managed hooks per run. My install had reached 9 identical entries for every event — meaning 9 `claude.sh` processes and 9 POSTs per Stop, and 9 concurrent 45s waits per PermissionRequest.

## Root cause

`mergeManagedHook()` filters existing managed entries out before pushing the fresh one, and `isManaged()` decides what counts as managed by substring-matching a marker. The marker is path-separator-normalized; the command it is compared against is not:

```js
function managedMarkerFor(command) {
  const m = command.match(/agent-hooks[\\/][^"'\s]+/);
  return m ? m[0].replace(/\\/g, "/") : "agent-hooks";   // -> "agent-hooks/claude.sh"
}

function isManaged(d, marker) {
  return !!d.hooks?.some(
    (h) => h.command.includes(marker) || h.command.includes("claude-signals"),
  );
}
```

On Windows `buildManagedHookCommand()` embeds a native path, so the stored command contains `agent-hooks\claude.sh` (backslash) while the marker says `agent-hooks/claude.sh` (forward slash). `includes()` is therefore always false, the previous entry is never recognized, the filter keeps it, and a new one is appended on top. On macOS/Linux the path already uses `/`, both sides agree, and the replace works — which is why this is Windows-only.

## Reproduction

Running the shipped helpers verbatim under the app's own Node (`ELECTRON_RUN_AS_NODE=1`):

```js
for (const [label, p] of [
  ["win32", "C:\\Users\\u\\.nodeterm\\agent-hooks\\claude.sh"],
  ["posix", "/home/u/.nodeterm/agent-hooks/claude.sh"],
]) {
  const command = buildManagedHookCommand(p);
  const marker = managedMarkerFor(command);
  console.log(label, marker, isManaged({ hooks: [{ command }] }, marker));

  let cfg = {};
  for (let i = 0; i < 5; i++) cfg = mergeManagedHook(cfg, command, ["Stop"]);
  console.log(label, "Stop entries after 5 installs =", cfg.hooks.Stop.length);
}
```

```
win32: marker="agent-hooks/claude.sh" recognised=false
win32: Stop entries after 5 installs = 5
posix: marker="agent-hooks/claude.sh" recognised=true
posix: Stop entries after 5 installs = 1
```

End to end: launch nodeterm on Windows N times, then look at `~/.claude/settings.json` — each of `SessionStart`, `UserPromptSubmit`, `Stop`, `StopFailure`, `Notification`, `PermissionRequest`, `SessionEnd`, `PreToolUse`, `PostToolUse` holds N byte-identical matcher objects. Mine had 9 each; the file was 23,916 bytes and dropped to 3,288 after deduplicating by hand.

## Impact

- Every hook event runs N copies of `claude.sh`: N `curl` POSTs to the hook endpoint per Stop, N payload temp files under `~/.nodeterm/pending/` per event, and N shells spawned on every `PreToolUse`/`PostToolUse` — the cost scales with how long the app has been installed.
- `PermissionRequest` is the worst case: with `NODETERM_PERM_WAIT_SECS=45`, N hook processes each poll for an answer file for up to 45 seconds, and each is entitled to emit a decision for the same prompt.
- `installHooksInto()` / `mergeManagedHook()` are shared by the Claude, Gemini, Copilot, Grok and Codex installers, so the same growth should affect every agent's config on Windows.
- Growth is unbounded — nothing ever prunes the stale entries.

## Suggested fix

Normalize both sides of the comparison, e.g.

```js
function isManaged(d, marker) {
  return !!d.hooks?.some(
    (h) =>
      h.command.replaceAll("\\", "/").includes(marker) ||
      h.command.includes("claude-signals"),
  );
}
```

Worth pairing with a one-time cleanup on upgrade, since existing Windows installs already carry the accumulated duplicates and the fix alone only collapses them on the next install run.

## Environment

- nodeterm 0.3.4 (`nodeterm.exe` ProductVersion 0.3.4.0), Windows 11 Pro 26200
- Claude Code CLI, hooks written to `C:\Users\<user>\.claude\settings.json`
- Managed script at `C:\Users\<user>\.nodeterm\agent-hooks\claude.sh`

Abhirup0:
## Problem
On Windows, `buildManagedHookCommand()` generates hook commands containing native backslash paths (`agent-hooks\claude.sh`), while `managedMarkerFor()` produces forward-slash normalized markers (`agent-hooks/claude.sh`).

Because `isManaged()` and `removeHooksFrom()` performed substring checks without normalizing `h.command`, the match returned `false`. This caused `mergeManagedHook()` to treat prior entries as foreign, appending duplicate hook definitions to `~/.claude/settings.json` on every app start.

## Solution
Normalize backslashes to forward slashes (`replaceAll('\\', '/')`) before checking `includes(marker)` in both `isManaged()` and `removeHooksFrom()`.

## Verification
- Added test in `install-helper.test.ts` verifying idempotency when merging Windows backslash hook commands.
- Added test verifying that existing accumulated duplicates are collapsed down to a single entry upon merge.
- Added tests verifying repair sweeps and `removeHooksFrom()` properly prune Windows backslash entries.
- Verified all 16 tests in `install-helper.test.ts` and `npm run typecheck` pass.

Fixes #558

### DIFF
--- a/src/core/agents/hooks/install-helper.test.ts
+++ b/src/core/agents/hooks/install-helper.test.ts
@@ -64,6 +64,27 @@ describe('mergeManagedHook', () => {
     )
     expect(out.hooks!.Stop).toEqual([{ hooks: [{ type: 'command', command: cmd }] }])
   })
+  it('is idempotent on Windows — drops prior managed entry when path uses backslashes', () => {
+    const winCmd = buildManagedHookCommand('C:\\Users\\dev\\.nodeterm\\agent-hooks\\claude.sh')
+    const once = mergeManagedHook({}, winCmd, ['Stop'])
+    const twice = mergeManagedHook(once, winCmd, ['Stop'])
+    expect(twice.hooks!.Stop).toEqual([{ hooks: [{ type: 'command', command: winCmd }] }])
+  })
+
+  it('collapses accumulated duplicate Windows managed entries down to one', () => {
+    const winCmd = buildManagedHookCommand('C:\\Users\\dev\\.nodeterm\\agent-hooks\\claude.sh')
+    const accumulated = {
+      hooks: {
+        Stop: [
+          { hooks: [{ type: 'command', command: winCmd }] },
+          { hooks: [{ type: 'command', command: winCmd }] },
+          { hooks: [{ type: 'command', command: winCmd }] }
+        ]
+      }
+    }
+    const out = mergeManagedHook(accumulated, winCmd, ['Stop'])
+    expect(out.hooks!.Stop).toEqual([{ hooks: [{ type: 'command', command: winCmd }] }])
+  })
 })
 
 describe('mergeManagedHook — repair sweep', () => {
@@ -81,6 +102,17 @@ describe('mergeManagedHook — repair sweep', () => {
     }
     const out = mergeManagedHook(before, cmd, ['Stop'])
     expect(out.hooks!.Stop).toEqual([{ hooks: [{ type: 'command', command: cmd }] }])
+    expect(out.hooks!.SubagentStop).toBeUndefined()
+  })
+
+  it('drops stale managed entries with Windows backslash paths from unmanaged events', () => {
+    const staleWin = buildManagedHookCommand('C:\\Users\\dev\\.nodeterm\\agent-hooks\\claude.sh')
+    const before = {
+      hooks: {
+        SubagentStop: [{ hooks: [{ type: 'command', command: staleWin }] }]
+      }
+    }
+    const out = mergeManagedHook(before, cmd, ['Stop'])
     expect(out.hooks!.SubagentStop).toBeUndefined()
   })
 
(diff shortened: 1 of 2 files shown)


eneskirca:
Thanks for this — the diagnosis is exactly right: `managedMarkerFor` normalized to `/` while `isManaged`/`removeHooksFrom` matched the raw stored command, so on Windows we never recognized our own `agent-hooks\claude.sh` entry and appended a fresh set every launch.

**The catch is timing: #606 (fix/558-windows-hook-dedup) landed on `main` yesterday and fixes the same issue (#558) by the same root cause.** It introduces a single shared `normalizeHookCommand` used on both sides of every managed-entry comparison, and goes further than the two `replaceAll('\\','/')` call sites here:

- `codex.ts` imports the same normalizer, so the two installers can never drift again;
- `mergeManagedHook` strips every managed entry per subscribed event before re-adding one, so a Windows `settings.json` that already accumulated N duplicates collapses to one on the next launch (self-repair, not just idempotent install);
- `removeHooksFrom` and the strip are handler-level, so a user who hand-merged our command into their own definition keeps their handler.

This PR conflicts with `main` now (same lines in `install-helper.ts`/`.test.ts`). I re-applied your added tests on top of current `main` in an isolated worktree and they pass 24/24 against the #606 implementation — so the behavior you're pinning is already guaranteed there.

**Suggested: close as resolved by #606**, or, if you'd like to keep something from here, the only net-new value is your extra test cases — happy to see those cherry-picked onto `main` if they cover a shape #606's suite doesn't. Either way, thank you for the fix and the clear write-up; the analysis matched #606's independently. (Merge decision is the maintainer's.)

Abhirup0:
Thanks @eneskirca for the detailed breakdown! Glad to see the fix in #606 landed with the self-repairing cleanup logic and shared normalization. Closing this in favor of #606.

PR #609, opened 2026-09-01
fix(hooks): hand curl a payload path it can open on Windows

### LINKED ISSUE #605: Windows: all sessions show status Unknown — the hook payload path is POSIX and native curl cannot open it
## What happens

Every session sits under **Unknown** in the Sessions panel's Status tab. Need attention 0, Unread 0,
Running 0, Idle 0 — three sessions, all Unknown, including ones that are actively working.

Status comes from the agent's own hooks (`agent:status` IPC), not from parsing terminal output, so
this is a hook-delivery failure rather than a display bug. Confirmed:
`%APPDATA%\node-terminal\agent-status.json` holds `"nodes": {}` — no node has ever reported.

## Root cause: a POSIX path handed to a native Windows curl

`src/core/agents/hooks/managed-script.ts:198` builds the payload temp file as

```sh
nt_payload_file="$HOME/.nodeterm/pending/payload-$$.tmp"
```

and hands it to curl at `:258`, `:267` and `:360` as `--data-urlencode "payload@${nt_payload_file}"`.

The generated script runs under `sh` (Git Bash / MSYS), where `$HOME` expands to a POSIX path —
`/c/Users/<user>` — but `curl` on Windows is a **native** binary (both `C:\Windows\System32\curl.exe`
and Git's own `mingw64\bin\curl.exe`). It cannot open `/c/...`, and MSYS's argument translation does
not rewrite it because the value is embedded after `payload@`.

Reproduced directly:

```
$ curl -sS --data-urlencode "payload@/c/Users/<user>/.nodeterm/pending/probe.txt" …
curl: Failed to open /c/Users/<user>/.nodeterm/pending/probe.txt
curl: option --data-urlencode: error encountered when reading a file

$ curl -sS --data-urlencode 'payload@C:\Users\<user>\.nodeterm\pending\probe.txt' …
(no error)
```

So every hook POST fails, the app never learns a state, and every session stays Unknown.

Traced end to end with `sh -x` on the installed `~/.nodeterm/agent-hooks/claude.sh`: the curl line is
built correctly, the POST fails, the script falls through `nt_candidates` looking for another
endpoint, finds none, and returns 1 — silently, by design (a hook must never break the agent).

## Everything else in the chain is fine

Worth stating, because it narrows this to one line:

- env is complete in the session — `NODETERM_NODE_ID`, `NODETERM_AGENT_ID`, `NODETERM_HOOK_ENDPOINT`
- the endpoint is listening — `127.0.0.1:62620`, owned by the nodeterm process
- `hook-endpoint.env` is present and well-formed (`PORT`, `TOKEN`, `VERSION=2`, `NODE_TOKEN_DIR`)
- the per-node token file exists for each of the three nodes
- `curl` itself is available

## Steps to reproduce

1. Run nodeterm on Windows with any agent node.
2. Sessions panel → Status tab.

**Expected** — sessions grouped under Running / Idle / Need attention.
**Actual** — everything under Unknown; `agent-status.json` has `"nodes": {}`.

**Frequency** — always.

## Why it hurts

The Status tab is the reason to have one: on a canvas of long-running agents it is how you see which
one needs you. With every session Unknown it answers nothing, and the failure is silent by design —
hooks swallow their errors so they cannot break the agent, which is right, but it means nothing
anywhere says the status pipeline is dead.

It also disables everything else keyed off agent state: the node status chips, "waiting for your
response", and the phone's notifications.

## Suggested fix

Convert the path before it reaches curl, keeping the script POSIX-portable:

```sh
nt_payload_arg="$nt_payload_file"
command -v cygpath >/dev/null 2>&1 && nt_payload_arg=$(cygpath -w "$nt_payload_file" 2>/dev/null || printf %s "$nt_payload_file")
```

and use `payload@${nt_payload_arg}`. `cygpath` ships with Git Bash and MSYS2 and is absent
elsewhere, so the guard keeps macOS and Linux byte-identical.

The temp-file design itself is right and should stay — the comment above `:198` explains why the
payload must stay off curl's argv (`/proc/<pid>/cmdline` exposure, and E2BIG dropping large
PreToolUse bodies silently). Only the path form is wrong.

Same generator produces `codex.sh`, `gemini.sh`, `grok.sh` and `copilot.sh`, so this is one fix for
all of them. It also matches something visible in Codex CLI output on this machine —
`hook: PreToolUse Failed`, repeatedly.

---

Windows 11 Pro 10.0.26200 · nodeterm 0.3.4 · Git Bash `sh` · code references against `main` @ 4b6cfd9

ezwep:
Fixes #605.

## The bug

Every session shows status **Unknown** on Windows — Running 0, Idle 0, Need attention 0, with
sessions that are actively working sitting under Unknown.

It is not a display bug. Status comes from the agent hooks, and
`%APPDATA%\node-terminal\agent-status.json` holds `"nodes": {}`: no node has ever reported.

`managed-script.ts` writes the hook body to a 0600 temp file and passes it as
`--data-urlencode "payload@$nt_payload_file"`. The temp file is right, for the reasons documented
above it — keeping the body off curl's argv (`/proc/<pid>/cmdline` exposure, and E2BIG silently
dropping a large PreToolUse). The path **form** is what breaks.

The script runs under `sh` — Git Bash / MSYS — where `$HOME` expands to `/c/Users/<user>`. `curl` on
Windows is a **native** binary, and cannot open that:

```
curl: Failed to open /c/Users/<user>/.nodeterm/pending/payload-484.tmp
curl: option --data-urlencode: error encountered when reading a file
```

MSYS argument translation does not rescue it — the value is embedded after `payload@`, which its
heuristic does not rewrite. The same path spelled `C:\Users\...` works.

So every hook POST fails. Silently, and rightly so: a hook must never break the agent. Which is also
why nothing anywhere says the whole status pipeline is dead. Traced with `sh -x` against an
installed `claude.sh` — the curl line is built correctly, the POST fails, the script walks its
fallback endpoints, finds none, returns 1.

## The change

`nt_payload_arg` carries the path curl is given: `nt_payload_file` everywhere, converted with
`cygpath -w` under MSYS. A missing cygpath, a failing one, and an empty result all fall back to the
original path, so the worst case is today's behaviour rather than an empty argument. Cleanup keeps
the POSIX path, because `rm` there is the MSYS one.

**The gate is `uname -s` reporting `MINGW*`/`MSYS*`, not merely that a `cygpath` exists** — and that
distinction earned its place twice:

- A name check would fire on a WSL or Linux box with anything called `cygpath` on `PATH`, converting
  a path its own POSIX curl could already open. That breaks a case that works today.
- `uname` is read from the environment rather than inherited from it, so the gate still holds when a
  caller **replaces** the env — which is exactly what this file's own executed tests do. An env-var
  gate (`MSYSTEM`) passed review and then silently disabled itself under test, turning five green
  tests red again. That is how the `uname` form was found.

`CYGWIN*` is outside the gate on purpose: Cygwin ships its own curl and that one takes POSIX paths.

One generator serves claude, codex, gemini, grok and copilot, so this is one fix for all five. It
also explains `hook: PreToolUse Failed` in Codex CLI output on Windows.

## Verification

Windows 11, Node 24.19.0, vitest 4.1.11, measured against `upstream/main` on the same machine:

```
upstream/main   11 failed | 44 passed (55)
this branch      6 failed | 54 passed (60)
```

`npm run typecheck` clean. **Zero new failures**, and five of the repo's own tests go red → green —
all of them about the POST actually delivering: `makes the event verified when this node has a token
file`, `sends both credentials on stdin, never on curl's command line`, `is keyed by
$NODETERM_NODE_ID`, and the two unverified-post cases. They were failing for exactly this reason:
the script could not post at all. The remaining six are fallback-endpoint tests that fail
identically on upstream and are untouched here.

## Tests

Two existing assertions move from `nt_payload_file` to `nt_payload_arg` — the same four POSTs, named
differently.

Four new tests **execute** the generated script with a fake curl, cygpath and uname, and assert the
argv curl actually received in each branch: converted under MSYS, untouched on POSIX even with a
cygpath present, both fallbacks, and a converted path containing spaces. Review's point, and a fair
one: a string-matching test would have stayed green if the conversion block moved below the curl
calls, which would hand every POST an empty argument.

## Three surfaces

- **Desktop (Windows)** — where the bug bites; this is the fix.
- **macOS / Linux / Server Edition** — byte-identical: `uname -s` reports Darwin or Linux and the
  branch is never entered.
- **Mobile** — not applicable directly, though the phone's notifications are keyed off the same
  agent state, so they were dark on Windows for the same reason.

https://claude.ai/code/session_01Pn6BXSQNTRMB5Pnt4K88xi


ezwep:
**Confirmed fixed on a live Windows machine.**

I applied the change from #609 by hand to the five installed scripts in `~/.nodeterm/agent-hooks/`
(`claude`, `codex`, `copilot`, `gemini`, `grok` — 12 curl invocations in total) and restarted
nothing else.

Before:

```
agent-status.json   "nodes": {}
curl: Failed to open /c/Users/<user>/.nodeterm/pending/payload-484.tmp
```

After, traced with `sh -x` on the same script:

```
+ cygpath -w /c/Users/<user>/.nodeterm/pending/payload-1470.tmp
+ curl … --data-urlencode 'payload@C:\Users\<user>\.nodeterm\pending\payload-1470.tmp'
```

and the status pipeline came alive immediately:

```
nodes reporting: 1
  term-…  state=working   updated 2 s ago
```

The Sessions panel's Status tab now shows the session under **Running** instead of Unknown, and it
tracks working → idle as the turn ends. Nodes that have not fired a hook since the patch stay
Unknown until their next event, which is expected.

So the diagnosis holds end to end: one path form, five generated scripts, the whole status surface.

eneskirca:
**Verdict: merge-ready.** Reviewed on Linux in an isolated worktree off `origin/main`; merges cleanly (no conflict with the #606 hook work that just landed).

- Root cause and fix are right: under Git Bash/MSYS the script is POSIX `sh` but `curl` is a native binary, so `$HOME`→`/c/Users/...` after `payload@` cannot be opened, and the `--data-urlencode "payload@<file>"` form is what breaks. `cygpath -w` under the `uname -s` MINGW*/MSYS* gate is the correct spelling.
- The `uname -s` gate (over an `MSYSTEM` env check or a bare `command -v cygpath`) is the load-bearing choice and it holds: it won't fire on a WSL/Linux box that happens to have a `cygpath`, and it survives a caller replacing the env — which the executed tests here actually depend on.
- Fallbacks are safe: a missing/failing/empty `cygpath` all fall back to `nt_payload_file`, never to an empty `payload@` argument. The four executed tests (fake curl+cygpath+uname, asserting real argv) verify that, and would catch the conversion block being moved below the curl calls — a string-match test would not.
- One generator, so this fixes all five agents at once, and `remote-hooks.ts` reuses `buildManagedScript`, so the SSH-remote hook path is covered too (remote hosts run POSIX curl, so the gate is inert there — correct).

Verified locally: `npm run typecheck` clean; the hooks suites + `no-electron` + `fs-atomic.guard` = 144 passed. CI (including `quality-windows` on a real runner) is green. The live-Windows confirmation in your comment closes the device-verification loop; I could not re-run that leg on this Linux box.

PR #611, opened 2026-09-01
fix(terminal): make project-jump Cmd+1..9 bubble to dispatcher

AlexanderWillner:
Extracted from #588 per review — standalone regression fix.

The Cmd/Ctrl+1..9 project-jump chord must prevent xterm's control-byte write (Linux/Windows Ctrl+2..8 are `^@ ^[ ^\ ^] ^^ ^_`) without calling `preventDefault`. A swallowed event is marked `defaultPrevented` and the window's bubble-phase dispatcher (`Canvas.projectJumpGesture`) bails, which is why the jump stopped working from a focused terminal/modal.

- `terminalKeyAction` now returns `'bubble'` for `ownsProjectJump` (returns `false` from xterm's handler, skips keymap, event still bubbles)
- JSDoc rewritten to match
- Tests updated to expect `bubble`

This was buried in #588's 819-line feature PR; extracting lets it land fast independent of the swimlane product decision.

Ref: https://github.com/eneskirca/nodeterm/pull/588#issuecomment-5498602893 point 3, and https://github.com/eneskirca/nodeterm/pull/588#issuecomment-5500480388

### DIFF
--- a/src/renderer/terminal/terminal-config.test.ts
+++ b/src/renderer/terminal/terminal-config.test.ts
@@ -550,9 +550,10 @@ describe('terminalKeyAction', () => {
     expect(
       terminalKeyAction(ev({ key: 'Enter', code: 'Enter', shiftKey: true }), false, false, true)
     ).toBe('shift-enter')
-    // project-jump swallow wins over bubble
+    // project-jump bubbles to the dispatcher (which performs the switch) — it must
+    // still beat a registry bubble's reason but reach the window as a bubble, not a swallow
     expect(terminalKeyAction(ev({ key: '1', code: 'Digit1', metaKey: true }), false, true, true)).toBe(
-      'swallow'
+      'bubble'
     )
     // keyup never bubbles — the dispatcher only acts on keydown
     expect(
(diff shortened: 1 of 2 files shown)


eneskirca:
Verified on an isolated worktree of this head on top of current `main` (post-#622): `npm run typecheck` clean; `terminal-config`, `globalKeybindings`, `projectJump`, `keybindings` and `keydown-intercept` suites pass (5 files / 267 tests); merges clean with `main`.

Checked against the keybindings contract in CLAUDE.md:
- Both consumers (`TerminalNode` and `ModalTerminal` `attachCustomKeyEventHandler`) return `false` for `'bubble'` **without** `preventDefault`, so `dispatchGlobalKeydown`'s `defaultPrevented` bail no longer eats the chord and `projectJumpGesture` runs on the bubble phase as designed. Returning `false` from xterm's handler also skips its keymap, so no `^@`/`^[` reaches the pty.
- Terminal-first stand-down is untouched: `ownsProjectJump` is already `false` under that policy, so the digit still reaches the PTY.
- Server Edition unchanged: `liveProjectJumpTarget` is `null` in a browser tab, so the action stays `'pass'` there (the browser owns ⌘1..9 anyway).

Looks good to me. Merge decision is @eneskirca's.

PR #612, opened 2026-09-01
feat(accounts): track which Claude account each session runs as; link existing config dirs; per-node account chip

howieyoungdahl:
## Why

Managed Claude accounts assume nodeterm launched the session: `data.accountId` is stamped at node creation and every account-aware reader keys off it. Two real workflows fall outside that model — a user who keeps a second login in their own config dir (`export CLAUDE_CONFIG_DIR=~/.claude-2; claude …` in a plain terminal), and any hand-launched `claude` whose identity diverges from the node's. For those panes nodeterm had no account at all: the transcript jail rejected `~/.claude-2/projects/**` (so the context meter and subagent cards stayed empty), the usage popover only saw `~/.claude`, and nothing on the canvas said which of several panes was on which account.

## What changed

- **Observed account.** The hook server derives which config dir a Claude session runs from out of the payload's `transcript_path` (`<configDir>/projects/<slug>/<session>.jsonl`) and attaches an `ObservedClaudeAccount` LABEL to the normalized event beside `verified` / `clientRevision` — one implementation, both shells inherit it, neither raw listener changes. Pure string classification (managed local, managed remote, linked, system, unknown), never throws, never reads the filesystem: a forged POST naming `~/.ssh/projects/x` gets `known: false` and nothing is opened. Recorded by the mirror (`MirrorEntry.account`) and the renderer agent-status store (persisted like `agentId`).
- **Linked accounts.** `ClaudeAccount.configDir` adopts a pre-existing local config dir as a first-class account (`claude-accounts:link`; Settings → Accounts → *Link existing config dir…*, or one click on a *Detected config dir*). `claudeConfigDirFor` resolves through a registered accounts source, so env injection, transcript roots, the transcript index, usage rows and the pickers all work for a linked id with no per-caller branch. The transcript jails accept `<linkedDir>/projects/**` for dirs from settings only. Removing a linked account only forgets the record: the `rm -rf` names the managed root directly and cannot reach the user's dir. The hook installer writes `settings.json` through a symlink, so a profile that symlinks `settings.json` to `~/.claude/settings.json` stays a symlink (pinned by a real-fs test).
- **Account chip.** One `AccountChip` component on the terminal header, kanban card, card modal and sessions-sidebar row. Shown for any non-system account, and for system panes only when ≥ 2 distinct accounts are live on the core, so two panes side by side are always told apart. An unlinked dir is named by its last path segment (`.claude-2`, dashed) with a tooltip pointing at Settings → Accounts. Readers (`readSessionName`, `context.ensure`, transcript search, ⌘M view) use the effective account (`data.accountId ?? observed`); spawn/env never does.
- Docs: `CLAUDE.md` (Managed Claude accounts → *Linked accounts*, *Observed account*), `docs/SERVER.md`, and the design plan `docs/superpowers/plans/2026-09-01-claude-account-tracking.md`.

## Three surfaces

- **Desktop:** full — link, chip, readers, usage row, pickers.
- **Server Edition:** full — same core service over the ws-bridge; `Browse…` degrades to the path field where `selectFolder` is unsupported.
- **Mobile:** N/A. `MirrorEntry.account` and `MirrorSettings.claudeAccounts[].dir` (now a linked dir when applicable) are additive; @eneskirca the phone could render the same chip from `nodes[<id>].account` if wanted.

## Verification

- `npm run typecheck` — passed (both projects).
- `npm run build` + `npm run server:build` — passed (with a lock-accurate `npm ci`; the first attempt on a sibling worktree's `node_modules` failed on monaco 0.52 vs 0.56 drift, unrelated to this branch).
- Targeted vitest (13 files: `claude-accounts-core`, `claude-config-dir`, `claude-accounts-service`, `claude-accounts-link-symlink`, `agents/hook-account-label`, `agent-status-mirror`, `server/agent-status`, `hook-verified-parity`, `renderer/lib/accountChip`, `AccountChip`, `AccountsSection.link`, `agentStatus.persist`, `ws-bridge.namespaces`) — 392 passed.
- Full suite `npx vitest run` — 9,750 passed, 40 skipped, 5 failed in 3 files, all pre-existing and reproduced on the unmodified base commit / documented: `node-pty-patch` (unpatched `node_modules`, per CONTRIBUTING), `git-service.worktree-remote` (5 s ssh timeout; passes at 30 s), `test/server/headless-e2e` (fails identically on `origin/main` in this environment).
- Mutation checks: dropping the hook-server `account` spread (6 red), the mirror capture (2), the linked-remove guard (1, the real fixture dir got deleted), the server jail widening (1), the write-through-symlink install (1), the chip visibility rule (5), the store persistence gate (2), `resolveObserved` (7) — each reddened its tests; all reverted.
- Isolated Server Edition smoke (own `--data-dir`, `--port 18443`, fake `HOME`, own `TMUX_TMPDIR`; never the live install): two plain terminal nodes; replayed real hook payloads with `transcript_path` under `<fakehome>/.claude-2/projects/…` and `<fakehome>/.claude/projects/…`. Before linking: node 1 header + sidebar row + kanban card show the dashed `.claude-2` chip, node 2 shows the system chip (visible because two identities are live). Settings → Accounts listed the dir under *Detected config dirs*; one click linked it (email captured from `.claude.json`, managed hook installed into the fixture dir, row shows `Linked` + an *Unlink* button whose confirm says the folder is kept). After linking, the very next hook event classified the pane as the linked id and the mirror advertised the linked dir; with the renderer re-resolution, the chips (node header, sidebar row, kanban card) flip to the linked label the instant the dir is linked — and back to the dashed unlinked chip (dir back in *Detected*) the instant it is unlinked — with no further hook event in either direction.
- The hook installer's write-through-symlink behaviour was confirmed on a real fixture where `settings.json` symlinks to another profile's file (the two-profile layout): still a symlink after `link`, target gained the hook, user's own keys preserved.

Test-hygiene fix included: the symlink test runs the REAL installer, whose managed script lives per machine at `$HOME/.nodeterm/agent-hooks/<agent>.sh`; it now runs under a scratch `HOME` so a developer's live hook script is never rewritten by the suite.

https://claude.ai/code/session_01EWZaBhXLTFnyyPh6eTtDav


eneskirca:
Pulled `8a920879` into an isolated worktree on current `main` (8be08bc1): `npm run typecheck` clean, full `npx vitest run` **9781 passed / 40 skipped / 0 failed** (`node-pty-patch` included — this machine's `node_modules` is patched). `git merge-tree` against `main` and against #319: no conflicts in either direction. Approved the fork CI runs.

What I verified and agree with:
- The label is computed once in `hook-server.ts` and neither raw listener changes; both jails (`src/main/index.ts`, `src/server/agent-status.ts`) widen per linked ROOT (`<dir>/projects/**`) from settings only. `hook-verified-parity`, `no-electron`, `fs-atomic.guard` all green.
- `remove` on a linked id returns before the `rm -rf`, and the `rm -rf` line is `accountConfigDir(userData, id)` — structurally unable to name a linked dir even if the renderer drops the settings row first (then the managed path simply does not exist; `force: true`). Good.
- `data.accountId` stays immutable; only readers use `data.accountId ?? observed`, spawn/env never does — consistent with CLAUDE.md.
- Link refuses `~/.claude`, the managed root and duplicates by normalized path; `stat` follows a symlinked dir (correct — the CLI resolves through it too); `..` cannot survive; `installClaudeHooksInto` keeps a symlinked `settings.json` (real-fs test). On Server Edition the handler runs as the server user against its own `$HOME` — the same trust boundary the fs API already exposes, so no escalation (but see 3).

**One thing to fix before merge:**

**1. Remote (SSH-project) panes leak into the local link flow.** `classifyClaudeConfigDir` is deliberately host-agnostic and the Canvas listener calls `cs.setAccount` for every node, so an SSH node whose hooks post `/home/u/.claude-2/projects/…` *on the host* lands in the store as `known: false`. From there:
- `unlinkedConfigDirs` lists that remote path under **Detected config dirs** with a **Link** button that runs the LOCAL `claude-accounts:link`, i.e. it `stat`s the desktop's `/home/u/.claude-2`. Usually "No such directory"; if a same-named dir exists locally (same username on both machines is the common case) it links an unrelated local dir, installs the hook into it and widens the local jail — for a pane that runs on another machine.
- The unlinked chip's tooltip tells the user to link it in Settings → Accounts, which cannot be done for a remote dir (link is local-only by design, as the API doc says).
- Classifier rule 3 matches a linked local dir against a remote pane's path by string equality, so after such a link the remote pane is labelled as the local linked account and `effectiveAccountId` hands that id to the local readers.

Suggested shape: gate where the label enters the renderer (or at least the detected list + the tooltip) on the node being remote — `isRemoteSessionNode(data)` (`shared/worktree.ts`) / the project being an SSH project, the same predicate the worktree gate uses, never `data.remote`. Managed-remote labels (rule 2) can stay, they carry the right id. A test "an SSH node's unknown dir is never offered for linking" would pin it.

Smaller:
2. Code comments and the PR body cite a design doc (`docs/superpowers/plans/2026-09-01-claude-account-tracking.md`, "design D1/D4/D7", "§3 verbatim") that is not in the diff. Either include it or rephrase so the next reader does not chase a file that is not there.
3. `docs/SERVER.md`: one sentence that on Server Edition *Link* stats and writes (`settings.json`, hook) into any directory the **server user** can reach — `~` there is the server's home, not the browser user's.

Everything else reads merge-ready to me; the merge call is @eneskirca's. #319 touches the same area; file overlap is only `CLAUDE.md` / `src/main/index.ts` and the two merge clean in either order.

howieyoungdahl:
Thanks — that is a real bug and the analysis is right. Fixed, plus the two smaller items. Pushed as `cf9bd2eb`.

**First, a heads-up you did not cause:** `main` has moved since your review (`8be08bc1` → `9e76faf8`) and this branch no longer merged — seven files conflicted. I resolved it with `git merge origin/main`, **not** a rebase, so the commit you reviewed (`8a920879`) is still in the history exactly as you read it. All seven conflicts were additive on both sides and both halves are kept; the load-bearing one is `isSafeLocalTranscriptPath`, where main added `grokHomeDir` and this branch added `linkedDirs` — both kept, main's in its original argument position so no upstream call site moves. Say the word if you would rather have a linear rebase.

**1. Remote panes in the local link flow.** Confirmed end to end, and one detail is worth stating because it decides where the fix belongs: `observedClaudeAccount` only ever sees the POST payload, so **core is structurally unable to know** whether the posting node is remote. `classifyClaudeConfigDir` being host-agnostic is correct and stays — rules 2 and 4 exist precisely to answer for remote hosts. The renderer is not merely the convenient place to gate; it is the only one.

So `ObservedClaudeAccount` gains an optional `remote`, stamped once where you pointed — the Canvas `agent:status` listener — by a new pure `observationIsRemote`, which ORs `isRemoteSessionNode(data)` (never `data.remote`) with the owning project being an SSH project. **A node it cannot place reads as remote**, because only the "local" guess is destructive.

From there: `unlinkedConfigDirs` skips remote observations (that is the Link button gone), `resolveObserved` refuses the dir match in both directions, and the `ext:` tooltip stops telling the user to do something they cannot do. Managed-remote labels are untouched — they resolve by id, never by path — pinned by its own test plus a mutation that deliberately breaks it. The flag round-trips through the agentStatus localStorage explicitly rather than relying on absence failing safe, because it does not.

Your test is in: **"an SSH node's unknown dir is never offered for linking"**, alongside the remote tooltip, the remote-vs-local-linked-dir collision, the managed-remote non-regression, and the Settings "Detected config dirs" omission. Seven mutation checks, each applied → red → reverted.

**2.** 27 citations of the plan document rewritten across 21 files: the invariant is now stated inline instead of pointing at a file you cannot open, and the D-number shorthand is gone.

**3.** One sentence in `docs/SERVER.md`: on Server Edition, Link stats and writes into any directory the **server user** can reach, and `~` there is the server's home, not the browser user's.

Gates: `npm run typecheck` clean; 11 targeted files 412 passed; `npx vitest run src/server src/core` 5 failed / 3718 passed, and I proved those 5 pre-existing by detaching to `origin/main` in the same worktree and reproducing the identical set. `server:build` OK.

Two things I did **not** change, flagged rather than hidden. `accountKey` still keys two unlinked observations of the same absolute path as one identity even when they are on different machines — it costs at most one chip, and closing it means keying on host, which is a bigger change than this fix deserves. And I think the right long-term home for the remoteness fact is `HookEventMeta` beside `verified`, with the predicate injected per shell — that would give the Server Edition and the mobile mirror the same fact for free instead of it living only in the renderer. Out of scope here; happy to open an issue.

eneskirca:
Re-verified `cf9bd2eb` in a fresh isolated worktree on current `main` (3411efdf): typecheck clean, full `npx vitest run` **10332 passed / 40 skipped / 0 failed**. `git merge-tree` vs `main`: 0 conflicts; vs #540 (shares `hook-server.ts` / `src/main/index.ts` / `CLAUDE.md`): 0 conflicts. Approved the fork CI runs. Checked the merge resolution rather than trusting it: `isSafeLocalTranscriptPath` now takes `codexHomeDir, grokHomeDir, linkedDirs` and both jail call sites (`src/main/index.ts`, `src/server/agent-status.ts`) pass `grokHomeDir()` and `linkedClaudeConfigDirs()`. Plan references are gone from the diff (the remaining `docs/superpowers` mentions are pre-existing, unrelated files), and the SERVER.md sentence is there. Keeping the merge commit rather than rebasing is fine by me.

The remote gate is the right shape — renderer-side because core cannot know, "cannot place ⇒ remote", `isRemoteSessionNode` OR project-is-SSH, the flag round-tripped through localStorage. Detected list, tooltip and both `resolveObserved` directions are covered.

**One branch is still open, and it is the collision case the new test names:** `classifyClaudeConfigDir` rule 3 in core is unchanged and host-agnostic, so when a LOCAL linked account holds `/home/me/.claude-2`, a remote pane posting that path arrives at the renderer already as `{known: true, accountId: 'lnk'}` — not as `known: false`. `resolveObserved` then takes the `accounts.some(a.id === accountId)` early return before `dirMatchable` is ever consulted, so the remote pane is labelled with the local linked identity and `effectiveAccountId` hands `lnk` to the local readers. The test at `accountChip.test.ts` "is never matched against a local account that happens to hold the same path" starts from `known: false`, which is not what the hook server emits in that situation, so it passes while the path stays open. The same-username layout you already call "not a corner case" is exactly the one that triggers rule 3.

Smallest fix in the layer you chose: in `resolveObserved`, when `observed.remote` and the matched account is a LINKED one (`configDir` set, no `host`), demote to `{accountId: null, known: false}` — a linked account is local by definition, so a remote observation can never legitimately resolve to it; managed-remote ids (`host` set) stay untouched. Plus a test whose input is `{known: true, accountId: 'lnk', remote: true}`. Alternatively, and cleaner long-term as you say, carry remoteness in `HookEventMeta` so core can skip rule 3 for remote posts.

Other than that this is merge-ready from my side; the call is @eneskirca's. `accountKey` keying two same-path observations across machines as one identity is fine to leave, as noted.

howieyoungdahl:
Your trace held in full, and it was the right thing to check. Fixed in `67017a2c`.

Confirming the mechanism in my own words, because the ordering is the whole point: core's `classifyClaudeConfigDir` rule 3 dir-matches linked accounts **host-agnostically at POST time**, while `remote` is stamped later, in the Canvas listener. So a remote pane reaches `resolveObserved` already as `{known: true, accountId: 'lnk', remote: true}`, the `accounts.some(...)` early return fires before `dirMatchable` is ever consulted, and `effectiveAccountId` hands `lnk` to the local readers. Exactly as you described.

**The fix is the one you suggested**, in the layer already chosen: a remote observation that resolves to a **linked** account (`configDir` set, no `host`) demotes to `{accountId: null, known: false}`. Managed-remote (`host` set) and managed-local (no `configDir`) are untouched.

**On the test that gave false confidence — I kept it and renamed it rather than deleting it.** Starting from `known: false` is not a wrong scenario; it is the *link-after-observation* ordering, which is real and reachable (the dir is observed first, the user links it afterwards). It was just not the ordering the hook server emits when the account already exists, which is the case your finding is about. It now says which ordering it pins, and two new tests cover the production shape — one whose input is literally `{known: true, accountId: 'lnk', remote: true}` — plus the pending / broken-configDir / managed edges.

Mutation checks: removing the demotion reds both new tests; widening it to include `host` accounts reds the managed-remote non-regression; adding a `!pending` condition reds the widening test. All reverted; targeted suites 71/71, typecheck clean.

**One thing you should know before you merge it, because it is a real cost of this shape and I do not think it is free.** `observationIsRemote(undefined)` returns `true` — "cannot place ⇒ remote" — so a **local linked** pane whose node the canvas can no longer place now also demotes: its chip falls back to the bare dir and its readers fall back to the system root. That is a visible regression for a node that has been deleted from the canvas while its session is still posting, or during a load where the lookup is momentarily empty.

I made the change anyway, and wrote the cost into the comment, because the alternative is worse: narrowing the demotion to "known remote" would reinstate the `unknown ⇒ local` guess that the whole `remote` flag exists to refuse, and a wrong *local* label is the destructive direction — it points the readers at another account's transcripts, which is the bug we started from. A wrong *unknown* label only costs a chip. If you would rather have it the other way I will take the change, but I think this is the right trade and it should be a decision rather than an accident.

Your longer-term suggestion still stands as the cleaner answer: carry remoteness in `HookEventMeta` so core can skip rule 3 for remote posts, which removes the ambiguity instead of resolving it defensively. Happy to open that as an issue rather than grow this PR.

eneskirca:
`67017a2c` is exactly the fix, and it goes further than what I described — you found the ordering I had not separated. Verified rather than read:

- **Mutation check as asked:** deleting the `observed.remote && isLocallyLinkedAccount(held)` branch reddens **2 tests** in `accountChip.test.ts`; reverted, 105 pass across `accountChip` / `agentStatus.persist` / `AccountChip`. So the new case genuinely discriminates the fix, which the old `known: false` case could not.
- **The split is the right one.** The old test pinned "linked AFTER the observation" (nothing claimed the dir at POST time); the new one pins "linked BEFORE" — core's rule 3 already stamped the id, `resolveObserved`'s id branch returns first, `dirMatchable` is never consulted. That second ordering is what the hook server actually emits for a user with a linked account and an SSH project, and it is the one that was open.
- **The non-regression holds:** managed-remote (`host` set, id from the path SHAPE) and managed-local (no `configDir`) still resolve; only an id that core derived from a path on THIS machine is demoted. Both directions covered by the added cases, including the `pending` and unusable-`configDir` widenings.
- **The stated cost is the right trade** and I would not narrow it: a local linked pane whose node the canvas can no longer place also demotes, because `observationIsRemote(undefined)` is true. One label fewer, never a wrong one — splitting "on a host" from "cannot place" would reinstate the unknown ⇒ local guess.

**Gates on your head, and on your head merged with current `main`** (your base is 58 commits back, and #540/#319 have landed since, so I ran both):

- branch head: typecheck clean, full `npx vitest run` **10333 passed / 40 skipped**, 1 unrelated failure — `codex-relay-daemon.test.ts`'s stale-control-port timeout, which passes 35/35 in isolation and touches no file this PR changes. Load-flaky, not yours.
- merged with `main` (a395e135): typecheck clean, full suite **10479 passed / 40 skipped / 0 failed**, no conflicts.

CI runs were sitting on fork approval; approved, so the checks should clear now. Nothing left from my side — merge is @eneskirca's call.

PR #614, opened 2026-09-02
feat(grok): link two grok nodes and read each other's transcript (context links)

xaviguardia:
## What this does

Two grok nodes on the canvas could not be linked. The canvas refused the gesture with
`not linkable (needs two context-capable agents)`, because `CONTEXT_LINK_CAPABLE` did not list
grok. This adds grok to that list and gives it the one thing membership requires: a locator.

That refusal is the whole feature seen from the user's side, so it is what this PR measures
against — see **Evidence** below, which shows the red before the green rather than only the green.

## The trap this exists to avoid

The locator resolves **`chat_history.jsonl`**, *not* the `updates.jsonl` that grok's own hook
payloads advertise. Routing through the advertised path does not raise: it hands the reader an
**empty transcript, with nothing logged**. A silent failure in a feature whose entire job is to
show you text is the worst shape available, so `grok-context-link.test.ts` pins both halves — the
file we do resolve, and the file we must not.

The session directory comes from the hook-fed `sessionId → dir` map and there is deliberately **no**
fallback that scans grok's sessions tree. A scan keyed on anything weaker than the session id is how
a node ends up reading a stranger's conversation.

## The path jail

`isSafeLocalTranscriptPath` is widened to `$GROK_HOME/sessions` — that subdirectory, and **not**
`$GROK_HOME`, because `auth.json` lives in the same tree. It honours a relocated `$GROK_HOME` in
**both** shells (`src/main/index.ts`, `src/server/agent-status.ts`): invariant 11 — the two raw hook
listeners change together, or Server Edition silently loses the feature.

Getting this wrong fails **closed**: the reader simply never reads, with no diagnostic. That is why
one of the three verified mutations is the dangerous one — widening the jail to the whole of
`$GROK_HOME`, which would expose `auth.json`.

## Consumers audited before joining the list

The `hasUsage` lesson (it opened three features when it looked like one). **Seven calls** across two
files — nine mentions, if the two imports are counted:

| Where | Calls | What it gates |
|---|---|---|
| `renderer/canvas/Canvas.tsx` | 6 | the link gesture; **both `verify` doors** (`:8959` target, `:8967` reviewer); node-kind descriptors for link targeting (`:2900`, `:8678`, `:9062`, `:9158`) |
| `renderer/nodes/TerminalNode.tsx` | 1 | tip wording only (`:1342`) |
| `renderer/lib/noteLink.ts` | 0 | no call — but it documents the concept, and the documentation was wrong |

**So joining the list makes a grok node eligible both to be reviewed and to review.** Two separate
doors:

- `:8959` — can this node be the **target** of a verify? *"not an agent session whose work can be read"*
- `:8967` — can this node be the **reviewer**? *"cannot read linked context, so it cannot review anything"*

That is intended, and it is written here rather than left to be discovered a year from now.

`noteLink.ts` claimed the capability was "(claude/codex/gemini)" and had been wrong since opencode
joined. It now points at `canContextLink` instead of restating a list that drifts.

## Three surfaces

- **Desktop** — yes.
- **Server Edition** — yes, same core, driven off the persisted `bridges[]` rather than the
  renderer's live edges.
- **Mobile** — N/A: linking is a canvas gesture and the phone has no canvas. The file it would read
  is the same one, so this is a surface gap, not a capability gap.

No `window.nodeTerminal` member was added — the renderer only asks the pure `canContextLink`, so
nothing new crosses the preload bridge. No `=== 'grok'` at any call site: membership plus one leaf.

## Docs

`docs/grok-agent.md` is corrected — including claims this branch did not introduce but **made
false**:

- the Server Edition column said `initContextLink` is never called from `src/server`. It is.
- `CHAT_CAPABLE` and `TRANSFER_SOURCE_CAPABLE` still blamed "the parser Task 10 is blocked on". That
  parser now exists. Both stay **not joined** — but as an *unmade decision*, not a missing piece.
  Scope is unchanged; the stated reason is now true.
- checklist item 21 is closed **by measurement**: `grok inspect --json` lists `get-linked-context`
  with `vendor: claude`, `compatibilityStatus: enabled`, and `externalCompat.cells` →
  `{surface: 'skills', enabled: true, source: 'default'}`. `source: 'default'` is the part that
  matters: grok reads `~/.claude/skills` with no config edit by the user. What item 21 still asks of
  a new device is narrowed to the two per-user failure modes a default-config machine cannot
  exercise.

## Evidence

Red and green were both captured on a real canvas with two real grok nodes, with literal commands and
outputs recorded so a reviewer can repeat them without repeating the session.

**Red** — the same command, against the currently shipped package (whose bundle carries
`CONTEXT_LINK_CAPABLE = ["claude","codex","gemini","opencode"]`):

    link: nothing linked — <node>: not linkable (needs two context-capable agents, ...)

**Green** — against a build of this branch, two grok nodes linked on the canvas, each reading the
other:

    === Grok Node Terminal Context Link Setup — full transcript (6 lines) ===
    ...
    user: <user_query>
    [nodeterm] You are now linked to "Grok". ...
    assistant: Linked to **Grok**. I'll pull its context only when needed; standing by.

The resolved `transcriptPath` ends in **`chat_history.jsonl`**, and the source file holds 7 JSONL
lines of which 6 are emitted — the omitted one is `type: "reasoning"`, skipped deliberately. The two
lines carrying `synthetic_reason: "system_reminder"` are emitted **as `system_reminder:`, not as
`user:`**, which is the task06 behaviour confirmed on a live session rather than on a fixture.

A "linked, 0 lines" result would not have been green — it is precisely the silent failure that
routing to `updates.jsonl` produces.

## Tests

`grok-context-link.test.ts` (new): the locator resolves `chat_history.jsonl` and **not**
`updates.jsonl`; it learns nothing without a hook; it returns undefined when no chat history exists;
the jail admits the grok root, honours a relocated `$GROK_HOME` **and stops admitting the default**,
and refuses `~/.ssh/id_rsa`, `~/.grok/auth.json` and `$HOME`; plus the capability invariant.

Three mutations verified red, each with a `diff` proving the intended line actually changed before
the result was read.

## Not verified

Stated rather than omitted, so nobody has to rediscover it:

- **A REMOTE grok node, over the SSH reverse tunnel, still will not resolve.**
  `isSafeRemoteTranscriptPath` — the remote analogue of the jail widened here — knows only claude's
  roots. This is **pre-existing and shared with codex and gemini**; it is not introduced by this
  change, and widening it is a separate decision with its own failure modes.
- **The live two-node proof ran on an instance without hook identity.** The macOS keychain dialog
  was cancelled, so the reader presented itself as `legacy` rather than `verified`. The result does
  not depend on that — the per-node token governs *who* may request a read, not *which* file is
  resolved or *how* it is rendered, and this change touches only what sits below that door — but the
  verified path itself was not exercised here.


eneskirca:
Reviewed at 9b00318 in an isolated worktree: typecheck clean, focal suites (`grok-context-link`, `context-link-render`, `config.capabilities`, `claude-accounts-core`, `context-link`, both `no-electron`, `server/agent-status`) 99/99, CI approved and green.

Verified the one claim I could measure here: `grok inspect --json` (1.0.13, fresh `GROK_HOME`, no login) lists both `get-linked-context` and `manage-nodeterm-canvas` from `~/.claude/skills`, with the `skills` compat cell `enabled:true, source:'default'` — so checklist item 21 is closed by a measurement that reproduces.

The jail widening to `$GROK_HOME/sessions` (not `$GROK_HOME`), the hook-fed `sessionId → dir` map with no tree scan, and the both-shells `grokHomeDir()` pass are all as they should be. Merge-ready from my side; the merge call is @eneskirca's.

PR #617, opened 2026-09-02
feat(grok): cross-agent transfer and the Cmd+M chat panel

xaviguardia:
> **Stacked on [#614](https://github.com/eneskirca/nodeterm/pull/614)** — review the top commit(s) only; the rest land with #614.

## What this does

Grok joins `TRANSFER_SOURCE_CAPABLE` and `CHAT_CAPABLE`. Both ride the reader the context-link PR
shipped, so each capability needs exactly one new leaf: a Markdown renderer for transfer, a message
builder for the ⌘M panel. Neither re-derives grok's line vocabulary — they build on the same
`grokParse`, so two views of one file cannot drift apart.

## The part that needs review: `CHAT_CAPABLE` had to be split

`CHAT_CAPABLE` carried **two** facts that coincided only while claude was its single member:

1. *we can render this agent's conversation ourselves*
2. *CLAUDE's transcript resolver can locate and parse this agent's file*

Grok satisfies (1) and not (2). And `transcriptGates.readsClaudeTranscript` — the gate for the find
bar's transcript index and the meter's mount-time rehydration — was reading `CHAT_CAPABLE` for
**(2)**, on a comment explaining that a separate list "would mean the same thing". It did, once.

So (2) now lives in `CLAUDE_TRANSCRIPT_READABLE` (claude only), and `CHAT_CAPABLE` keeps (1).

**Why this is not bikeshedding.** `resolveTranscript` falls back to `transcriptPathForCwd` when its
sessionId leg misses — which a grok id *always* does, since no `<that id>.jsonl` exists under
`~/.claude/projects` — and returns **the newest claude transcript for the node's directory**. A grok
node left in the merged list would show a stranger's conversation in its find bar and rehydrate its
meter from that session.

This repo already survived exactly this bug once: codex and gemini joined `USAGE_CAPABLE` while
`hasUsage` gated those two features, and a codex node metered and searched an unrelated claude
session. `transcriptGates.ts` documents it in full. Merging the lists back reintroduces it pointed at
a new agent — and it fails **open**: it shows data rather than hiding it.

`config.capabilities.test.ts` therefore pins the pair together:

    expect(canChat('grok')).toBe(true)
    expect(readsClaudeShapedTranscript('grok')).toBe(false)

If a future change simplifies one list away, that line fails first.

## Transfer

`renderGrokTranscript` sits beside the claude/codex/gemini renderers, registered in the same two
maps. One behaviour is deliberate and worth naming: harness-injected text gets its own heading —
`## Injected (system_reminder)`, never `## User`.

Grok files skill reminders, compaction notes, project instructions and subagent completions under the
`user` role, marked only by `synthetic_reason`. A handoff document is read by the **receiving** agent,
which treats a `## User` heading as an instruction from the person. Mislabelling one hands it
somebody's tooling as a command.

The **destination** list is untouched: `transferTargets` derives targets from `BUILTIN_AGENT_IDS` and
never consulted `TRANSFER_SOURCE_CAPABLE`, so grok was already a valid target. Unifying the two would
light a menu over an operation that does nothing.

## Chat

`chat:read-transcript` takes an `agentId` and routes on it. The channel is registered through the
CorePlatform seam, so **both shells** serve it and both route the same way.

## Consumers audited before touching either list

| List | Consumers | What they ask |
|---|---|---|
| `CHAT_CAPABLE` | `TerminalNode.tsx:1344`, `transcriptGates.ts:36` | two **different** questions — which is how the conflation surfaced |
| `TRANSFER_SOURCE_CAPABLE` | `transferItems.tsx:113` | the menu only; targets come from elsewhere |

## Tests

New: `render-grok.test.ts`, `grok-chat.test.ts`, the agent-routing block in `transcript-ipc.test.ts`,
and the paired invariant in `config.capabilities.test.ts`. The routing tests write a claude transcript
and a grok history for the *same* cwd and assert the claude one never comes back.

One existing test changed rather than being deleted: `transferItems.test.tsx` used grok as its example
of a non-transfer-capable agent. It now uses opencode, whose renderer genuinely is unwritten — an
example that stops asserting anything the moment its subject gains the capability is worse than no
example.

Three mutations verified red, each with the diff printed before the result was read: merging the two
lists back, disabling the agent routing (the leak assertion fires by name), and unregistering the
transfer renderer. **The first two passed green over nothing until the tests in this change existed.**

## Not verified

- The chat panel was exercised through its IPC handler, not through a mounted React panel on a live
  grok node.
- Transfer was exercised through `buildHandoff`'s capability path and the renderer's output; a
  full canvas transfer from a grok node into a claude node was not performed.


eneskirca:
Reviewed the top two commits at ec19d2f (on #614) in an isolated worktree: typecheck clean, focal suites (`render-grok`, `grok-chat`, `transcript-ipc`, `config.capabilities`, `transferItems`, `handoff/index`, both `no-electron`) 77/77, CI approved and green.

The `CHAT_CAPABLE` / `CLAUDE_TRANSCRIPT_READABLE` split is the right call and the pinned pair (`canChat('grok') && !readsClaudeShapedTranscript('grok')`) is exactly the guard the codex/gemini episode needed. Routing `chat:read-transcript` by agent before the claude resolver runs closes the cwd-fallback leak; making `agentId` a required prop is the correct fix for the wiring-drop failure. Both shells serve the channel through `registerTranscriptIpc`.

One accepted degrade worth stating in the description: an SSH-project grok node answers `found:false` in ⌘M (the locator reads the local hook-fed map), which the panel renders as missing rather than as an empty conversation — correct, just not remote-capable yet.

Merge-ready from my side, after #614; the merge call is @eneskirca's.

PR #618, opened 2026-09-02
feat(grok): a context meter on the denominator grok states, and its own tool vocabulary

xaviguardia:
> **Stacked on #614 → PR4** — review the top commit only; the rest land with their own PRs.

## What this does

Two capabilities that were declared impossible, both on numbers grok has been writing to disk all
along: a per-node **context meter**, and the phone's **"what it's doing now"** line.

## The meter, and the comment it replaces

`config.ts` explained grok's absence from `USAGE_CAPABLE` by blaming a missing parser for
`updates.jsonl`, concluding there was no used count to divide. **Both halves were false**, and the
first is the one worth remembering: the counters were never in that file. They are in
`~/.grok/sessions/<cwd>/<id>/signals.json`, so the blocker it declared was not the real blocker even
on the day it was written — a comment that explains an absence with a false cause sends the next
reader to build a parser nobody needed. It survived because nobody could check it: the integration
was written with no grok binary on the machine.

Measured across **22 real sessions** (grok 1.0.13, 2026-09-02):

| Key | Present | Meaning |
|---|---|---|
| `contextTokensUsed` | 22/22 | the numerator |
| `contextWindowTokens` | 22/22 | the denominator, **stated** — never inferred from the model id |
| `contextWindowUsage` | 22/22 | the percentage grok already computed |

The third one is a **free oracle** and is pinned as a test: grok is the only agent that states the
answer as well as both operands, so a changed rounding rule or key meaning fails a test instead of
leaving a meter quietly off by a percent.

`grokSignalsParse` returns `null` unless both numbers are usable — **no denominator, no meter**, and
no fallback window from anywhere. `signals.json` carries 66 keys (latencies, GCS queue counters,
lines touched, peak RSS); exactly three are read, and a test rewrites the other 63 to prove nothing
else reaches the meter.

## One new flag on the tail, and why it is not a tuning knob

`signals.json` is a whole JSON document **rewritten in place**, not a JSONL appended to. The existing
offset read would hand the parser the bytes past the previous read — a fragment, which never parses.
The meter would fill once and then **freeze, with nothing anywhere to say why**.

`wholeFile: true` re-reads from zero. Both shells create the tail identically (invariant 11), and a
test asserts the failure directly: without the flag the rewritten document's new number never
arrives.

## Consumers audited before joining `USAGE_CAPABLE`

This list once gated **three** features, and joining it for codex/gemini handed a node a stranger's
claude session as both its meter and its search index.

| Consumer | Verdict |
|---|---|
| `TerminalNode.tsx:1343` | the per-node meter — intended |
| `sessionList.ts:310` | `usesContext` on the session list — intended |
| `ContextMeter.tsx` | the popover, already agent-aware |
| `transcriptGates.ts` | reads `CLAUDE_TRANSCRIPT_READABLE`, **not** this list |

So grok joins the meter **without** joining claude's transcript readers, and
`config.capabilities.test.ts` pins that the two disagree for grok.

## The activity line, and a second false doc claim

`docs/grok-agent.md` §8.3 said grok's file hooks *"never send `PreToolUse`"*, so the
`recordRawToolEvent` call was a no-op and was deleted.

**Grok does publish the event.** It spells it `pre_tool_use` — its own snake_case, in both field
dialects — and `recordRawToolEvent` gates on the exact string `PreToolUse`. The blocker was a
**spelling**, not an absence, which is exactly why deleting the call looked correct and closed the
door on a working feature. The shells now translate at the boundary where `grokRawFields` already
decodes grok's dialect, rather than loosening a claude-shaped gate.

`toolActivity` gains grok's **fifteen** tool names, from `signals.json.toolsUsed` across the same 22
sessions — measured, not read off the docs. Their argument keys are a separate question and only two
were ever observed in a captured payload (`read_file.target_file`, `run_terminal_command.command`),
so those two lines carry a detail and the other thirteen name the action and stop. A phrase with no
detail is honest; one built on a guessed key renders wrong forever and nobody notices.

Both vocabularies share one function because claude's names are PascalCase and grok's are
snake_case — but only while the match stays **exact**. `grep`/`Grep` and `write`/`Write` differ by
case alone, and a test asserts all four separately, so adding case-folding fails loudly instead of
reading grok's argument keys out of a claude payload.

An MCP call arrives already resolved as `server__tool` (grok's `use_tool` dispatcher never appears in
a payload), so it is named from the string itself — no per-server vocabulary to keep in sync.

## Tests

New: `grok-signals.test.ts` (including the oracle and the 63-key isolation), the `wholeFile` block in
`context-tail.test.ts`, and grok's vocabulary plus the case-collision guard in
`agent-status-mirror.test.ts`.

Three mutations verified red, each with the diff printed before the result was read: dropping the
`null` when the window is missing, putting grok into the claude-transcript readers, and removing
`wholeFile`. **The third was green over nothing until this change added a test for it.**

## Surfaces

Desktop and Server Edition both create the tail and both translate the tool event. The phone reads
the mirror, which is agent-agnostic, so both features arrive there with no phone-side work — but the
activity line is *rendered* by `nodeterm-ios`, a private repo, so @eneskirca may want to confirm it
displays as expected there.

## Not verified

- No live grok node was watched filling a meter end-to-end on a canvas; the tail was exercised
  through its own tests and the parser against real files.
- The 22 sessions all ran on one machine and one account, so the 500k window is the only one
  observed. The reader takes whatever the file says, so a different window needs no code change —
  but it has not been seen.


eneskirca:
Reviewed the top two commits at e0a46fa (on #617) in an isolated worktree: typecheck clean, focal suites (`grok-signals`, `context-tail`, `agent-status-mirror`, `config.capabilities`, `server/agent-status`, both `no-electron`, `hook-verified-parity`) 249/249, CI approved.

On the question that decides whether a grok meter is allowed at all: the denominator is `contextWindowTokens`, **grok's own stated number** in `signals.json`, never inferred from the model id, and `grokSignalsParse` returns null without it. The fixture agrees with itself (381615 / 500000 → 76, matching `contextWindowUsage`). That satisfies rule 6 of "Adding a new agent". `wholeFile: true` on a rewritten-in-place document is right, and the tail's push is deduped on value change, so the 1 s whole-file re-read costs nothing visible. Both shells create the tail identically.

Two fixes before merge:

1. **`CLAUDE.md` is ahead of this PR's code.** The grok bullet now lists `SESSION_ID_CAPABLE` among grok's memberships, but at this commit `SESSION_ID_CAPABLE = ['claude', 'copilot']` — grok joins it in #619. If #618 lands alone the doc lies. Please move that word to #619's doc commit.

2. **Merge conflict with #592.** #592 moves the shells' grok branch into `core/grok-hook-session.ts` (`applyGrokHookSession`); this PR extends the inline branch in both shells (tail `track`/`untrack`, the `pre_tool_use` → `recordRawToolEvent` translation). Whichever lands second rebases, and the additions belong inside `applyGrokHookSession` — duplicating them in `src/main/index.ts` and `src/server/agent-status.ts` is the drift #592 exists to remove.

Not measurable here (no logged-in grok on this box): the 22-session claim and the fifteen `toolsUsed` names. The fixture's shape (64 keys, `primaryModelId: grok-4.6`) is consistent with a real capture, and only three keys are read, so the residual risk is the vocabulary going stale, which fails as "Using <tool>" rather than as a wrong line.

PR #619, opened 2026-09-02
feat(grok): mint a session id, feature-detected from grok's own --help

xaviguardia:
> **Stacked on #614 → PR4 → PR5** — review the top commit only; the rest land with their own PRs.

## What this does

A grok node learned its session id only from a hook. A node whose agent never ran a tool — or whose
SSH tunnel was down — had nothing to resume with. `SESSION_ID_CAPABLE`'s own comment records the
cost: measured on one host after a reboot, **18 of 40 agent nodes relaunched as blank conversations**
while their transcripts sat intact on disk.

Grok now mints its own id at node creation, gated on a probe of **grok's** CLI.

## The probe belongs to the agent it probes

`core/grok-cli.ts` reads `grok --help`. Not `claude --version`: the two CLIs are installed and
upgraded independently, so claude's answer is not even correlated with grok's, and borrowing it would
be a guess dressed as a measurement.

`supportsSessionIdFlag` takes grok's answer as a **required** third argument. A default would let a
caller silently omit it and get "grok never mints" with a green typecheck — the same shape as the
optional prop that let the chat panel's agent id be dropped in the previous PR. The lesson taken from
that one is that this family is fixed by a type that cannot express the broken state, not by another
test somebody has to remember to write.

**Feature-detection, never a version floor.** An unknown flag makes grok *exit*, so a floor guessed
too low kills every launch below it instead of degrading.

The word-boundary anchor is load-bearing twice:

- `--session-id-file` must not answer yes for `--session-id`.
- grok's **own** help mentions `` `--session-id` `` inside the description of `--fork-session`. The
  prose case is not hypothetical — grok writes it.

Both are pinned as tests.

## The flag goes before the `--`

grok's separator is **end of options**, and this is the failure that does not look like one: a flag
placed after it is not rejected, it is swallowed as part of the **prompt**. The node launches, looks
healthy, mints an id nodeterm never learns, and the agent reads `--session-id <uuid>` as the first
words of its instructions.

Nothing in the exit code says so, so the test asserts on **position in the assembled line**, not on
whether the launch worked.

## The id must not already exist

Measured on 1.0.13: `--session-id` *"must be a valid UUID and must not already exist under the target
session directory"*. Handing grok a taken id is a **launch error**, not a resume — grok exits and the
user sees a dead terminal.

`mintFreeGrokSessionId` (pure, in `shared/`) re-mints against the ids already on disk, and after
three collisions returns **undefined rather than a taken id** — degrading to the pre-minting command
line instead of to a node that will not start. `listGrokSessionIds` (core, node-side) reads directory
*names* only.

Two more grammar differences from claude, both measured and both respected by omission rather than by
code: `--session-id` combines with `--resume`/`--continue` only alongside `--fork-session`, and
`--resume` also accepts a **title**, failing as ambiguous on duplicates — so nothing here resumes grok
by title.

## A new bridge member, implemented on every surface

`window.nodeTerminal.grok` — preload (Electron), ws-bridge (Server Edition, backed by a **real**
server-side handler: `registerGrokCliIpc` runs in both shells), and the relay stub, which answers the
fail-open caps rather than rejecting because the launch path reads it synchronously.

The taken-id lookup crosses that boundary as an **array**: a `Set` arrives as `{}` and would report
"nothing taken" for every cwd, which is the one wrong answer that call can give.

## Tests and mutations

New: `grok-cli.test.ts`, `grok-session-mint.test.ts`, grok's block in `workspace.session-id.test.ts`,
the composed-line tests in `launch.test.ts`, and the both-shells registration test in
`server/handlers/index.test.ts`.

**Five mutations verified red**, each with the diff printed before the result was read:

| Mutation | Caught by |
|---|---|
| drop the probe's word-boundary anchor | `--session-id-file` and the prose case |
| move the flag after the `--` | the composed-line position assertion |
| return a taken id instead of re-minting | the mint chooser's tests |
| point the launch site at a constant instead of grok's probe | `workspace.session-id.test.ts` |
| delete the server shell's handler registration | `E_NO_HANDLER` in the server test |

The last two are the wiring family this branch keeps meeting — a call site and a registration that
the typecheck validates and no test exercises. **Both passed green before the tests in this change
existed.**

Two existing tests changed rather than deleted: `resume.test.ts` and `workspace.session-id.test.ts`
both used grok as their example of a *non*-capable agent. They now use agents whose leaf is genuinely
unwritten, plus a positive assertion for grok — an example that stops asserting anything the moment
its subject gains the capability is worse than no example.

## Not verified

- No grok node was launched from a canvas with a minted id and then cold-restored; the composed line
  and the minting were exercised through their own tests.
- The **first** grok node in a cwd mints without consulting the disk (the per-cwd set is warmed on
  demand, and node creation is a synchronous factory). Documented in §8; the residual risk is a v4
  UUID collision, not the reuse this guard exists for.


eneskirca:
Reviewed the top commit at 5e9665f (on #618) in an isolated worktree: typecheck clean, focal suites (`grok-cli`, `grok-session-mint`, `workspace.session-id`, `launch`, `resume`, `server/handlers/index`, `grok-session`, `config.capabilities`, both `no-electron`) 141/141, CI approved and green.

The feature-detect claim is now verified against the real binary, not the PR's fixture: I installed `@xai-official/grok@1.0.13` here and ran `grok --help`. It prints `-s, --session-id <SESSION_ID>` (wording matches what the PR quotes: valid UUID, must not already exist under the target session directory) and, as you say, also mentions `` `--session-id` `` in the `--fork-session` prose. `grokCliCapsFrom` answers `true` on the real output and `false` on the same output with the `-s` line removed while the prose mention remains — so the word-boundary anchor does what the tests claim.

The rest holds up: the flag goes before grok's `--` (asserted on the composed line), a taken id degrades to no flag rather than a dead terminal, the taken-id lookup crosses IPC as an array, and `registerGrokCliIpc` runs in both shells with the relay stub answering the fail-open caps.

One follow-up, not a blocker and shared with claude's minting: an SSH project consults the **local** grok probe and the local sessions tree. A host whose grok lacks `--session-id` would get the flag anyway and exit on launch. Same shape as the claude path today, so it belongs in a separate change (the remote probe the permission-mode gate already does over the ControlMaster), but worth a line in §8.

Merge-ready from my side, after #614 → #617 → #618 (and the #592 rebase noted there); the merge call is @eneskirca's.

eneskirca:
Merged `origin/main` into this branch (7feb3e8) to clear the conflict after #614/#617 landed. The only conflict was the import block in `src/renderer/state/workspace.ts` (main's `nodeSizing` imports vs this branch's grok session-id imports); both sides kept, no logic touched. Typecheck clean, the grok suites green (28 files / 427 tests).

Issue #583, opened 2026-09-01
Windows: GIT_ENV unconditionally prepends POSIX paths with ':' delimiter, corrupting PATH for git subprocesses


Abhirup0:
## What happens

In `src/core/git-service.ts:42-46`:

```ts
const GIT_ENV: NodeJS.ProcessEnv = {
  ...process.env,
  PATH: `/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin${process.env.PATH ? `:${process.env.PATH}` : ''}`,
  GIT_TERMINAL_PROMPT: '0'
}
```

On Windows (`process.platform === 'win32'`), `PATH` entries are separated by `;`, and Windows paths contain drive letter colons (`C:\...`).

Unconditionally prepending `/opt/homebrew/bin:/usr/local/bin:...:` with a `:` delimiter fuses the Unix paths and the first Windows path entry into a single malformed token:

```
/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:C:\Windows\System32
```

When Windows splits `PATH` on `;`, this entire first token is treated as a single invalid directory.

## Impact

Every local Git operation (`git status`, `git diff`, `git push`, `git fetch`, `git clone`, `git credential fill`) executed via `git-service.ts` uses `GIT_ENV`.

On Windows, this causes:
1. The first entry in the user's `PATH` to become unreachable for Git subprocesses.
2. Git child processes that rely on tools in `PATH` (such as `git-credential-manager.exe`, OpenSSH `ssh.exe`, or GPG signing binaries) to fail resolving their executables.

## Reproduction

Run the following in Node on Windows:

```js
const corruptedPath = `/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin${process.env.PATH ? `:${process.env.PATH}` : ''}`
console.log(corruptedPath.split(';')[0])
// Logs: "/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:C:\..." (invalid path)
```

## Suggested Fix

Guard the POSIX path prepending so it only applies on non-Windows platforms, or use `path.delimiter`:

```ts
const EXTRA_POSIX_PATHS = ['/opt/homebrew/bin', '/usr/local/bin', '/usr/bin', '/bin']

const GIT_ENV: NodeJS.ProcessEnv = {
  ...process.env,
  ...(process.platform !== 'win32'
[...]

PR #623, opened 2026-09-02
fix(github): resolve gh executable path in runGitHubCliCommand

Abhirup0:
Unifies `gh` executable resolution across `git-service.ts` and `github/credentials.ts` by extracting `ghPath()` into a standalone `src/core/gh-path.ts` module.

### Rationale

1. **Single unified resolver for both `gh` call sites**: `git-service.ts` resolved through `ghPath()` while `credentials.ts` invoked literal `'gh'`. Routing both through `ghPath()` prevents drift between git operations and credential checks.
2. **Coverage beyond `gitEnv()`'s fixed directories**: `ghPath()` delegates to `findExecutableSync`, which consults the cached login-shell `PATH` first. This allows resolving `gh` installed under `~/.local/bin`, `mise`, `asdf`, or `scoop`, which `gitEnv()`'s four hardcoded POSIX directories cannot reach.
3. **Clean architectural separation**: Extracting `ghPath()` into `src/core/gh-path.ts` avoids importing `git-service.ts` into `credentials.ts`, keeping the credential module decoupled from `GitService` and its IPC/spawn surface.

### Changes
- Extracted `ghPath()` and `_resetGhPathForTest()` into `src/core/gh-path.ts`.
- Replaced local `ghPath()` in `src/core/git-service.ts` with import from `./gh-path`.
- Updated `runGitHubCliCommand` in `src/core/github/credentials.ts` to execute `ghPath() ?? 'gh'`.
- Added unit tests in `src/core/gh-path.test.ts` and updated `src/core/github/credentials.test.ts`.

### DIFF
--- a/src/core/gh-path.test.ts
+++ b/src/core/gh-path.test.ts
@@ -1,0 +1,45 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { ghPath, _resetGhPathForTest } from './gh-path'
+import * as execPath from './exec-path'
+
+describe('ghPath', () => {
+  beforeEach(() => {
+    _resetGhPathForTest()
+    vi.restoreAllMocks()
+  })
+
+  it('memoizes on hit so repeated calls do not re-probe', () => {
+    const spy = vi.spyOn(execPath, 'findExecutableSync').mockReturnValue('/usr/local/bin/gh')
+
+    expect(ghPath()).toBe('/usr/local/bin/gh')
+    expect(ghPath()).toBe('/usr/local/bin/gh')
+    expect(spy).toHaveBeenCalledTimes(1)
+  })
+
+  it('does not memoize on miss so a newly installed gh is picked up', () => {
+    const spy = vi.spyOn(execPath, 'findExecutableSync')
+      .mockReturnValueOnce(null)
+      .mockReturnValueOnce('/opt/homebrew/bin/gh')
+
+    expect(ghPath()).toBeNull()
+    expect(spy).toHaveBeenCalledTimes(1)
+
+    expect(ghPath()).toBe('/opt/homebrew/bin/gh')
+    expect(spy).toHaveBeenCalledTimes(2)
+
+    // Now cached because it was a hit
+    expect(ghPath()).toBe('/opt/homebrew/bin/gh')
+    expect(spy).toHaveBeenCalledTimes(2)
+  })
+
+  it('probes with common bin dirs fallbacks', () => {
+    const spy = vi.spyOn(execPath, 'findExecutableSync').mockReturnValue('/usr/bin/gh')
+
+    expect(ghPath()).toBe('/usr/bin/gh')
+    expect(spy).toHaveBeenCalledWith('gh', [
+      '/opt/homebrew/bin/gh',
+      '/usr/local/bin/gh',
+      '/usr/bin/gh'
+    ])
+  })
+})
--- a/src/core/gh-path.ts
+++ b/src/core/gh-path.ts
@@ -1,0 +1,34 @@
+import { findExecutableSync } from './exec-path'
+
+const COMMON_GH_BIN_DIRS = [
+  '/opt/homebrew/bin/gh',
+  '/usr/local/bin/gh',
+  '/usr/bin/gh'
+]
+
+/**
+ * Resolves the path to the `gh` executable.
+ *
+ * Windows resolves a bare command against PATHEXT (`gh.EXE`, typically under
+ * `C:\Program Files\GitHub CLI\`).
+ * On macOS, GUI apps don't inherit the shell PATH, so the shared resolver checks
+ * the login-shell PATH first, followed by well-known POSIX locations.
+ *
+ * MEMOIZED-ON-HIT rather than computed at import: a miss is re-probed so a gh
+ * installed while the app is running is picked up, and the async login-shell PATH
+ * probe that lands after module load is no longer raced.
+ */
+let cachedGh: string | null | undefined
+
+export function ghPath(): string | null {
+  if (cachedGh) return cachedGh
+  const found = findExecutableSync('gh', COMMON_GH_BIN_DIRS)
+  // Only a HIT is cached. Caching a miss here would freeze the answer for the process lifetime.
+  if (found) cachedGh = found
+  return found
+}
+
+/** Reset the cached path (for testing). */
+export function _resetGhPathForTest(): void {
+  cachedGh = undefined
+}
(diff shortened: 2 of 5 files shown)


howieyoungdahl (commented):
This is now overtaken by `main`. Commit `e36c5e89` ("fix(core): stop building a POSIX PATH for git
subprocesses on Windows", closing #583 and #565) landed the same change: `src/core/git-env.ts` with
`gitEnvFrom(base, platform)` + `gitEnv()`, `git-service.ts` reduced to `const GIT_ENV = gitEnv()`,
and `github/credentials.ts` switched to `env: gitEnv()`. `git merge-tree` this PR against current
`main` conflicts in all four files it touches — `git-env.ts` and `git-env.test.ts` as add/add, plus
`git-service.ts` and `github/credentials.ts`.

Recording what I checked, since the diagnosis in this PR and in #583 is the same and it is correct:

- The POSIX branch is byte-identical to the old literal — `prefix` when `PATH` is empty, otherwise
  `${prefix}:${PATH}` — so nothing changes on macOS or Linux. Both this PR and the merged commit get
  that right.
- Omitting the PATH key on win32 rather than setting it is the right call for a reason the merged
  version writes down and this one does not: Windows env vars are case-insensitive and `process.env`
  there usually spells it `Path`, so an added `PATH` beside an inherited `Path` is a second entry for
  one variable. Same code, and the comment is the part worth keeping.
- Both versions add `GIT_TERMINAL_PROMPT: '0'` to `runGitHubCliCommand`, which did not have it
  before. That is a genuine POSIX behaviour change (`gh` shells out to `git`), and desirable, but it
  is not mentioned in either description — worth stating wherever this is finally recorded.
- The tests here are the right shape: `platform` is a parameter, so they run on the `windows-latest`
  job as well as everywhere else and the Windows branch is verified without a Windows machine.
  `main`'s version has the same property, plus a case for `tmux-hint.ts`'s `findCommand`, which is
  the same class of bug and which this PR does not cover.

**The one thing here that is NOT on `main`, and is still worth having.** `main`'s
`runGitHubCliCommand` still calls `execute(command, …)` with the literal `'gh'`; this PR exports
`ghPath()` and uses `ghPath() ?? 'gh'`. Now that `exec-path.ts` resolves candidates through PATHEXT,
`ghPath()` finds `gh.EXE` on Windows and the login-shell PATH on macOS, so a `gh` installed
somewhere other than the four hardcoded POSIX dirs becomes reachable from the credential path. That
is a real improvement and it survives the supersession.

If you re-open it as that change alone, one shape note: `github/credentials.ts` importing
`git-service.ts` points a small credential module at the module that owns `GitService`, the clone
single-flight and `remote-ssh/remote-git`. `ghPath()` is a five-line memoised wrapper around
`findExecutableSync`, which already lives in `exec-path.ts` — moving it there (or into its own
`gh-path.ts`) lets both callers import it without the layering inversion, and keeps `git-service`
free to grow import-time state.

Non-blocking: the added `runGitHubCliCommand` test asserts the pre-existing
`command !== 'gh'` guard rather than anything the PR changes, and there is a stray blank line at the
end of `credentials.test.ts`.

**What I could not verify:** no Windows machine, so the `;`-vs-`:` corruption and the `Path`/`PATH`
duplicate reasoning are read from the code and issue #583's own reproduction, not observed. I did
not run the suite. I did not check whether the author was already told this had landed — as of the
read, the PR had no comments and no reviews.

Abhirup0:
Thank you @howieyoungdahl for the thorough and constructive review!

I've rebased onto current `main` (dropping the superseded `git-env` additions that landed in #630) and restructured the remaining `ghPath` improvement according to your feedback:

1. **Standalone `gh-path.ts` module**: Moved `ghPath()` out of `git-service.ts` into `src/core/gh-path.ts` so both `git-service.ts` and `github/credentials.ts` import from it without any layering inversion.
2. **Path resolution in credentials**: `runGitHubCliCommand` now invokes `ghPath() ?? 'gh'` instead of the literal `'gh'`.
3. **Tests & cleanup**:
   - Added `src/core/gh-path.test.ts` testing hit memoization and miss re-probing.
   - Updated `src/core/github/credentials.test.ts` to assert that `runGitHubCliCommand` actually queries `ghPath()`.
   - Cleaned up the trailing blank lines.

eneskirca:
Reviewed on an isolated worktree at `2e166074` merged with current `main`: merges clean, `npm run typecheck` green, `gh-path` / `github/credentials` / `exec-path` suites green, and your own CI (including `quality-windows`) is green. `_resetGhPathForTest` matches house style (`_resetForTest`, `_resetStaticCacheForTest`). **Merge-ready.**

One correction to the description, because it changes what this PR is for and someone will read it later as the rationale.

**The ENOENT premise is stale — #630 already closed it.** `runGitHubCliCommand` passes `env: gitEnv()`, and Node resolves the *program name* against the **child** environment's `PATH`, not the parent's. Measured on this box:

```
execFile('ghtest', [], { env: parentEnv })                  -> ERR ENOENT
execFile('ghtest', [], { env: { ...parentEnv, PATH: fake + PATH } }) -> FOUND
```

`gitEnv()` prepends `/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin` on POSIX — the same three directories `ghPath()`'s fallback list holds — so the GUI-blind-PATH case on macOS is already covered on `main`, and on Windows libuv walks `PATH`+`PATHEXT` from the child env and finds `gh.exe` from the bare name anyway.

**What the change is still worth, and what I'd put in the body instead:**

1. **One resolver for both `gh` call sites.** `git-service.ts` resolves through `ghPath()` while `credentials.ts` passed the bare string — two answers for one question, which is exactly the drift CLAUDE.md warns about. After this they cannot disagree.
2. **Coverage beyond `gitEnv()`'s four fixed directories.** `ghPath()` goes through `findExecutableSync`, which consults the cached *login-shell* PATH first — so a `gh` under `~/.local/bin`, mise/asdf shims or scoop resolves, and those are the installs `gitEnv()` cannot reach.
3. **The layering fix you already describe.** Importing `git-service.ts` from `credentials.ts` would pull `GitService` and its spawn/IPC surface into the credential path; a standalone module is right.

That is a real improvement — just not the one the first paragraph claims.

**Ordering with #655.** #655 also rewrites this region of `git-service.ts` (it wraps every `gh` call in a `ghInvocation()` helper) and it needs a rebase for unrelated reasons. I composed both locally: this one first, then #655's `ghInvocation` stays in `git-service.ts` and only swaps its import to `./gh-path` — a three-line resolution, typecheck and tests green. So **this should land first.**

Abhirup0:
Thank you @eneskirca for the review and the detailed explanation regarding child environment resolution!

I have updated the PR description with the accurate rationale:
1. Unifying both `gh` call sites to avoid drift between `git-service` and `credentials`.
2. Coverage beyond `gitEnv()`'s fixed directories via `findExecutableSync`'s login-shell PATH lookup (`~/.local/bin`, `mise`, `asdf`, `scoop`).
3. Clean architectural separation without layering inversion.

PR #627, opened 2026-09-02
feat(grok): a session map that survives a restart, and a $GROK_HOME a GUI app can see

xaviguardia:
## What this carries

Five commits, on top of the grok work merged in #619. They travel together because
each one is small on its own and they share the same files (`config.ts`, the two raw
hook listeners, `docs/grok-agent.md`); splitting them would mean three PRs that
conflict with each other for no reviewer benefit.

| Commit | What |
|---|---|
| `b613c1da` | The `sessionId → dir` map survives a restart, and `$GROK_HOME` is probed from the login shell so a GUI-launched app sees it |
| `c9097cf9` | Per-launch model override — grok gets ITS catalogue, never the gateway's |
| `4b154273` | Subagent cards, keyed by `subagentId` |
| `60fea72d` | Two doc claims closed by measuring them instead of leaving them "unverified" |
| `d544ffb0` | A test pinning the whole composed launch line, where `--session-id` and `--model` meet |

Read them in that order; each section below covers one.

---

Two pieces of session continuity for grok, on top of the session-id work merged in #619.

Two more silent gaps, both about **where a grok session lives**.

## The map survives a restart (§8.5)

`sessionId → session directory` was in-memory: after an app restart a grok node's name went blank
until that session fired another hook. Correct, and silently unhelpful for a session sitting idle.

It is now written to `grok-session-dirs.json` under userDataDir — **atomically** (a torn file here is
a map that resolves nothing *and* cannot be repaired), debounced so a hook burst is one write, and
never awaited on the hook path.

**Persisting is not scanning, and that is the whole design.** A scan searches a tree for a session id
— that is what claude does, and it is the behaviour that made nodes adopt each other's names, which
is why grok derives instead. This file holds only what a hook already *told* us, so a restart
recovers facts we were given rather than guessing at facts we were not.

Every entry is re-validated with `isSafeGrokSessionId` on load: the file is ours, but it sits in a
directory the user can edit and its values become filesystem paths. A corrupt or hand-edited file
yields an **empty** map, never a partial one built from whatever happened to parse.

## The remote seam (§8.4)

`readAgentSessionName` now takes a `grokRemoteSummary` reader, shaped exactly like claude's
`setRemoteTranscriptReader`. Its answer is **final** for a session it owns: an unreadable host yields
*no name*, never the local directory's.

"Could not read the host" and "there is nothing" are different facts, and answering the first with
the second is how a node ends up wearing another session's name.

**Still owed, and stated rather than half-wired:** the desktop's implementation of that reader. It
needs the host's `$GROK_HOME`, which `RemoteHooks` computes at install time and discards.

## `$GROK_HOME` is probed (§8.9) — the most silent bug in the integration

A desktop app launched from Finder/Dock/`.desktop` never sourced the user's rc; the grok CLI, started
by the shell inside a tmux pane, did. For a user whose only `export GROK_HOME=…` lives in `.zshrc`,
nodeterm wrote the hook file under `~/.grok` and grok read somewhere else: **no badge, no unread dot,
no notification, no session name, ever — and no diagnostic anywhere.**

`ensureGrokHomeProbed` asks the login+interactive shell once (the same spawn `resolveShellPath`
already makes for `$PATH`), and `installManagedAgentHooks` re-writes grok's hook file if the answer
moves the target.

| Property | Why it is not incidental |
|---|---|
| an explicit `process.env.GROK_HOME` wins, and the shell is **not even asked** | that spawn costs hundreds of ms and can hang on a slow dotfile |
| async and fail-open | boot never waits; a hung shell means today's behaviour |
| the fallback **records itself** (`grokHomeFallbackWasSilent`) | the bug was never the wrong path — it was that a wrong path said nothing |

Checklist item 31 asked whether anyone sets the variable at all *before* this got built. Answered on
this device: unset in the environment **and** absent from every shell rc. Which is precisely why the
probe had to cost nothing when it finds nothing.

`summary.json` carries `grok_home` inside it (29 of 29 local sessions, all agreeing) — a better
source for reads, but it cannot bootstrap the *install*, since reading it means already knowing the
sessions directory. Recorded for a future reader rather than used.

## Mutations

Five more verified red, each with the diff printed before the result was read: removing the debounced
write, letting a remote failure fall back to the local directory, accepting any id from the persisted
file, deleting the probe's call site in the installer, and letting the probe override an explicit
environment value.

**Two of the five passed green on the first attempt** — the wiring family again: a debounced write
every test called explicitly, and a probe call site no test exercised. Both are now pinned.

## Not verified (this half)

- The remote reader has a seam and tests, but no SSH implementation, so a remote grok node still
  degrades to no name. Item 26 is likewise untouched: that probe still runs over a plain ssh exec
  channel.
- The login-shell probe was exercised with an injected answer and through a mocked `resolveShellEnvVar`,
  not against a real GUI launch on a machine that exports `GROK_HOME` — this one does not export it.


eneskirca:
Reviewed at `d544ffb0` in an isolated worktree (main checkout untouched). `npm run typecheck` clean; `src/shared/agents`, `src/core/agents`, `src/core/grok*`, `src/renderer/nodes`, `src/renderer/state`, plus the `fs-atomic` / `no-electron` / `hook-verified-parity` guards — 1886 tests green. grok **1.0.13 is installed on this machine** (`~/.grok/bin/grok`, just not on PATH), so several of your measurements reproduce here; I have said which ones do not.

## Four of the five commits are ready

- **`$GROK_HOME` probe** — same shape as `resolveShellPath` it says it mirrors: 5 s timeout, cached per name, `win32` and a non-identifier name short-circuit, explicit `process.env` wins and the shell is not asked, async and fail-open. And the wiring test is not decorative: deleting `installGrokHooks()` from the `.then()` in `installManagedAgentHooks` turns `hooks/index.test.ts` red, which I ran.
- **Persisted map** — `writeFileAtomic` + mode 0600, debounced, `unref`'d, every entry revalidated with `isSafeGrokSessionId`, corrupt file ⇒ EMPTY map. Correct against the fs-atomic guard.
- **Model override** — `-m, --model <MODEL>` confirmed in `grok --help` here, and the composed line puts it before `--`. `grokModelsFrom` parses this machine's real `grok models` output exactly as your fixture claims; the only difference is the auth banner line, which the parser correctly ignores. `modelGatewayEnv` returning `{}` for grok, with the reason stated, is the right call.
- **Doc claims closed by measuring** — fine, except where they run ahead of the code (below).

## Blocking: the subagent commit (`4b154273`) cannot produce a card

Two independent legs are missing. I measured both rather than reading them.

**1. The hooks are never subscribed.** `GROK_HOOK_EVENTS` (`src/shared/agents/hook-events.ts`) is untouched by this PR and carries no `SubagentStart` / `SubagentStop`. That list *is* what `installGrokHooks` writes, so those events never fire. The file this repo installed on my machine says so:

```
$ python3 -c "import json;print(list(json.load(open('/root/.grok/hooks/nodeterm-status.json'))['hooks']))"
['SessionStart','UserPromptSubmit','Stop','StopFailure','Notification','SessionEnd',
 'PreToolUse','PostToolUse','PostToolUseFailure']
```

The branch you added to both raw listeners is therefore unreachable.

**2. `normalizeGrok` still returns `null` for both.** Its closing line is still *"Not subscribed in v1: PermissionDenied, SubagentStart/Stop, PreCompact/PostCompact."* — and the card is created only at `Canvas.tsx:11068 case 'subagent-start'`, off the normalized event. So even with the hooks subscribed, nothing renders:

```ts
normalizeGrok({ nodeId:'n1', agentId:'grok',
  payload:{ hookEventName:'subagent_start', subagentId:'child-1', subagentType:'explore', cwd:'/x' }})
// → null      (expected kind 'subagent-start')
```

Joining `SUBAGENT_CAPABLE` is not inert either: `fanoutCapable` (`TerminalNode.tsx:1511`) now puts the hide-fan-out control on every grok node, for fan-out that cannot exist.

And `docs/grok-agent.md` states the feature as shipped twice — the capability row (*"grok fires native SubagentStart/SubagentStop"*) and the surfaces table (*"Subagent cards | yes — one card per `subagentId` | yes — the same branch in `src/server/agent-status.ts`"*). That is exactly the drift CLAUDE.md's canvas-control rule warns about, one layer up: the doc describes a product that does not exist yet.

The three new `config.capabilities.test.ts` cases are the tell. *"shows one card per subagent INSTANCE, not one per type"* asserts only `canSubagent('grok') === true`. A test named for a behaviour whose body checks list membership cannot fail when the behaviour is absent.

**Either** add the two events to `GROK_HOOK_EVENTS` and give `normalizeGrok` its `subagent-start` / `subagent-end` branches (the raw-listener half and the doc rows are then correct as written), **or** drop `4b154273` and revert those doc rows to "not joined". The other four commits stand on their own and I would take them today.

## Two small notes, non-blocking

- `grokHomeFallbackWasSilent()` is exported with *"Exposed so a diagnostic can say so instead of leaving the user with a dead node and no explanation"* — but nothing in the codebase reads it. Either wire the diagnostic or say the reader is owed; as written the sentence promises something the user never sees, which is the failure mode that function exists to close.
- In `exec-path.ts` the new function landed between `shellPathNow`'s docstring and `shellPathNow`, so *"The cached login-shell PATH…"* now documents `cachedShellVars`.

## Not verified on this machine

Everything needing an authenticated grok session: that `subagentId` differs between two same-type children, that the start's `sessionId` is the parent's, and that `/rename` rewrites `generated_title`. `subagentId` / `subagent_id` do appear as strings in the 1.0.13 binary here, and `SessionEnd` carrying `subagentType` for a child session is documented at `~/.grok/docs/user-guide/10-hooks.md:104` — so your child-end guard is corroborated. The rest rests on your capture.

xaviguardia:
Both legs are now in, at `b760e352` — the blocking point was real and the diagnosis was exact.

**What was wrong.** You were right that nothing could render. The capability was joined, the raw-listener branch existed, the doc rows described the feature, and the two things that make it *happen* were missing. The three `config.capabilities.test.ts` cases are the honest tell you named: a test called "one card per INSTANCE" that asserts `canSubagent("grok") === true` cannot fail when the mapping is absent.

**Leg 1 — subscription.** `GROK_HOOK_EVENTS` now carries `SubagentStart` / `SubagentStop`, with **no matcher** (a subagent is not a tool call, so grok's tool matcher would never apply). The installer tests pin the pair in the file that actually gets written — and the SSH installer writes the same list, so a grok node on a remote host gets cards too, which `remote-hooks.test.ts` now asserts.

**Leg 2 — the mapping.** `normalizeGrok` maps both to `subagent-start` / `subagent-end`, keyed on `subagentId`, carrying `description` as the task and `lastAssistantMessage` as the result.

**On what you could not verify here** — the capture answers all three, and it is in the branch at `tasks/grok-full-support/evidence/grok-subagent-payloads.jsonl` (1.0.13, two parallel `explore` children):

```
subagent_start  sessionId=parent-1  subagentId=child-a  subagentType=explore  description=...
subagent_start  sessionId=parent-1  subagentId=child-b  subagentType=explore  description=...
subagent_stop   sessionId=child-a   subagentId=child-a  subagentType=explore  lastAssistantMessage=alfa
subagent_stop   sessionId=child-b   subagentId=child-b  subagentType=explore  lastAssistantMessage=beta
```

Two children of the same type, different `subagentId` — hence per instance. The start's `sessionId` is the parent's; **the stop's is the CHILD's own**, equal to its `subagentId`. That asymmetry is why neither branch emits a `sessionId` at all: passing the stop's through would move the node's session onto the child and take the context meter with it.

**Tests.** They assert the mapped event, not membership. Mutation-checked rather than asserted: disabling the branch turns 5 of them red, and the file was restored byte-identical (same md5).

**The two nits, both taken.** `grokHomeFallbackWasSilent` now has its reader — `installManagedAgentHooks` warns with the consequence and the fix, since a flag nobody reads leaves precisely the unexplained dead node it was written to prevent. And `shellPathNow`'s docstring is back on `shellPathNow`.

**Verification.** Typecheck clean on both configs; 119/119 across the grok suites plus the `no-electron` / `fs-atomic.guard` / `hook-verified-parity` guards. On the full suite, 8 files stay red — all environment (real-tmux, session-host, node-pty patch), none touched by this branch, and the one that looked new to me (`managed-script`, a cygpath case) fails identically on a clean `upstream/main`, so it is yours, not ours.

The doc line saying these events were "not subscribed" is corrected in the same commit rather than left to a follow-up — that sentence was the drift you flagged, one layer down.

eneskirca:
Heads-up on ordering: this now needs a rebase, and the reason is that the rest of the grok stack landed while it sat here.

`main` has since taken #592, #618 and #619 (plus #537, which is where the new file below comes from). Merging `origin/main` into this head conflicts in seven files — `src/main/index.ts`, `src/server/agent-status.ts`, `src/shared/agents/normalize.ts`, `src/shared/agents/hook-events.ts`, `src/core/agents/hooks/grok.test.ts`, `src/main/remote-ssh/remote-hooks.test.ts` and `docs/grok-agent.md` — which is exactly the overlap the earlier review flagged: the natural home for #618's tail track/untrack and `recordRawToolEvent` calls is *inside* `applyGrokHookSession`, not beside it in both shells.

CI is red for a smaller, mechanical reason worth fixing in the same pass: `src/server/headless-node-factory.test.ts` (new on main) stubs `grokCaps: async () => ({ sessionIdFlag: false })`, which no longer satisfies the `GrokCliCaps` this branch widens — `error TS2322` in both `quality` and `quality-windows`.

Not asking for anything else; the review itself was positive and grok 1.0.13 is installed here, so the measured claims reproduce. Ping when it is rebased and we will re-verify and land it.

xaviguardia:
Rebased onto current `main` (`17e728f1`) — `36286a0e`, MERGEABLE, and the CI stub is fixed. Ready for re-verification.

**The `GrokCliCaps` break, as you diagnosed it.** `headless-node-factory.test.ts` stubs `grokCaps` and this branch widens the interface with `models`. Stated as `models: []`, which is grok's own "no catalogue" answer (a failed or absent `grok models`) and therefore the pre-feature behaviour: no model switching offered, never a partial list. Typecheck clean on both configs.

**The seven-file overlap resolved as you predicted**, and putting the two branches in one tree turned up something worth flagging, because it changes what this PR is for:

**#592 already subscribes `SubagentStart`/`SubagentStop` and already normalizes them — but that mapping cannot produce a card.** It emits no `toolUseId`, which is the field every agent's card store keys on (claude by `tool_use_id`, codex by `agent_id`), and it spreads `...base`, so `sessionId` rides along. On the **stop** that id is the CHILD's own, so start and end file under different keys and the started card never closes — a badge lit forever, no error. Replaced with the measured mapping, which emits no `sessionId` at all.

A test pinned the shape that made it unreachable: *"normalizes SubagentStart and SubagentStop without inventing an instance id"* builds payloads with **no** `subagentId` and asserts the result carries `sessionId`. Neither matches the wire — all four captured 1.0.13 payloads carry `subagentId`, so the id-less shape does not occur, and `sessionId` is the trap. Rewritten to assert the id-less input maps to **nothing** rather than to a card no store can key. Same failure mode you named in the first review, one layer along: a test whose name describes a behaviour while its body checks something that survives the behaviour's absence.

The capture is in the branch at `tasks/grok-full-support/evidence/grok-subagent-payloads.jsonl`:

```
subagent_start  sessionId=<parent>  subagentId=<child-a>  subagentType=explore  description=...
subagent_start  sessionId=<parent>  subagentId=<child-b>  subagentType=explore  description=...
subagent_stop   sessionId=<child-a> subagentId=<child-a>  lastAssistantMessage=alfa
subagent_stop   sessionId=<child-b> subagentId=<child-b>  lastAssistantMessage=beta
```

Two children of the same type, different ids — one card per instance. Mutation-checked rather than asserted: disabling the branch reds 5 tests, and the file was restored byte-identical.

**Conflict resolutions worth naming**, since a rebase is where a PR quietly changes meaning:
- Both raw listeners: your `if (plan.forgetSessionId) grokContextTail.untrack(...)` kept, my obsolete `forgetGrokSession` + untrack dropped (`applyGrokHookSession` owns it now). The child-teardown guard stays **before** it, so a subagent ending cannot untrack the PARENT's tail.
- `hook-events.ts`: took yours — the list lives in `grok.ts` as `GROK_EVENTS` now, and it already carries the pair.
- Doc: took your `Notification` write-up and your `SubagentEnd` alias note, adding only what the live capture contributes. Dropped my "Not subscribed: PermissionDenied, PreCompact, PostCompact" line, which #592 made false.

**File set: 31 → 29.** Three files left the diff because #592 already makes those changes (`hook-events.ts`, `grok.test.ts`, `remote-hooks.test.ts`); one entered, the CI stub.

**Verification:** 216/216 across the ten grok / parity / guard suites. Full suite: 10 red files of 783, all environment — five real-tmux/session-host/node-pty, two remote-atomic, and the three that fail on any Mac over "this Mac" vs "this computer" and pass on your Linux CI. None touched by this branch.

xaviguardia:
@eneskirca Ping as requested. Head `cb43ac61`, rebased onto current `main` (includes #666/#668/#669), MERGEABLE, CI green; typecheck clean, 335/335 across the grok, parity and guard suites.

Your three asks: `SubagentStart`/`SubagentStop` subscribed (`GROK_EVENTS` in `grok.ts`), mapped in `normalizeGrok` keyed on `subagentId`, `grokCaps` stub fixed. Nothing open from our side.

eneskirca:
Re-verified `cb43ac61` in an isolated worktree, merged onto current `origin/main` — **the rebase landed clean** (29 files, no conflicts) and the `GrokCliCaps` stub is fixed (`grokCaps: async () => ({ sessionIdFlag: false, models: [] })`). `npm run typecheck` clean on both projects; 753 test files / 10 650 tests green across `src/main` `src/renderer` `src/server` `src/core` `src/shared`. Two files reddened once under parallel contention — `agents/hooks/managed-script.test.ts`'s failover case and `LicenseSection.test.tsx` — and both pass alone and on a full rerun; neither is touched by this branch.

grok 1.0.13 is installed on this machine, so I measured rather than trusted:

- **`-m, --model <MODEL>`** and the `models` subcommand are both in `grok --help` on this build.
- **`grok models` prints exactly your fixture's shape here.** The only difference is the banner line ("You are not authenticated." instead of "You are logged in with grok.com."), and `grokModelsFrom` is unaffected because it reads only the indented bullets after `Available models:` — the reason for not treating `Default model:` as data holds for the banner too.
- **`subagentId` is a real hook payload field on 1.0.13**, not an inference: the binary's interned hook-payload field table carries `… toolUseId toolInput toolInputTruncated toolResult durationMs isBackgrounded level subagentId phase …` beside `subagentType`, and the task tool's own help says a backgrounded task *"returns immediately with a subagent_id"* — i.e. per instance, which is what the card keys on.
- **Grok's shipped docs on this machine corroborate the stop's session id.** `~/.grok/docs/user-guide/10-hooks.md:101` — `SubagentStop` *"fires once, **in the subagent**, with stop decision control"* — which is exactly why neither branch may pass a `sessionId` through.
- The three earlier nits are closed: `grokHomeFallbackWasSilent` has its reader (`hooks/index.ts` warns with the consequence and the fix), the `shellPathNow` docstring is back above `shellPathNow`, and the session map is `writeFileAtomic` + mode 0600 with `isSafeGrokSessionId` re-validated per entry on read.

**Two things before merge.**

**1. The capture is not in the branch.** The PR body and `normalize.ts`'s own comment both point at `tasks/grok-full-support/evidence/grok-subagent-payloads.jsonl`. That path does not exist at `cb43ac61`, and `git log --all -- <path>` has never seen the file — so the subagent mapping is pinned only by payloads we wrote by hand. That is precisely what `__fixtures__/grok/hook-payloads.json`'s provenance note exists to prevent: *"every earlier grok test asserted against payloads WE wrote from the shipped docs, so they pinned our reading of the documentation rather than grok's behaviour."* The machinery is already on main — please add the four captured payloads to that fixture and pin start/stop in `normalize.grok.capture.test.ts` (a same-type pair with different `subagentId`s is the case worth having there). If the capture is genuinely lost, say so and drop both references instead: a comment naming a file nobody can open is the exact defect the previous round found one layer along.

**2. CLAUDE.md line 1250 now contradicts the branch.** The grok bullet still reads *"NOT in `SUBAGENT_CAPABLE` — subagent cards still need the `spawn_subagent` PreToolUse/PostToolUse payload, which nobody has captured."* That sentence landed on `main` after your branch was written, and this PR falsifies both halves (grok joins the list, and through the native `SubagentStart`/`SubagentStop` hooks rather than the spawn tool call). Same-PR doc sync, as with `docs/grok-agent.md`.

Everything else is verified and ready; neither item is a change to the mechanism.

eneskirca:
Re-verified `cb43ac61` merged with current `main` in an isolated worktree: the rebase is clean (29 files), the `GrokCliCaps` stub in `src/server/headless-node-factory.test.ts` is fixed, typecheck clean, 753 files / 10,650 tests green. grok 1.0.13 is installed on this box, so the claims were measured rather than read: `-m, --model` and `models` exist; `grok models` output matches your fixture (only the auth banner differs, which the parser already ignores); `subagentId` IS a 1.0.13 hook payload field (it sits in the binary's interned field table beside `subagentType`/`toolUseId`/`description`, and the task tool's own help says it "returns immediately with a subagent_id"); and grok's `10-hooks.md:101` confirms `SubagentStop` fires *inside* the subagent, which is what your claim about the stop's sessionId rests on. The three earlier nits are closed.

Two things left, both small and neither about the mechanism:

1. **The evidence file does not exist.** The PR body and a comment in `normalize.ts` both cite `tasks/grok-full-support/evidence/grok-subagent-payloads.jsonl`, and it is not in the branch and has never been in the history. So the subagent mapping is pinned only by hand-written payloads — precisely what the provenance note on `__fixtures__/grok/hook-payloads.json` exists to prevent. Adding the four payloads to that fixture and pinning them in `normalize.grok.capture.test.ts` is enough; if they were composed rather than captured, say so there in the same words the existing note uses.
2. **CLAUDE.md line 1250 still says grok is NOT in `SUBAGENT_CAPABLE`** and that nobody has captured the payloads. This PR falsifies both halves, so it needs updating in the same change — same rule the repo applies to the agent-facing skill text.

Everything else reads merge-ready to me; ping when those two land and I will take it.

xaviguardia:
@eneskirca Both remaining items landed at `52942f29`.

**1. The capture is in the fixture now.** The jsonl lived under `tasks/` (gitignored), which is why `git log --all` never saw it. The four live payloads from that 2026-09-02 run — two parallel `explore` children, keys and shapes unchanged, ids/paths redacted like the rest of the file — are in `__fixtures__/grok/hook-payloads.json`. `normalize.grok.capture.test.ts` pins start/stop against those bytes: same type, different `subagentId`s; start.`sessionId` is the parent; stop.`sessionId` equals the child's `subagentId`. Comments that named the missing path now point at the fixture.

**2. CLAUDE.md grok bullet** now lists `SUBAGENT_CAPABLE` and names the native `SubagentStart`/`SubagentStop` path (keyed on `subagentId`, not the spawn tool call). The subagent-visualization section got a grok paragraph next to the existing Codex one.

Nothing in the mechanism changed. Ping when you take it.

PR #628, opened 2026-09-02
feat(browser): say why a page did not load, and show that one is loading

basmilius:
A `<webview>` node that failed to load told you nothing useful, and the two surfaces that host one
disagreed about what "nothing useful" looked like.

Chromium paints its own error page inside the guest for a failed main-frame navigation. So a web
node pointed at a dev server that is not running sat on another product's chrome, in the middle of
the canvas, with no way back. The browser surface did slightly better: a red strip along the bottom
carrying the raw `ERR_CONNECTION_REFUSED`, over that same page. The web node had no `did-fail-load`
listener at all.

Neither surface showed that a navigation was in flight either, beyond the browser toolbar's reload
button turning into a stop.

## What it does

Both surfaces now cover the guest with one plate (`WebviewErrorPlate`) that names the failure in a
sentence, shows the URL it was aiming at, adds a line about what to check where there is one, keeps
Chromium's own symbol, and offers Try again. The symbol stays because it is the one part of the
plate that is searchable and quotable in a bug report.

The wording lives in one module, `webviewError.ts`:

| Code | Says | Adds |
| --- | --- | --- |
| `ERR_CONNECTION_REFUSED` | Nothing is listening at that address. | If this is a dev server, it may not be running yet. |
| `ERR_NAME_NOT_RESOLVED` | That host name could not be resolved. | Check the spelling, and whether this machine is online. |
| `ERR_CERT_AUTHORITY_INVALID` | The certificate is not from a trusted authority. | Usual for a self-signed certificate on a local server. |
| anything not in the table | The page could not be loaded. | nothing |

Two rules that table exists to enforce. An unknown code degrades to the generic sentence plus the
symbol rather than to an invented explanation, because a confident wrong cause is worse than none.
And Try again is withheld only where the address itself is the problem (`ERR_INVALID_URL`,
`ERR_UNKNOWN_URL_SCHEME`), since re-running that exact navigation cannot succeed; a certificate or
a refused connection keeps the button, because the user is the one who knows whether they just
fixed the server.

A subframe failure is ignored: covering a working page because an ad did not load hides the content
the reader came for. `ERR_ABORTED` is ignored too, which is what a navigation the app replaced, and
a download, both look like. Both live in `isReportableFailure`, so the rule is named rather than
inlined as `!== -3`.

Loading is an indeterminate bar across the top of the view. Indeterminate on purpose: a guest
reports `did-start-loading` and `did-stop-loading` and nothing between, so any percentage would be
invented. It honors `prefers-reduced-motion`.

## The trap

**The plate clears on `did-start-loading` and nowhere else.** Chromium navigates to its own error
page under the URL that just failed, so clearing from `did-navigate` wipes the plate the instant it
appears. A new load starting is the one signal that means "we are trying again", and it is
order-independent, which no comparison against the failed URL would be. `webview-error-parity.test.ts`
pins it at source level for both surfaces, because the tempting simplification is exactly the one
that breaks it.

## One more split

The browser surface's `failed` state also held "Enter a URL or search term". That is a complaint
about what was typed in the address bar, not about a page, and it was rendered in the same red
strip. It keeps the strip as `notice`, so it can never cover a page that is perfectly readable,
while a real load failure gets the plate.

## On `webviewReload`

Try again is a reload, so the helper lands here with its first consumer rather than as its own
change. It carries two rules worth having in one place: Shift bypasses the cache and never silently
falls back to a cached re-fetch (the gesture means "I do not trust what is on screen", and a
downgrade is indistinguishable from a page that genuinely did not change), and both methods throw
before the guest attaches, which is correct to swallow because a guest that does not exist has
nothing to reload.

The second commit then makes it the *only* reload path. Two others in the same file still reached
past it, the toolbar button and pressing Enter on the URL already loaded, so the rules applied to
the plate's button and not to the one directly beside it. Routing them through it also gives the
toolbar the Shift bypass it never had, and `webview-reload-parity.test.ts` pins both surfaces at
source level: a direct `.reload()` works in the happy path while losing both rules, which is exactly
the simplification a later refactor reaches for.

The web node had no reload at all, on the one node kind whose whole content is a page someone else
is serving. A page that failed could only be retried from the plate, and a stale one not at all. It
gets the same button, in the glyph idiom of the two beside it, shown exactly when there is a guest.

## Surfaces

- **Desktop:** the feature. The kanban card modal gets it for free, since it hosts the same
  `BrowserSurface`.
- **Server Edition:** N/A by construction. There is no `<webview>` in a browser, so there is
  nothing to degrade.
- **Mobile companion:** N/A. No canvas, no browser nodes.

## Testing

28 unit tests plus `npm run typecheck`, all renderer-side and pure.

- `webviewError.test.ts` (11): the named codes, the unknown-code degrade, the missing-description
  fallback, the two non-retryable codes, subframe and aborted filtering.
- `webview-error-parity.test.ts` (8, four rules over both surfaces): both describe through the
  shared module, both filter through `isReportableFailure` and no longer inline `!== -3`, both
  render the shared bar, and both clear the plate from `onStart` and never from `onNav`.
- `webviewReload.test.ts` (5): the two reload modes, no fallback on a forced reload, no guest, and
  the not-yet-attached throw.
- `webview-reload-parity.test.ts` (4, two rules over both surfaces): every reload goes through the
  helper carrying the Shift modifier, and neither surface calls the guest's methods directly.

image

image

### DIFF
--- a/src/renderer/nodes/BrowserSurface.tsx
+++ b/src/renderer/nodes/BrowserSurface.tsx
@@ -259,7 +259,7 @@ export function BrowserSurface({
     // A navigation with an initiator: whatever it navigates to is not a restore echo.
     restoringNavRef.current = null
     locationRef.current = safe
-    if (safe === src) ref.current?.reload()
+    if (safe === src) reloadWebview(ref.current, false)
     else setSrc(safe)
   }
 
@@ -274,8 +274,8 @@ export function BrowserSurface({
         </button>
         <button
           className="browser-node__btn"
-          onClick={() => (loading ? ref.current?.stop() : ref.current?.reload())}
-          title={loading ? 'Stop' : 'Reload'}
+          onClick={(e) => (loading ? ref.current?.stop() : reloadWebview(ref.current, e.shiftKey))}
+          title={loading ? 'Stop' : 'Reload (Shift to bypass the cache)'}
         >
           {loading ? '✕' : '⟳'}
         </button>
--- a/src/renderer/nodes/BrowserSurface.tsx
+++ b/src/renderer/nodes/BrowserSurface.tsx
@@ -4,6 +4,10 @@ import { BrowserStartPage } from './BrowserStartPage'
 import { useBrowserHistory } from '../state/browserHistory'
 import { useDiscardWhenHidden, webviewAudible } from './useDiscardWhenHidden'
 import { DiscardedPlate } from './DiscardedPlate'
+import { reloadWebview, type ReloadableWebview } from './webviewReload'
+import { describeLoadFailure, isReportableFailure, type WebviewLoadFailure } from './webviewError'
+import { WebviewErrorPlate } from './WebviewErrorPlate'
+import { WebviewLoadingBar } from './WebviewLoadingBar'
 
 // Minimal typing for the Electron <webview> element methods/events we use.
 type WebviewEl = HTMLElement & {
@@ -70,7 +74,11 @@ export function BrowserSurface({
   const [loading, setLoading] = useState(false)
   const [canBack, setCanBack] = useState(false)
   const [canFwd, setCanFwd] = useState(false)
-  const [failed, setFailed] = useState('')
+  /** A complaint about what was TYPED in the address bar, not about a page. Rendered as a strip
+   *  under the toolbar, so it never covers a page that is still perfectly readable. */
+  const [notice, setNotice] = useState('')
+  /** A main-frame load failure, which DOES cover the guest — see {@link WebviewErrorPlate}. */
+  const [failure, setFailure] = useState<WebviewLoadFailure | null>(null)
   // Memory saver (see `useDiscardWhenHidden`): the page is released while hidden and rebuilt on
   // reveal. `loadingRef` mirrors the `loading` state because the hook reads it at fire time, from
   // a callback that must not force the observer to be re-created.
@@ -92,6 +100,10 @@ export function BrowserSurface({
     const onStart = (): void => {
       loadingRef.current = true
       setLoading(true)
+      // The one unambiguous "we are trying again" signal, and the ONLY place the plate clears.
+      // `did-navigate` is not usable for it: Chromium navigates to its own error page under the
+      // very URL that just failed, so clearing there would wipe the plate the instant it appears.
+      setFailure(null)
     }
     const onStop = (): void => {
       loadingRef.current = false
@@ -115,7 +127,7 @@ export function BrowserSurface({
       restoringNavRef.current = null
       setAddress(u)
       locationRef.current = u
-      setFailed('')
+      setNotice('')
       if (echo) return
       // A genuine navigation re-opens the title gate below: the first title of the NEW page always
       // records, however it compares to the old page's.
@@ -140,11 +152,16 @@ export function BrowserSurface({
       if (lastUrlRef.current) useBrowserHistory.getState().record(lastUrlRef.current, title)
     }
     const onFail = (e: Event): void => {
-      const ev = e as unknown as { isMainFrame: boolean; errorCode: number; errorDescription: string }
-      if (ev.isMainFrame && ev.errorCode !== -3) {
+      const ev = e as unknown as {
+        isMainFrame: boolean
+        errorCode: number
+        errorDescription: string
+        validatedURL?: string
+      }
+      if (isReportableFailure(ev.isMainFrame, ev.errorCode)) {
         // A restore that never landed has no echo to swallow — disarm, or the next navigation pays.
         restoringNavRef.current = null
-        setFailed(ev.errorDescription || 'Failed to load')
+        setFailure(describeLoadFailure(ev.errorCode, ev.errorDescription, ev.validatedURL || locationRef.current))
       }
     }
     wv.addEventListener('did-start-loading', onStart)
@@ -196,9 +213,10 @@ export function BrowserSurface({
     onDiscard: () => {
       setDiscarded(true)
       setSrc('')
-      // A failure banner belongs to the page we just released; the restore re-navigates and will
-      // raise its own if the load fails again.
-      setFailed('')
+      // A failure belongs to the page we just released; the restore re-navigates and will raise
+      // its own if the load fails again.
+      setFailure(null)
+      setNotice('')
       onGuestDiscarded?.()
     },
     onRestore: () => {
@@ -233,11 +251,11 @@ export function BrowserSurface({
   const go = (): void => {
     const safe = searchOrUrl(address)
     if (!safe) {
-      setFailed('Enter a URL or search term')
+      setNotice('Enter a URL or search term')
       return
     }
     setAddress(safe)
-    setFailed('')
+    setNotice('')
     // A navigation with an initiator: whatever it navigates to is not a restore echo.
     restoringNavRef.current = null
     locationRef.current = safe
@@ -293,12 +311,16 @@ export function BrowserSurface({
               setSrc(u)
               setAddress(u)
               locationRef.current = u
-              setFailed('')
+              setNotice('')
             }}
           />
         )}
         {discarded && <DiscardedPlate />}
-        {failed && <div className="browser-node__error">{failed}</div>}
+        {failure && !discarded && (
+          <WebviewErrorPlate failure={failure} onRetry={() => reloadWebview(ref.current, false)} />
+        )}
+        {loading && <WebviewLoadingBar />}
+        {notice && <div className="browser-node__error">{notice}</div>}
       </div>
     </div>
   )
--- a/src/renderer/nodes/BrowserSurface.tsx
+++ b/src/renderer/nodes/BrowserSurface.tsx
@@ -217,6 +217,9 @@ export function BrowserSurface({
       // its own if the load fails again.
       setFailure(null)
       setNotice('')
+      // The guest is gone, so no did-stop-loading is coming to end a navigation left in flight.
+      loadingRef.current = false
+      setLoading(false)
       onGuestDiscarded?.()
     },
     onRestore: () => {
--- a/src/renderer/nodes/WebNode.tsx
+++ b/src/renderer/nodes/WebNode.tsx
@@ -5,6 +5,10 @@ import type { CanvasNode } from '../state/workspace'
 import { httpUrl } from './webUrl'
 import { useDiscardWhenHidden, webviewAudible, type AudibleWebview } from './useDiscardWhenHidden'
 import { DiscardedPlate } from './DiscardedPlate'
+import { reloadWebview, type ReloadableWebview } from './webviewReload'
+import { describeLoadFailure, isReportableFailure, type WebviewLoadFailure } from './webviewError'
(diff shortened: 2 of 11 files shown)


howieyoungdahl (commented):
Read this as a `<webview>` change first and a UX change second, because the keep-alive contract in
`CLAUDE.md` fails silently when it is broken: a guest that gets moved in the DOM dies and takes a
full page reload with it, and nothing on screen says so. `git merge-tree` against `main`
(`db39636e`) is clean. I did not run the suite or the app.

## What I verified and agree with

- **Nothing in the diff touches the pool.** `lib/webviewKeepAlive.ts`, `state/webviewKeepAlive.ts`,
  `overlayKeepAliveData` and `Canvas.tsx` are all absent from the eleven changed files, so the
  append/remove order discipline and the ghost merge are untouched by construction rather than by
  agreement.
- **Neither surface moves the guest.** In `BrowserSurface`, `{!discarded && <webview/>}` stays the
  first child of `.browser-node__view` and the plate, bar and renamed `notice` are all appended
  after `DiscardedPlate` — no kept child's relative order changes, so `lastPlacedIndex` never
  schedules a move. In `WebNode` the scarier-looking edit is wrapping the element in a fragment, and
  it is safe for two reasons worth stating: the `<webview>` is the fragment's index-0 child with
  both conditionals after it, and the fragment is what that position always renders when live, so
  there is no runtime transition between element types.
- **`live` is equivalent to the condition it replaced.** `live = !!src && !discarded && !restoring`,
  but the outer ternary has already taken `discarded || restoring`, so inside the else branch it
  reduces to `!!src` — the old test exactly. Deriving it once and using it for both the body and the
  header button is the right call; two copies is how those two drift.
- **No node data is written.** Every new fact is component-local `useState`, so there is no
  `updateNodeData` that would need routing through `updateGhostData` when the node is a ghost, and a
  ghost keeps the same mounted element, so the plate and bar survive a project switch with it.
- **The clear-on-`did-start-loading` rule is right and is the one that would have been "simplified"
  away.** Chromium navigates to its own error page under the failed URL, so `did-navigate` would
  wipe the plate as it appeared; pinning it at source level in `webview-error-parity.test.ts` is the
  same discipline as `hook-verified-parity.test.ts` and `fs-atomic.guard.test.ts`, and it is the
  right tool for an invariant a type cannot carry.
- **`isReportableFailure` is the same predicate `BrowserSurface` already had** — `isMainFrame &&
  code !== -3` — now named, and `WebNode` gets it for the first time.
- **The error table is accurate.** I checked the codes against `net_error_list.h` values: -2/-6/-7/
  -20/-21/-100/-101/-102/-104/-105/-106/-109/-118/-137/-200/-201/-202/-300/-301/-302/-310/-324/-501
  all map to the right symbol, and the unknown-code degrade returns the generic sentence plus the
  raw symbol rather than a guess.
- Server Edition N/A is correct — no `<webview>` in a plain browser, and the changed files are
  renderer-only.

## Two things I would change

**1. `WebNode` now knows its page is loading and does not tell the memory saver.** Its
`useDiscardWhenHidden` options still pass `isLoading: () => grantingRef.current`, which only covers
the `nt-media://` grant round-trip; the hook reads `isLoading()` at fire time to decide whether to
discard. `BrowserSurface` feeds the page-load state in (`isLoading: () => loadingRef.current`) and
keeps the ref specifically for it — its own comment says *"`loadingRef` mirrors the `loading` state
because the hook reads it at fire time, from a callback that must not force the observer to be
re-created"* — and this PR adds the same signal to `WebNode` as plain state, with no ref, so it
cannot be read there even if someone wanted to.

Concretely: a web node pushed off-screen mid-navigation is now discardable in a state
`BrowserSurface` would refuse. Nothing is permanently lost (the restore re-navigates from
`data.url`), which is why I would call this small rather than a bug — but it is the same
"a memory lever must ask what it is ending" question the repo asks of every other lever, and the
answer currently differs between the two surfaces for no stated reason. `loadingRef` + `isLoading:
() => grantingRef.current || loadingRef.current` is the whole fix.

**2. `webview-error-parity.test.ts` can silently skip its own strongest assertion.** The last case
ends:

```ts
const nav = src.indexOf('const onNav = ')
if (nav > -1) expect(src.slice(nav, src.indexOf('const onNavInPage'))).not.toContain('setFailure(')
```

Rename `onNav` in `BrowserSurface` — or change its declaration to `function onNav(` — and the guard
against clearing the plate from `did-navigate` evaporates with the suite still green. That is the
shape `CLAUDE.md` already flags for `getMenuItemById`: the lookup answers nothing for a typo and the
fail-safe is to do nothing, which is indistinguishable from the feature working. Since the surfaces
genuinely differ (`WebNode` has no `onNav`), the cheapest fix is a per-surface expectation — assert
the handle is FOUND for `BrowserSurface.tsx` and absent for `WebNode.tsx` — rather than a silent
skip. The `slice(start, start + 400)` window in the same test fails loudly if it drifts, so that one
is fine as is.

## Smaller notes, non-blocking

- `BrowserSurface`'s `onDiscard` clears `failure` and `notice` but not `loading`, while `WebNode`'s
  clears `loading` too; and the plate is `!discarded`-gated on both surfaces while the bar is not.
  The hook's `isLoading` gate means neither can currently strand a bar over the `DiscardedPlate`, so
  this is consistency rather than a defect — but one of those two asymmetries is redundant and the
  other is load-bearing, and it is not obvious from the code which.
- `WebNode` keeps a third failure presentation: the grant-refused / unusable-scheme `error` still
  renders as a bare `<span>{error || 'No source'}</span>`. The comment says that split is
  deliberate, and I think it is right — just confirming it is a decision, since the PR's stated
  premise is that one failure should not read two ways.
- The reload-parity regex `/\.current\??\.reload(IgnoringCache)?\s*\(/` pins the `.current` idiom
  specifically; a future call through a local variable would pass it. Fine for what exists today.
- `reloadWebview`'s "a forced reload never falls back" rule means a Shift+click on a guest exposing
  only `reload()` does nothing at all, silently. Real `WebviewTag`s always have both, so this is
  theoretical — the rationale in the docstring is convincing enough that I would not change it.

Good change. My asks before merge are (1) and (2); everything else is a note.

**What I could not verify:** I did not run Electron, so I have not seen an opaque `position:absolute;
inset:0` plate actually composite over a *live* `<webview>` guest. The existing
`.browser-node__error` strip and `BrowserStartPage` are precedents for DOM over a mounted guest, and
`.editor-node__image` / `.browser-node__view` are both positioned containing blocks (`styles.css:3040`
and the pre-existing `.browser-node__error` rule), so the layout should hold — but full-cover over a
painting guest is a stronger claim than a 20px strip and is the thing I would want a screenshot of.
I also did not run the 28 new tests, and I have not confirmed the `prefers-reduced-motion` branch
renders as intended.

basmilius:
Thanks — both asks are addressed, and the screenshots you were missing are already in the PR body.

**(1) `WebNode` and the memory saver** — fixed in `7b1135de`. `loadingRef` mirrors the guest's load
lifecycle and `isLoading` is now `() => grantingRef.current || loadingRef.current`, the same shape
`BrowserSurface` uses and for the same reason: the hook reads it at fire time from a callback that
must not re-create the observer.

**(2) The parity test's silent skip** — fixed in `5230634f`. The `if (nav > -1)` is gone; each
surface is asserted rather than probed. `BrowserSurface` must have the handle and must not clear the
plate inside it, `WebNode` must have no `did-navigate` listener at all. I verified it the way you
would want: renaming `onNav` to `onNavigated` now reds the suite instead of quietly passing.

**The `loading` asymmetry you flagged as a note** — also fixed, in `f224b6c5`. `BrowserSurface`'s
`onDiscard` now clears `loading` too. The guest is gone, so no `did-stop-loading` is coming to end
it; the `isLoading` gate means it could not strand a bar today, but that was the redundant half
rather than the load-bearing one.

**The plate over a live guest** — the two screenshots at the bottom of the description are exactly
that, sorry, they are easy to miss with no caption. Both are `ERR_NAME_NOT_RESOLVED`, which means
Chromium had already navigated the guest to its own error page: that page is painting underneath,
and the fact that you cannot see any of it is the compositing you wanted confirmed. What they do not
show is the loading bar or the reduced-motion branch.

**The third failure presentation in `WebNode`** — confirming it is a decision. The bare `<span>` is a
source that never reached a guest (unusable scheme, refused grant), not a page that was attempted and
failed, so it is not the same kind of thing the plate speaks for.

eneskirca:
Reviewed at `5230634f` in an isolated worktree (main checkout untouched). `npm run typecheck` clean; `src/renderer/nodes` green (17 files, 126 tests), and the `fs-atomic` / `no-electron` / `source-hygiene` guards pass. **Merge-ready.**

**Both surfaces, checked rather than assumed.** `WebNode` and `BrowserSurface` both import `webviewError`, both filter through `isReportableFailure`, both render `WebviewErrorPlate` and `WebviewLoadingBar`, and both clear only from `onStart` — and the kanban card modal picks it up for free (`CardModal.tsx:425` hosts the same `BrowserSurface`). The two parity tests are real source pins in the house idiom, not vacuous: the `did-navigate` case is asserted per surface with the handle looked up first, so a renamed handler goes red instead of silently passing.

**The trap is the right one and it is stated in the right place.** Chromium navigating to its own error page under the URL that just failed is precisely why `did-navigate` cannot clear the plate; `did-start-loading` is order-independent and a URL comparison would not be. Worth having pinned.

**Guest-supplied text is safe.** `failure.url` and `failure.raw` are React text nodes — no `dangerouslySetInnerHTML`, no `href`, nothing clickable — and `.browser-node__failed-url` is `nowrap` + `text-overflow: ellipsis`, so a hostile redirect target cannot break the plate's layout out of the node. It is muted secondary text under a "could not be loaded" sentence, not an actionable surface, so I do not think the URL needs suppressing.

**Details that are right and easy to get wrong:** the plate positions against `.editor-node__image` (`position:absolute; inset:0`), so it covers the body and not the header; `--tint-rgb` / `--accent` are `:root` tokens so the bar themes correctly in both; `loadingRef` is cleared in `onDiscard` on both surfaces, so a guest released mid-navigation cannot leave the bar spinning with no `did-stop-loading` coming; and splitting "Enter a URL or search term" out into `notice` is the correct fix — it never had any business covering a readable page.

**Two notes, neither blocking:**

- The `<webview>` event wiring itself has no behavioural test on either surface (the 28 are pure + source-grep), and it cannot have one here — `<webview>` only exists under Electron. So the one thing worth a device pass is the **first** load: `WebNode` attaches its listeners in a `useEffect` keyed on `live`, which runs after the commit that mounts the guest with its `src` already set. `BrowserSurface` has shipped that same shape, so I expect it to be fine (the guest is created asynchronously after attach), but "does the bar appear on a node's very first navigation, and does a dev-server-down node show the plate rather than Chromium's page" is the check I would want run on a real build.
- `describeLoadFailure` does not cap `raw`/`url` length. Harmless today — `errorDescription` is Chromium's own short symbol — and the CSS handles the URL. Only worth a line if you ever render `raw` somewhere without the ellipsis rule.

Surfaces call is right: Desktop is the feature, Server Edition is N/A by construction (no `<webview>` in a browser, and the new code is inert when the ref is null), mobile has no canvas.

PR #644, opened 2026-09-02
test(tmux): give every vitest run a private TMUX_TMPDIR

### LINKED ISSUE #629: Shared tmux server dies abruptly under a full vitest run from inside a node: every node prints [server exited unexpectedly]
## Summary

Every Claude/agent node on the canvas died at once, twice today (12:37 and 14:46:43 local), each pane printing tmux's `[server exited unexpectedly]` followed by nodeterm's `[process exited with code 1]`. Root: the single shared tmux server (`-L node-terminal`) died abruptly; every node is a pane on it, so one server death is a total outage. Both deaths happened while a full `vitest run` of this repo (688 files / ~9,400 tests, 5 workers on a 10-core Mac) was executing **from inside a nodeterm node**, on a machine already in memory crisis (16 GB swap 93% used, a jetsam event at 14:07 jettisoned 245 processes). Two other identical full runs (11:07, 14:12) finished in ~60 s with no death, so the trigger is resource-related and non-deterministic, not a specific test.

Not a Claude Code issue: a session still on Claude 2.1.257 died the same way as the 2.1.258 ones, and Claude only exits because its pane got SIGHUP.

## Evidence

- `[server exited unexpectedly]` is tmux's `CLIENT_EXIT_LOST_SERVER` (client.c): the client lost the socket **without** `MSG_SHUTDOWN`. `kill-server`, SIGTERM and SIGINT all go through `server_send_exit()` and print `[server exited]` instead. So the server was SIGKILLed, crashed, or hit a tmux `fatal()`. No crash report for tmux exists in `~/Library/Logs/DiagnosticReports` or `/Library/Logs/DiagnosticReports`, and no JetsamEvent at either death time.
- tmux 3.7c `server_accept()` calls `fatal("accept failed")` for any errno other than EAGAIN/EINTR/ECONNABORTED/ENFILE/EMFILE (e.g. ENOMEM/ENOBUFS). nodeterm's PtyManager polls the server constantly (`display-message`, `list-panes`, …), so the server is doing `accept()` many times per second exactly while the suite spawns thousands of short processes. This is the mechanism that fits all the evidence; it was not reproduced on purpose (a repro kills every live node).
- New tmux server `-L node-terminal` (pid 37335) started 14:47:14; nine nodes re-registered in `agent-status.json` at 14:47; all sessions in `tmux -L node-terminal ls` are created ≥ 14:47:14.
- The QA node's run log (`scratchpad/base/suite.txt`) stops being written at 14:46:43, mid-run. Two orphan test servers `tmux -L nt-pr3-*` (from `agent-message.realtty.test.ts`) started at 12:36:49 and 14:46:39 are still alive, with 8 `exec -a claude bash -i` fake agents: their `afterAll` never ran because the vitest process died with its pane. 128 `nt-e2e-*` temp dirs are left in `$TMPDIR` from the same kind of interrupted runs.
- Ruled out: no agent ran `pkill`/`kill -9`/`kill-server` against the shared server at either time (all Claude transcripts of the day searched); the 11 tmux-touching test files (`*.realtmux`, `*.realtty`, `*.realsh`, `host-destroy-tmux`, `pane-owner`, `phone-target-e2e`) were run against a decoy server named `node-terminal` in a private `TMUX_TMPDIR` with `TMUX` pointing at it: decoy survived, 147/152 passed, same 5 env failures as the QA runs.
- `vitest.config.ts` already documents fd pressure at 10 workers ("whole files fail in ways that never reproduce alone"); the same pressure reaches the tmux server, which is not a vitest worker and has no retry.

## Also found (not the killer, but hazards)

- `src/main/remote/host-destroy-tmux.test.ts` creates and destroys sessions **on the real `node-terminal` socket** by design, and wires a real `PtyManager` (with reap/snapshot timers) to it.
- Some suites (`account-env.realtmux`, `control-master.realsh`, `pane-owner`, `tmux-paste.realtmux`) start servers in the real `/tmp/tmux-<uid>/` dir even when `TMUX_TMPDIR` is set (a `nodeterm-rmt` server appeared there at 14:46 and again during the decoy run).

## Asks

1. Resilience: when the shared server dies, nodes currently end as dead panes. Either recreate the server and sessions and relaunch each agent with its resume id, or stop sharing one server across every node/project (per-project socket at least).
2. Test-suite guard: refuse to run (or auto-isolate with a private `TMUX_TMPDIR`) any real-tmux suite when `$TMUX` points at the `node-terminal` server, and make `host-destroy-tmux.test.ts` use a private socket.
3. Cap the suite's process/fd burst when run from a node (lower `maxWorkers`, or `fileParallelism: false` for the real-process suites) so a full run cannot starve the shared server.
4. Cleanup for killed runs: orphan `nt-pr3-*` servers and `nt-e2e-*` dirs survive a SIGHUP'd run forever.

## Environment

nodeterm 0.3.4 · tmux 3.7c (Homebrew) · macOS 26.3 (Darwin 25.3.0), 24 GB RAM, 10 cores · Claude Code 2.1.257/2.1.258 · vitest 4.1.11, `maxWorkers: '50%'`.

eneskirca:
## What this addresses

Ask 2 of #629 (and half of ask 4). It does **not** claim to fix the server death — see "What this does not claim".

## The measurement

Every test file that spawns a real `tmux` was run individually with `TMUX_TMPDIR` pointed at a probe directory, and the probe was then inspected for socket files. Exactly one suite binds a production socket name:

| suite | socket | own `TMUX_TMPDIR`? |
|---|---|---|
| `src/main/remote/host-destroy-tmux.test.ts` | `node-terminal` (`TMUX_SOCKET`) | **no** — inherits, i.e. `/tmp/tmux-<uid>` |
| `src/core/agents/pane-owner.test.ts` | `nodeterm-rmt` (hardcoded in the bytes under test) | yes |
| the other 6 `*.realtmux` / `*.realtty` suites + 26 tmux-adjacent files | private per-pid names | yes, or a faked `child_process` |

So on a machine where nodeterm is running — which is how this repo is developed — `host-destroy-tmux.test.ts` created sessions on the shared server carrying the developer's entire canvas and drove a real `PtyManager` at it, and `pane-owner.test.ts` was one `TMUX_TMPDIR` mistake away from the same.

## The fix

`tmux` resolves `-L <socket>` to `$TMUX_TMPDIR/tmux-<uid>/<socket>`, falling back to `/tmp` when the variable is unset. Re-pointing the variable therefore re-points **every** socket name at once, including the real ones and including a suite nobody thought about.

- `test/setup/tmux-sandbox.ts` (`globalSetup`) — creates one private directory per run, and on teardown kills every server still bound inside it before removing it. That last part is the shape of #629's leftovers: `nt-pr3-*` servers whose `afterAll` never ran, each still holding its fake agents. `rm -rf` alone unlinks the socket and leaves the server alive forever.
- `test/setup/tmux-worker-env.ts` (`setupFiles`) — re-asserts it inside each worker and **throws** if it is missing. Redundant today (workers inherit the main process's env), deliberately kept: that inheritance is vitest's implementation detail, not its contract, and the silent failure mode is "every test is back on the live server with a green suite".
- `enterSandbox` also strips `TMUX`/`TMUX_PANE` — a suite run from inside a nodeterm terminal inherits a live client's, and production strips both from every child for the same reason.
- `host-destroy-tmux.test.ts` keeps the socket name (it has no choice — `PtyManager` binds `TMUX_SOCKET` itself and the suite exists to watch the manager's own kill land) but now **refuses to start** unless the sandbox is in effect.

## The guard, and what it cannot do

`src/core/tmux-socket-isolation.guard.test.ts` pins the invariant three ways because each leg covers what the others cannot:

1. **Structural** — the sandbox. The only leg that covers a test nobody thought about.
2. **Behavioural** — it starts a server on the real socket *name* and asserts the socket file appeared inside the sandbox. Asserting the environment variable would only prove we set a variable; the resolution rule is a property of tmux.
3. **By review** — an allowlisted source scan (two entries, each with its reason), same shape as `fs-atomic.guard.test.ts`.

The scan is the weakest leg **on purpose** and the file says so: a test can still escape by handing a real tmux an `env` object it built from scratch with no `TMUX_TMPDIR` in it, and no regex sees that reliably. It also carries a positive control, since a scan nobody proved can match anything reads exactly like a scan with nothing to find — it caught its own two known offenders on the first run.

## What this does not claim

**The server death in #629 was not traced to a test, and this does not fix it.** The reporter's evidence points at tmux's `server_accept()` calling `fatal()` under the suite's process/fd burst on a memory-starved machine, and two identical full runs finished clean. Sharing a server with the user's live sessions is a hazard whatever kills it; this removes the hazard, not a proven cause. Asks 1 (resilience when the shared server dies) and 3 (capping the suite's burst) stay open.

## Verification

- Full suite green on Linux with the sandbox in place (~10,000 tests, 3 workers); `/tmp/tmux-0/` byte-identical before and after, and the sandbox directory removed at teardown. Before the change the same run created `/tmp/tmux-0/node-terminal`.
- `npm run typecheck` clean.
- Mutation-tested: dropping `globalSetup` from `vitest.config.ts` makes every file fail with the sandbox-missing error; the scan's own file was its first hit.

### Device checklist (not verified here)

This machine is Linux; the reporter's is macOS, and the path-length arithmetic is the macOS-only part.

1. `npm test` from inside a nodeterm terminal on macOS: suite green, and no new socket appears in `/tmp/tmux-<uid>/`.
2. The sandbox fits under macOS's 103-character `sun_path` limit — `makeTmuxTmpdir` falls back from `/private/var/folders/…` to `/private/tmp`, which is the branch that only runs there. A failure here is a thrown error naming the number, not a silent fallback.
3. `src/main/remote/host-destroy-tmux.test.ts` passes on macOS (it self-skips without tmux at one of the three fixed paths).
4. After a full run, `ls /tmp/ntvitest-*` is empty — the teardown removed the sandbox.
5. Kill a run mid-flight (the #629 scenario) and confirm the leftovers are now confined to one `ntvitest-*` directory instead of the shared tmux dir.

Fixes #629

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2870,6 +2870,13 @@ hardcode `-L nodeterm-rmt`; re-spelling it would judge different bytes) and
 the one file that reached the live server by construction — measured, not inferred — and it now
 refuses to start unless the sandbox is in effect.
 
+**What the sandbox does NOT do:** two suites naming the same socket inside it still share one tmux
+server, so a `kill-server` there is still a shared-server kill — it has just been moved somewhere
+harmless. Measured on CI the day this landed: the guard test's own `kill-server` on
+`node-terminal` ended `host-destroy-tmux.test.ts`'s session mid-assertion. A suite kills its OWN
+sessions by exact target (`-t =<name>`, since a miss falls through to prefix matching), or it owns
+a socket name nothing else uses.
+
 The guard has three legs on purpose, and the weakest one is the scan: a test can still escape by
 handing a real tmux an `env` object it built from scratch with no `TMUX_TMPDIR` in it, which no
 regex sees. So the structural leg is the sandbox, the behavioural leg actually **starts a server on
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2846,6 +2846,43 @@ also hold an exclusive candidate lock until the rename and cleanup finish. Never
 those back to `<target>.tmp` / `<target>.part` or a read-only "does the destination exist?" check —
 the overlap tests exercise the resulting race.
 
+## The test suite never touches a live tmux server
+
+This repo is developed from inside nodeterm, so `tmux -L node-terminal` and `-L nodeterm-rmt` are
+not fixture names on a contributor's machine — they are the servers holding every terminal they have
+open. A test that binds one shares a process with the user's whole canvas, and the failure mode is
+not a red test: it is every pane printing `[server exited unexpectedly]` (issue #629).
+
+**Every vitest run gets a private `TMUX_TMPDIR`.** tmux resolves `-L <socket>` to
+`$TMUX_TMPDIR/tmux-<uid>/<socket>` and falls back to `/tmp` when the variable is unset, so
+re-pointing the variable re-points every socket name at once — including the real ones, including
+in a suite nobody thought about. `test/setup/tmux-sandbox.ts` (`globalSetup`) creates the directory,
+kills whatever is still bound inside it and removes it; `test/setup/tmux-worker-env.ts`
+(`setupFiles`) re-asserts it inside each worker and **refuses to run** if it is missing, because
+vitest's env inheritance into workers is an implementation detail and a silent fallback would put
+every test back on the live server. `enterSandbox` also strips `TMUX`/`TMUX_PANE` — a suite run from
+inside a nodeterm terminal inherits a live client's, and production strips both for the same reason.
+
+Two suites deliberately name a real socket, and both are allowlisted with their reason in
+`src/core/tmux-socket-isolation.guard.test.ts`: `agents/pane-owner.test.ts` (the production bytes
+hardcode `-L nodeterm-rmt`; re-spelling it would judge different bytes) and
+`main/remote/host-destroy-tmux.test.ts` (`PtyManager` binds `TMUX_SOCKET` itself). The latter was
+the one file that reached the live server by construction — measured, not inferred — and it now
+refuses to start unless the sandbox is in effect.
+
+The guard has three legs on purpose, and the weakest one is the scan: a test can still escape by
+handing a real tmux an `env` object it built from scratch with no `TMUX_TMPDIR` in it, which no
+regex sees. So the structural leg is the sandbox, the behavioural leg actually **starts a server on
+the real socket name and proves the socket file landed inside the sandbox** (asserting the env var
+would only prove we set a variable — the resolution rule is a property of tmux), and the scan exists
+to make a third allowlist entry a decision somebody signs for. Same shape as the `fs.rename` guard,
+for the same reason: nobody reading one file can see this.
+
+**What this does not claim.** #629's server death was not traced to a test — the reporter's evidence
+points at tmux's `server_accept()` calling `fatal()` under the suite's process/fd burst on a
+memory-starved machine, and two identical runs finished clean. Sharing a server with the user's live
+sessions is a hazard whatever kills it; this removes the hazard, not a proven cause.
+
 ## Conventions
 
 - **Two docs, two audiences — keep both.** This file holds the deep invariants with their
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -310,6 +310,16 @@ Where a behaviour can only be verified on hardware we do not have in CI (a Mac, 
 GPU), say so explicitly rather than implying coverage. Several docs carry numbered device
 checklists for exactly this.
 
+**A test never touches a live tmux server.** You will most likely run `npm test` from inside a
+nodeterm terminal, where `-L node-terminal` and `-L nodeterm-rmt` are the servers holding every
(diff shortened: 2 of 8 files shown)


eneskirca:
Follow-up commit, worth calling out because CI found the same class of bug this PR is about, one level in.

The behavioural leg starts a server on the real socket *name* to prove the sandbox catches it, and tore it down with `kill-server`. Inside the sandbox that name is shared with `host-destroy-tmux.test.ts`, which binds it too (no choice — `PtyManager` hardcodes the name), so the kill ended that suite's session mid-assertion. Green locally at 3 and 2 workers; red on CI's 1.

**The sandbox moves the shared server somewhere harmless. It does not make `kill-server` safe.** Two suites naming one socket inside it still share one tmux server. A suite kills its own sessions by exact target (`-t =<name>` — a miss falls through to fnmatch and then to prefix matching), or it owns a socket name nothing else uses. That is now in the module header and in CLAUDE.md rather than only in this thread.

Re-verified: full suite green at 1 and 2 workers, `/tmp/tmux-0/` byte-identical throughout.

PR #652, opened 2026-09-02
fix(codex): resume sessions after a shared-daemon reset

howieyoungdahl:
Extracted from #537 on its own, per the staging discussion there — this is that branch's **first** commit and is unrelated to canvas control. It merges clean on current `main`.

## The problem

A shared Codex daemon restart is not a terminal-session restart. tmux survives and the rollout survives, but every `codex --remote unix://` TUI attached to that account's app-server socket exits together — so a pane that was mid-conversation is simply gone, with the thread still on disk and nothing reattached to it.

## The change

`buildCodexLauncherScript` stays in the pane as a bounded transport supervisor after binding a thread, instead of `exec`ing the remote TUI.

On an abnormal client exit it resumes that exact thread **only** when `app-server daemon version` no longer reports `status: running`, or the known control-socket inode changed. The same healthy generation returns the original status, so a deterministic CLI error cannot relaunch forever. A reconnect never replays the launch prompt or options — that would duplicate the user's turn — and three rapid resets stop with a manual `codex resume <thread>` receipt rather than looping.

Preflight probes live protocol health before lifecycle start, because Codex's PID ownership record can go stale while the shared process is still responsive, and killing that "orphan" fans one bookkeeping failure out across every node on the canvas.

## Testing

The generated shell is exercised under a real `/bin/sh` — replaced-socket, missing-daemon, healthy-client-error and responsive-orphan cases. The healthy-client-error case is the mutation guard: drop the status check and it goes red.

Both shells wire the same spine. Electron and Server Edition arm the same signed record secret, thread start/bind handlers, capability refresh and UI identity events; the server composition is isolated in `server/codex-shared-identity.ts` and behavior-tested. This is deliberate — the previous Server Edition answer ("deliberate plain Codex") bypassed the launcher entirely, so a reconnect fix could have been perfectly green in the launcher while every headless pane still fell back to its shell.

Gates on current `main`: `npm run typecheck` clean; `codex-launcher-sh.test.ts` + `codex-shared-identity.test.ts` 37/37; `git merge-tree` against `main` reports no conflicts.

## Coordination

The launcher lines here will conflict with #540 once both land — I will rebase whichever goes second.

https://claude.ai/code/session_01G79wE44T1imWesbY83z4UZ


eneskirca:
Reviewed on an isolated worktree at `pull/652/head`, merged onto current `origin/main` (no conflicts even after #540 landed). `tsc` clean both projects; `codex-launcher-sh.test.ts` + `codex-shared-identity.test.ts` + the `no-electron` / `fs-atomic` guards **94/94** green.

**Verdict: changes requested.** The supervisor idea is right and the resume-gating logic is careful. But leaving the launcher in the pane changes what `#{pane_current_command}` answers, and nothing in the PR accounts for that. I measured it.

## Blocking — the pane now reports `sh`, and six call sites read that field

I ran **this PR's own generated `nt_run_shared`** under real `/bin/sh` (dash) in tmux 3.4 on Linux, supervising a live client that holds the tty:

```
pane_current_command = [sh]

  PID     PGID  ARGS
  1481372 1481372 -bash                     <- pane root
  1481392 1481392 /bin/sh .../nodeterm-codex <- supervisor, owns the fg pgrp
  1481413 1481392 sleep 300                  <- the client, same pgrp
```

Before this PR the pane `exec`'d the TUI, so the same read answered `codex` / `node`. Now `isShellCommand('sh')` is **true while the TUI is alive and reading the tty**. That is a live-session contract, not a cosmetic one:

1. **`agent-restart.ts:237`** — `performExitPhase` writes `/quit`, then polls `#{pane_current_command}` and takes `if (isShellCommand(pane)) return 'exited'` **immediately, on the first 250 ms poll**, before Codex has quit. `performResumePhase` then KILL_LINEs and echo-delivers the resume line into the *live* TUI, where it lands as prompt text and is submitted as a turn. This is the "rename splice = lost launch" shape the repo already has an invariant for, and it hits the node menu's Restart, the bulk restart-idle-agents action, and Eco hibernation's exit sweep.
2. **`TerminalNode.tsx:3604`** — hibernation wake's `!isShellCommand(pane)` guard passes for the same reason.
3. **`core/trigger-delivery.ts:96`** — the shell-owned-pane gate that exists precisely to stop a delivery landing in a live agent TUI now reads a supervised Codex pane as shell-owned.
4. **`core/remote-ssh/agent-resync-decide.ts:22`** — `isShellCommand(paneCommand) ? 'ended'` marks a live remote Codex session as ended on reconnect resync.
5. **`core/pane-process.ts:28`** and **`core/pty-manager.ts:4070`** both return the not-an-agent answer for every managed Codex node.

`isAgentPane` (`pane-owner-predicate.ts`) is fine — it scans every member of the foreground group, finds `codex`, and its `agentPidIn` doc already anticipates exactly this shape ("a wrapper loop makes the second case ordinary"). So the kernel-truth predicate survives the change; the six `isShellCommand` readers do not.

**A measured one-line fix exists.** `set -m` in the supervisor gives the client its own process group and makes it the tty's foreground pgrp, restoring the old answer — measured on the same host:

```
set -m  => pane_current_command = [sleep]     # would be codex/node for the real client
1453093 1453093 1453094 /bin/sh .../nt-jobctl
1453094 1453094 1453094 sleep 120             # tpgid == the client's own pgid
```

dash supports `set -m` in a script. It also means tty SIGINT reaches only the client, which is what the existing `130` case already wants. Please add it (or an equivalent), and pin it with a test that fails if the supervisor ever becomes the foreground group again — this is not something a reader will re-derive.

## Non-blocking

- **The Server Edition switch-on is a second change riding along.** Reversing "deliberate plain Codex" to a live shared app-server on a whole surface is independently risky and is not required by the reconnect fix — the launcher change stands on its own on desktop. Your justification (a fix green in the launcher while every headless pane bypasses it) is fair, but given #537 was split for staging reasons this is the same argument one level down. I would take it as its own PR; if it stays, say so in the title.
- **The rapid-reset stop is not a ceiling.** `nt_rapid_resets` resets to 0 after any run ≥ 10 s, so a daemon flapping every 11 s resumes forever. Defensible for a supervisor, but the PR body's "three rapid resets stop" reads as a hard bound and it isn't.
- **Inode reuse fails safe, correctly.** A restarted daemon that lands on the same inode while healthy yields no resume — a false negative, which is the right direction. Worth one line in the comment so nobody later "fixes" it into a positive.
- `nt_socket` is assigned in `nt_app_server_generation` without being local (that function is `{ }`, not `( )`), so it leaks into the caller's scope. Harmless today, one rename from not being.

Separating measurement from assumption for the record: the `pane_current_command` and `set -m` results above are **measured on this host** (Linux, tmux 3.4, dash, codex-cli 0.151.0) with the PR's own generated script. What I did **not** verify on a device: a real shared-daemon restart driving an actual `codex --remote` reconnect, and the Server Edition wiring against a live app-server.

howieyoungdahl:
Thanks — that measurement is exactly right, and I want to say plainly: I shipped the launcher without checking what job control does to `#{pane_current_command}`, and you caught a real live-session hazard, not a cosmetic one.

**Fixed at `dfd8a896`: `set -m` at the top of the generated script**, before `nt_run_shared` is even defined, so job control is on for the very first launch too, not only a resume issued after a daemon reset. Verified the same way you did — a real pty, the `ps`-based foreground-group read this repo already gates on (`pane-owner.ts`), not a string assertion:

```
# before the fix (job control off)
pgid == the supervisor's own pid, foreground argv[0] == the supervisor

# after (set -m)
pgid == the client's own pid (its own new group), foreground argv[0] == the client
```

Pinned with `src/core/codex-launcher-foreground-group.realtty.test.ts`: a control case proving the same shape *without* `set -m` still shows the old, broken answer (so the fix case isn't vacuous), the fix case, and a structural check that the generator still emits `set -m` early enough that it's live before `nt_run_shared` can ever be called. Gates: `tsc` clean; `codex-launcher-sh.test.ts` + the new file both green; `vitest run src/server src/core` — same two pre-existing failures as base (`opencode.test.ts`'s 4, reproduced identically with my commit stashed out; `pty-single-user.test.ts` under full-suite contention), nothing new.

On the three non-blocking points:

- **`nt_socket` scoping** — changed `nt_app_server_generation` from `{ }` to `( )` so it's actually local, not only accidentally-harmless-today.
- **The rapid-reset stop is not a ceiling** — agreed, and I should have said so in the PR body rather than let "three rapid resets stop" read as a hard bound. Added a comment on the counter stating plainly that any run ≥10s clears it, so a daemon flapping every 11s resumes forever; this only stops a *rapid* burst.
- **Inode reuse fails safe** — added a comment on the generation check saying so, so nobody "fixes" it into a false positive later.

On the Server Edition wiring riding along: fair, and I don't think it's actually separable from *this* fix the way #537's ask 1 hunk was. The reconnect bug only exists for a session running against the shared app-server, and on Server Edition that plumbing didn't exist before this PR (`registerCodexIdentityIpc(() => UNKNOWN_CODEX_IDENTITY_CAPS)` — a deliberate permanent "no shared identity" answer). Landing the launcher fix without it would ship a fix nobody running headless benefits from, and reverting `src/server/*` here would leave Server Edition's Codex nodes on the old degrade indefinitely. If you'd rather review it as two PRs anyway, say so and I'll split `src/server/index.ts` / `handlers/index.ts` / `codex-shared-identity.ts` out — I'd rather ask than guess wrong on this one given the guidance from #537.

What I did **not** verify on a device, same as last time: an actual `codex --remote` reconnect against a real shared-daemon restart, and the Server Edition wiring against a live app-server. The pty measurement above is real, but it's a stand-in for the client, not a live Codex.

eneskirca:
Re-verified `dfd8a896` in an isolated worktree, merged onto current `origin/main` (clean). `npm run typecheck` clean on both projects; `src/core` + `src/server` green — 259 files, 3839 passed, 15 skipped, including the new `codex-launcher-foreground-group.realtty.test.ts`. None of the failures you reported at base reproduce here.

**The blocking finding is closed, and I measured it again rather than reading the diff** — this time with the real script SHAPE (the client launched from inside a function inside the `while` loop, dash `/bin/sh`, under a real pty), and then through the field the six call sites actually read:

```
without set -m   ps -t <tty>                  tpgid == the supervisor's own pgid, foreground argv = /bin/sh run.sh
                 tmux #{pane_current_command} = sh
with set -m      ps -t <tty>                  client has its own pgid and owns tpgid, foreground argv = sleep 30
                 tmux #{pane_current_command} = sleep
```

(Linux, tmux 3.4, dash, on an isolated socket.) So the pre-supervisor answer is restored for the whole script's life rather than only inside a post-reset resume, and the unconditional generator test pins that `set -m` is emitted before `nt_run_shared` can be defined. The CONTROL case in your realtty test is what keeps the fix case from going vacuous — right call, and it is the reason I could re-measure the claim instead of taking it.

The three non-blocking notes are all taken: `nt_app_server_generation` is a `( )` subshell so `nt_socket` is genuinely local, the rapid-reset counter now says plainly that any run ≥ 10 s clears it (so an 11 s flap resumes forever and this only stops a rapid burst), and the inode-reuse case is named as a fail-safe false negative with "do not fix this into a positive".

**One thing before merge, and it is the rebase's doing rather than yours.** `main` has since gained a CLAUDE.md paragraph describing the exact degrade this PR removes (~line 1555): *"…`src/server/handlers/index.ts` deliberately registers neither — so on the Server Edition the file is byte-identical … that shell answers `shared: false` (`UNKNOWN_CODEX_IDENTITY_CAPS`) … It turns into a real gap only when that edition grows the shared app-server, and the fix is the two registrations."* This branch makes exactly those registrations, so after a merge CLAUDE.md asserts the old behaviour in one section and the new one in yours. One paragraph, in this PR — that file's own same-PR rule.

On splitting the Server Edition wiring into its own PR: still @eneskirca's call, not mine. Your argument is on the record and I have no new objection to it.

eneskirca:
Verified again on `dfd8a896` merged with current `main`: typecheck clean, `src/core` + `src/server` green (259 files / 3839 tests). Your `set -m` fix was re-measured here with the real script shape (function-scoped while loop, dash, a real pty) against the field six call sites read: without it `#{pane_current_command}` reports `sh` and tpgid is the supervisor; with it the client is back in its own process group and the answer matches pre-supervisor behaviour. The three non-blocking notes are addressed too.

One thing was left, and I pushed it maintainer-side rather than sending the PR back for a paragraph — `48e85f8e` on your branch, your commits untouched. `main` has since grown a CLAUDE.md paragraph stating that `src/server/handlers/index.ts` "deliberately registers neither" handler and answers a flat `shared: false`; this branch makes exactly those registrations, so on merge the contract would have contradicted its own code. The paragraph now describes what you built (the spine wired after the boot secrets exist, and `registerCodexIdentityIpc()` answering from the live resolver so an early browser caller waits for the refresh instead of being pinned to a false "plain Codex" answer), and keeps the remote leg as the part that stays desktop-only.

Splitting the Server Edition half into its own PR is the owner's call and he has not asked for it; merging as one.

PR #656, opened 2026-09-02
Add AGY, Pi, and Goose agent support

laulpogan:
## Summary

- add built-in AGY, Pi, and Goose agents across creation menus, settings, keybindings, process recognition, and custom-agent launch handling
- use each CLI's interactive launch contract: `agy --prompt-interactive`, positional Pi prompts, and Goose `session` / `run --interactive --text`
- enable linked context sharing for all three agents in local and SSH sessions
- install the `get-linked-context` skill into Claude-compatible, Agent Skills, and Antigravity skill directories
- preserve Goose launch arguments when restarting linked sessions and select the remote context shim for SSH nodes

## Validation

- 234 focused tests pass
- `npm run typecheck` passes
- `npm run build` passes
- production Electron build launched real AGY 1.1.24, Pi 0.84.4, and Goose 1.45.0 TUIs
- Pi loaded `get-linked-context`; Goose discovered it through `goose skills list`
- authenticated context shim listed linked AGY and Goose nodes and read their live terminal output
- live run recorded zero console errors and zero failed requests

Full local suite: 10,172 passed, 25 failed, 37 skipped. All 11 failing test files are unchanged by this branch; failures are existing macOS/environment-sensitive real-shell, socket-path, localStorage, account-environment, and platform-copy expectations.

## Scope

Context sharing uses the terminal-output fallback for AGY, Pi, and Goose. Structured transcript parsing for these CLIs is not included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ then **Send** (nothing auto-submits). Your voice never leaves the machine.
 ### Node kinds
 
 🖥 **Terminal** (xterm + tmux, AI naming) · 🤖 **Agent** (Claude Code / Codex / Gemini /
-GitHub Copilot / opencode / Grok / custom) · 📝 **Sticky note** (link to an agent as context) · 🗂 **Group**
+GitHub Copilot / opencode / Grok / AGY / Pi / Goose / custom) · 📝 **Sticky note** (link to an agent as context) · 🗂 **Group**
 (bind to a **git worktree** for agent-per-branch) · ✏️ **Editor** (Monaco, ⌘S) ·
 🔀 **Diff** · 🌐 **Web / Video**
 
--- a/src/core/agent-launch.test.ts
+++ b/src/core/agent-launch.test.ts
@@ -177,6 +177,23 @@ describe("prepareAgentLaunch logical argv", () => {
     });
   });
 
+  it("builds AGY, Pi, and Goose interactive launches", async () => {
+    await expect(
+      prepareAgentLaunch(start("agy", { prompt: "fix it" }), "posix", builtinContext("agy")),
+    ).resolves.toEqual({ command: "'agy' '--prompt-interactive' 'fix it'" });
+    await expect(
+      prepareAgentLaunch(start("pi", { prompt: "fix it" }), "posix", builtinContext("pi")),
+    ).resolves.toEqual({ command: "'pi' 'fix it'" });
+    await expect(
+      prepareAgentLaunch(start("goose"), "posix", builtinContext("goose")),
+    ).resolves.toEqual({ command: "'goose' 'session'" });
+    await expect(
+      prepareAgentLaunch(start("goose", { prompt: "fix it" }), "posix", builtinContext("goose")),
+    ).resolves.toEqual({
+      command: "'goose' 'run' '--interactive' '--text' 'fix it'",
+    });
+  });
+
   it("builds every builtin resume grammar and derives shared identity from trusted context", async () => {
     await expect(
       prepareAgentLaunch(
--- a/src/core/agent-launch.test.ts
+++ b/src/core/agent-launch.test.ts
@@ -249,6 +249,27 @@ describe("prepareAgentLaunch logical argv", () => {
     ).resolves.toEqual({ command: "'agent' '--prompt' ''" });
   });
 
+  it("preserves a Goose base harness through the trusted session-host path", async () => {
+    const custom = {
+      id: "custom:goose-wrap",
+      launchCmd: "goose-wrap",
+      promptInjectionMode: "argv",
+      baseAgent: "goose",
+    } as const;
+    await expect(
+      prepareAgentLaunch(start(custom.id), "posix", customContext(custom)),
+    ).resolves.toEqual({ command: "'goose-wrap' 'session'" });
+    await expect(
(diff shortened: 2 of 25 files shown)


eneskirca:
Reviewed on a clean worktree off `origin/main`. `npm run typecheck` clean; the touched suites (agents, agent-launch, context-link core + server, noteLink, keybindings, workspace, remote-hooks, plus all of `src/renderer/lib` and canvas-control-core) are green — 1915 tests, no regressions against main. The boundary/guard tests (`no-electron` both shells, `fs-atomic.guard`, `config.capabilities`) pass.

This is careful work. The discipline I most wanted to see is here: the three agents join exactly one capability list, and `config.capabilities.test.ts` pins the *negative* side (`hasHooks`/`canResume`/`hasUsage`/`canChat`/`canReadTitle`/… all false) rather than leaving it to inference. `resolveLinkTranscript` has no locator for them, so it returns `''` and the reader says "no transcript yet" — fails closed, and the PR description says so instead of implying structured transcripts work. `assembleLaunchCommand`/`assembleResumeCommand` stay byte-identical for every existing agent (empty `launchArgs` collapses through `filter(Boolean)`), and both launch paths — the shell assembler and `core/agent-launch.ts` — got the new fields, which is the parity that usually gets missed. The `contextLinkShimPath` change also fixes a real pre-existing bug: SSH endpoints were being handed the *local* shim path.

Three things to address.

**1. `verify` gates on `canContextLink`, not `hasHooks` — and this PR is the first to make those two lists differ.**

Until now every `CONTEXT_LINK_CAPABLE` agent was also in `AGENT_HOOK_TARGETS`, so `verify`'s gate was accidentally sufficient. With agy/pi/goose in the list, `verify --node <agy node>` and `verify --agent goose` both pass the two `canContextLink` checks in `Canvas.tsx` (~9811 / ~9819), then arm the reviewers and the judge through `armAfter`. `depSatisfied` (`renderer/lib/pendingLaunch.ts`) requires `state === 'done'`, and these agents emit no hook events at all, so the state stays unknown forever — deliberately not treated as satisfied. Result: the whole panel sits QUEUED with only the manual ▶ escape, and nothing tells the user why.

Note the `--after` flag path already refuses exactly this, with a named error (`resolveAfter` → `hasHooks`: "not an agent session that reports when it is done"). `verify` should refuse on the same predicate — for the target *and* `--agent` — rather than building a panel that cannot fire.

**2. Unconditional skill installs into two more directories, local and remote.**

`installSkill()` now creates `~/.agents/skills/get-linked-context/` and `~/.gemini/antigravity-cli/skills/get-linked-context/` on every boot, on every machine, whether or not Pi/Goose/AGY are installed — and `RemoteHooks.installContextLink` now does three sequential remote writes per SSH connect instead of one. Two asks: (a) is creating third-party tool directories that don't exist intended, or should the write be conditional on the parent existing? (b) `~/.agents/skills` is a *shared* directory — a skill placed there is discoverable by tools that are not ours and not in `CONTEXT_LINK_CAPABLE`; worth a line in the comment saying that's understood.

**3. Custom-agent launch parity between the two paths.**

`custom-agent.ts` inherits `launchArgs`/`promptLaunchArgs`/`promptFlag` from the base harness, so the shell path gives a `baseAgent: 'goose'` custom agent `goose-wrap run --interactive --text '…'` (nicely covered by your new test). But `resolveTrustedAgentConfig` in `core/agent-launch.ts` returns none of those three for the custom branch, so the session-host path emits the bare command with a defaulted `--prompt`. Either thread them through or say in a comment why the trusted path deliberately doesn't.

**Minor, no change requested:** the one-shot context-link note is idle-gated on `status[selfId]?.state === 'working'`, which is inert for an agent with no hook state — for these three the note can land mid-turn. It's user-initiated on edge draw so the blast radius is small, but a comment noting the gate is a no-op here would be honest.

**What I could not verify:** none of `agy`, `pi`, `goose` are installed on my machine, so every launch grammar (`agy --prompt-interactive`, positional Pi, `goose session` / `goose run --interactive --text`) and both discovery-path claims (`~/.agents/skills`, `~/.gemini/antigravity-cli/skills`) are author-reported here, not reproduced. Your validation notes name versions and say the TUIs were launched and `goose skills list` found the skill, which is the right level of evidence — I'm recording it as device-verified by you, unverified by me.

Verdict: **changes requested**, item 1 being the one that matters. The rest is close to merge-ready.

laulpogan:
Addressed all three requested changes in `073d660a` and rebased onto current `origin/main`.

- `verify` now requires completion hooks for both the target and reviewer. AGY, Pi, Goose, and hookless custom agents fail with a named error instead of remaining queued forever.
- Optional AGY/Pi/Goose skill roots are written only when their parent directories already exist, locally and over SSH. The shared `~/.agents/skills` visibility is documented.
- Session Host custom agents now inherit their base agent’s `launchArgs`, `promptLaunchArgs`, prompt mode, `promptFlag`, and separator. A Goose wrapper test pins `session` and `run --interactive --text`.
- The context-link comment now states that hookless agents can receive the user-triggered note mid-turn.

Verification: 289 focused tests passed (7 platform skips), TypeScript clean, production build passed. The broader shortcut/keybinding set also passed 87 tests after expanding its synthetic key pool for the three new commands.

PR #657, opened 2026-09-02
fix(canvas): a workspace save that stops working now says so, and retries

howieyoungdahl:
A canvas can stop persisting **indefinitely and with no signal**. The user keeps working, the tab looks normal, and nothing reaches disk. I hit this in the field on a Server Edition instance: `.nodeterm/project.json` and `workspace.json` were frozen for **2h40m** while cards were opened and closed, then resumed on their own.

There are two independent defects, both on current `main`.

## The failure kind is "never sent", and that is provable

`workspace-store.ts` writes the index unconditionally near the end of `saveNow`, outside every early-out. So a frozen `workspace.json` proves `saveNow` was never entered — which rules out "the server rejected it" and "the reply was lost". The save never left the renderer.

## Defect A — a conflict latches the debounce off, and only two buttons can unlatch it

`Canvas.tsx`, the debounce effect: `if (!dirty || conflict) return`, with deps `[dirty, conflict, persist, resaveTick]`. The comment above it is accurate — "resolving it (either button clears it) re-arms the save" — and that is exactly the problem: **nothing else clears `conflict`.** If the user never presses Reload or Keep-mine, saves are off for the life of the tab.

Reaching that state does not require a git pull or a second machine. Any server-side write that removes cards and then broadcasts the change is classified as a conflict-while-dirty, and the bar appears. The bar's copy said only *"Project file changed on disk (git pull or another machine)"* — which reads as informational. It does not say that the canvas has stopped saving, so there is no reason to press anything.

## Defect B — one rejected save ends persistence for the tab, permanently

`writeDisk` does `await api.workspace.save(...)` with **no catch**, and is invoked as `void persist()`. On a rejection the throw skips both `setDirty(false)` and the `setResaveTick` nudge, so `dirty` stays true and **no dep of the debounce effect ever changes again** (`persist` is a stable callback). The 800 ms timer is never re-armed. A single transient WS failure is enough, and it is silent.

This is the more dangerous of the two: A at least renders a bar.

## The fix

- New pure `lib/savePersistence.ts`, in the shape `pendingLaunch`/`launchDelivery` already established for this class of problem (bounded backoff to a terminal `failed` state with a Retry), rather than a new idiom.
- `writeDisk` catches, records the failure and re-arms, so a transient rejection self-heals instead of ending persistence.
- A red `SaveFailureBar` for the terminal state.
- The conflict bar now always says the thing that matters: **your canvas is not being saved until you choose.**

## Surfaces

- **Desktop:** affected — `WorkspaceWatcher` (`main/index.ts`) reaches the same path on a git pull. Fixed.
- **Server Edition:** affected, same renderer code. Fixed.
- **Mobile:** N/A — no canvas.

## What I did not do, so you can weigh it

**The reproduction is seam-level, not end-to-end.** I drove the trigger and the consequence against the real modules and pinned the source, with mutation checks on each new test — but I did not stand up a browser and watch a canvas freeze. This repo has no render harness for `Canvas` (its own `canvas-wiring.test.tsx` says as much), and building one was out of scope for a fix this size. If you would rather see an end-to-end reproduction before merging, that is a fair ask.

Three things about the original incident are unknowable after the fact and I am not claiming them: whether the bar was actually on screen, whether `dirty` was true at the freeze boundary, and what ended the freeze. This change adds a log line on path B so the next occurrence answers the first two.

Also honest: canvas-control's "source node is not in exactly one saved project" is downstream of the same cause. Removing the silence and self-healing transport failures makes it far less likely, but it does not make canvas control resolve an unsaved node — that needs its own change.

Gates: `npm run typecheck` clean; renderer + shared suites 5056/5056; targeted files 34/34. The three `src/server`/`src/core` failures on my machine also fail on pristine `origin/main` (verified by stashing).

https://claude.ai/code/session_01G79wE44T1imWesbY83z4UZ

### DIFF
--- a/src/renderer/canvas/Canvas.tsx
+++ b/src/renderer/canvas/Canvas.tsx
@@ -145,6 +145,12 @@ import type { DictationTarget } from '../components/DictationOverlay'
 import { describeOs, REPO_URL } from '../lib/bugReport'
 import { shouldReleasePaneFocus } from '../lib/paneFocus'
 import {
+  autosaveDelay,
+  nextSaveDelivery,
+  type SaveDelivery
+} from '../lib/savePersistence'
+import { SaveFailureBar } from '../components/SaveFailureBar'
+import {
   adoptedNodesNotice,
   decideExternalChange,
   mergeIncomingNodes
@@ -884,6 +890,11 @@ export function Canvas() {
   // give the debounced-autosave effect a dependency that CHANGES in that case — `dirty` stays true
   // throughout, so without it the effect would never re-arm. Rare, so a re-render costs nothing.
   const [resaveTick, setResaveTick] = useState(0)
+  // What the last whole-workspace save did, when it did not land. `undefined` is the ordinary
+  // state. It is a dep of the autosave effect BECAUSE a refusal must re-arm the debounce: `dirty`
+  // never goes false on a failed save, so nothing else in that dep list can change and the timer
+  // would never be scheduled again (the 2026-09-02 silent freeze).
+  const [saveDelivery, setSaveDelivery] = useState<SaveDelivery | undefined>(undefined)
   // The active project's .nodeterm file changed on disk while we have unsaved local edits AND it
   // changed something we also hold (the user must pick a side for that half). `added` counts the
   // nodes that arrived with it and were already adopted onto the canvas — they are never part of
@@ -2196,8 +2207,13 @@ export function Canvas() {
       // persists the conversion, so the next load is a no-op.
       const projects = ws.projects.map(migrateProjectTags)
       useProjects.getState().hydrate({ ...ws, projects })
-      // Upgrade the on-disk format (e.g. v1 -> v2 migration) right away.
-      void api.workspace.save(useProjects.getState().toWorkspace())
+      // Upgrade the on-disk format (e.g. v1 -> v2 migration) right away. Reported like every
+      // other save: a `void` here used to make the very first write of the session the one write
+      // that could fail with no signal at all — including the format migration.
+      api.workspace.save(useProjects.getState().toWorkspace()).catch((err: unknown) => {
+        console.warn('[canvas] initial workspace save failed', err)
+        setSaveDelivery((prev) => nextSaveDelivery(prev, Date.now()))
+      })
     })
     return () => {
       cancelled = true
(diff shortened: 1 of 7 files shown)


eneskirca:
Reviewed on an isolated worktree at `pull/657/head`, rebuilt against current `origin/main`.

**Verdict: ready to merge.** The mechanism is right and the gates reproduce here: `tsc -p tsconfig.web.json` clean, renderer + shared **5057/5057** green, targeted suites 77/77. Three notes, none blocking.

### The fix itself holds up

Defect B is real and is the dangerous one. `writeDisk` now catches, records, and — because `saveDelivery` is a dep of the debounce effect — re-arms at the backoff delay. Two things I checked specifically because they are where this shape usually goes wrong:

- **The retry is genuinely bounded.** `saveRetryDelay(6)` is `undefined` ⇒ `failed` ⇒ `autosaveDelay` returns `null`. A whole-canvas last-writer-wins write is exactly the thing that must not be retried forever, and it isn't.
- **The terminal `failed` state self-heals on any explicit action.** There are ~7 direct `void persist()` call sites (project add/close, ⌘K Save, Keep-mine), all of which run through the same `writeDisk` and clear `saveDelivery` on success. So a user who never presses **Retry now** still recovers on their next project switch. That is worth stating in the PR body — it materially lowers the cost of the bounded schedule.

The conflict gate is unchanged (`autosaveDelay(true, true, …) === null`), so Reload / Keep-mine behave exactly as before. `onKeepMine`'s `void persist()` is now covered by the catch rather than being another bare `void`.

### 1. "only two buttons can unlatch it" is not accurate — and the truth fits your incident better

There is a third clear path, `Canvas.tsx` ~2696:

```ts
useEffect(() => { setConflict(null) }, [activeProjectId])
```

A project switch drops the conflict, which re-arms the save. That doesn't weaken defect A — a single-project Server Edition tab has nowhere to switch to — but it is the one mechanism that explains the part of the field report you correctly declined to claim: **a freeze that ends on its own after 2h40m is what a project switch clearing a latched conflict looks like.** Please correct the sentence and add the switch as a third unlatch; the next reader will otherwise go looking for a path that is already there.

### 2. "a frozen `workspace.json` proves `saveNow` was never entered" is one step too strong

`writeAtomic(this.indexPath, …)` is unconditional, but it is not unreachable-by-throw: `writeDataFile`, `reconcileSsh` and `mirrorSshCache` all `await` above it, and an exception from any of them skips the index write *and* rejects the promise. So a frozen index is equally consistent with "entered and threw", which is still defect B — the conclusion survives, the proof does not. "Consistent with never-sent, and in either case the rejection was discarded" is the claim the evidence supports.

### 3. The dep-array literal is the brittlest of the source-text assertions

```ts
expect(SRC).toContain('}, [dirty, conflict, persist, resaveTick, saveDelivery])')
```

goes red on any future dep addition or a prettier reflow, and the failure will read as "the fix was removed" when it wasn't. The repo does have precedent for source-level pins (`hook-verified-parity.test.ts`), so I am not asking you to drop them — but this one asserts formatting, not the invariant. Matching just `saveDelivery]` inside the effect's dep list keeps the guard and survives an added dep.

The seam-level-not-end-to-end caveat is the right call to have made, and stated in the right place.

PR #659, opened 2026-09-03
Add a messaging default for new projects

laulpogan:
## Summary

- add a machine-local setting that enables messaging for newly created projects
- keep existing projects unchanged and preserve clone confirmation
- restore local tmux pane ownership across app restarts with a signed, generation-bound proof
- keep unsupported or unverifiable warm sessions fail-closed

Existing local panes created before this upgrade need one terminal-session restart to establish a proof.

## Verification

- npm run typecheck
- 153 focused tests across settings, project creation, messaging ownership, and terminal persistence
- npm run build
- built Electron app exercised with Playwright: setting toggled successfully, no console or network errors
- independent correctness and security review: no remaining major findings

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -417,8 +417,11 @@ project's nodes only.** The contract:
 
 ## Terminal session continuity (tmux)
 
-`src/core/pty-manager.ts` runs each terminal inside a persistent tmux session
-(`tmux new-session -A -D -s nt-<nodeId>`) on a dedicated socket (`-L node-terminal`) with
+`src/core/pty-manager.ts` runs each terminal inside a persistent tmux session on a dedicated socket
+(`-L node-terminal`). Ownership-sensitive local creates use a detached create-only command that
+returns the tmux generation, then attach a painter client; warm opens use attach-only. Each side
+recovers once if the session appears or disappears across the asynchronous probe. Other tmux paths
+still use attach-or-create where no ownership proof is minted. All paths use
 a generated config (`-f <userData>/tmux.conf`, so the user's `~/.tmux.conf` never
 interferes; status bar off, **mouse on**, 50k history, `set-clipboard on` + `terminal-features
 ",*:clipboard"`, and the copy-mode mouse bindings). Because the tmux *server* outlives the app,
--- a/src/core/agents/agent-message-decide.ts
+++ b/src/core/agents/agent-message-decide.ts
@@ -39,9 +39,9 @@ export type NotPermittedReason =
    *  the per-project grant cannot be attributed (PR #237 review I-1); its own word because the
    *  human's fix is de-duplicating ids, not moving nodes. See `agent-message-scope.ts`. */
   | 'ambiguous-target-node-id'
-  /** No RUNTIME proof of which project spawned the target pane: the ledger has no entry (never
-   *  spawned this run, or only re-attached after a restart), or its owner disagrees with the sole
-   *  store claimant (a project.json listing a pane it did not spawn). The store's node-set is
+  /** No proof of which project spawned the target pane: the ledger has neither a fresh-spawn fact
+   *  nor a valid generation-bound restart proof, or its owner disagrees with the sole store
+   *  claimant (a project.json listing a pane it did not spawn). The store's node-set is
    *  attacker-writable, so ownership is proven at spawn, not read from the file (PR #237 fix
    *  round 2 — the confused deputy driven end-to-end); unprovable ⇒ refuse. Not retryable: the
    *  pane must be freshly (re)spawned by its real owner before it can be messaged. */
--- a/src/core/agents/agent-messaging.ts
+++ b/src/core/agents/agent-messaging.ts
@@ -80,8 +80,9 @@ export interface AgentMessagingDeps {
    */
   messagingEnabled(projectId: string): boolean
   /**
-   * The project that PROVABLY spawned the target node's pane this run, or `undefined` when
-   * unproven (runtime ledger, `core/agents/pane-ownership.ts`). The delivery gate trusts THIS,
+   * The project that PROVABLY spawned the target node's pane, either this run or through a signed
+   * machine-local restart proof; `undefined` when unproven (`core/agents/pane-ownership.ts`). The
+   * delivery gate trusts THIS,
    * not the persisted store's node-set, to decide whose grant applies — the store is
    * attacker-writable (`project.json` lists any node id) and cannot tell a real owner from a
    * project that merely listed a live pane it never spawned (PR #237 fix round 2). Undefined ⇒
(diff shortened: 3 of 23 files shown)


eneskirca:
Reviewed on a clean worktree off `origin/main`. `npm run typecheck` clean. The suites you named are green (51 files / 783 tests, including `pane-ownership-persistence`, `pty-single-user`, `pty-session-host`, `projects`, `default-settings`, and the `fs-atomic` / `no-electron` guards).

The crypto half is good, and I want to be specific about why rather than wave at it. `nt-pane-owner-v1|` is properly domain-separated from `nt-node-auth-v1|`, which is the extension `node-auth-token.ts` explicitly anticipated ("the same secret can later mint other capability CLASSES"), so an agent holding its own node token cannot repurpose it as an ownership proof. The generation tuple `#{pid}|#{session_id}|#{session_created}|#{pane_id}|#{pane_pid}` is entirely tmux-server state — nothing a pane's shell can set, unlike the tmux session-env variable the old comment rightly ruled out — so a replacement session under the same name cannot inherit authorization, and the strict regex means a vanished target that makes tmux exit 0 with empty fields resolves to `undefined` rather than to a match. The new tests exercise the forgeries directly (hand-edited `projectId`, wrong secret, wrong generation, `../` ids, null secret), and `restorePaneOwner` requires the renderer's claim *and* the signed record *and* the live generation to agree — none of the three is sufficient alone. Worth noting for the record that on desktop the root secret is sealed through the platform, and the shell that stores it as a raw 0600 file (Server Edition) is the one where messaging never runs, so the disk proof does not widen the same-uid surface in the shell that matters.

Four things.

**1. Blocking: two tests that pass on `origin/main` fail on this branch.**

`src/core/pty-project-settings.test.ts`:
- `project settings at the spawn — LOCAL leg > adds the project keys to tmux update-environment so a shared server copies them`
- `project settings at the spawn — SHELL precedence > feeds the project's shell to the tmux session program too`

Confirmed green on `origin/main`, red here (`expected "/bin/fish", received "nt-node-1"`). Cause is the design change, not a flake: with `ownerProjectId` present the local tmux path no longer node-pty-spawns `new-session`, so `spawns[0]` is now `attach-session -t nt-node-1`. As far as I can tell the *feature* still works — the project `-e` keys and the shell ride the `execFileSync` create command — but that means the regression test for the project-shell/project-env contract is now pointed at the wrong process. Please move the assertions onto the create call rather than deleting them; losing coverage of "the project's shell actually reaches the session" would be the worse outcome. Your validation ran 153 focused tests, which is what let these through.

**2. `new-session -A` is gone from the default local path. What happens on the race?**

`localTmuxMode` is set whenever `ownerProjectId` is present, i.e. for essentially every local project node, so this is the main path, not an edge. Create-mode drops `-A` (create-only) and attach-mode is attach-only. `-A` previously absorbed a wrong `has-session` answer in *both* directions; now a session that appears between the probe and the create fails the whole open with "Could not create the terminal session atomically", and one that disappears before the attach fails the attach. What is the intended behaviour when two windows, a co-attach, or the reaper race that window — and should the create fall back to an attach when tmux reports the session already exists? A hardening that turns a recoverable state into a dead terminal on the default path needs that answered explicitly.

**3. `execFileSync` on the create path has no timeout.**

It is consistent with this file's existing style, so not a blocker — but the existing synchronous calls are `source-file` / `show-options` / `set-option`, and this one is `new-session`, which starts the user's shell and its rc files, synchronously on the main process, on every fresh terminal open. A timeout at minimum, and a note on the measured cost, would be good.

**4. The seeded `capabilityAck: 'kept'` answers a question that has not been asked.**

I verified the claim in the comment for the folder path: `openOrAdoptFolder` probes and routes an existing `.nodeterm/project.json` to `adoptProject`, which never reaches this factory — so a cloned project file does still raise the notice. Good.

But `project-capability-consent.ts` states the invariant this is in tension with: the ack exists because "a switch that arrives from a clone is not the same thing as a switch the user set", and a `declined` answer deliberately re-raises a *new* notice when a hostile `true` re-arrives via `git checkout`. Pre-writing `'kept'` at creation converts the ack from "the user saw this file's switch and kept it" into "the user once set a machine-wide preference", and a `true` arriving later then meets a standing `kept` and grants in silence — C1 reached from the other direction. Two concrete ways in:

- `openSshProject` → `addProject` → `createProject` does **not** probe the remote `.nodeterm/project.json` first (unlike the folder path). A fresh SSH project pointed at a remote folder that already holds someone else's project file gets its ack pre-answered before `reconcileSsh` pulls that file.
- Locally: a virgin folder gets `kept` at creation; a later `git checkout`/pull that introduces a `project.json` with `agentMessaging: true` inherits the standing answer and the notice never fires.

Both need the machine default switched on, and pane ownership still gates delivery, so this is not a cross-project escalation — but it is silent consent for a file the user never saw. Suggestion: bind the seeded answer to the fact that *this machine authored the file*, or re-arm the ack when the project's lineage changes (adoption / SSH reconcile re-key), rather than writing a blanket `kept` up front.

**Minor:** `pane-owners/<sha>.json` is only removed in `destroySession`, so proofs for node ids that no longer exist accumulate in the data dir. They fail closed on generation mismatch, so it is hygiene rather than security, but a sweep alongside the existing token sweeps would be tidy.

**Also:** the ledger comment in `pane-ownership.ts` and `CLAUDE.md`'s tmux section both describe the old contract (`tmux new-session -A -D -s nt-<nodeId>` as one atomic command). Since the create path is now two steps, the CLAUDE.md line should move with it in this PR.

Verdict: **changes requested** — item 1 is a hard blocker, item 2 needs an answer before this lands on the default local path, and item 4 is the one I'd most like your view on. Product call on the default itself (which ships OFF here, correctly) is @eneskirca's.

laulpogan:
Addressed the review in `af63ae8c` and rebased onto current `origin/main`.

- The two project-settings regressions now assert against the synchronous tmux create call. They still pin the project env value, argv secrecy, project shell, socket, and session name.
- The local tmux path now recovers once in both race directions: create-only falls back to a warm attach only after a strict generation probe; a vanished warm attach is discarded and retried as a cold create. The create-race result is `fresh: false`, so it cannot mint pane ownership.
- Synchronous `new-session` and its race confirmation have a 6-second timeout. On tmux 3.6a, 20 detached `/bin/sh` creates measured 7 ms median, 13 ms p95, 18 ms max.
- Global-default consent now carries machine-local `local-default` provenance. A watcher-confirmed local checkout/pull or external SSH reconcile re-arms only that seeded answer, persists the change, and raises the normal notice. Explicit per-project answers remain intact. This covers both concrete paths from the review.
- Updated the tmux continuity and pane-ownership documentation for the two-step contract.

I left stale proof cleanup out: it remains fail-closed hygiene and is separate from these correctness/security fixes.

Verification: 267 focused tests passed, TypeScript clean, production build passed, and a fresh read-only security review found no blocking issue.

PR #664, opened 2026-09-03
fix(canvas-control): an agent opening a node no longer moves the user's view

basmilius:
## What's wrong

I work in several projects at once. My skills render their output as HTML in a webview, so a report, a PR review or a slide deck arrives as a node on the canvas. Every time one of those opened, nodeterm threw me into that project. I would be typing in a terminal in project 1, an agent in project 4 would finish, and suddenly I was in project 4.

## Why

Canvas control routes by source node, and React Flow only holds the active project's nodes. So when an agent in a background project calls a verb, its node is not on the live canvas: `routeControlSource` resolves the owning project and, for anything `needsLiveCanvas` says yes to, travels there first.

That travel exists to give the verb a canvas to act on. Only five verbs were exempt (`list`, `send`, `reply`, `sticky`, `open-project`), each for exactly this reason. The rule was never extended to the verbs that create a node, and those are the ones an agent reaches for when it has something to show you.

## What changed

A verb whose whole effect is putting a node on the owning project's canvas does not need that canvas on screen. It needs a place to put a node, and the project's serialized nodes are one. That is already how the `--project` cold-open branch writes into another project's store.

`OFF_CANVAS_VERBS` (`show-image`, `show-video`, `show-web`, `open-browser`, `open-terminal`, `open-claude`, `open-agent`) now run against a staged copy of the owning project's nodes and commit through `applyNodeMutation` plus a new `appendProjectEdges` for the ropes and context bridges. `commitCanvas` replaces both arrays wholesale from the live canvas, so it cannot serve a project React Flow does not hold.

You get a notice instead of a trip: `An agent opened a node in "api". That project is not on screen.` It stays up until you dismiss it, and its button takes you to the node.

## Rules worth reviewing

**The handler's verb body reads and writes through shims** (`readNodes` / `applyNodes` / `addRopeEdges` / `addBridgeEdges` / `readLinkEdges` / `touchCanvas`), which are pass-throughs on the on-screen path. A verb reaching for `nodesRef.current` or `setNodes` compiles, passes every other test, and quietly answers about the wrong project. `canvas-wiring.test.tsx` pins their absence at source level because nothing else can see it. That pin earned its keep during the rebase: the `color` verb landed upstream meanwhile, written the old way, and is now routed through the shims too.

**The acting project is the source's, not the active one.** `ctlProject` read `activeProjectId`, which was only correct because the travel had just made the two the same project. Off canvas it decides another project's ssh flag, browser session key, managed account, permission mode and launch override.

**A node created off canvas is re-armed with `armForColdOpen`.** `initialCommand` is never serialized, so a launch left there is silently dropped and the session never starts. The reply corrects `queued` and `queuedIds` to the whole batch and carries `offCanvas: true` beside the sentence, because an orchestrator must not report a session as started when its `--after` deps merely happened to be satisfied. A refusal commits nothing.

**A worktree-bound `--group` still travels.** `cwdForNewNodeIn` subtracts the worktree store's `staleGroupIds`, which is scoped to the active project. Off canvas that subtraction cannot be made, and failing open would hand the new node a checkout that was deleted outside the app.

Everything that reads or drives something live still travels: `browser` needs a mounted webview guest, `write` and `close` need a running pane, `focus` and `goto` move a camera that belongs to the active project by definition, and the layout verbs are refinements a human is watching.

## Smaller things in here

`.announce-banner` was `12.5px`. Half pixels are not a size this app uses, so it is `13px` now, which touches all five banners. The notice's button uses the app's default button rather than the accent treatment `.announce-banner__btn` gives an install or a reload; `.announce-banner__action` joins `.confirm__btn`'s selector instead of restating the recipe, so the two cannot drift.

## Tests

- `renderer/lib/offCanvasControl.test.ts` covers which verbs qualify, the worktree refusal including a nested frame and a parent cycle, and the notice copy.
- `renderer/state/projects.mutation.test.ts` covers `appendProjectEdges`, including that a retried call is idempotent.
- `renderer/canvas/canvas-wiring.test.tsx` pins the shims, the acting project, the cold-open re-arm, the sticky notice and the travel target.

## Surfaces

Desktop and Server Edition are identical. This is renderer code plus the projects store, with no new IPC or bridge member. Mobile is not applicable: it attaches to tmux sessions over the transport protocol and has no canvas.

## How to try it

Open two projects, A and B, with a terminal node in B. From B's terminal:

```sh
(sleep 15; sh "<userData>/canvas-control/nodeterm.sh" show-web --url https://example.com) &
```

Switch to A and keep working. You stay in A, the notice names B, and the node is waiting there when you go.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1861,6 +1861,47 @@ else, and its context links must keep classifying across restarts).
   so reporting it as "the error" would be a confident wrong fact. Reading the text, and the
   *failed-to-start* watchdog (a station that never emits ANY hook event — the opposite failure,
   which hangs dependents honestly rather than firing them wrongly), stay open.
+  **Opening a node never moves the human's view (`lib/offCanvasControl`, 2026-09).** Routing is by
+  SOURCE, and React Flow holds only the ACTIVE project's nodes, so a verb declared
+  `needsLiveCanvas` used to TRAVEL to the calling agent's project before it ran — which meant a
+  background agent rendering its output as a node (`show-web` for a report, `show-image` for a
+  screenshot, `open-claude` for a fan-out) yanked the user out of the project they were typing in,
+  at a moment they did not choose. Same G5 objection `send`/`reply`/`sticky`/`open-project` are
+  already exempt for; those verbs simply never needed a canvas, while these needed a PLACE TO PUT A
+  NODE — which the owning project's SERIALIZED nodes can be, exactly as the `--project` cold-open
+  branch already writes into another project's store. `OFF_CANVAS_VERBS` (`show-image`,
+  `show-video`, `show-web`, `open-browser`, `open-terminal`, `open-claude`, `open-agent`) are now
+  answered against a staged copy of that project's nodes, committed through
+  `applyNodeMutation` + the new `appendProjectEdges` (ropes and bridges — `commitCanvas` replaces
+  both arrays wholesale from the live canvas and cannot serve a project React Flow does not hold),
+  and the human gets a NOTICE naming the project with a "Go there" button. Rules a refactor must
+  not undo:
+  - **The handler's verb body reads and writes through the shims** (`readNodes` / `applyNodes` /
+    `addRopeEdges` / `addBridgeEdges` / `readLinkEdges` / `touchCanvas`), which are pass-throughs
+    on the on-screen path. A new verb reaching for `nodesRef.current` or `setNodes` compiles,
+    passes every other test, and quietly reads or writes the WRONG project;
+    `canvas-wiring.test.tsx` pins the absence at source level because nothing else can see it.
+  - **The acting project is the SOURCE's, not the active one.** `ctlProject` used to read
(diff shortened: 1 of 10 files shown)


eneskirca:
Closing as already fixed — but thank you for this, the diagnosis was exactly right and it was reported independently by the maintainer the same day.

The fix landed in #665 (merged 2026-09-03). Same root cause you identified: `routeControlSource` resolves the owning project and Canvas travelled there because React Flow only holds the active project's nodes. The shipped fix reuses the cold-open path that the `--project` targeted opens already use — for `open-terminal` / `open-claude` / `open-agent` whose source project is open but not active, the node is written into that project's serialized store (`armForColdOpen` + `applyNodeMutation` + `writeDisk`) instead of travelling; every other verb still travels because it reads live state the serialized copy does not carry. A closed project is written to without reopening its tab.

Two things from your PR that are worth checking against what shipped, and if they are missing we will take them: your `offCanvasControl.ts` split, and anything in your handling of ropes/links for a cold-opened node. If you would like to re-open this as a follow-up on top of #665 for those, please do.

PR #674, opened 2026-09-05
feat(packaging): ship a .rpm for Fedora/RHEL

Almighty-Shogun:
This PR adds a `.rpm` to the Linux release, so Fedora and RHEL get a package they can install
instead of a `.deb` they cannot and an AppImage that needs a library Fedora no longer ships by
default. It also corrects two places that named the Linux packages wrong, and writes down what a
Fedora checkout needs before it can build at all.

**Why the release job changes with the target.** electron-builder bundles fpm but shells out to the
system `rpmbuild`, so the rpm target dies on a bare runner with "Need executable 'rpmbuild' to
convert dir to rpm". The install step and the `dist/*.rpm` upload glob land in the same commit as
the target, so no release can be cut from a tree whose job builds a target it has no tool for.

**The package is called `node-terminal`, not `nodeterm`.** electron-builder derives the deb and rpm
package name from package.json's `name`, not `productName` (only an `@scope/...` name falls back to
the product name), so the artifacts are `node-terminal_0.3.4_amd64.deb` and
`node-terminal-0.3.4.x86_64.rpm`, while the AppImage and the `/opt/nodeterm` install prefix use the
product name. Two existing references assumed otherwise and therefore never matched anything: the
README's `sudo apt install ./nodeterm-*.deb` line, and `uninstall.sh`'s `dpkg -s nodeterm` probe,
which meant every `.deb` user reached the end of the uninstall script without being told the
package was still on their machine. Both now say `node-terminal`, and the probe covers rpm.

**Rejected: renaming the package to match the app.** `linux.packageName: "nodeterm"` would have made
the existing text correct instead, but apt and dnf would then treat it as a different package from
the one already shipped, turning an upgrade into a rename for current `.deb` users. Documenting the
real name is cheaper than that.

**Rejected for now: the static AppImage runtime.** The default runtime is still the FUSE2 one, so on
a stock Fedora install the AppImage exits with a `libfuse.so.2` error until `fuse-libs` is present.
`toolsets.appimage: "1.0.3"` drops that requirement, but it is upstream-flagged beta and the static
runtime stops passing the `--no-sandbox` launcher argument the FUSE2 path adds, which is a
behaviour change for every Linux user rather than a Fedora fix. The README names the one-line
workaround instead.

**Building on Fedora needs `CFLAGS=-D_GNU_SOURCE`.** smart-whisper's vendored
`whisper.cpp/ggml/src/ggml.c` calls `CPU_ZERO`, `CPU_SET_S`, `pthread_getaffinity_np` and `getcpu`
without ever defining `_GNU_SOURCE` (upstream ggml gets it from its CMake build, which node-gyp does
not use), and GCC 14 promoted implicit declarations from a warning to an error. A bare `npm install`
therefore fails inside smart-whisper's own install script, before this repo's `postinstall` runs, so
nothing here can patch around it. `libxcrypt-compat` is needed too: fpm's bundled Ruby links
`libcrypt.so.1`, which Fedora's glibc dropped, and without it both the deb and the rpm target fail.

**Measured on Fedora 44** (GCC 16.2.1, rpm 6.0.2, Node 24.18.1): `CFLAGS=-D_GNU_SOURCE npm install`
builds node-pty and smart-whisper clean against Electron 42's ABI; `npm run dist:linux` produces all
three artifacts (AppImage 176 MB, deb 135 MB, rpm 111 MB); `dnf install` of the rpm resolves with
nothing extra pulled in, so the default `Requires` set (gtk3, libnotify, nss, libXScrnSaver,
`(libXtst or libXtst6)`, xdg-utils, at-spi2-core, `(libuuid or libuuid1)`) is already satisfiable
there; the postinstall scriptlet correctly leaves `chrome-sandbox` at 0755 on a userns kernel; and
the installed app launches and renders.

**Not verified:** the rpm leg on the `ubuntu-latest` runner, which stays unexercised until a tag is
cut; any RHEL, openSUSE or Fedora-derivative host; the full `npm test` suite (typecheck is clean and
the packaging-adjacent suites pass); and the AppImage on a Fedora box with `fuse-libs` installed.
Auto-update needed no code change, since `isManualUpdatePlatform` already treats a Linux install
without `APPIMAGE` in the environment as manual-download, which is exactly what an rpm install is,
but that path was read rather than exercised.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -144,7 +144,7 @@ jobs:
 
   # Linux build: unsigned, so it runs on a plain GitHub runner (fast + free tier), in
   # parallel with the self-hosted mac job — it must never queue behind notarization.
-  # Publishes AppImage + deb + latest-linux.yml to the SAME tag's GitHub Release; the
+  # Publishes AppImage + deb + rpm + latest-linux.yml to the SAME tag's GitHub Release; the
   # nodeterm.dev update feed serves latest-linux.yml alongside the mac files.
   release-linux:
     runs-on: ubuntu-latest
@@ -155,6 +155,12 @@ jobs:
         with:
           node-version: 22
           cache: npm
+
+      # electron-builder ships its own fpm, but the rpm target shells out to the system
+      # rpmbuild ("Need executable 'rpmbuild' to convert dir to rpm"). Installed explicitly
+      # rather than relying on whatever the runner image happens to carry.
+      - name: Install rpmbuild
+        run: sudo apt-get update && sudo apt-get install -y rpm
 
       - name: Install dependencies
         run: npm ci
@@ -177,6 +183,7 @@ jobs:
           files: |
             dist/*.AppImage
             dist/*.deb
+            dist/*.rpm
             dist/latest-linux.yml
 
   # Windows build: UNSIGNED BETA (#454), on a plain GitHub runner. Publishes the NSIS
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -3225,6 +3225,43 @@ the same script hand-packs `build/icon.icns` (size-checked frames — issue #369
 zip, `--publish never`). Production release signing/notarization and the update-feed hosting are
 handled outside this repo.
 
+**Linux ships AppImage + deb + rpm**, all unsigned, all built by `release-linux` on a plain
+`ubuntu-latest` runner (`npm run dist:linux` locally). Three things about it are easy to get wrong:
+
+- **The packages are named `node-terminal`, the AppImage is named `nodeterm`.** electron-builder's
+  `linuxPackageName` is package.json's `name`, not `productName` (it only falls back to the product
+  name for an `@scope/…` name), so the artifacts are `node-terminal_<version>_amd64.deb`,
+  `node-terminal-<version>.x86_64.rpm` and `nodeterm-<version>.AppImage`, the launcher is
+  `/usr/bin/node-terminal`, and the install prefix is `/opt/nodeterm` (that one IS the productName).
+  Anything that matches on the package name (`scripts/uninstall.sh`'s `dpkg -s` / `rpm -q` probes,
+  the README's install lines) must say `node-terminal` or it silently never fires. Renaming the
+  package to match the app would strand existing `.deb` installs on a package apt no longer tracks,
+  which is why the mismatch is documented rather than fixed.
+- **The rpm target shells out to the system `rpmbuild`.** electron-builder bundles fpm, but not
+  rpmbuild, so `release-linux` installs it explicitly (`apt-get install -y rpm`); without it the
+  target dies with "Need executable 'rpmbuild' to convert dir to rpm". The default `Requires` set
+  electron-builder emits (gtk3, libnotify, nss, libXScrnSaver, `(libXtst or libXtst6)`, xdg-utils,
+  at-spi2-core, `(libuuid or libuuid1)`) resolves on Fedora 44 with nothing extra pulled in;
+  verified by installing the built rpm, which also proved rpm 6.0.2 accepts fpm's spec.
+- **Auto-update is already correct for rpm and needs no work**: `isManualUpdatePlatform`
+  (src/shared/update-platform.ts) keys off the absence of `APPIMAGE` in the environment, so a deb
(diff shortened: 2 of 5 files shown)


eneskirca:
**Verdict: MERGE-READY — with one release-workflow decision for @eneskirca (below).**

I re-measured the load-bearing claims on a Linux box rather than taking them on trust, and they hold:

- **`name` vs `productName`.** `package.json` really is `name: "node-terminal"`, `build.productName: "nodeterm"`. So `dpkg -s nodeterm` in `scripts/uninstall.sh` and `sudo apt install ./nodeterm-*.deb` in the README could never have matched anything. That is a real, currently-shipping bug being fixed here, independent of the rpm work — every `.deb` user reached the end of the uninstall script without being told the package was still installed. Your reasoning for documenting the mismatch rather than setting `linux.packageName` is also right: renaming strands existing installs on a package apt no longer tracks.
- **`apt-get install -y rpm` really does provide `rpmbuild`.** Checked the Ubuntu 24.04 package contents directly: `rpm 4.18.2+dfsg-2.1build2` ships `./usr/bin/rpmbuild`. The install step is correct and correctly placed *before* `npm ci`.
- **Auto-update needs no change, and for the reason you give.** `isManualUpdatePlatform(platform, hasAppImage)` is `platform === 'linux' && !hasAppImage`, and its docstring already names `.rpm` explicitly. An rpm install has no `APPIMAGE` in the environment, so it lands on the manual-download card. Nothing to do — verified by reading, same as you.
- **Target order is safe and matters.** `rpm` is appended *after* `AppImage` in `build.linux.target`, so the AppImage stays first and `latest-linux.yml` keeps pointing at it. `deb` has been in that list for a long time without disturbing the feed, which is the strongest available evidence that a third non-AppImage target will not either. Worth a one-line comment in `package.json` or CLAUDE.md that the **order is load-bearing**, so nobody tidies the array alphabetically later.

`bash -n scripts/uninstall.sh` is clean and no test pins the package-name probes. The `elif` chain is fine (a box with both `dpkg` and `rpm` is vanishingly rare, and dpkg-first is the right precedence). The CLAUDE.md and README additions are the honest kind — they say what was measured, on what, and name the two "rejected for now" options with their actual costs rather than hand-waving them.

### The decision for the maintainer

`release-linux` currently has **no Linux packaging smoke**, so the rpm leg is first exercised *at tag time*. If `rpmbuild` or fpm misbehaves on `ubuntu-latest`, the failure lands mid-release, in the job that also publishes the AppImage and `latest-linux.yml` — the worst possible moment. @eneskirca, three options in increasing cost:

1. **Merge and accept the risk.** The mitigation is cheap (rerun the job), and the first failed release is recoverable.
2. **Run `release.yml` once against a throwaway pre-release tag** before the next real one.
3. **Add a `linux-package-smoke.yml`** — `workflow_dispatch`-only, build-and-never-publish — mirroring the existing `win-package-smoke.yml`. That is this repo's own established pattern for exactly this situation, and it is what I would pick: it makes the rpm leg checkable on demand forever, not just once.

Nothing about that is a change request on this PR — it is a release-process call, and the PR is honest that the runner leg is unverified ("stays unexercised until a tag is cut"). The RHEL/openSUSE/derivative hosts are likewise stated as unverified, which is the right way to leave them.

Nice piece of work: the Fedora build notes (`CFLAGS=-D_GNU_SOURCE` for ggml's unguarded GNU affinity calls under GCC ≥14, `libxcrypt-compat` for fpm's Ruby needing `libcrypt.so.1`) are the kind of thing that costs a newcomer an evening and is nowhere findable. Good that it is written down.

Almighty-Shogun:
Thanks for re-measuring those rather than taking them on trust.

Added the target-order note in `docs(claude): record that the Linux target order is load-bearing`. It says AppImage stays first because electron-builder writes the update feed from the first publishable target, that `deb` and now `rpm` sit behind it for that reason, and that `package.json` is JSON and can't carry the comment, so the rule lives in CLAUDE.md. It ends with a plain "do not alphabetize or otherwise re-sort that array".

On the release-workflow question, that's yours to make and I'm happy with any of the three. For what it's worth I'd also pick the smoke workflow, since `win-package-smoke.yml` already establishes the pattern and it makes the rpm leg checkable on demand rather than once. Say the word if you want me to add it in this PR, otherwise I'll leave it as it stands.

No merge needed here, the branch is still clean against `main`.

Issue #547, opened 2026-08-31
Subagent cards vanish when a new prompt starts a turn while the subagent is still running


ezwep:
nodeterm 0.3.4, macOS desktop.

### What happens

A terminal node spawns background subagents. While they are still running, I send a new
prompt in that same node. Every subagent card for that node disappears immediately — and
permanently for the rest of the run, while the agents keep working.

At the moment the prompt is submitted the footer still reads `Waiting for 2 background
agents to finish`, and the two status rows under the prompt keep advancing — one of them
climbing from 132.3k tokens onward over the following ten seconds. The card is gone.

Tying the vanished card to a still-running agent: the card stood at 1m 2s elapsed and named
the same task as one of the two status rows, which stood at 1m 1s; the other row was at
3m 51s. That row then kept running and kept reporting progress after the card was gone. The
correspondence is by elapsed time and task subject — I cannot see the internal ids from the
UI, so that identification is inference, not proof.

### Root cause

`src/renderer/canvas/Canvas.tsx:10077`:

    if (e.newTurn) an.clearForParent(e.nodeId) // genuine new turn -> drop the previous fan-out

`src/renderer/state/agentNodes.ts:183` then deletes every entry for the parent, with no
regard for whether it finished:

    clearForParent: (parentNodeId) =>
      set((s) => {
        const ids = Object.keys(s.byId).filter((id) => s.byId[id].parentNodeId === parentNodeId)

Both verified in the shipped `app.asar`, not only in main.

Because `start()` only fires from live hook events and a subagent already past its
`PreToolUse` emits no second one, nothing rehydrates `byId` — so the card is gone for the
rest of the run. Same permanent loss as #402, different trigger.

### Why the #402 fix doesn't cover this

The fix for #402 reasoned that "an unmount says the component went away, not that the work
did", and stopped unmount from clearing. It left `newTurn` as an owner of clearing, on the
assumption that a new turn means the previous fan-out is over. That holds only if every
subagent from the previous turn has finished — and with background agents it does not,
since the entire point of `Waiting for N background agents to finish` is that they outlive
the turn that spawned them.

One sentence from that commit applies verbatim here: long-running subagents — exactly the
ones a card is for — are the ones lost.

### On "the previous fan-out is stale by definition"

That is true of a *finished* card and false of a working one, and the store already draws
exactly that line: `SubagentViz.state` is `'working' | 'done'`, `finish()` sets it and is
[...]

Issue #402, opened 2026-08-24
Subagent cards vanish when the parent terminal unmounts (project switch) while the agent is still running


ezwep:
## Summary

Subagent cards — the small "working / <task> / <elapsed>" nodes roped to their parent terminal — disappear permanently when the parent terminal node unmounts, e.g. on switching projects. The subagent itself keeps running, and the card never comes back.

## What I see

Before: the parent session shows `gsd-research-synthesizer (Synthesize research) — Backgrounded agent`, with its card on the canvas roped to the node: *working · Synthesize research · 18s*.

After switching away and back: the card is gone, while the same session still prints

```
* Waiting for 1 background agent to finish
> ○ gsd-research-synthesizer   Retrying SUMMARY.md write … 4m 4s · ↓ 105.2k tokens
```

So the agent is alive and progressing — 4 minutes in, 105k tokens — with nothing on the canvas to show it. The only remaining trace is inside the terminal's own output.

## Expected

The card stays on the canvas, roped to its parent, for as long as the subagent is running — a project switch should not affect it. It should disappear when the work is actually over: the card reaching `done` and being cleared by the lifecycle events that already handle that (`newTurn`, `sessionPhase === "end"`), or the parent node being deleted.

The rope matters as much as the card: it is what tells me *which* session spawned the agent. A card without its parent visible would be only half the fix.

## Root cause

The terminal node component clears every subagent card for itself on unmount:

```js
reactExports.useEffect(
  () => () => {
    useAgentNodes.getState().clearForParent(id2);
  },
  [id2]
);
```

`clearForParent` deletes the cards from `byId` along with their `activityById`, positions, sizes and expanded state. Switching projects unmounts the previous project's nodes, so that cleanup runs even though nothing about the subagent has ended.

Nothing can restore them afterwards. `byId` is only ever written by `start(toolUseId, viz)` and `finish(toolUseId, result)`, which come from live hook events — so a subagent that already started emits no further `start`, and the card stays gone for the rest of its run. There is no rehydration path: `saveLoopOverrides` persists to `localStorage` only for ids beginning with `loop-`, and even that covers position/size/expanded, never `byId` itself.

The other two `clearForParent` call sites look correct and I would not touch them — `e.newTurn` (a new turn supersedes the previous turn's subagents) and `sessionPhase === "end"`. Those are lifecycle events. Unmount is not: it says the component went away, not that the work did.

## Why it is hard to reproduce

[...]

PR #682, opened 2026-09-05
feat(agents): optionally hide a subagent's card when it finishes

Almighty-Shogun:
This PR adds an opt-in setting, off by default, that removes a subagent's card from the canvas as
soon as that subagent finishes, instead of leaving it there until the parent agent's next turn
clears it. Off reproduces the previous behavior exactly: both new lines are a single early return,
so with the flag false every touched function's body is byte-identical to before.

**Why the gate is in the store rather than in Canvas.** Two paths mark a card done and only one of
them passes through Canvas: `finish()`, which already absorbs the async task-notification tail,
codex's SubagentStop and SessionEnd, and `sweepStaleWorking()`, the decay for a subagent whose end
never arrives. Gating at the Canvas listener would have covered the first and silently missed the
second, which is the case that matters least often and is hardest to notice. Both now call the
existing `dropCards` after their own early-return guards, so an already-done or not-yet-stale card
is untouched, and since cards and their edges come out of one memo the edge goes with the node.

Only a finished card is ever dropped, which is what leaves the existing rules alone. #547 (a new
turn must keep the cards of subagents that are still running) is unaffected, because
`clearFinishedForParent` is unchanged and simply finds nothing left to drop. Eco's `liveSubagents`
counts cards that are not done, and an absent card and a done card are equally non-live, so
hibernation behaves identically either way.

**Rejected: a grace delay before removal.** It would need a timer owner plus cancellation on
`clearForParent`, project switch and unmount, and it reintroduces the lingering card the setting
exists to remove. The trade-off is real and points the other way for anyone who reads a card as it
finishes, so this is the part most likely to want revisiting; if a delay is wanted later it belongs
on Canvas's existing 60 s sweep rather than a per-card timeout.

**Surfaces.** Desktop and Server Edition are identical: this is pure renderer plus the settings
store, with no bridge member and no IPC channel added. Loop and cron cards are untouched, coming
from `agentStatus` rather than this store. Mobile: N/A, the transport carries no per-node card
concept.

**Not verified.** No test renders the Settings toggle or round-trips the value through
`settings.json`, so persistence rests on the same generic `update(patch)` path as its neighbours
rather than on coverage added here. The link between the setting and the store is pinned by a
source read in `canvas-wiring.test.tsx`, matching that file's existing idiom; it was
mutation-checked (deleting the effect reds it) but it is the weaker kind of pin CONTRIBUTING warns
about, and converting the effect into a pure decision would be the stronger fix. The behavior
itself was confirmed by hand over several days of daily use on a local build, not end to end in
CI, and no measurements were taken: nothing here is on a hot path and the change removes work
rather than adding it.


eneskirca:
**Verdict: MERGE-READY once rebased — plus one copy question.**

It is `CONFLICTING` right now: `AgentsSection.tsx` gained the `controlConfirm` row on `main` in the same two spots you add `autoHideFinishedSubagentCards`. Both conflicts are purely "two people appended adjacent entries"; I resolved them locally (keep both, in either order) and then verified the merged tree:

- `tsc -p tsconfig.web.json` clean
- full suite green — **11179 passed / 40 skipped**, including your 7 new `agentNodes.turn-boundary` cases, `canvas-wiring.test.tsx` and `default-settings.test.ts`

I also checked the three claims the design rests on, rather than taking them from the description:

1. **`finish()` and `sweepStaleWorking()` really are the only two writers of `state: 'done'`** (`agentNodes.ts:264` and `:291`). Gating both is therefore complete, and your argument for gating in the store rather than at the Canvas listener is right — the listener only sees the first one.
2. **Eco is genuinely untouched.** `buildHibernationCandidates` builds `liveParents` by adding a parent only for a card present with `status !== 'done'`. An **absent** card adds nothing, so "dropped" and "done" are the same answer to `liveSubagents`. The `#402` regression this store was hardened against was `clearForParent` wiping **working** cards; you only ever drop done ones, so it stays closed.
3. **`dropCards` takes the whole footprint** — `byId`, `activityById`, `positions`, `sizes`, `expanded`, and it nulls `selectedId` when the dropped card held it. No orphaned override can come back attached to a later card reusing the id.

The `=== true` coercion on the settings read is the right instinct for a hand-editable `settings.json`, and default-OFF plus a pinned `DEFAULT_SETTINGS` test is the correct compatibility promise. Your own honesty about the `canvas-wiring.test.tsx` source-pin being the weaker kind of test is accurate and I would not hold the PR for it — the effect it pins is two lines and mutation-checked.

### The one substantive question: the stale-decay path

The setting is labelled *"Remove a subagent's card from the canvas **as soon as that subagent finishes**"*, and `finish()` matches that exactly. But `sweepStaleWorking()` is the opposite case — it fires precisely because the subagent's end **never arrived** (crashed CLI, killed pane, slept machine). With the flag on, those cards are deleted too, so a fan-out whose subagents silently died leaves no trace on the canvas at all: no card, no edge, nothing that says a subagent was ever launched. That is the one situation where a visible stale card is carrying the most information.

For Eco it makes no difference (done ≡ absent, per point 2), so this is purely about what the user can see. Two ways out, either fine by me:

- **Don't gate the decay** — let `sweepStaleWorking` keep marking `done` and let the turn boundary take it. One line less code, and the setting then does exactly what its label says.
- **Keep it as is and widen the description** — e.g. "…as soon as that subagent finishes, or stops responding", so the label does not promise a narrower rule than the code applies.

I lean to the first, but it is a product call, not a correctness one — say which you prefer and I have no further objection.

### Two nits

- The new `useEffect` in `Canvas.tsx` leaves a double blank line after it (there was already one). Cosmetic.
- Your "rejected: a grace delay" paragraph is the right analysis, and I would keep it in the commit message — it is the question the next person will ask, and the answer ("if a delay is wanted it belongs on Canvas's existing 60 s sweep, not a per-card timeout") is the useful part.

Rebase, pick one of the two options above, and this is good to go.

Almighty-Shogun:
Good catch, and I've taken your first option.

`sweepStaleWorking` no longer consults the setting. It goes back to marking `done` and letting the turn boundary take the card, exactly as before this branch. Your reasoning is the deciding part: the sweep fires precisely because the end never arrived, which is the opposite of what the label promises, and it's the one case where a visible card carries the most information. A fan-out whose subagents died silently shouldn't leave the canvas looking like nothing was ever launched.

So the setting gates `finish()` and nothing else, and the description needs no widening.

The test that asserted the drop is now its inverse with the reasoning in a comment, plus one asserting the turn boundary still takes the decayed card, so the old path is pinned end to end. `main` is merged in, with both `AgentsSection.tsx` conflicts resolved the way you did (keep both, either order), and the double blank line is gone.

`tsc -p tsconfig.web.json` clean; `agentNodes.turn-boundary` + `canvas-wiring` + `default-settings` green at 75, and the full settings-section suite green at 330 since the merge touched `AgentsSection.tsx`.

On the "rejected: a grace delay" paragraph, agreed, it stays in the commit message.

PR #684, opened 2026-09-06
Feature: Remember the window's size, position and maximized state

Almighty-Shogun:
This PR adds persistence for the main window's size, position and maximized state. The window was created at a hard-coded 1400x900 on every launch and nothing ever read or wrote its geometry, so anyone who works maximized re-maximized it every single time they opened the app. Electron persists none of this on its own. It surfaced on Fedora but was never platform-specific — macOS and Windows forgot it too.

State lands in `<userData>/window-state.json` through `renameAtomicSync` with a per-call unique temp, like every other store. `main/window-state.ts` imports no electron, in the `keydown-intercept.ts` shape: pure decisions over plain rectangles, a structural window for the tracker, and `screen.getAllDisplays()` called at the seam in `index.ts`. That was the point of the split — the interesting behaviour here only happens on hardware (a second monitor, a Mac in fullscreen), and this way it can be pressed by a test instead. What gets passed in is the displays' **work areas**, not their raw bounds: a saved position has to be judged against the space a window can actually occupy.

The refusals are the feature, because each is a way the naive version is worse than the fixed size it replaces. **`getNormalBounds()`, never `getBounds()`** — while maximized the latter returns the maximized rectangle, so saving it makes the next un-maximize hand back a screen-sized window: state that looks right and behaves wrong, and only for the users who maximize. **An unreachable position is dropped, not clamped** — a laptop undocked from the monitor its window was on would otherwise reopen the app off-screen, running and focusable from the dock but visible nowhere, with no gesture that rescues it; reachability is a real overlap with some work area, judged against the *clamped* size, and dropping keeps the user's size while letting the platform place a window it knows how to place. **No capture while minimized or fullscreen** — `isMaximized()` is false while a macOS window is fullscreen, so capturing there records `maximized: false` and erases exactly the preference this exists to remember; the last non-fullscreen state stands, which also means the app never reopens *into* fullscreen, deliberately, since that mode is usually temporary and much harder to escape on first launch. Maximizing happens before the first paint while the window is still `show: false`, saves are debounced through a drag and flushed synchronously on `close` (an awaited write there races the process exit), every field is re-validated as a number on read because the file is hand-editable and its values reach the `BrowserWindow` constructor before there is a window in which to report a failure, and `NT_MULTI` is excluded so a dev sandbox sharing the real userData cannot move the window of the app being developed.

Alternatives I rejected: clamping an off-screen position into the nearest display, which invents a corner the user never chose and is indistinguishable from a bug when it happens; and restoring fullscreen alongside maximized, which is trivial to add and unpleasant to land in on a first launch.

Surfaces — **Desktop**: the whole feature. **Server Edition**: not applicable, a browser tab's geometry belongs to the browser, which is why none of this is in `src/core`. **Mobile**: not applicable, *nodeterm mobile* has no window concept; nothing for @eneskirca to carry over.

Verified: `npm run typecheck`, and the `src/main` + `src/shared` suites plus the atomic-write guard (2525 passing). 37 unit tests cover the module. Following the mutation habit in CONTRIBUTING.md I inverted twelve of the guards above; three survived the first version of the suite — a fixture that could not tell the clamped size from the saved one, a timer fake that hid a missing debounce cancel, and no case for an overflowing exponent (`1e999` parses to `Infinity`, so `typeof === 'number'` only *looks* sufficient). All twelve are killed now.

Not verified: I tested this by hand only on Fedora 44 / GNOME, where size and maximized restore correctly across quit and relaunch. The macOS fullscreen path, the Windows path, and multi-monitor undocking are covered by unit tests against synthetic work areas and by nothing else. One known limit: a native-Wayland client cannot set its own position, so x/y is honoured under XWayland and ignored otherwise — size and maximized restore either way, which is what the drop-don't-clamp rule already degrades to.


Abhirup0:
Very clean PR and great adherence to repo invariants (the pure-decision module split, `renameAtomicSync`, `NT_MULTI` isolation, and pre-paint maximize are all spot on).

One edge case in `reachableOn` (`src/main/window-state.ts:108`):

The JSDoc for `MIN_VISIBLE_HEIGHT` notes that *"a title bar the user cannot reach is a window they cannot move, and on a stored position from a monitor that is now gone there is no gesture left that rescues it."*

Because `reachableOn` computes symmetric height intersection:
```ts
const h = Math.min(r.y + r.height, area.y + area.height) - Math.max(r.y, area.y)
return w >= MIN_VISIBLE_WIDTH && h >= MIN_VISIBLE_HEIGHT
```
if a user was previously on an external monitor positioned **above** their laptop screen (negative `y`), say `x: 200, y: -500, width: 1000, height: 700`, and unplugs it:
* Against a laptop screen at `x: 0, y: 27, width: 1920, height: 1053`, the bottom 173px of the window overlaps the top of the laptop screen (`y: 27..200`).
* `h = 173 >= 80`, so `reachableOn` returns `true` and the position is kept.
* The window opens at `y: -500`, leaving the title bar and tab drag strip >500px above the screen. With `titleBarStyle: 'hiddenInset'`, the user cannot grab the window to move it down.

Checking that the top region/title bar itself lands within the work area (e.g. requiring `r.y >= area.y` or checking intersection of the top ~40px of the window with the work area) would close this case.

Minor note: on Linux, Electron docs note `getNormalBounds()` is only supported on certain desktop environments. When `!win.isMaximized()`, falling back to `win.getBounds()` would ensure unmaximized windows save reliably across all Linux WMs.

eneskirca:
**MERGE-READY** (one trivial doc conflict to resolve first).

Measured here (merged onto `main` @422767af in an isolated worktree): `npm run typecheck` clean; `window-state.test.ts` green, and — the two that mattered — `src/main/window-raise.guard.test.ts` and `src/core/fs-atomic.guard.test.ts` both green with the new `win.maximize()` and the new store in place (48 tests total).

Storage is on the right side of the line: `<userData>/window-state.json` is machine-local, which is what window geometry is. Nothing here goes near `.nodeterm/project.json`, and the module is deliberately Electron-free with `screen.getAllDisplays()` called at the seam in `index.ts` — so the refusals are testable rather than only reproducible by someone with two monitors.

The four decisions I would have asked for are already made and already justified in the code:
- `getNormalBounds()` rather than `getBounds()`, so a maximized session does not hand back a screen-sized window on the next un-maximize.
- An unreachable saved position is **dropped**, not clamped. This is the one that turns the feature from "nice" into "safe" — a laptop undocked from the monitor it was last on would otherwise reopen the app running, focusable from the dock and visible nowhere.
- No capture while minimized or fullscreen (`isMaximized()` is false during macOS fullscreen, so capturing there would erase exactly the preference this exists to remember).
- `renameAtomicSync` + a per-call unique temp, per docs/atomic-writes.md, and a synchronous flush on `close` because an awaited write there races process exit.

Interaction with what landed on main since you branched: #737 made the show path `win.once('ready-to-show')`. Maximizing inside `createWindow()` while the window is still `show: false` does not disturb that, and the raise guard agrees.

**Conflict:** `CLAUDE.md` only — your "Window geometry is REMEMBERED" bullet lands on the same lines as main's new "Nothing raises the window without a user action" bullet. Both are additions; keep both, in either order.

Not verified here (headless Linux server, no display): the multi-monitor undock path, macOS maximize-before-first-paint, and the Wayland x/y caveat the doc block already states. Those are the device checklist.

Almighty-Shogun:
Thanks both.

@Abhirup0 you're right about `reachableOn`, and I reproduced your case before changing anything: `{x: 200, y: -500, width: 1000, height: 700}` against a work area of `{x: 0, y: 27, width: 1920, height: 1053}` gives `h = 173`, which clears `MIN_VISIBLE_HEIGHT`, so the position was kept and the title bar opened 527px above the screen. Under `titleBarStyle: 'hiddenInset'` there's nothing else to grab, which is exactly the failure the module's own doc block claims to prevent.

Fixed in `fix(window): drop a saved position whose title bar lands off the screen`. `reachableOn` now also requires the window's top edge to sit on the work area, within a new `TOP_OVERHANG_SLACK` of 24px. The slack is there because window managers report decorations inconsistently and a few pixels of overhang shouldn't cost the user their position on every launch; it's far smaller than a title bar, which is the thing being protected. The rule is per display, like the overlap it joins, so a window at `y: 0` on a second screen whose work area starts at `y: 0` is still reachable even though it sits above the laptop's `y: 27`.

Your second point is in too, as `fix(window): read getBounds while un-maximized, for Linux WMs without getNormalBounds`. Capture now reads `getBounds()` unless the window is maximized. The maximized case still needs `getNormalBounds()` for the reason the original commit gives, so it still depends on the WM there, but that record restores by re-maximizing rather than by its size, so it matters less. I said that in the docstring rather than implying the whole thing is now WM independent.

Three new tests for the first fix, one for the second. I mutation-checked all four: reverting the top-edge rule reds the bottom-edge test, reverting the fallback reds the `getBounds` test. `tsc -p tsconfig.node.json` clean, and `window-state.test.ts` + `window-raise.guard.test.ts` + `fs-atomic.guard.test.ts` green at 52.

@eneskirca conflict resolved as you described, both bullets kept, yours first. The two CLAUDE.md sub-bullets describing the old `getNormalBounds` rule and the old reachability rule are updated inside the commits that changed the behaviour.

Still unverified here (no second monitor, Linux): the actual undock, and macOS maximize before first paint. Both were already on the device checklist and these changes don't remove them.

PR #687, opened 2026-09-06
Feature: Remember the canvas lock across restarts

Almighty-Shogun:
This PR adds an opt-in setting, off by default, that keeps the bottom-left canvas lock as you left it across a restart. The lock is a `useState(false)` in Canvas and nothing ever read or wrote it, so it survived project switches within a session and died on every reload. Off reproduces the previous behavior exactly: the initializer is unchanged, and with the flag false nothing is read from storage or written to it.

**Why it is opt-in rather than simply persisted.** The lock was transient by design, and the comment saying so is the argument against the naive version: "a lock that survives restart reads as 'the app is frozen' to whoever opens it next". The canvas will not pan, and the only thing on screen explaining why is a small lit button the user may not have armed. `state/viewMode.ts` cited that same reasoning as its own precedent for why the Omni board *does* persist, so both sentences are corrected here rather than left contradicting the code. Default off means anyone who locks once by accident is unaffected, and because nothing is written while the setting is off, a user who opts in later finds no stale lock from months ago waiting to restore itself.

**The restore cannot be a `useState` initializer, and that is the shape of the whole change.** `useSettings` starts on `DEFAULT_SETTINGS` and hydrates from disk asynchronously, while `<Canvas />` is mounted before that lands, so an initializer reads the shipped default (off), never looks again, and the feature ships inert with a green suite and no error anywhere. It is gated on `hydrated`, the same way the first-launch notification consent is. One effect then does both jobs and the order is what makes it safe: the first run after hydration restores and latches, every run after that persists. Without the latch, switching the setting on mid-session would reach into storage and lock a canvas someone is working in, instead of starting to remember from where the canvas is now. `CONTRIBUTING.md` gains that trap as a house rule, since it is not specific to this feature: any future opt-in read from settings at mount hits it.

**The bit is one global flag, not one per project.** `<Canvas />` is mounted once and is not keyed by project, so the lock has always carried across project switches within a session; one global flag in `localStorage` (`nodeterm.canvasLocked`, beside the view mode and the explorer pin) extends that rather than introducing a second rule. The setting says whether to remember and belongs in `settings.json`. The lock itself is one person's view state and belongs in neither `settings.json` nor the git-shared `project.json`.

**Rejected: a per-project map.** It changes existing in-session behavior, since the lock currently follows you between projects, and it would owe the dead-id pruning `sidebarCollapsedItems` does. **Rejected: gating `autoPanOnNodeDrag` and `autoPanOnConnect` on the lock.** Neither is passed to `<ReactFlow>`, so both hold React Flow's `true` default, and dragging a node or a connection to the viewport edge still scrolls the camera while locked. The comment now records that, because it is a camera move getting through a camera lock, but gating them changes how a node drag behaves while locked and is a separate decision from this one. That comment also carried a pre-existing false claim, that the lock froze node dragging and edge connecting, which `nodesDraggable` and `nodesConnectable` contradict along with the JSX comment at the prop site, the button's own tooltip and CLAUDE.md; it is corrected here.

Surfaces: **Desktop**, the whole feature. **Server Edition**, works unchanged, pure renderer plus the settings store with no IPC channel, preload or bridge member added; the two tiers do split differently there and the docs say so, since the setting is the server's `settings.json` while the bit is per browser profile, so two profiles on one machine can legitimately disagree about the lock, which is correct for view state. **Mobile**, not applicable, there is no canvas; nothing for @eneskirca to carry over.

Verified: `npm run typecheck` clean, and 372 tests pass across `renderer/lib`, `renderer/canvas`, `settings` and `state`. `canvasLock.test.ts` covers the parse rule, the round trip in both directions (unlocking has to persist too, or a lock could never be cleared for the next launch), a lock left by a previous launch, throwing storage, and the method-less `localStorage` global from #412. Following the mutation habit in CONTRIBUTING.md I removed each line of the effect in turn: whole-block removal, both gated branches, the hydration gate and the latch all go red, but dropping `canvasLocked` from the dependency array survived every assertion while silently breaking button-driven persistence, since the effect then stops re-running on a toggle and only a settings flip would ever write. There is no ESLint here, so `exhaustive-deps` cannot catch it either; that case now has its own pin.

Tested by hand on a packaged Fedora 44 build from this branch, across all four paths: with the setting on a lock survives a restart, and an unlock survives it too, so a lock can still be cleared for the next launch; with the setting off, which is the shipped default, the canvas opens unlocked whatever it was left as; and turning the setting on mid-session does not retroactively lock the canvas from a stored value.

Not verified: the Server Edition split above was reasoned from the shared-renderer path and read rather than exercised in a browser, and the build was run only on Fedora, though nothing here is platform-specific. The effect ordering was traced against the real store rather than driven through a mounted Canvas, so the unit coverage of that ordering is source-level pins plus the hand test rather than a rendered one. No measurements were taken: this is one boolean, read once at hydration and written on a click.


eneskirca:
**MERGE-READY** (one trivial conflict to resolve first).

Measured here (merged onto `main` @422767af in an isolated worktree): `npm run typecheck` clean; `canvasLock.test.ts` + `canvas-wiring.test.tsx` + `viewMode.test.ts` green (60), and the whole `src/shared` + settings sections suite green (99 files / 1494 tests).

Where the state lives is the decisive question for this one, and it is right: the **bit** goes to `localStorage` (`nodeterm.canvasLocked`, through `lib/localStore`), the **preference** to `settings.json`, and nothing at all to `.nodeterm/project.json`. A canvas lock is one person's view state, so it must never ride the git-shared file — this respects that, and the module comment says so where the next reader will find it. Default OFF keeps an untouched install byte-identical.

Two things I checked because they are the ways this class of change usually ships inert:
- The hydration gate is real. `<Canvas />` mounts on `DEFAULT_SETTINGS`, so a `useState` initializer would have read `false` forever; gating on `settingsHydrated` and latching the first run is the correct shape, and the wiring pin in `canvas-wiring.test.tsx` covers the deps array, which is the one line a refactor can delete while staying green.
- The comment correction on `canvasLocked` is accurate — main's block comment there claimed node dragging and edge connecting were frozen, which contradicts both CLAUDE.md and the actual `panOnDrag`/`panOnScroll` props further down. Good catch.

**Conflict:** `src/renderer/canvas/Canvas.tsx` conflicts with main, comment-only (the same block comment, edited on both sides). Resolve by keeping this branch's version. `CLAUDE.md` / `CONTRIBUTING.md` auto-merge.

Nit, non-blocking: the new wiring tests match source text. There is precedent for that in the same file (the `frameNode` pin) and the rationale is stated, so I am not asking for a change — just noting that they will need an edit whenever those three lines are reformatted.

Not verified here (headless Linux server): nothing device-specific in this change, so there is no device debt beyond a quick "lock, quit, relaunch" pass with the setting on and off.

Almighty-Shogun:
Conflict resolved keeping this branch's version of the comment, as you suggested, and `main` is merged in.

`tsc -p tsconfig.web.json` clean; `canvasLock.test.ts` + `canvas-wiring.test.tsx` + `viewMode.test.ts` green at 60, matching what you measured.

On the nit: agreed, those wiring tests match source text and will need an edit whenever those three lines are reformatted. Leaving them as they are, for the reason you named.

PR #688, opened 2026-09-06
fix(windows): harden native module bootstrap

FabricioCasali:
## Summary
- fail fast when the selected Visual Studio C++ toolchain lacks the x64 Spectre-mitigated runtime required by node-pty's /Qspectre build
- override the clang/lld LTO variables inherited from official Node 26 builds before npm ci, preventing node-gyp 12 from handing /opt:lldltojobs to MSVC link.exe
- run the same bootstrap toolchain probe in quality-windows and document the supported Windows setup path

## Why
A machine can have the Visual Studio C++ workload while missing its separately installed Spectre libraries. The current bootstrap accepts that setup and node-pty fails much later with MSB8040.

Official Node 26 Windows builds also carry clang/lld ThinLTO values in process.config. node-gyp 12 copies them into the generated MSVC project, where link.exe rejects /opt:lldltojobs with LNK1117. The bootstrap now supplies the two gyp overrides before npm ci instead of rejecting a Node version allowed by package.json.

The bootstrap still installs nothing machine-wide: it names the missing Visual Studio component and exits.

## Verification
- full bootstrap on Windows with Node 26: npm ci completed and node-pty plus smart-whisper rebuilt against Electron's ABI
- npm run typecheck
- Windows-relevant CI suite: 492 passed, 6 skipped
- bootstrap suite: 11 passed
- git diff --check

## Surfaces
Desktop Windows developer and build setup only. No Server Edition or mobile runtime behavior changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,10 @@ jobs:
           node-version: '22'
           cache: 'npm'
 
+      - name: Verify Visual Studio C++ toolchain
+        shell: cmd
+        run: call "%GITHUB_WORKSPACE%\bootstrap-windows.bat" --check-vs-build-tools
+
       - name: Install dependencies
         run: npm ci
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -3247,7 +3247,14 @@ winget hints (it never installs machine-wide tools itself, and the full bootstra
 elevated) and runs `npm ci`. Its `--check-vs-build-tools` mode is the narrow exception used by
 `quality-windows`: it branches before the elevation refusal, runs only the VS C++ probe, and exits
 before the Node / Python / `npm ci` steps. Fixture injection additionally requires the explicit
-`NODETERM_BOOTSTRAP_TESTING=1` sentinel. `.github/workflows/win-package-smoke.yml` is a
+`NODETERM_BOOTSTRAP_TESTING=1` sentinel. The C++ probe also verifies the x64 Spectre runtime that
+`node-pty`'s `/Qspectre` build requires; the workload alone can be present while that optional
+component is absent, which otherwise fails only after the full install with `MSB8040`. Before
+`npm ci`, the bootstrap sets `npm_config_enable_thin_lto=false` and
+`npm_config_enable_lto=false`: official Node 26 Windows builds carry clang/lld ThinLTO settings in
+`process.config`, and node-gyp 12 copies them into an MSVC addon project where `link.exe` rejects
+`/opt:lldltojobs` with `LNK1117`. These are gyp overrides, not a reason to reject a Node version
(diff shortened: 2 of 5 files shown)


eneskirca:
**Verdict: MERGE-READY — pending one CI run that has not happened yet.**

Verified locally (merged onto `main`): merges cleanly, `bootstrap-windows.test.ts` skips cleanly off Windows (11 skipped, as designed), nothing else in the tree is touched. Both failure modes are diagnosed precisely and both fixes are the narrow ones — the Spectre probe names the exact Individual Component instead of installing anything, and the LTO overrides are gyp variables rather than a `package.json` engines bump that would have rejected a Node version we allow. Good judgement on both.

A bonus worth noting: `CLAUDE.md` on `main` *already* claims `--check-vs-build-tools` "is the narrow exception used by `quality-windows`" — but `ci.yml` never called it. This PR is what makes that sentence true. That is a doc-ahead-of-mechanism drift closed, not just a new step.

### The one thing to settle before merging

This adds a **hard gate to `quality-windows`**, and no CI has run on this branch (`gh pr checks 688` → "no checks reported"). If `windows-latest` does not carry *MSVC v143 – VS 2022 C++ x64/x86 Spectre-mitigated libs*, every PR's Windows job goes red on a probe, with a message telling the runner to open Visual Studio Installer. The 492-passing run in the description was on your machine, which is the one environment where the probe is known to pass.

So: @eneskirca, please approve/run CI on this PR once and confirm `quality-windows` is green. That single run is the whole risk — the change is otherwise self-contained, and the gate is genuinely worth having (it catches a runner-image regression before it becomes an `MSB8040` deep inside `npm ci`).

### Two nits

1. **The new LTO test is Windows-gated but does not need to be.** `describeWindows` is `process.platform === 'win32' ? describe : describe.skip`, and `'disables inherited Node 26 LTO settings before npm ci'` only does `readFileSync` + `indexOf` on the batch file. It would run everywhere. As written, the ordering invariant is only checked on the Windows runner — which is exactly where it is hardest to notice it stopped running. Moving that one `it` outside `describeWindows` costs nothing. (CLAUDE.md's rule is to platform-gate only what *can't* run elsewhere, and to say why.)

2. **x64-only probe.** `for /d %%D in ("%VS_INSTALLATION%\VC\Tools\MSVC\*") do if exist "%%~fD\lib\spectre\x64\vcruntime.lib"` is right for today (`dist:win` is x64 NSIS), but on an ARM64 Windows dev box it demands the x64 Spectre libs and refuses an otherwise-usable checkout. One sentence in the batch comment saying "x64 because that is the only Windows arch we ship" would keep the next reader from treating it as an oversight.

Neither blocks. Merge once the Windows job has actually been seen green.

FabricioCasali:
@eneskirca Could you please approve and run the CI for this PR, especially quality-windows? There are currently no checks reported.

PR #694, opened 2026-09-07
ci: skip electron-rebuild in the quality jobs

Temikus:
@eneskirca some drive-by changes for the issues I noticed during my PR runs.

`package.json's` postinstall runs `patch-node-pty.mjs` followed by `electron-rebuild -f -w node-pty,smart-whisper`, so every npm ci in CI does a full native compile of `node-pty` and `whisper.cpp`. 

E.g. on run `34057272144` that compile accounted for 220s of the 266s quality-windows job and 79s of the quality job.

That compiled output is never exercised by the test suite. electron-rebuild targets Electron's ABI, so a plain-node `vitest` run cannot load the resulting binding: the `*.realtty.test.ts` and `*.realtmux.test.ts` suites already self-skip for exactly this reason, the Windows job's session-host tests only `import type { IPty } from 'node-pty'` (erased at compile time), and `electron-vite` build keeps `node-pty` external anyway.

The only effect of the install step is `patch-node-pty.mjs`, which rewrites node-pty's C++ sources with the version-pinned markers `src/main/node-pty-patch.test.ts` asserts. 

Both quality jobs now run `npm ci --ignore-scripts`, then invoke the patch script directly, then `npm rebuild esbuild` (`esbuild`'s postinstall links its own platform binary and would otherwise be skipped). 
`release.yml` and `win-package-smoke.yml` are untouched, since those jobs ship or verify real native binaries.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,8 +24,19 @@ jobs:
           node-version: '22'
           cache: 'npm'
 
+      # electron-rebuild (via postinstall) targets Electron's ABI, so vitest cannot load the
+      # binding it produces — the *.realtty / *.realtmux suites self-skip, and the Windows
+      # session-host tests only `import type`. Skip the compile; keep the two install effects
+      # that are actually load-bearing: patch-node-pty.mjs writes the markers
+      # src/main/node-pty-patch.test.ts asserts, and esbuild needs its platform binary linked.
       - name: Install dependencies
-        run: npm ci
+        run: npm ci --ignore-scripts
+
+      - name: Patch node-pty sources
+        run: node scripts/patch-node-pty.mjs
+
+      - name: Link esbuild platform binary
+        run: npm rebuild esbuild
 
       - name: Type check
         run: npm run typecheck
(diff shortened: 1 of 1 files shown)


eneskirca:
The change itself is one we want — a full `node-pty` + `whisper.cpp` native compile on every `npm ci` is real CI time for jobs that never load either module.

Two things block it right now, and both need a push from you:

1. **It conflicts with `main`.** Please rebase.
2. **`quality` is failing** on the current head (`quality-windows`, CodeQL and dependency review all pass). The run is old enough that its log has expired on our side, so a rebase + re-run is also how we find out whether the failure is yours or just a stale base.

One thing to confirm while you are in there: the quality jobs must still be able to import anything that needs the rebuilt native modules. `src/main/node-pty-patch.test.ts` asserts the patch marker is present in `node_modules/node-pty`'s SOURCES rather than in a built artifact, so it should be unaffected — but please say so in the PR body once the run is green, since that test is the one guard that stops a node-pty upgrade from silently dropping the ptmx-leak patch.

Temikus:
@eneskirca thanks, all three addressed:

- Rebased onto main. The only conflict was the new VS toolchain check in quality-windows, which I kept.
- quality failure was mine. On Linux, node-pty has no prebuild, and ~20 suites import it at runtime (`Failed to load native module: pty.node`). The jobs now run `npm rebuild node-pty esbuild` after the patch step, so node-pty compiles for Node from the patched sources. Only whisper.cpp is skipped. Both jobs are green and install time is 76s→15s and 254s→21s.
- node-pty-patch guard: confirmed it ran and passed in the quality log. I added a note to the PR body.

PR #696, opened 2026-09-07
feat(canvas): save a named layout and restore it after switching screens

basmilius:
## Why: undocking costs you the arrangement you had

A canvas is arranged for the screen it was arranged on. Unplug the ultrawide and nine terminals
spread across 3440px are being read through a 1512px window. Zoom out and they are unreadable;
rearrange by hand and you have lost the layout you had on the big screen. Plug back in and you pay
it again in the other direction.

Node geometry already persists, so nothing is lost by moving nodes. What is lost is the *previous*
arrangement. This saves that under a name.

## What it adds

A **Layouts** button in the Dock's view cluster, after *Fit view*. Its menu lists the saved layouts
with the window size they were saved at and their node count, restores one on click, and carries
three actions per row: update it to the arrangement now on screen, rename it, delete it. `⌘K` mirrors it with *Save canvas layout…* and one *Restore layout:
`<name>`* per layout. No new keybinding: a chord that opens a menu earns little, and a per-layout
chord cannot be registered for names that do not exist yet.

Updating exists because the three-step alternative already worked: save, retype the name you are
looking at, confirm the replace. Re-typing a name is friction rather than a decision. It reuses
`saveLayout`'s replace-by-id path, so `createdAt` survives and `updatedAt` moves, and it re-captures
the window size and camera because you are updating from this screen.

Update and delete both confirm, restore does not, and the split is the undo stack. `Cmd+Z` replays
node arrays and a layout lives beside them rather than in them, so those two are gone the moment they
run while a restore is one undo away. Both dialogs come from `layoutIsShared` plus
`deleteLayoutMessage`/`updateLayoutMessage`, one definition of who else a destructive edit reaches. I
had that wrong at first by gating on `cwd` alone, which told an SSH project's user their edit was
private when their `project.json` sits on the host for everyone.

## A layout is geometry, and only geometry

Position, size, collapse state, camera. Restoring never creates, deletes, renames, recolors,
reparents or respawns a node, and never touches a tmux session. That is what makes it safe to click
without thinking about it first.

## Storage: content is shared, the camera is not

A layout is a project in miniature, so it splits on the seam the project itself uses.

| Half | Lives in | Why |
|---|---|---|
| name, node rects, collapse, the author's window size | `ProjectFileV1.layouts` in `.nodeterm/project.json` | Content, git-shared like `kanban`. Node ids are shared, so a teammate's clone resolves the same nodes and clicking "Ultrawide" gives them the arrangement you built. |
| this machine's camera per layout | `IndexEntryV3.layoutViewports` in `workspace.json` | Where I am looking is not something a repo shares. The rule that moved `viewport` out of the shared file. |

I had this the wrong way round at first. The initial design put layouts entirely in machine-local
storage, on the `viewport`/`breadcrumbs` precedent. That precedent does not reach. Those are camera
facts, and nobody else's canvas moves when I pan; node geometry is already shared content in
`project.json`, and a restore writes it. Keeping the snapshot private would have handed the team all
of the churn with none of the benefit.

What sharing costs, so nobody finds it out later:

- Saving, renaming or deleting a layout is a commit-able diff. Correct here. It is a deliberate,
  named act, not a gesture, which is what the viewport rule was about.
- Deleting removes it for everyone with that project, and the confirm dialog says so.
- Name uniqueness holds at the point of save, not as a global invariant. A merge can land two
  "Ultrawide"s. Both are shown and either can be renamed, because silently dropping a teammate's
  layout on load is the worse failure.
- A layout with no local camera falls back to a framing derived from its own rects, so a teammate's
  layout opens on the nodes rather than on empty canvas.
- A build older than this one drops `layouts` on its first save, because `projectToFile` builds an
  explicit object. Mild, since a layout is re-savable from what is on screen, but real, and it wants
  a line in the release note.

## Restore

One pure transform, `applyLayout` in `renderer/lib/canvasLayout.ts`. The rules that carry weight:

**A node the layout does not know is left exactly where it is.** Not moved, not tidied, not stacked
at the origin. Two holes in this turned up during review and both are closed. An out-of-layout child
used to ride a placed frame, because it kept its parent-relative position; and it could be clamped by
`extent: 'parent'` when the frame's saved rect did not contain it. Pass one now resolves a target
root for every node, and a placed frame grows, never shrinks and never re-centers, to contain its
out-of-layout descendants. A restore applied twice is identical the second time, and that is a test.

**One transform, not N placements.** A child's parent-relative position comes from its parent's
target rect when the parent is in the layout, and from its current rect when it is not. Placing
nodes one at a time would let each ancestor re-fit move the frame the next call measures against.

**A layout entry whose node is gone is skipped**, never recreated. Geometry is not content.

**`premaxRect` survives**, because a restore is a placement and not a maximize, and collapse
round-trips with `expandedHeight` intact, so expanding a restored node does not snap it to a stale
size.

**The camera goes through `setViewport`, never `fitView`.** That is the "go to node" invariant from
#663: a fit resolves later, against whatever is measured by then, and on an unmeasured canvas it
collapses the bounds to the origin.

Restore then says what it did, in the shape `Restored "Ultrawide": 12 nodes moved, 2 not in this
layout.` The sentence is built from the transform's own return value, so it cannot drift from what
happened. Undo needed no work, since the existing debounced history effect picks up the `setNodes`.

The recorded window size is a label, not a matcher, and there is no auto-apply when a display
changes. Moving someone's canvas because they plugged in a monitor is the loudest surprise this
feature could produce.

## Hostile input

`project.json` is cloned, hand-editable and git-shared, so `sanitizeLayouts` runs at both serializer
seams, `fileToProject` and `projectToFile`. That is the `normalizeNodeIcon` rule: validating one
direction passes every round-trip test while leaving the other open, and what we write is what the
next machine trusts. It drops rather than repairs. Any non-finite number drops the layout whole,
because a `NaN` reaching `node.position` is dereferenced by React Flow's `adoptUserNodes`, which is a
white-screen crash from a file anyone can commit, and `JSON.stringify` writes it back as `null`.
Capped at 20 in both directions.

## Surfaces

Desktop is full. Server Edition is full **with no new IPC**, since the whole feature rides
`WorkspaceStore`, which both shells boot, plus renderer code; nothing is stubbed. Relay tabs are
refused with a reason ("Layouts are managed on the host"), disabled rather than hidden, because a
relay tab is a live connection and never a workspace on this disk. Kanban is N/A, as the board shows
cards rather than geometry and the Dock already sits under its overlay. Mobile is N/A for v1:
*nodeterm mobile* has no canvas, so surfacing layouts means extending the transport protocol.
Follow-up in `nodeterm-ios`, @eneskirca.

## What I did not build

Cross-project apply, where one click applies the same-named layout in every open project. It would
need the transform to run against serialized `CanvasNodeState[]` as well as live `CanvasNode[]`, so
two copies of one rule. It is additive if it is ever wanted.

A private layout tier, for which the pattern would be the shared `.nodeterm/settings.json` plus a
machine-local `localSettings` overlay. Nobody has asked for one, and it would put two kinds of row in
the same menu with nothing telling them apart.

Restoring node existence. `⇧⌘T` already reopens closed sessions and owns that job.

## Testing

`npm run typecheck` is clean. 294 tests across the seven touched suites, including the new
`shared/canvas-layout.test.ts`, `renderer/lib/canvasLayout.test.ts`,
`renderer/lib/canvasLayoutView.test.ts` and `renderer/state/projects.layouts.test.ts`. The transform,
the sanitizers and the store actions are pure and carry their own; the persistence legs extend the
suites that already cover `closedSessions` and `breadcrumbs`.

Exercised by hand on a running instance: saving, restoring, updating, renaming and deleting through
the Dock menu all behave.

What that pass did not cover, so the tests are the only evidence for these four:

1. A nested group frame with a child created *after* the save, which is the clamp case above.
2. Collapse three nodes, save, expand them, restore, then expand one. It has to come back at its real
   height rather than a stub.
3. Save, add two nodes and delete one, restore. The new nodes must not move, and the counts in the
   notice must match what happened.
4. Hand-edit `"x": null` into a layout in `.nodeterm/project.json` and reopen. The layout drops, the
   project still loads, and the bad value is not written back.

## Docs

`CLAUDE.md` gains a `Canvas layouts` bullet under *Canvas interaction & panels* with the invariants
and their reasoning. `CONTRIBUTING.md` gains the short version, per the two-files rule.

image

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2923,6 +2923,69 @@ the Settings section and ShortcutsPanel start disagreeing about what a chord mea
     shells boot — no new bridge member); mobile is N/A (no canvas, no camera); the kanban board is
     likewise N/A, and a project that activates ON the board neither shows nor spends its
     once-per-run resume card (it would sit invisible under the opaque overlay).
+- **Canvas layouts** (`@shared/canvas-layout`, `renderer/lib/canvasLayout.ts` capture/apply +
+  `renderer/lib/canvasLayoutView.ts` for the menu wording and the fallback camera; Dock button
+  after "Fit view", mirrored in ⌘K) - a NAMED SNAPSHOT OF NODE GEOMETRY (position, size,
+  collapse state) per project, so an arrangement built for the ultrawide can be restored after a
+  day on the laptop screen. Load-bearing rules:
+  - **Geometry only, and that is the whole safety story.** A restore never creates, deletes,
+    renames, recolors, reparents or respawns a node, and never touches a tmux session - so the
+    worst a bad layout can do is move things, which ⌘Z takes back (the debounced history effect
+    picks up the `setNodes` like any other placement, which is why restore has no confirm and
+    delete does).
+  - **The content half is git-shared, the camera half is machine-local.** `Project.layouts` rides
+    `.nodeterm/project.json` beside `nodes` and `kanban`, because node geometry is already shared
+    content in that file and a restore writes exactly those fields. `Project.layoutViewports` (this
+    machine's camera per layout) rides `IndexEntryV3` in `workspace.json`. **The camera precedent
+    (`viewport`, `breadcrumbs`) deliberately does NOT reach the geometry**: those are facts about
+    where one person was looking, and nobody else's canvas moves when I pan - but where the nodes
+    SIT is the canvas itself, and sharing an arrangement with the repo is the point.
+  - **Restore rules** (`applyLayout`, pure + tested): a live node the layout does not mention is
+    left exactly where it is, never tidied or stacked at the origin; an entry whose node is gone is
+    skipped and counted, never resurrected; a frame the layout addresses gets the rect the user
+    saved and is NOT re-fitted afterwards (the fit would overwrite it and the arrangement would
+    drift a little on every restore), while a frame the layout does not mention but whose
+    descendant just moved IS re-fitted, deepest first, so an out-of-layout child cannot be clamped
+    by `extent:'parent'` into an inverted range. The whole layout lands in ONE transform as an
+    explicit two-pass (resolve every target root origin, then emit) - a reduce over N placements
+    re-fits an ancestor between two of them, so the second node is measured against a frame the
+    first just moved and the result is differently wrong depending on array order. `collapsed` is
+    restored with the CHROME height on the node and the real one in `expandedHeight` (the
+    `flowToNodeStates` rule: the stored height is ALWAYS the expanded one, or a
+    save-while-collapsed shrinks the node permanently). `premaxRect` and every non-geometry field
+    ride the spread untouched - a restore is a placement, not a maximize.
+  - **The camera is applied with `setViewport`, NEVER `fitView`** - the "Go to node" invariant
+    above, and a canvas that has just been rearranged is exactly the unmeasured state where a
+    queued fit collapses the bounds and flies to the origin. This machine's recorded camera wins;
+    a layout restored here for the first time (a teammate's, or one saved on another machine) gets
+    `layoutFramingViewport`, which is the core `framingViewport` rule rewritten locally because the
+    renderer has no import path into `src/core` and because a layout's rects are ROOT-space, so
+    unlike `CanvasNodeState` positions every entry anchors the camera.
+  - **The window size is a LABEL, never a matcher.** `CanvasLayout.window` is the author's window
+    at save time, shown in the menu so the user can tell the two arrangements apart. Nothing
+    auto-applies a layout on a display change: the file travels, so matching on it would pick a
+    stranger's monitor for this user's screen - and a canvas that rearranges itself when you plug
+    in a projector is worse than one that does not.
+  - **Cap 20** (`CANVAS_LAYOUTS_CAP`; a full node-geometry list in a file that is committed and
+    cloned), name capped at 60, and **both sanitized on BOTH serializer seams** (`fileToProject`
+    on the way in, `projectToFile` on the way out) - the same two-seam rule `normalizeNodeIcon`
+    and `sanitizeNodeTriggers` follow, because live node data is reachable by a peer canvas
+    mutation and whatever we write is what the next machine trusts. `sanitizeLayouts` DROPS a
+    layout whole rather than repairing it: a repaired layout no longer describes the arrangement it
+    is named after, and a non-finite coordinate is a white-screen crash (`adoptUserNodes`
+    dereferences the position unguarded) that `JSON.stringify` then writes back as `null`.
+  - **Downgrade note:** a build older than this one drops `layouts` on its first save, because
+    `projectToFile` builds an explicit object and simply does not know the field. The layouts are
+    gone from that checkout's file until someone on a current build saves again; nothing else
+    breaks, and the machine-local cameras are pruned to match on the next load.
+  - **Surfaces:** Desktop full; **Server Edition full with NO new IPC** (pure renderer plus
+    `workspace.save`, which both shells already boot); **relay tabs REFUSED with the reason** -
+    the Dock button is disabled with "Layouts are managed on the host" and the ⌘K entries are
+    omitted (the palette has no disabled row), because a relay tab is a live connection to another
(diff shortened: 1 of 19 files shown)


eneskirca:
**MERGE-READY**, with one product question for @eneskirca (below) and two cosmetic nits.

Measured here (merged onto `main` @422767af in an isolated worktree): `npm run typecheck` clean, and the **entire** suite green — 825 files / 11269 tests, 0 failures. The six suites this touches are 252 of those.

## Overlap with the zones work (#394 / #408) — measured, and there is none in state

I checked this first because it was the one thing that could have sunk the PR. `renderer/lib/nodeZones.ts` is a **stateless placement verb**: `zoneTargetRect` computes a rect from the live viewport, `placeNodeInRect` writes plain absolute node geometry, and nothing is stored under a name. So the two features never write the same state and cannot disagree.

What they *do* share is a roadmap. #394 was phased as v1 (keyboard + menu snap, shipped in #408), v2 (drag-time overlay), v3 (**saved layouts per project**) — and this PR is v3. The v1 design note even pre-decided the storage question the same way you have: saved layouts go in `project.json`, git-shared. Worth linking #394 from this PR and saying whether it now narrows to the v2 overlay alone.

## Storage — this is the part I pushed hardest on, and it is right

The split is exactly the seam CLAUDE.md's State & persistence model draws:

- `Project.layouts` → `.nodeterm/project.json`, beside `nodes` and `kanban`. Defensible because **node geometry is already shared content in that file** and a restore writes precisely those fields. A layout is not a second, competing copy of the canvas — it carries no `cwd`, no command, no agent id.
- `Project.layoutViewports` → `IndexEntryV3` in `workspace.json`. Machine-local, same rule as `viewport` / `breadcrumbs`, and `projectToFile` explicitly refuses to emit a field of that name.

The "I had it the wrong way round at first" paragraph in the description matches what the code does, and `layoutIsShared` answering true for an **SSH** project as well as a folder one is the detail most people get wrong — an SSH project's `project.json` sits on the host where everyone reads it, so gating on `cwd` alone would have told those users their delete was private.

I also checked the thing this could have broken: `recordLayoutViewport` calls `markWorkspaceDirty()`, but since `layoutViewports` never reaches the file, `WorkspaceStore.save`'s skip-unchanged path means a camera-only change writes **no** project file and bumps **no** rev. So restoring a layout does not produce a git diff for your teammates. Good.

Sanitization on both serializer seams, plus the inline (cwd-less) branch in `WorkspaceStore.load` and the `migrateLegacy` path, is thorough — those two extra branches are exactly where this class of bug hides, and they are covered.

## Transform

`applyLayout` is the right shape: an explicit two-pass (resolve every target root origin, then emit) rather than N placements, root-space rects, drop-whole-on-bad-input, and the frame-grow rule that keeps `extent: 'parent'` from clamping an out-of-layout child. Restore returning the **same array** when it addressed nothing live is the small detail that stops a no-op from writing a shared-file revision. The `setViewport`-never-`fitView` rule is honoured.

## Nits (non-blocking)

1. `src/renderer/components/Dock.tsx` — the `<Tooltip label="Update to the arrangement on screen" …>` line starts at column 0 while its siblings are indented. Stray formatting.
2. `saveCanvasLayout` and `renameCanvasLayout` omit `promptDialog` from their `useCallback` deps (the others are complete). Harmless if it is stable, but it is the one inconsistency in an otherwise exact set.

## The one thing that is not mine to decide

**Should a named layout be git-shared at all?** Your reasoning is sound and I am not asking you to change it. But the counter-case is real and the description half-concedes it: a layout named "Ultrawide" is named after *my* monitor, and it lands in everyone's `git status` when I save, rename or delete one. You have already disclosed all of it (commit-able diff, delete reaches everyone, merge can land two same-named layouts, older builds drop the field on first save), so this is a product call, not a defect — and if it turns out to be noisy, the escape hatch you describe (a machine-local tier) is additive.

Not verified here (headless Linux server): the four hand-test gaps you list — the nested-frame clamp case, collapse round-trip, the counts in the restore notice, and the hand-edited `"x": null` drop. The transform and the sanitizers carry unit tests for all four, so the debt is a device pass, not a hole.

PR #715, opened 2026-09-09
feat(agents): choose the default gateway model

dkattan:
New canvas agent sessions can inherit a chosen discovered gateway model. Settings distinguish the CLI default through the gateway, a configured gateway model, and subscription routing, while migrating the existing subscription preference.

Depends on https://github.com/eneskirca/nodeterm/pull/380 for subscription relaunch support. GitHub cannot base this upstream PR on that fork-only branch, so its existing commits remain visible until #380 lands.

This draft was opened before splitting the original feature branch. It now contains only default-model selection; the other intents are separate PRs:

- https://github.com/eneskirca/nodeterm/pull/716 — Scan live agent session health and identity.
- https://github.com/eneskirca/nodeterm/pull/717 — Report whether command delivery actually submitted.
- https://github.com/eneskirca/nodeterm/pull/722 — Verify foreground agent exit and replacement startup.
- https://github.com/eneskirca/nodeterm/pull/718 — Allow selecting text in subagent transcripts.
- https://github.com/eneskirca/nodeterm/pull/719 — Configure and validate the model discovery path.
- https://github.com/eneskirca/nodeterm/pull/720 — Inspect masked session environments with accurate freshness.
- https://github.com/eneskirca/nodeterm/pull/723 — Apply discovered context limits with a scoped launch catalogue.
- https://github.com/eneskirca/nodeterm/pull/724 — Route Claude subagents to a served gateway model.
- https://github.com/eneskirca/nodeterm/pull/725 — Preserve the launch model and context window across meter views.
- https://github.com/eneskirca/nodeterm/pull/726 — Group affected sessions for recovery and verify each repair.

All eleven PRs target upstream `main`. The dependent changes retain their prerequisite commits, so their GitHub diffs remain cumulative until those prerequisites land. Each dependent PR links its prerequisites and a comparison showing only its own change.

Rebased onto upstream `main` at `e2e3e3479f24`. The prerequisite #380 also includes the maintainer-requested guard that preserves inherited credentials in plain terminals; this PR carries that guard into the three-way launch mode.

Validation: 67 focused launch/settings tests passed on this branch, including plain-terminal credential regressions. The combined stack passed typecheck and focused rebase tests; fresh Linux and Windows hosted quality checks (including builds), CodeQL, and dependency review pass on the final published head. Remaining full-suite Mac/tmux failures reproduced on unchanged current `main`.

---
This is **part 2 of 7 in a stack** made with GitButler:
[...]

PR #716, opened 2026-09-09
feat(diagnostics): scan live agent session health

dkattan:
Adds `npm run diagnose:agents --` to correlate persisted agent nodes with tmux panes, foreground processes, launch environment, and visible context limits. The scan reports exited agents, mismatched models/windows, and actionable terminal errors; optional JSONL output excludes captured screen and argv text.

Reads authoritative inline project files before compatibility caches and verifies executable positions instead of treating any argument named after an agent as process proof. Process inspection requires POSIX `ps`.

Split from #715 and independent of gateway-default settings.

Validation: 15 scanner tests passed, including regressions for stale workspace data and unrelated foreground arguments; `npm run typecheck` and `git diff --check` passed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "typecheck": "tsc --noEmit -p tsconfig.node.json && tsc --noEmit -p tsconfig.web.json",
     "test": "vitest run",
     "test:unit": "vitest run test/remote",
+    "diagnose:agents": "node scripts/agent-session-scan.mjs",
     "rebuild": "node scripts/patch-node-pty.mjs && electron-rebuild -f -w node-pty,smart-whisper",
     "server:build": "esbuild src/server/main.ts --bundle --platform=node --format=cjs --outfile=out/server/main.cjs --external:node-pty --external:ws --external:smart-whisper --tsconfig=tsconfig.node.json",
     "host:build": "esbuild src/session-host/host.ts --bundle --platform=node --format=cjs --outfile=out/session-host/host.cjs --external:node-pty --tsconfig=tsconfig.node.json",
--- a/scripts/agent-session-scan.mjs
+++ b/scripts/agent-session-scan.mjs
@@ -1,7 +1,8 @@
 #!/usr/bin/env node
 
 /**
- * Local, read-only field diagnostic for Nodeterm agent sessions.
+ * Local field diagnostic for Nodeterm agent sessions. Scanning does not mutate Nodeterm state;
+ * the explicit `--append` option writes the metadata-only report chosen by the caller.
  *
  * It joins four independently useful facts:
  *   - the node's persisted launch record (workspace/project files),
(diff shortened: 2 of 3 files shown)


eneskirca:
## Verdict: MERGE-READY (two small follow-ups, neither blocking)

Reviewed against CLAUDE.md. **Measured** on this machine: merges clean onto `main` (422767af), `tsc -p tsconfig.node.json` and `tsconfig.web.json` both exit 0, and `src/core/agent-session-scan-script.test.ts` passes **15/15**.

### On the leak surface (the reason this got a close look)

This is the right shape, and better than a denylist:

- `SAFE_ENV_KEYS` is a 4-entry **allowlist**, so a credential-shaped variable is never read at all — not merely masked.
- `reportLogLines` destructures `screen` and `foreground` **out** before serializing, so terminal text provably cannot reach `--append`. The header comment and the code agree.
- `--append` writes `mode: 0o600`.
- `spawnSync` with an argv array throughout — no shell, no interpolation.
- `capture-pane -t "<session>:"` vs `show-environment -t "=<session>"`: the comment explaining why the two targets differ is correct, and is the kind of detail usually found the hard way.
- `ok:false` is preserved end to end and exits 1, distinct from `--fail-on-findings`' exit 2. This is the rule CLAUDE.md states most often, and it is honoured here.
- `loadWorkspaceNodes` mirrors `isInlineProjectFileId` and follows the v3 index through both `inline-projects/<id>.json` and `<cwd>/.nodeterm/project.json`, in the same precedence as `WorkspaceStore.loadV3`.

`readProcesses`/`readScreen` returning `[]`/`''` on failure looked like `ok:false` laundering, so I traced it: an empty process list lands on `state: 'unknown'`, and an empty screen produces no signals and no window comparison. Neither fabricates a finding. Fine as written.

### Follow-ups (not blocking)

1. **An idle machine exits 1.** `tmux list-panes -a` exits non-zero with *"no server running on …"* when no server exists, and `scan()` maps that to `ok:false` → exit 1. CLAUDE.md's session-memory section makes exactly this distinction (a socket with **no tmux server** is an ANSWER, not a failure — `isNoServerError`). This errs in the safe direction, but it makes the script unusable in CI on a quiet machine. Reusing that classifier would give `ok:true` with zero sessions.

2. **Only one socket, and the default silently under-reports.** `DEFAULT_SOCKET = 'node-terminal'` misses `nodeterm-rmt`, and `--socket` takes a single value so both cannot be scanned in one run. CLAUDE.md documents that both sockets coexist on one machine (a host running `nodeterm-server` while being SSH'd into is exactly that case) and that a sweep owes both. Iterating the shared socket list would close it.

### Minor, take or leave

- `defaultUserDataDir()` hardcodes `node-terminal`, correct for `electron-vite dev` (verified — `~/.config/node-terminal` exists here) but not for a packaged build, where electron-builder's `productName: "nodeterm"` applies. `--user-data` is the escape hatch; worth one line in `HELP` saying so, since "field diagnostic" implies the packaged app.
- `classifyProcesses`' `'plain-terminal'` branch is unreachable from `scan()` (`if (!expectedAgent) continue` returns first), so the test covering it pins something production never does.
- `title` and `cwd` do land in the durable log. Defensible for a local 0600 admin tool, but a `/rename`d title is somewhat conversation-derived — worth a sentence in the header comment, which currently reads as if nothing content-derived is retained.
- `appendFileSync`'s `mode` applies on **create** only; an existing looser-permissioned file is not tightened.

PR #717, opened 2026-09-09
fix(terminal): report command submission outcome

dkattan:
Command delivery now reports whether the final Enter write succeeded. Cancelled deliveries and failed writes report failure, allowing callers to distinguish submitted text from text still sitting in a terminal. The completion callback runs exactly once even when the callback itself throws.

Independent split from https://github.com/eneskirca/nodeterm/pull/715. The verified-restart draft consumes this result.

Validation: 18 focused delivery tests, TypeScript checks, and the production build passed; the patch applies cleanly to current main.

### DIFF
--- a/src/renderer/terminal/command-delivery.test.ts
+++ b/src/renderer/terminal/command-delivery.test.ts
@@ -1,3 +1,4 @@
+import type { DeliveryOutcome } from './command-delivery'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import {
   DELIVERY_ATTEMPTS,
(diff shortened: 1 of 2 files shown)


eneskirca:
## Verdict: MERGE-READY — one wording point before you merge

**Measured**: merges clean onto `main` (422767af), both typechecks exit 0, and `command-delivery.test.ts` + `agent-restart.test.ts` + `session-ready-signal.test.ts` pass **114/114**. Mutation check: forcing `outcome = 'submitted'` in the catch gives **1 failed / 20 passed**, so the modified test genuinely discriminates rather than restating the source.

The change is correct and mirrors a leg that was already right — `src/session-host/session.ts:565-568` has always done `if (!write('\r')) return; finish()`. `'cancelled'` matches its documented meaning ("or the transport threw"). `return () => finish()` also closes a real hole: `finish` has a defaulted `outcome` parameter, so handing the raw function out let any caller that passed an argument inject an outcome.

### The one thing to fix, most likely in the PR body

The stated motivation — *"a rejected Enter auto-dismissed a restart as successful"* — is **not closed by this PR**. `src/renderer/terminal/agent-restart.ts:312` declares `const settle = (): void => {` with no parameter, so `performResumePhase` discards the outcome and still returns `'resumed'`. The only reader of the outcome anywhere is `src/renderer/nodes/TerminalNode.tsx:3325`, and it acts on `'line-too-long'` only.

So the report is now accurate but nothing consumes the accuracy yet. Either take `settle(outcome)` and fail the restart on `'cancelled'` in this PR, or reword the body so it claims the (real, useful) groundwork rather than the fix.

### Test nit

`command-delivery.test.ts:246` ("invokes a throwing settlement callback exactly once") passes on `main` too — the old code also called it once and threw. Adding `expect(f.writes.at(-1)).toBe('\r')` would pin what actually changed: Enter written *before* the callback throws.

PR #718, opened 2026-09-09
fix(subagents): allow transcript text selection

dkattan:
Enable drag selection and copying in subagent transcript text by overriding the canvas node's selection suppression. The text cursor follows the selectable content.

Independent CSS-only split from https://github.com/eneskirca/nodeterm/pull/715.

Validation: inspected the six-line selector change and passed git diff --check. No runtime code changes.

### DIFF
--- a/src/renderer/styles.css
+++ b/src/renderer/styles.css
@@ -6077,6 +6077,12 @@ button.group-node__setup:disabled {
   color: #d6d6d6;
   white-space: pre-wrap;
   word-break: break-word;
+  /* The transcript is TEXT, not a terminal emulator: drag-select + copy must work (the node
+     root's canvas-drag suppression must not swallow it — keep `nowheel` for scroll, but the
+     selection below is what makes the content usable). */
+  user-select: text;
+  -webkit-user-select: text;
+  cursor: text;
 }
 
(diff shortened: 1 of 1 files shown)


eneskirca:
## Verdict: MERGE-READY

**Measured**: merges clean onto `main` (422767af); web typecheck exits 0.

The fix is genuinely needed rather than cosmetic — `@xyflow/react/dist/base.css:178-182` sets `.react-flow__node { user-select: none }`, which is what was swallowing the selection inside an expanded subagent transcript.

It matches three existing precedents in the same stylesheet verbatim in shape (`.term-md__content` at `styles.css:2454-2457`, and the chat panel at `:8242-8245`). Omitting `-moz-user-select` matches those precedents and is harmless — Firefox aliases the unprefixed property.

Nothing is swallowed as a side effect: `.subagent-node__term` already carries `nodrag nowheel` (`SubagentNode.tsx:84`) and has no click handler; only `__head` does (`:66`).

No test. The repo does have a CSS-guard pattern (`styles.pills.test.ts`, `styles.theme.test.ts`), so a small guard would be in-house style — optional for six lines of CSS, not a blocker.

PR #719, opened 2026-09-09
feat(gateway): configure the model discovery path

dkattan:
Gateways can expose their model catalogue at /v1/models, /openai/v1/models, or a custom path under the saved gateway URL. Settings now provide that choice, preview the effective endpoint, and refresh discovery when the path changes. Path validation keeps credentials on the configured host and falls back for unsafe input.

Independent split from https://github.com/eneskirca/nodeterm/pull/715.

Validation: 21 focused route/settings tests, TypeScript checks, production and Codex relay builds, and diff checks passed using the lockfile dependencies. Regression coverage verifies that selecting a path persists it and updates the displayed endpoint.

### DIFF
--- a/src/core/agent-env-ipc.ts
+++ b/src/core/agent-env-ipc.ts
@@ -28,7 +28,7 @@ async function discoverModels(
   saved: ModelGatewaySettings | undefined,
   storedSecret: string | null
 ): Promise<ModelDiscoveryResult> {
-  const routes = modelGatewayRoutes(settings?.baseUrl ?? '')
+  const routes = modelGatewayRoutes(settings?.baseUrl ?? '', settings?.discoveryPath)
   const apiKeyField = settings?.apiKey ?? ''
   // SECURITY GATE — this handler is reachable by relay peers and Server Edition WS clients, and
   // `settings` (URL included) is caller-supplied. `${secret:…}` / `${env:…}` references resolve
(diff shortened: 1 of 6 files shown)


eneskirca:
## Verdict: CHANGES-NEEDED (small — the security core is sound)

**Measured**: merges clean onto `main` (422767af), both typechecks exit 0, 39 tests pass.

I ran an adversarial probe of 17 inputs through `modelGatewayRoutes`. **Every one resolved to `host=gw.example.test`** — absolute URL, `..`, `%2e%2e` (blocked because `%` is not in the character class), `\`, `?`, `#`, NUL, >500 chars, and no-leading-slash all degrade to `/v1/models`. `//evil.test/x` and `///evil.test` pass through, but only as a *path* after the authority; the host is unchanged. The design is right: path-suffix-only, so the host can never move, which is why the credential trust gate is correctly left untouched — `agent-env-ipc.ts:31` still gates `${env:}`/`${secret:}` resolution on baseUrl byte-identity, and the new comment at `:40-44` is accurate.

Both shells get it automatically: the change is `src/shared` + `src/core/agent-env-ipc.ts`, and `registerAgentEnvIpc` is booted from both `src/main/index.ts:1305` and `src/server/index.ts:235`. No migration is needed — `mergeSettings` (`src/core/settings-store.ts:19-22`) spreads `saved?.modelGateway` over the default, so an old settings.json simply has no key, and the default is byte-identical (pinned by test).

### Two things to fix

1. **An invalid custom path is a silent no-op.** `sanitizedGatewayDiscoveryPath` is exported but has **zero callers outside `model-gateway.ts`** — `ModelGatewaySection.tsx` writes `patchGateway({ discoveryPath: e.target.value || undefined })` straight through. The only feedback is the preview quietly showing `/v1/models`, and since the preview is `{routes ? … : null}`, an empty `baseUrl` means it is not rendered at all — so the rejection is *fully* invisible. Call the exported sanitizer in the section and say "not a valid path — using /v1/models".

2. **CLAUDE.md is not updated.** The "Model gateway / switcher" bullet still describes discovery as a fixed derivation. One sentence covers it: the suffix is configurable via `settings.modelGateway.discoveryPath`, path-suffix-only, host never movable.

### Note for the stack

`#723`–`#726` carry a **stale squashed copy** of this work (`24095c34`) and are the reason those PRs conflict. This branch is the current version and merges clean — land it from here and drop that commit when rebasing the stack, rather than reconciling two versions of the same change.

Tests are real behaviour tests (they assert the degradations), not text-restating. Two safe-but-unpinned cases worth a line: no-leading-slash (`v1/models`) and `//host`.

PR #720, opened 2026-09-09
feat(agents): inspect masked session environments

dkattan:
Click a session context meter to inspect its environment. Fresh local sessions expose a masked capture; warm local and SSH sessions read their tmux environment. Pending or warm session-host attachments, and sessions with no live backend identity, report unavailable. Reopening the drawer or changing sessions clears previous results and ignores late responses.

Credential names and sensitive values are masked before leaving core. The tmux reader parses complete shell-format output without evaluating it, keeping multiline values together until masking and refusing malformed records. Captures become visible only after the backend confirms a fresh generation; a reattach cannot mislabel its new client environment as the older pane's spawn environment.

Validation: 159 focused tests passed on Node 22; TypeScript check and production/Codex relay builds passed. The local full-suite failures from the initial run were reproduced on unchanged `main` and traced to the local Node/native/macOS/tmux setup. Hosted checks run against this independent draft.

Split from https://github.com/eneskirca/nodeterm/pull/715. This branch targets `main` and has no gateway-launch or recovery prerequisites.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1198,6 +1198,14 @@ else, and its context links must keep classifying across restarts).
   the shared mapping. Desktop and Server Edition use the same core handler; relay tabs deliberately
   do not apply this machine's gateway to another core. Mobile needs a settings/model-picker surface
   before it can expose the feature.
+  **Spawn-environment visibility:** `pty.envInfo(persistKey)` (`pty:env-info`) supplies the context
+  meter on canvas nodes, kanban cards and card modals with the environment for that session. A
+  current local spawn returns the masked core capture (`source: 'spawn'`); an older local or SSH
+  session falls back to its tmux session environment (`source: 'tmux'`); a missing session or failed
+  read returns the explicit `unavailable` state. Credential-shaped names and values are irreversibly
+  masked in `core/pty-env-info.ts` before crossing IPC. The fallback parses `show-environment -s`
+  without evaluating its shell output, including multiline values and tmux's `unset NAME;` form
+  (the plain output spells the same removed-variable fact as `-NAME`).
 - **Grok** (`@xai-official/grok` 1.0.0, builtin since 2026-08) — in `AGENT_HOOK_TARGETS`,
   `RESUMABLE_AGENTS`, `RENAME_CAPABLE`, `PERMISSION_MODE_CAPABLE`, `CANVAS_CONTROL_CAPABLE`,
   `CONTEXT_LINK_CAPABLE`, `CHAT_CAPABLE`, `TRANSFER_SOURCE_CAPABLE`, `USAGE_CAPABLE` and
--- a/src/core/pty-env-info.test.ts
+++ b/src/core/pty-env-info.test.ts
@@ -1,0 +1,160 @@
+import { describe, expect, it } from 'vitest'
+import {
+  isSecretEnvKey,
+  isSecretEnvValue,
+  maskPtyEnv,
+  maskPtyEnvVar,
+  parseTmuxSessionEnv
+} from './pty-env-info'
+
+describe('isSecretEnvKey', () => {
+  it('marks the credential-shaped names the env layers actually write', () => {
+    expect(isSecretEnvKey('ANTHROPIC_AUTH_TOKEN')).toBe(true)
+    expect(isSecretEnvKey('ANTHROPIC_API_KEY')).toBe(true)
+    expect(isSecretEnvKey('OPENAI_API_KEY')).toBe(true)
+    expect(isSecretEnvKey('COPILOT_PROVIDER_API_KEY')).toBe(true)
+    expect(isSecretEnvKey('MY_CUSTOM_OAUTH_SECRET')).toBe(true)
+    expect(isSecretEnvKey('my_project_password')).toBe(true)
+  })
+
+  it('does NOT mark plain config names — including suffix look-alikes', () => {
+    // The gateway / autocompact / hook vars carry no credential.
+    expect(isSecretEnvKey('ANTHROPIC_BASE_URL')).toBe(false)
+    expect(isSecretEnvKey('CLAUDE_CODE_AUTO_COMPACT_WINDOW')).toBe(false)
+    expect(isSecretEnvKey('CLAUDE_AUTOCOMPACT_PCT_OVERRIDE')).toBe(false)
+    expect(isSecretEnvKey('CLAUDE_CODE_SUBAGENT_MODEL')).toBe(false)
+    expect(isSecretEnvKey('CLAUDE_CONFIG_DIR')).toBe(false)
+    expect(isSecretEnvKey('NODETERM_HOOK_ENDPOINT')).toBe(false)
+    // Socket/pid names end in SOCK/PID and must not mask.
+    expect(isSecretEnvKey('SSH_AUTH_SOCK')).toBe(false)
+    expect(isSecretEnvKey('SSH_AGENT_PID')).toBe(false)
+  })
+})
+
+describe('isSecretEnvValue', () => {
+  it('finds credentials hidden behind ordinary URL and connection-string names', () => {
+    expect(isSecretEnvValue('postgres://app:super-secret@db.internal/app')).toBe(true)
+    expect(isSecretEnvValue('jdbc:postgresql://app:super-secret@db.internal/app')).toBe(true)
+    expect(isSecretEnvValue('Server=db;User Id=app;Password=super-secret')).toBe(true)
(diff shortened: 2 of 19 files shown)


eneskirca:
## Verdict: CHANGES-NEEDED (one measured leak, one undiscussed exposure; the architecture is right)

**Measured** on this machine: merges clean onto `main` (422767af), both typechecks exit 0, and the six touched suites pass **159/159**.

The core design decision is correct and worth keeping: masking happens in `src/core/pty-env-info.ts` **before** the session record is written, so even the `tmux show-environment` fallback cannot emit a raw value, and there is no reveal path in the popover. `remoteShowEnvironmentArgs` interpolates `sessionName(persistKey)`, which is already sanitized to `[A-Za-z0-9_-]`, so the remote leg is not a shell-injection surface. All good.

Two things should change before this ships.

### 1. BLOCKING — `maskSecret` reveals most of a short secret

`maskSecret` shows `slice(0,3)` + `(len - 8)` bullets + `slice(-4)`, which means **7 characters are always visible regardless of length**. I ran it:

```
SHORT_SECRET len=9  -> abc•fghi          (8 of 9 characters visible)
API_KEY      len=11 -> hun•••pass        ("hunter2pass")
DB_PASSWORD  len=11 -> Tr0•••or&3        ("Tr0ub4dor&3")
MY_TOKEN     len=12 -> abc••••ijkl
GH_TOKEN     len=20 -> ghp••••••••••••G7h8
```

For anything in the 9–16 character band the majority of the credential is on screen, in a popover that is routinely screenshotted and screen-shared. Separately, the bullet run is `value.length - 8`, so the mask publishes the secret's **exact length** in every case.

Suggested: mask entirely below a threshold (~20 chars), and use a fixed-width bullet run so length is not disclosed. Showing only the last 4 above the threshold would match common practice; the leading 3 buys diagnosis very little.

### 2. BLOCKING (decision, not code) — a relay guest reads the host's environment

`relay-api.ts` builds `pty: { ...real.pty, … }`, so `pty.envInfo` on a relay tab resolves against the **host's** core. A guest in a shared session therefore gets:

- every environment variable **name** on the host, unmasked (names alone can carry customer/host/project identity — `ACME_PROD_*`), and
- every non-secret **value** in full: `PATH`, `HOME`, `PWD`, internal endpoint URLs, `AWS_PROFILE`, and so on.

That may well be acceptable for how relay sessions are actually used, but it is a real widening of what a guest can see and the PR does not discuss it. Either gate the popover off for relay tabs, or state the decision in the PR body and in the CLAUDE.md paragraph this PR adds.

### Non-blocking notes

- **The denylist's stated rationale does not match what it does.** The comment argues name-matching means "a future env layer … can introduce a key name with no ancestor in this file and it still masks" — true against an exact-name list, but a *new credential with an unfamiliar name still prints in full*. Measured miss: a JWT value (`eyJhbGci…`) is not caught by `isSecretEnvValue`, so e.g. `GH_PAT` (no listed substring, no URL/PEM/`k=v` shape) renders verbatim. The value-level check does real work — `postgres://u:p@h/db` and `redis://:pw@h:6379` are both caught, `https://api.example.com/v1` correctly is not — it just cannot be the backstop the comment implies. Worth softening the comment, and consider adding high-entropy known prefixes (`ghp_`, `github_pat_`, `xoxb-`, `sk-`, bare JWT shape).
- **Substring matching over-masks.** Measured: `MONKEY_CONF`, `TURKEY_MODE` and `LC_KEYBOARD` all classify as secret (`key`). Safe direction, but it degrades the diagnostic the feature exists for. Word-boundary matching (`_KEY`, `KEY_`, `APIKEY`) would keep the coverage and drop these.
- `SECRET_NAME_EXEMPT` contains `NODETERM_HOOK_SOCK`, which matches no entry in `SECRET_NAME_PARTS` — dead entry.
- Discoverability: `ContextMeter` returns `null` until usage data exists and only renders for `USAGE_CAPABLE` agents, so "inspect the spawn environment" is unreachable on a plain terminal, on grok/copilot, and on any agent node before its first turn. If this is meant as a general diagnostic, the meter is a narrow home for it.
- `console.warn` in the tmux failure path prints `err.message`, and `promisify(execFile)` folds stderr into it — low risk here, but it is the pattern CLAUDE.md calls out elsewhere.

eneskirca:
## Review — take as is (two optional hardening fixes)

**Verdict: mergeable as-is.** This is the one PR in the group I'd take without changes.

### Structural note (matters for how this gets merged)

**#720 is not part of the #722→#756 stack.** Its two commits (`acb33828`, `510b51f7`) do not appear in #722 or anything above it:

```
#720: 510b51f7 fix(terminal): preserve environment source accuracy
      acb33828 feat(terminal): inspect session environment
#722: d6e21105 fix(agents): allow verified resume fallback
      a9253ff8 fix(terminal): make agent restarts verifiable     <- no #720 commits
```

The stack is #722 → #723 → #724 → #725 → #726 → #756. #720 stands alone against `origin/main` and can merge independently, in any order. It does share 11 files with #722, so whichever lands second may need a trivial rebase.

### What I ran

Isolated worktree at `510b51f7` off `origin/main` (`e43d3a54`):

- `npm run typecheck` — clean (both tsconfigs).
- `vitest run` on `pty-env-info`, `ContextMeter`, `pty-single-user`, `pty-session-host`, `control-master`, `preload/index` — **6 files, 165 tests, all passing.**

### Security review (the part I looked hardest at)

This PR reads session environments, so I checked exposure rather than taking the header comment's word for it.

- **Masking happens in core before the record is written** (`maskPtyEnv` at the `pendingSpawnEnv` assignment in `pty-manager.ts`, and `maskPtyEnvVar` inside `parseTmuxSessionEnv`). Both query paths are covered. Correct placement.
- **No reveal path.** The popover renders `variable.value` only, and the `title` for a secret is the literal string `'masked credential'`, not the value.
- **A relay guest cannot reach this.** `src/main/remote/host-service.ts` `onRpc` is a closed allowlist (`pty.attach`/`pty.kill`/`pty.scroll`, fs, git, `projects.*`) with `default: socket.respond(req.id, false, …)`. There is no `pty.envInfo` verb, so nothing new is exposed to a paired device.
- **No credential is introduced into argv.** `remoteShowEnvironmentArgs` interpolates only `sessionName(persistKey)`, which sanitises to `[A-Za-z0-9_-]`. Given node ids are attacker-controlled through `project.json`, I specifically checked this — it's safe.
- **Nothing sensitive is in the env nodeterm itself injects**: `hookServer.buildPtyEnv` deliberately carries no bearer and no per-node token (the 2026-08-13 argv-leak fix), so the only credentials visible here are ones the user's own shell already exported.
- **`spawn` vs `tmux` source is handled correctly.** `pendingSpawnEnv` is only promoted to `spawnEnv` when the pre-spawn probe proved `fresh`, and SSH/detached paths publish no candidate at all. A warm re-attach correctly falls back to the tmux session env rather than publishing the freshly-composed *client* env. That is a subtle distinction and it's right.

### Optional fixes (non-blocking)

Both are defence-in-depth. The surface shows the user their own environment, which they could read with `env` in the very pane being inspected — so neither is a trust-boundary break.

1. **`SECRET_NAME_PARTS` has reachable gaps.** `GPG_PASSPHRASE`, `RAILS_SECRET_KEY_BASE`-style `…_SALT`, `…_SIGNING_…` and `…_PRIVATE_…` names don't match any part, and their values don't trip `isSecretEnvValue` either. Adding `passphrase`, `private`, `signing`, `salt` costs nothing and the list is explicitly designed to be name-based rather than exhaustive. A value-prefix check for the common token shapes (`sk_live_`, `sk-`, `ghp_`, `gho_`, `xox[bap]-`, `AKIA`) would close most of the rest.

2. **`maskSecret` leaks length, plus first-3/last-4.** `'•'.repeat(value.length - 8)` makes the exact secret length visible, and for a 12-character value 7 of 12 characters are shown in clear. A fixed-width mask (`value.slice(0,3) + '••••••••'`) keeps the "is this the key I think it is?" diagnostic value without the length oracle.

### One design question

`ContextMeter` returns `null` before the env block when there's no usage (`if (!usage) return null`). So "Spawn environment" is only reachable on a `USAGE_CAPABLE` agent that has *already emitted usage data*. A plain terminal, an SSH node, or a grok node can never inspect its environment — and those are arguably the cases where you'd most want to (is `NODETERM_HOOK_ENDPOINT` set? did `CLAUDE_CONFIG_DIR` land?).

Not a blocker, and I understand the meter popover is a convenient existing surface. But the coupling is incidental rather than intentional, and it will read as a bug the first time someone tries to debug a non-agent node. Worth either a separate render gate or its own home in a follow-up.

*Reviewed against `origin/main` @ `e43d3a54`.*

PR #722, opened 2026-09-09
fix(agents): verify process exit and restart completion

dkattan:
Agent restarts now stop only the verified foreground agent process, await session recycling, and confirm that the replacement agent actually starts. Desktop, SSH, preload, and browser transports share the same explicit outcomes. Cold restore uses upstream transcript-presence checks and the missing-conversation fallback, then verifies the actual replacement process. The acknowledgement deadline covers both resume and fresh-fallback delivery.

A resume-delivery timeout cancels pending delivery and fails instead of reporting success while a command can still submit later. Refusals retain a concrete reason for callers.

Prerequisites: https://github.com/eneskirca/nodeterm/pull/380, https://github.com/eneskirca/nodeterm/pull/715, and https://github.com/eneskirca/nodeterm/pull/717. This PR targets `main`; its GitHub diff includes prerequisite commits until those changes land. [Review this change alone](https://github.com/dkattan/nodeterm/compare/1dfa5c16cbbde043b5933519256d3e9bebbc704b...6d67dcd5678db8e365a8e5d54d39fa487ef81e2c).

Replaces https://github.com/dkattan/nodeterm/pull/7. Split from https://github.com/eneskirca/nodeterm/pull/715.

Validation: focused delivery, restart, process-proof, acknowledgement, cold-transcript/fallback, readiness, and remote-delete suites passed after rebasing onto current main. The combined stack passed typecheck; fresh Linux and Windows hosted quality checks, CodeQL, and dependency review all pass on this published head. The local Mac full-suite failures remaining after dependency and source-assertion corrections reproduced on unchanged current main.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -687,20 +687,29 @@ Lifecycle, by intent:
   agent-restart.ts` restarts the agent CLI *inside* the pane and leaves the PTY, the tmux session
   and its scrollback untouched. It exists for **new-model pickup** — a freshly released model only
   shows up in a CLI's model list on a fresh launch, and doing that by hand means closing and
-  re-resuming every agent node on the canvas. Choreography: write the CLI's own exit line (`/exit`
-  for claude, `/quit` for codex — that table is also the gate, an agent not in it can never be
-  restarted in place), poll `pty:pane-command` (`#{pane_current_command}`, local tmux socket or the
-  project's SSH ControlMaster; any failure reads as "not a shell yet") every `RESTART_POLL_MS`
-  (250 ms) until a SHELL owns the pane, then echo-deliver `resumeCommand(...)` — the same
-  `claude --resume` / `codex resume` the cold restore uses. **Nothing is ever killed**: if the CLI
-  has not quit within `RESTART_EXIT_TIMEOUT_MS` (6 s) the run reports `exit-timeout` and leaves the
-  session running. A `working` **or `blocked`** session is refused — `/exit` typed into a
-  permission prompt would ANSWER it, not quit — and a node is held one-restart-at-a-time until the
-  resume line has actually LEFT the pane (an un-submitted line is where a second `/exit` would be
-  spliced in). The bulk action runs the same per-node closure sequentially over every idle agent
+  re-resuming every agent node on the canvas. Choreography: preflight `pty:pane-command`, then call
+  the ONE core stop mechanism, `pty.terminateForeground(nodeId, sourceAgentId)`: `paneOwner` proves
+  the expected harness from full argv, `agentPidIn` retains the exact PID that verdict rested on,
+  and the local/SSH kill leg re-proves that PID is still live in the foreground group immediately
+  before SIGTERM. **Nothing is typed into the pane.** The two PIDs are distinct facts: `PaneOwner.panePid` is the
+  long-lived login shell tmux spawned, while `agentPidIn(PaneOwner)` is the current harness PID;
+  the latter is re-read at the action boundary rather than persisted, because an exited process's
+  PID can be reused. If a successful kernel probe finds the expected agent absent, core returns
+  `already-exited`: local restart force-recycles the whole session without misidentifying its
+  current foreground owner as the agent, then cold restore resumes the conversation. An unreadable/unknown probe still
+  refuses (`lost-foreground`). Poll every `RESTART_POLL_MS` (250 ms) until a shell owns the
+  pane, then echo-deliver `resumeCommand(...)` — the same `claude --resume` / `codex resume` cold
+  restore uses. If the group does not release within `RESTART_EXIT_TIMEOUT_MS` (6 s), report
+  `exit-timeout` and do not resume. A `working` **or `blocked`** session is still refused by policy:
+  ordinary restart must not discard active work or a pending decision merely because the PID stop
+  is safe. The node is held one-restart-at-a-time until the resume line has actually LEFT the pane,
+  so a second restart cannot terminate the replacement or splice another launch line into it. The
+  bulk action runs the same per-node closure sequentially over every idle agent
   node in canvas order and reports one summary line. `performRestartResume` is now a COMPOSITION of
   `performExitPhase` + `performResumePhase` (2026-08-12, behavior-pinned split) — hibernation
-  drives the halves separately; each half refuses independently.
+  drives the halves separately; each half refuses independently. As of 2026-09-01 every production
+  caller, including Eco, supplies that same identity-gated terminator; `performExitPhase` has no
+  in-band exit-write path.
 - **Agent hibernation ("Eco", 2026-08-12, OPT-IN default off)** → `settings.agentHibernationEnabled`
   (+ `agentHibernationIdleMinutes`, default 30; Settings → Agents): a 60 s renderer sweep
   (`Canvas`) exits the CLI of up to **2** agent nodes per pass that are hook-idle in state `done`,
@@ -713,7 +722,7 @@ Lifecycle, by intent:
   share ONE `guardConcurrentRestart` set. Load-bearing rules a refactor must not undo:
   (1) **recurring fact is durable** — both loop-card dismiss surfaces route through
   `lib/loopCard.ts`, which HIDES a cron/schedule card but retains `agentStatus.loop`
-  (`dismissed: true`); clearing it would let Eco `/exit` a CLI whose cron wakeup lives in that
+  (`dismissed: true`); clearing it would let Eco terminate a CLI whose cron wakeup lives in that
   process. (2) **Fire-time re-asks**: still-offscreen, remote, eligibility — a plan-time verdict
   is stale by seconds. (3) `hibernated` **self-heals** on live hook states + SessionStart (never
   on `done` — a late Stop POST must not undo a just-performed hibernate); cold restore (`fresh`)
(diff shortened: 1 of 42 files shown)


eneskirca:
## Verdict: MERGE-READY — the premise holds; I measured it

**Measured** on this machine: merges clean onto `main` (422767af), both typechecks exit 0, and the touched renderer suites pass **693/693** across 51 files.

This PR replaces the typed-`/exit` restart with an identity-verified SIGTERM of the agent's foreground process group. That contradicts a load-bearing sentence in the old contract (*"Nothing is ever killed"*), so the two questions worth answering were whether the new mechanism is safe and whether the contract was updated honestly. Both check out.

### The contract was updated in the same PR

`CLAUDE.md` is in the diff, and the rewrite is thorough rather than cosmetic: the restart choreography, the `panePid`/`agentPidIn` distinction, `already-exited` vs `lost-foreground`, the Eco wording, the model-switch paragraph, and — correctly — the gemini `EXIT_SEQUENCES` note now saying the table entry survives but production never types it. That is the discipline the Conventions section asks for.

### The risk I could actually test

The thing SIGTERM-instead-of-`/exit` could plausibly break is transcript durability: if the CLI only checkpoints on a clean exit, `--resume` afterwards would silently lose the last turn. So I ran it end to end on this machine.

**Claude Code 2.1.268, Linux, real tmux pane:**

1. Started `claude`, sent one turn, confirmed the reply.
2. Located the foreground process group from the pane tty (`ps -o pid,pgid,stat` → `claude` as `Sl+`, pgid ≠ the login shell's) and sent `kill -TERM -<pgid>` — the same signal to the same target this PR's `terminateForeground` uses.
3. The pane returned to `bash` within ~4s; `#{pane_current_command}` read `bash`.
4. The transcript **grew from 36 to 40 lines across the SIGTERM**, the final record being a `cost-state` entry — Claude Code installs a signal handler and flushes on the way out.
5. `claude --resume <id>` brought the session back with the full conversation intact.

So the premise is verified, not assumed: on Claude, the exact-PID SIGTERM is clean and the conversation survives it. Given that the old path could type `/exit` into a permission prompt (which answers it) or into a shell that already owns the pane, this looks like a net safety improvement rather than a regression.

### Also good

- `model-respawn-trace.ts` explicitly bans argv, env values, credentials, gateway URLs and raw errors from the trace stream, and `settle()` deliberately keeps the user-facing `detail` out of the log even though today's callers only pass controlled sentences. That is the right instinct given how much of this repo's history is credentials leaking into somewhere readable.
- The respawn ack is generation-checked with a finite deadline, so a superseded attempt resolves rather than leaking a pending entry.
- The lifecycle effect still keys on `[data.respawnNonce, offscreenEpoch]`, and the ~609 lines of `TerminalNode.tsx` change land inside the spawn/resume body rather than in the mount-stable observer effects the file's invariants protect.

### Device-verify owed (not blocking this review)

The SIGTERM evidence above is **Claude on Linux only**. The same path now stops codex, gemini, grok and copilot. Each should get the same three-step check (turn → SIGTERM → resume) before this is trusted in the field, and macOS is unverified throughout. Gemini is the one I would check first, since its `/quit` history is precisely the "exit destroyed the thing we meant to resume" hazard.

eneskirca:
## Review — take with one named fix

**Verdict: the direction is right and the implementation is careful, but one behaviour change costs the user their scrollback in a common case. Fix that and I'd take it.**

### What I ran

Isolated worktree at `d6e21105` off `origin/main` (`e43d3a54`):

- `vitest run src/renderer/terminal/ src/core/pty-manager.pane-owner.test.ts src/core/pty-single-user.test.ts src/core/pty-manager-platform.test.ts src/core/remote-ssh/control-master.test.ts src/shared/model-respawn-trace.test.ts src/renderer/nodes/session-ready-signal.test.ts src/preload/index.test.ts` — **40 files, 721 tests, all passing.**

### On the architectural change

Replacing the typed `/exit` with an identity-gated SIGTERM is, I think, the right call, and it is consistent with what CLAUDE.md already blessed for model switching ("A model switch SIGTERMs the pane's foreground non-shell process group (never types `/exit`)"). The identity gate is genuinely careful: `paneOwner` reads full argv, `agentPidIn` retains the exact PID the verdict rested on, and the kill legs re-prove that PID before signalling. `unknown` refusing while `not-agent` is treated as positive evidence is the correct asymmetry. The `working`/`blocked` refusal is preserved as policy rather than quietly dropped now that the stop is PID-safe — I'm glad that reasoning is spelled out.

I also checked the two things most likely to be silently broken by a change this shape, and both are fine:

- **Shell parity holds.** `agentProcess` and the widened `terminateForeground` have real ws-bridge implementations (with a sensible `value === true` back-compat shim), and the core handlers are registered in `pty-manager.ts`, which both shells boot.
- **The `recycle` cast→request switch is safe.** `origin/main` already registers `ptyRecycle` with *both* `handleWithSender` and `onWithSender`, so `client.request` has something to answer it. No hang on Server Edition.

### Blocking: "Restart agent (resume)" now destroys scrollback when the agent already exited

**Path.** The user's `claude` has already exited — they typed `/exit` themselves, or it crashed — and a shell owns the pane. They pick **Restart agent (resume)**.

1. `performExitPhase` reads the pre-flight pane: `before = "bash"`. It is traced at `exit.preflight` and then **never consulted again**.
2. `terminateForeground` probes, `isAgentPane` returns `not-agent` (a shell is not the expected harness), and core returns `'already-exited'` (`pty-manager.ts`, `finish('already-exited', 'expected-agent-absent')`).
3. `performRestartResume` sees `already-exited` **and** `onAlreadyExited`, which `TerminalNode.tsx` supplies for every non-relay session, so it takes the force-recycle branch.
4. `recycleAndAwaitRespawn('agent-already-exited', …)` calls `transport.recycle(id)` — the tmux session is killed and respawned.

**Wrong result.** The pane's scrollback is gone. On `origin/main` the identical input resumes in place: the first poll's `isShellCommand(pane)` returns `'exited'` and the resume line is delivered into the existing pane, so the history survives.

This is also the one case where the PR leaves CLAUDE.md self-contradictory. The lead sentence of that paragraph is unmodified context in your diff:

> restarts the agent CLI *inside* the pane and leaves the PTY, the tmux session and its scrollback untouched

while the text you added says local restart "force-recycles the whole session".

**Why the stated justification doesn't hold.** The comment says the recycle avoids "misidentifying its current foreground owner as the agent". But when a *shell* owns the pane there is nothing to misidentify — `isShellCommand` already answers that question, and it's the same predicate the polling loop below uses to decide the pane is ready for the resume line. The recycle is a strictly larger hammer for a case the code can already recognise.

**Fix.** On `already-exited`, consult the pane before recycling:

- if `isShellCommand(before)` (or a fresh `paneCommand()` read) — skip the recycle and fall through to `performResumePhase`. This is exactly the path you already implement for relay sessions, where `onAlreadyExited` is undefined.
- reserve `recycleAndAwaitRespawn` for a **non-shell** owner (vim, top, a foreground build), where typing a resume line would genuinely be wrong.

That keeps every safety property of the new terminator and restores the scrollback guarantee. It should also let the lead sentence of the CLAUDE.md paragraph stand unamended.

### Non-blocking notes

1. **Transcript flush on SIGTERM is an unverified assumption.** Restart is refused while `working`, so in practice no turn is in flight and the transcript should already be on disk — but "SIGTERM leaves the resumable transcript intact" is now load-bearing for claude/codex/grok/gemini and isn't measured anywhere in the PR. Per CLAUDE.md rule 16 this deserves a numbered device-checklist item per agent rather than silence.

2. **The `working`-state bypass reads correctly, for the record.** `TerminalNode.tsx` lets a `working` node through when `(targetModel || restartShell)` and `agentProcessProofWithin` returns `not-agent`. I checked that ordinary restart can't reach it — it can't, both flags are false there — and overriding a stale *status* with a kernel *proof* is defensible. Worth a line in CLAUDE.md since it is a documented refusal being conditionally lifted.

3. `EXIT_SEQUENCES` is now a support-allowlist that is never written. The comment says so clearly, which I appreciate; just flagging that the name now means something different from what it says, and a future reader may re-wire it.

*Reviewed against `origin/main` @ `e43d3a54`. Incremental diff reviewed, not the cumulative stack.*

eneskirca:
## Addendum — main moved, and #824 now overlaps this PR's thesis

`main` advanced from `e43d3a54` to `00c0697b` (PRs #809, #822, #824, #697) and this branch is now `CONFLICTING`. Flagging what that means for the review above, since a mechanical rebase won't cover it.

### My review still stands as written

`src/renderer/terminal/agent-restart.ts` is **untouched** by everything that landed, so `performExitPhase` / `performRestartResume` are byte-identical to what I reviewed. The scrollback finding is unaffected by the rebase.

Your conflicts are in `TerminalNode.tsx`, `agentStatus.ts`, `shared/types.ts` and `pty-manager-platform.test.ts`.

### The part that isn't mechanical: #824 solves an overlapping problem differently

PR #824 ("prove the agent owns the pane before quitting it, and re-check that pane on wake") merged two days ago and added `src/renderer/terminal/wake-identity.ts` — a pure decision module, 172 lines with 211 lines of tests — plus 178 lines of `TerminalNode.tsx`. It builds on **the same primitive your PR does**, `isAgentPane(owner, agentId, binaries)`, and reaches a different conclusion:

```ts
export type ExitContextVerdict = 'agent-owns-pane' | 'not-in-this-pane' | 'unreadable'
```

> Fails CLOSED on `unknown`, unlike most of this app's probes: everywhere else a probe that cannot
> answer degrades to the bare, pre-feature behaviour, but here **the pre-feature behaviour is to type
> `/exit` into whatever is there**.

So on current `main`, Eco/hibernation **still types `/exit`** — now gated on `isAgentPane === 'agent'`. Your PR's position is that nothing should ever be typed, and your CLAUDE.md edit states that as a rule ("Never type a CLI exit sequence into a pane") while also asserting "As of 2026-09-01 every production caller, including Eco, supplies that same identity-gated terminator." That sentence is no longer true of `main`, and the rebase will put your CLAUDE.md rewrite directly on top of #824's freshly-merged rationale for the opposite choice.

I don't think that settles which design is right — yours has a real argument, and #824's own comment concedes typing into a pane is the risky part. But it is now a **design decision for @eneskirca**, not a merge conflict, and resolving the textual conflict without raising it would quietly overwrite a policy he merged on 2026-09-19. Two verdict vocabularies for "does the agent own this pane" (`ExitContextVerdict` vs your `TerminateForegroundOutcome`) living side by side is also exactly the duplicated-rule drift CLAUDE.md warns about most.

### And it strengthens the blocking finding

Note what `main` now does when the probe *succeeds* and says the agent is not there:

> `'not-in-this-pane'` is **terminal for this node until something changes**

That is the same input as your `'already-exited'` — a successful kernel read showing the expected agent absent. `main`'s newly-merged answer for an irreversible action in that state is **refuse and do nothing**. Your restart path's answer is to recycle the tmux session and drop the scrollback.

These are adjacent paths rather than the same one (hibernate vs restart), so this isn't a contradiction I'd hold you to directly. But it means the house rule for "probe answered, agent isn't here" was set two days ago and set *conservatively* — which is more reason to take the in-place resume on that branch rather than the bigger hammer.

### Suggested order

Rebase onto `00c0697b`, then decide with @eneskirca whether the terminator replaces #824's gated-typing for Eco or coexists with it — before reconciling the CLAUDE.md paragraphs, since the prose follows from that call. The scrollback fix is independent of all of it and can land either way.

*Re-checked against `origin/main` @ `00c0697b`.*

PR #732, opened 2026-09-09
feat(tabbar): tabs that read as tabs, with an options button on every one

basmilius:
The project tabs now read as tabs. The strip was a 34px segmented pill floating inside the 44px bar, which read as a control sitting on a surface rather than as the surface's own edge. Tabs now run the full height of the bar, and the active one is filled with `--canvas-bg` so its body continues into the canvas below it. Every tab also gets its options button, so reaching another project's menu no longer means switching to it first.

The project tab strip: a full-height active tab open into the canvas, an options button on every tab

## What changed, and why

**The active tab is a hole in the bar.** To let it break the separator between bar and canvas, that line moves off `.tabbar`'s `border-bottom` onto a `z-index: auto` `::after` that the raised tab paints over. In the light theme `--panel` and `--canvas-bg` are one value apart, so the active tab carries an inset hairline on its two sides. Those hairlines rendered two grades darker than the bar's own rule, because `--border` is an alpha and the bar composites it over `--panel` while the tab edges sit over `--canvas-bg` (#000 in dark). A second shadow pair in `--panel`, listed last so it paints behind, gives them the same ground and the same rendered colour. The top hairline is gone: a lid on a shape whose point is that it opens into the canvas.

**Tabs touch.** `gap: 0`, since the hairlines are the seams and a strip of bar between two tabs read as a segmented control again. The `+` keeps an 8px margin so it stays a separate action rather than the next tab in a gapless row, and it moves 6px closer to the strip it ends.

**One font weight.** The active tab was 600 against 400 elsewhere, so switching projects reflowed the whole strip. 500 for every tab keeps the metrics still. The fill already says which tab is active, and the label stays neutral rather than the project colour, which was barely legible as 13px type on a near-black strip for a dark blue or deep red project.

**The options button is on every tab.** It used to render only for the active tab, at opacity 0 until hovered, so nothing on screen said the menu existed. It now shows on every tab at a resting opacity and brightens on hover, with a tinted plate under the glyph the way a node's header buttons do. The glyph is a vertical ellipsis. A chevron reads as "expand this", and a horizontal ellipsis beside the tab name reads as a separator. Opening the menu on an inactive tab does not switch to it.

**The two tab actions are one pair.** The view toggle and the options button share one recipe (plate on hover, 14px glyph, the same opacities) and sit in a `.tab__actions` cluster with a fixed 22px square each. The board glyph runs its bars edge to edge while the three dots stand alone in the middle, so padding around each glyph gave them hover plates of two different widths. The toggle also carries an aria-label, and a comment records why it renders on the active tab only: on an inactive tab it would flip that project's view without taking you there.

**The menu closes on a press anywhere outside it.** Its dismiss backdrop sits below the bar on purpose (z 25 against the bar's 30) so a click on another tab switches in one go instead of being swallowed. That left the bar itself, its empty stretch, the brand and the `+` with no way to close the menu. A capture-phase `pointerdown` listener covers that gap without consuming the event; the caret is excluded because it owns its own toggle.

**The drop line is visible again at both ends.** Reordering worked but stopped saying where the tab would land. The line hung at `left: -3px`, which the scrolling `.tabbar__tabs` clipped for the first tab; with `gap: 0` it sat on the seam the active tab's `z-index: 1` painted over; and the end-of-strip zone never had a marker, since every other target draws its line as the `::before` of the tab it lands in front of. Now `left: 0`, `z-index: 2`, and a `.tab__dropline` flex item whose -1px side margins cancel its width so nothing shifts when it appears.

## Tests

`TabBar.test.tsx` gains drag-reorder tests (drop on a tab, drop at the end, both markers), tests for the options button on inactive tabs and for the outside-press dismiss. Two older suites in that file had been written against module-scope imports the file no longer has and are moved onto the per-test `load()`.

## Verified

Typecheck and the TabBar suite pass. I built the branch and drove it with Playwright on macOS: three projects, all three tabs measure 44px against a 44px bar, all three at weight 500, an options button on each, opening the menu on an inactive tab leaves the active tab where it was, and a press on the empty bar closes it. Not verified on a real device: the light theme (the hairline reasoning is from the CSS, screenshots were dark only), Windows and Linux, and the Server Edition, though nothing here touches a bridge or IPC.

### DIFF
--- a/src/renderer/components/TabBar.test.tsx
+++ b/src/renderer/components/TabBar.test.tsx
@@ -264,3 +264,57 @@ describe('TabBar drag-reorder', () => {
     expect(onReorder).toHaveBeenCalledWith('p1', null)
   })
 })
+
+describe('TabBar options button', () => {
+  let root: Root
+  let host: HTMLElement
+
+  beforeEach(async () => {
+    ;(globalThis as { ResizeObserver?: unknown }).ResizeObserver = NoopResizeObserver
+    Element.prototype.scrollIntoView = (): void => {}
+    ;(window as unknown as { nodeTerminal: any }).nodeTerminal = {}
+    host = document.createElement('div')
+    document.body.appendChild(host)
+    useProjects.setState({
+      projects: [project(), project({ id: 'p2', name: 'Beta' })],
+      activeProjectId: 'p1'
+    })
+    root = createRoot(host)
+    await act(async () => {
+      root.render(
+        <TabBar
+          onSwitch={vi.fn()}
+          onReconnect={vi.fn()}
+          onReorder={vi.fn()}
+          onOpenWelcome={vi.fn()}
+          onRename={vi.fn()}
+          onSetFolder={vi.fn()}
+          onCloseProject={vi.fn()}
+          onRemoteAccess={vi.fn()}
+          onSetDefaultAccount={vi.fn()}
+          onSetDefaultPermissionMode={vi.fn()}
+          onOpenProjectSettings={vi.fn()}
+        />
+      )
+    })
+  })
+
+  afterEach(() => {
+    act(() => root.unmount())
+    host.remove()
+    useProjects.setState({ projects: [], activeProjectId: '' })
+  })
+
+  it('sits on every tab, so the menu is reachable without activating it first', () => {
+    expect(host.querySelectorAll('.tab__caret').length).toBe(2)
+  })
+
+  it('opens the menu for the tab it belongs to, active or not', async () => {
+    const inactive = host.querySelectorAll<HTMLButtonElement>('.tab__caret')[1]
+    await act(async () => {
+      inactive.dispatchEvent(new MouseEvent('click', { bubbles: true }))
+    })
+
+    expect(document.querySelector('.tab-menu')).not.toBeNull()
+  })
+})
--- a/src/renderer/components/TabBar.test.tsx
+++ b/src/renderer/components/TabBar.test.tsx
@@ -116,7 +116,6 @@ describe('TabBar caret menu', () => {
     expect(document.querySelector('.tab-menu')).toBeNull()
   })
 })
-
 describe('TabBar New-project pin', () => {
   let root: Root
   let host: HTMLElement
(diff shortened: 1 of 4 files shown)


eneskirca:
**MERGE-READY.**

Measured here (merged onto `main` @422767af in an isolated worktree): `npm run typecheck` clean, `TabBar.test.tsx` 11/11 green, `styles.theme.test.ts` green.

I went looking for the four behaviours a tab-strip restyle usually breaks, and all four survive:
- **Drag-reorder** works and is now actually covered — drop-on-tab and drop-at-end, plus both markers. The `left: 0` / `z-index: 2` reasoning is right: at `left: -3px` the first tab's line fell outside the `overflow-x: auto` scroller, and with `gap: 0` the seam is exactly where the active tab's `z-index: 1` would have swallowed it.
- **Board toggle** stays active-tab-only, with the reason recorded (on an inactive tab it would flip another project's view without taking you there). `commandTooltip` is preserved.
- **Caret menu** still opens/closes, and the new capture-phase `pointerdown` correctly excludes `.tab-menu` (which covers the nested `.tab-menu__sub` submenus too) and `.tab__caret`, so a second tab's caret hands the menu over instead of closing and reopening it.
- **Unread dot / count badge** untouched.

The macOS integrated-title-bar rule also holds: `.tabbar` is already `position: relative`, so the new `::after` separator anchors to the bar and not to some outer element; and the new per-tab options button is a `<button>` inside `.tabbar`, so it inherits the existing `-webkit-app-region: no-drag` rule with no new entry needed.

Two things for @eneskirca to look at rather than changes I am asking for:

1. **The draggable surface shrinks a little on macOS.** `.tabbar__tabs` was already `no-drag`, but it was 34px inside a 44px bar, so the ~5px above and below it belonged to `.tabbar__projects` and was draggable. With the tabs at full height that strip is gone, and once the tabs overflow the bar the only drag regions left are the brand and the traffic-light inset. Still draggable, but it is a conscious trade worth making on purpose.
2. **The active tab is now a flat rectangle** — `.tab` loses its `border-radius: 14px` and nothing replaces it. That is consistent with "a hole in the bar", but browser tabs conventionally keep rounded top corners, so it is a design call, not a defect.

Not verified here (headless Linux server): I could not press the light theme, Windows or Linux rendering, which matches what you already flagged. The hairline compositing argument reads correctly from the CSS, and nothing in this change touches a bridge or IPC, so the Server Edition should follow the desktop.

Issue #569, opened 2026-08-31
canvas-control: no way to start a QUEUED node, rename steals focus, and rename/close are canvas-scoped


ezwep:
## Summary

I drive nodeterm from an external orchestrator: a local dashboard opens agent nodes for repositories, and those sessions then work autonomously. Three separate things I ran into today share one root cause — **the canvas-control CLI can do less than the UI, and two verbs have UI side effects you cannot opt out of.**

Running v0.3.4 on macOS.

## 1. A queued node can be created but not started (no CLI equivalent of ▶)

`open-agent` into a project that is not the currently viewed one creates the node correctly — right project, right cwd, agent assigned — and nodeterm labels it `QUEUED ▶`. It then stays queued indefinitely; nothing runs in it.

Measured in that state: the node has no cwd in the session map, there is no process for that directory, and my agent's lock file is untouched. The moment I click the node, it flips to `RUNNING`, the agent boots and its startup command runs exactly as intended — so the whole chain is fine, it just never gets kicked off.

The UI has a ▶ button on the queued node ("run now"). The CLI verb list has no equivalent:

```
list open-terminal open-claude open-agent show-image show-video show-web open-browser
group ungroup move arrange align link verify spawn-team open-worktree close-worktree
branch rename write close board assign send reply notify sticky browser open-project
```

**Why it matters:** autonomous workers only work for projects the user happens to have in view. Everything else silently waits for a glance. From the outside there is also no way to tell the difference — the open call succeeds either way — so an orchestrator will report the session as started and route work to a receiver that does not exist. I now work around that by polling the session map to see whether a process actually appeared, but that is inferring state that the app already knows.

**Ask:** a `run` (or `start`) verb taking `--node`, or a `--run-now` flag on `open-agent`. Even just returning a machine-readable `queued: true` in the response would let callers tell the truth.

## 2. `rename` pulls the node into the foreground

`rename --node <id> --title "..."` brings that node (and its project) to the front, even when you are actively working in a different node. That is disruptive when it fires from automation rather than from a user gesture.

My agents rename themselves on startup so they are identifiable in cross-node lists, and they re-run that on every context reset — so this happens often, and always at a moment the user did not ask for.

I have reduced it by reading the current title first and only renaming when it actually differs, but that only lowers the frequency; the first rename still steals focus.

**Ask:** a `--no-focus` flag, or simply do not focus on a programmatic rename. Renaming does not seem like an action that needs the node visible.

[...]

PR #734, opened 2026-09-10
feat(mesa): hierarchical swarm with a live-execution contract

MasterMind-SL:
## Summary
- Adds Mesa (1→4→16 roles, Table view, TUI, worktree-gated writers) on Desktop and Server Edition. Default adapter is **mock**; live CLIs are `NODETERM_SWARM_ADAPTER=live`.
- Live-execution contract: finishing a turn only **collects** a result; only **validated** Mesa JSON for that `executionId` may **accept** the task; neither signal grants extra permissions (no `--yolo`, no blanket auto-approve).
- Host owns completion (hooks / process-exit → `needs_review` without fabricated JSON). Unobservable panes stay held and go to review after 10 minutes — the CLI is never retyped. Prompt transport is per harness (`file-argv` / `file-flag` / `stdin-after-start`) with hard size caps.

This is **not** autonomous-ready. Opt-in smokes against real grok/claude/gemini/opencode in a throwaway repo are still owed. Mesa viewers keep `mayLaunch={false}` and never type a worker CLI into O’s TUI.

## Test plan
- [x] `npx tsc --noEmit -p tsconfig.node.json` and `tsconfig.web.json`
- [x] Contract suite: fake CLI (exit 1, no JSON, wrong `executionId`, output cap, unobservable timeout), harness transports, gemini stdin-after-start, 8s still `starting` (90s grace), scoped approvals
- [x] Mesa table / inspector / TUI unit tests
- [x] `workspaces.test.ts` + `tui-tmux.realtmux.test.ts` (unsandboxed)
- [ ] Restart the app (Cmd+Q) and run `npm run dev:mesa:mock` — Nueva misión, activate A, confirm mock accepts only structured results
- [ ] `npm run dev:mesa:swarm` in a **throwaway** folder — one grok/claude worker writes jailed JSON; idle/turn-end without JSON stays `needs_review`
- [ ] Gemini: launch line has no `$(cat …)`; brief is typed after the pane is `gemini`
- [ ] Approve C1 without a local folder stays blocked; a non-`write-workspace` approval stays pending
- [ ] Mesa card viewers do not launch a second CLI (`mayLaunch={false}`)

Made with [Cursor](https://cursor.com)


eneskirca:
## Review: CHANGES-NEEDED — and first, a scope decision

Thanks for the volume of work here. I merged this into current `main` in a scratch worktree (5 conflicts, all mechanical), resolved them, and ran the suites. `tsc --noEmit` is clean on both projects; `src/core/swarm`, `src/shared/swarm`, `src/tui` are 148/148 green; the table/mesa renderer suites are 81/81 green; `no-electron` (both shells) and `fs-atomic.guard` pass. So the code builds and its own tests pass. That is not the problem.

### What this is, in two paragraphs

Mesa is a second orchestration system living beside the one the repo already has. A *mission* is a persisted JSON document under `<userData>/swarm-missions/` holding a goal, a fixed 21-slot org chart (`O` → `A B C D` → `A1…D4`), a budget, approvals and artifacts. Roles are dormant until activated; activating one materialises a canvas terminal node, launches an agent CLI into its tmux pane, and the host polls `pane_current_command` to observe it. The genuinely good idea is the **live-execution contract**: a turn ending is *not* a result. Only a JSON file the worker writes — matching `taskId`, `executionId` and every acceptance criterion, and explicitly not the host's own placeholder (`isFabricatedTurnResult`) — can move a task to `accepted`; everything else lands in `needs_review`. `UNOBSERVABLE_REVIEW_MS` holds an unobservable pane for 10 minutes and then sends it to review without ever retyping the CLI. Those two claims from the description are true and are pinned by tests.

Around that sits a large amount of new surface: a third view mode (`ProjectView` gains `'table'`, a 1,287-line `TableView.tsx` plus 1,287 lines of CSS) that is a *third session surface* rather than a third view of the same nodes — it has its own model picker, rename, delete, group, move, spawn and cell-to-cell messaging paths, wired as `*FromTable` callbacks parallel to the existing `*FromKanban` ones; a **third loopback HTTP server** (`tui-channel.ts`) serving a ~130-line ESM REPL emitted as a template string and run via `ELECTRON_RUN_AS_NODE=1`; a new top-level source root `src/tui/` that is neither core, shell, nor shared and is **circularly** bound to core (`src/core/swarm/swarm-service.ts` imports `src/tui/mesa-cli`, while `src/tui/mesa-orchestrator.ts` imports `src/core/swarm/{budget,context,scheduler}`); and a new scheduler/router/budget/persistence layer.

### Scope: this cannot be reviewed as one PR

14,216 lines across 113 files with no prior design thread. CLAUDE.md is touched by **3 lines**, and those 3 lines are about a kanban z-index, not about mesa — so a subsystem with its own runtime, network channel, persistence format and permission policy arrives with no entry in the file that holds this repo's invariants, and nothing in CONTRIBUTING.md. Before a line-by-line review is worth anyone's time, the design question has to be settled: **should this extend the existing machinery or sit beside it?** Concretely, it currently duplicates:

| Exists on `main` | Mesa's parallel |
|---|---|
| `renderer/lib/pendingLaunch.ts` — `--after` deps, arms a node, fires on `done`, exactly-once via `launchInFlight`, refuses non-hook agents | `core/swarm/scheduler.ts` + `runtime.ts doTick` — a second dependency resolver, a second concurrency cap, a second exactly-once mechanism |
| `canvas-control-core.ts` `spawn-team` + `parseTeamSpec` — one agent per role, labelled group, rope **and context bridge** per member | `renderer/lib/swarmRules.ts planSwarm()` — hard-coded 2 workers + 1 synthesizer, no ropes, **no context bridges**, so the 2026-07 fan-in is absent and the synthesizer is told to read the working tree instead |
| `renderer/lib/verifyPanel.ts` — lens-per-reviewer, armed behind the target, bridged, judge on `extraLive` | `runtime.maybeAdvancePipeline()` — auto-adds role `D` once the `C*` writers are accepted |
| `core/agents/agent-message*` — node-id-scoped, idle-gated, traced to the board log, app-owned envelope | the talk bar calls `api.pty.sendText(talk.toId, payload, {enter})` directly (`TableView.tsx`). ESC is stripped downstream by `PtyManager.sendText`, so this is *not* an escape hole — but there is no idle gate (the target can be mid-turn, which is the splice failure #569 was about), no trace and no board-log entry |
| `renderer/state/permissionMode.ts` + `withPermissionMode` + the claude `--version` `auto` gate | `MESA_FULL_ACCESS_MODE` (below), plus a re-implementation of the version gate in `cli-launch.launchPermissionMode()` |
| `core/agents/hook-server.ts` + the `node-auth-*` per-node identity | a third loopback server with a single machine-wide bearer |

If the answer is "extend", most of `src/core/swarm` and `src/tui` evaporates and the Table view becomes a view over the existing model. That is a conversation, not a diff.

---

## Blockers

**B1 — The app is renamed and `userData` is moved. This is fork branding in an upstream PR.**
`package.json`: `appId: com.nodeterm.app → com.nodeterm.mod`, `productName: nodeterm → nodeterm-mod`. `src/main/index.ts` adds `app.setName('nodeterm-mod')` and `app.setPath('userData', join(app.getPath('appData'), 'nodeterm-mod'))`, with the comment *"This repo is the nodeterm-mod fork."* Merged as-is, the next release is a different application identity: every existing user's `workspace.json`, `settings.json`, license, device id, managed Claude/Codex account dirs, tmux.conf, scrollback store and safeStorage-encrypted GitHub token become unreachable, and the electron-updater feed identity no longer matches. This must come out entirely.

**B2 — Every mesa mission launches its CLIs with permission prompts disabled, overriding the user's setting, undisclosed in the UI — and the description says otherwise.**
`src/shared/swarm/schemas.ts`: `export const MESA_FULL_ACCESS_MODE = 'bypassPermissions'`. It is the default in three independent places (`runtime.createMission`, `router.resolveTaskRoute`, `Canvas.addTerminalFromTable`) and is sent explicitly by the UI (`TableView.tsx`, `mesaModels.workerLaunchSpec`). Run through this repo's own `shared/agents/approval-mode.ts`, that emits `--approval-mode yolo` for gemini and `--ask-for-approval never` for codex. The description says *"neither signal grants extra permissions (no `--yolo`, no blanket auto-approve)"* — true of the *completion signals*, false of the mission default, which is exactly a blanket auto-approve and is literally `--yolo` for one agent. It also silently overrides `settings.claudePermissionMode` and `project.defaultPermissionMode`, which the tab menu warns about when a human sets it by hand. `withPermissionMode` itself is used correctly; the *input* is the problem.

**B3 — The new loopback channel has one machine-wide bearer, no per-caller scope, and it gates the human approval.**
`tui-channel.ts`: `listen(0, '127.0.0.1')`, `authorize()` is `header === 'Bearer ' + token`, and `missionId` is read from the request **body** with only an `isSafeNodeId` shape check — never bound to the token. `/tui` runs `handleCliLine`, whose verbs include `/approve <id>` (the write-workspace gate), `/activate <role>` (spawns another bypassPermissions agent) and `/goal`. The token file is `0600` and correctly kept out of argv — but every agent CLI mesa itself launches runs as the same uid and the path is deterministic (`<userData>/swarm/tui-endpoint.env`). So a worker that is prompt-injected by repo content can read the file and POST `{"missionId":"…","line":"/approve <id>"}` to materialise the worktrees the human was supposed to approve. This repo built `node-auth-*` and `docs/node-identity.md` precisely because *"the shared bearer proves 'a session on this machine', never which session"*; mesa reintroduces that model and puts an approval behind it. The server also starts **unconditionally at boot in both shells** with a 2 s `setInterval`, no setting, and `close()` never called — a user who never opens mesa still gets the port, the timer and the 0755 script on disk.

## Major

- **The "worktree-gated writers" guarantee covers 4 of 21 roles.** `roleWritesFiles` is `role.startsWith('C') && role !== 'C'`, i.e. C1–C4. The other 17 get `kind: 'shared'`, `pathHint: repoRoot`, no approval — and, per B2, no prompts. `workspaces.ts` is honest that *"a worktree is collision control, not a security sandbox"*.
- **Results are written into the user's git-shared folder.** `swarmWorkspaceResultsDir` → `<repo>/.nodeterm/swarm-results/<executionId>.json`. `.nodeterm/` is the folder that carries `project.json` and is documented as git-shareable; nothing gitignores this, nothing mentions it in the workspace docs, and an agent running `git add -A` commits it.
- **`git init` in a folder the user merely attached.** `ensureMesaGitRepo` runs `git init` plus an empty Mesa-authored commit when the workspace is not a repo. `isDangerousGitInitRoot` refuses only `/` and exactly `$HOME` — compare `isDangerousWorktreeRemovalPath` on `main`, which also refuses the repo itself and any *ancestor* of these.
- **No cap on the agent-written result file.** `fs.readFileSync(file,'utf8')` uncapped in `cli-launch.ts` / `cli-file.ts`, no string or array limits in `parseStructuredTaskResult`, and `payload` is stored raw into `mission.artifacts`, re-serialised and broadcast every 2 s forever (missions are never deleted, only cancelled). The "hard size caps" claim covers the prompt and the typed line, not this.
- **`mesaTypedLineError` checks length but not control characters.** `workspace.pathHint` is accepted as any string by `parseTask` and only `isAbsolute` + `existsSync`-checked; a `\n` in it splits a line that is submitted with Enter. `oneLine()` in `agent-message-envelope.ts` exists for exactly this class.
- **The whole user-facing surface is Spanish** (~200 accented characters across `TableView.tsx`, `MissionInspector`, `schemas.ts`, `mesa-orchestrator.ts`, `tui-channel.ts`: `Orquestador`, `Nueva misión`, `No se pudo entregar…`, `Usá /activate A … D4`). CLAUDE.md Conventions: *"Code comments, UI strings, and identifiers are all in English."*
- **Behaviour changes for users who never open mesa.** `defaultProjectView` and `viewMode.defaultView` both flip `'canvas' → 'table'`, so every user who never touched the setting boots into Mesa; the two-state `toggle()` becomes a three-state `cycle()` at the TabBar button, ⌘⇧B and ⌘K; and `data.callsign` is written by every node factory and persisted into git-shared `project.json` while being rendered only inside the Table view — it is also the one new node field with no counterpart to `sanitizeNodeTriggers` on the read path.
- **`swarm:note-idle` is a dead channel**: typed in `shared/types.ts`, `ipc.ts`, preload, ws-bridge and stubs, with a real ~35-line handler that mutates mission state — and zero callers. Its only test asserts that nobody calls it (`expect(SRC).not.toContain('api.swarm.noteIdle')`).
- **`shells.parity.test.ts` is ~20 `expect(file).toContain('<literal source string>')` assertions** against `src/main/index.ts` and `src/server/index.ts`. It proves both files contain a string, not that either behaves — and one of them *pins* `permissionMode: MESA_FULL_ACCESS_MODE`, i.e. it locks B2 in rather than catching it.
- **Unrelated work is bundled in.** The opencode `PATH` prepend (`exec-path.ts` + `pty-manager.ts`) is a real bug fix with nothing to do with mesa, and it changes the environment every opencode session spawns into. `MODEL_FLAG_CAPABLE` / `grok-models.ts` collides with `main`, which has since put grok **into** `MODEL_SWITCH_CAPABLE`; after my merge, `config.capabilities.test.ts` fails on *"grok takes --model natively without joining the gateway list"*. That is a product disagreement that deserves its own PR and its own argument.

## Credit where it is due

`relay-api.ts` correctly routes `swarm` to `stub.swarm` with a stated reason — a relay guest cannot drive the viewer's runtime. Every new write uses `writeFileAtomic`; there is **no bare `fs.rename`** anywhere in the new code. No `electron` import crosses into `src/core`. The token is on disk, never in argv. `MODEL_FLAG_CAPABLE` has its leaf and its test. And `claimNodeLaunch` solves a real problem — three possible first-typers into one pane — that would fix a latent kanban race on its own (it needs a release on node destroy/respawn; today a claimed id is never re-launchable).

## Suggested split

Four things in here are independently valuable and could land quickly — about 440 lines in total, none of which changes behaviour for a non-mesa user:

1. **opencode `PATH` fix** — `exec-path.ts`, its test, `pty-manager.ts`. ~40 lines. A real bug, unrelated to mesa.
2. **grok native `--model`** — `MODEL_FLAG_CAPABLE` + `grok-models.ts` + tests, ~55 lines, rebased against what `main` now does.
3. **Single-writer discipline for a pane** — `claimNodeLaunch` + `ModalTerminal` `mayLaunch` / `whenShellSettled` + `TerminalNode`'s claim branch, ~210 lines. Needs a CLAUDE.md line, because it reverses a documented invariant.
4. **Callsigns** — ~135 lines, once one non-mesa surface renders them.

Then: **5.** `src/shared/swarm/` as a pure spec (~1,330 lines, zero existing files touched, reviewable on its own); **6.** the host runtime with the mock adapter only, no IPC and no UI; **7.** the live adapter + IPC + both shells, behind a setting that is **off** by default, with `permissionMode` resolved through `activePermissionMode` / `resolvePermissionMode`; **8.** the Mesa view, with `defaultProjectView` left at `'canvas'`.

Phases 6–8 should not be read line-by-line until the design question at the top is answered, because if the answer is "extend `pendingLaunch` + `spawn-team` + `verify`", most of 6 and 7 does not need to exist.

*Not device-verified: this was reviewed on Linux with no Electron launch, so nothing in the opt-in checklist in your description (`dev:mesa:mock`, `dev:mesa:swarm`, the gemini launch line, the C1 approval) was exercised here.*

PR #742, opened 2026-09-10
fix(copilot): preserve selected model on restart

dkattan:
Copilot launches and restarts omitted the startup model selection. The provider metadata variables do not replace `--model` or `COPILOT_MODEL`, and an in-place restart reuses its existing shell. See [GitHub's model configuration documentation](https://docs.github.com/en/copilot/how-tos/copilot-cli/customize-copilot/use-byok-models).

Pass the selected internal model ID explicitly on launch and resume. For example, `openai/gpt-5.5` now resumes with `--model 'gpt-5.5'`, while the gateway still receives `openai/gpt-5.5` through `COPILOT_PROVIDER_WIRE_MODEL`. Copilot-based custom agents use the same mapping.

Validation:

- 186 targeted tests passed, including real `/bin/sh` checks for launch, resume, quoting, and wire-model separation.
- `npm run typecheck` and `git diff --check` passed.
- Live authenticated Copilot/gateway validation remains outstanding.

Desktop and Server Edition share this command builder. No mobile API or UI change is required.

### DIFF
--- a/docs/copilot-agent.md
+++ b/docs/copilot-agent.md
@@ -12,7 +12,8 @@ CLI reference:
 - `copilot --interactive <prompt>` opens the normal interactive TUI and submits the initial prompt;
   `--prompt` is non-interactive and exits after the response.
 - `--session-id=<uuid>` creates a caller-addressable session, `--resume=<uuid>` resumes it, `/exit`
-  cleanly returns to the shell, and `--model` selects a GitHub-routed model.
+  cleanly returns to the shell, and `--model` (or `COPILOT_MODEL`) selects the startup model,
+  including in BYOK mode.
 - BYOK mode activates through `COPILOT_PROVIDER_BASE_URL`. The implementation also sets
   `COPILOT_PROVIDER_TYPE`, `COPILOT_PROVIDER_API_KEY`, `COPILOT_PROVIDER_MODEL_ID`, and
   `COPILOT_PROVIDER_WIRE_MODEL`; GPT-5-family OpenAI models use
@@ -24,6 +25,7 @@ CLI reference:
 Primary references:
 
 - <https://docs.github.com/en/copilot/reference/copilot-cli-reference/cli-programmatic-reference>
+- <https://docs.github.com/en/copilot/how-tos/copilot-cli/customize-copilot/use-byok-models>
 - <https://docs.github.com/en/copilot/reference/hooks-reference>
 - <https://docs.github.com/en/copilot/how-tos/copilot-cli/customize-copilot/add-custom-instructions>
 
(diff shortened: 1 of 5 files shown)


Abhirup0 (commented):
LGTM — cleanly resolves the lost model on in-place restarts, and aligning `--model` with `COPILOT_PROVIDER_MODEL_ID` matches GitHub's BYOK contract.

Two non-blocking suggestions:

1. In `withAgentModel` (`src/shared/agents/model-gateway.ts:363`), if `copilotModelParts` receives a trailing slash (e.g. `'openai/'`), `modelId` resolves to `''` and emits `--model ''`. A quick `if (!modelId) return cmd` prevents passing an empty flag to Copilot.
2. In `src/shared/agents/launch.test.ts:170`, `it.skipIf(process.platform === 'win32')` skips both the real `/bin/sh` run and the command builder assertions. Splitting the string output check from the subshell execution would let the command matrix run on Windows CI too.

eneskirca:
## Verdict: CHANGES-NEEDED — one blocking gating bug

**Measured**: merges clean onto `main` (422767af), both typechecks exit 0, 104 tests pass. A verbose run confirms all four parameterized shell cases **actually execute** (not skipped), including the injection case `custom/o'model; $(exit 42)` run under a real `/bin/sh` — the quoting holds. That is a strong test in house style.

The architecture is right. **Rule 17 passes**: everything lives in the base-harness mapping (`shared/agents/model-gateway.ts`, dispatching on `capabilityAgentId`), nothing spells `copilot` in a frontend, and `model-gateway.test.ts` proves a custom agent with `baseAgent:'copilot'` inherits it (`copilot-wrapper --model 'gpt-5.5'`). **Rule 18 passes**: only a model id reaches argv, never a key, and the SIGTERM + session-recycle path is untouched. `docs/copilot-agent.md` is corrected in the same PR from the stale "after a clean `/exit`" to the recycle wording — the right habit.

### BLOCKING — `--model` and the BYOK env are gated by different conditions

Probe output on the merged branch:

```
gateway OK             model=openai/gpt-5.5  envKeys=6  PROVIDER_MODEL_ID=gpt-5.5  cmd=copilot --model 'gpt-5.5'
gateway URL missing    model=openai/gpt-5.5  envKeys=0  PROVIDER_MODEL_ID=-        cmd=copilot --model 'gpt-5.5'
gateway key missing    model=openai/gpt-5.5  envKeys=0  PROVIDER_MODEL_ID=-        cmd=copilot --model 'gpt-5.5'
gateway env ref unset  model=openai/gpt-5.5  envKeys=0  PROVIDER_MODEL_ID=-        cmd=copilot --model 'gpt-5.5'
```

`modelGatewayEnv` (`model-gateway.ts:307-311`) correctly returns `{}` when routes or key are missing, but `withAgentModel` (`:351-361`) checks only `normalizedAgentModel` — so it still emits `--model` with the **provider prefix stripped**. A Copilot node whose gateway is unusable therefore launches asking *GitHub's own catalogue* for a gateway-side id (`gpt-5.5`), where on `main` it launched bare on ordinary GitHub routing.

This is reachable without any misconfiguration: `agentModel` is persisted to `.nodeterm/project.json` (`workspace.ts:1938`/`:2018`), which is git-shared, and the gateway key may be `${env:VAR}`. Clone a repo whose canvas has a gateway-model Copilot node, without configuring the gateway, and the launch line changes. It contradicts the invariant this PR's own doc file restates ("Copilot BYOK is not activated merely because gateway settings exist … preserves ordinary GitHub Copilot routing for untouched nodes") and CLAUDE.md rule 14 ("a guess must degrade to nothing, never to something wrong").

The fix needs gateway state threaded into `assembleLaunchCommand`/`assembleResumeCommand` (`launch.ts:219-231`, `:263-266`), which do not receive it today. Shape is your call, but it should not ship as is.

### Minor — a still-valid argv guard was deleted

`src/core/pty-single-user.test.ts:327` replaced `expect(spawnArgs[0].args.join(' ')).not.toContain('--model')` with a comment. I re-added the assertion on the merged branch and ran it: **26 passed** — it still holds, because the flag rides the typed launch command and never the tmux client argv. Please keep the assertion (the explanatory comment can sit beside it); it guards against a future change putting the flag into argv, which is the failure mode this repo has already shipped once.

### Not measured

The PR's central premise — Copilot requires `--model`/`COPILOT_MODEL` at startup, and `COPILOT_PROVIDER_MODEL_ID` alone does not select it — could not be verified here: `copilot` is not installed on this host. Per CLAUDE.md rules 5 and 16 this wants a device check — does a BYOK Copilot session started with the provider env but no `--model` actually come up on the wrong model?

PR #749, opened 2026-09-11
test(renderer): normalize path separators in the machineName guard on Windows

### LINKED ISSUE #746: Windows: machineName.guard.test.ts fails because ALLOWED map lookup does not normalize path separators
## Environment
- Windows 11 (or any Windows host)
- Node: v22+ / v24+
- Revision: `main` (commit `af49f2e7` and later)

## What happens
Running `npm test -- src/renderer/lib/machineName.guard.test.ts` fails on Windows:

```text
FAIL  src/renderer/lib/machineName.guard.test.ts > no user-visible copy calls this machine a Mac > finds none outside the documented exemptions
AssertionError: use thisMachine()/thisMachineCap()/machineNoun() from lib/machineName: expected [ …(2) ] to deeply equal []

- []
+ [
+   "src\\renderer\\components\\onboarding\\OnboardingFlow.tsx:458: which is what makes the capsule sit flush on your Mac.",
+   "src\\renderer\\components\\PtyPressureBanner.tsx:120: title=\"Raises this Mac’s pty-device limit (kern.tty.ptmx_max) now and after every restart. macOS will ask for your password.\"",
+ ]
```

## Why
In `src/renderer/lib/machineName.guard.test.ts:60-63`:
```ts
const rel = relative(process.cwd(), file)
if (ALLOWED.has(rel)) continue
```

`ALLOWED` defines keys with POSIX forward slashes (`src/renderer/...`), whereas `relative(process.cwd(), file)` produces Windows backslashes (`src\renderer\...`). 

Because `ALLOWED.has(rel)` does a raw string lookup without path normalization, the check fails on Windows for all three exempt files. The test proceeds to scan `OnboardingFlow.tsx` and `PtyPressureBanner.tsx`, finding the legitimate macOS-specific copy and failing the suite.

This hits the rule in `CONTRIBUTING.md:107`: *"Normalize BOTH sides of a path comparison, through one function."*

## Fix
Normalize `rel` with forward slashes before querying `ALLOWED.has(...)`, matching the pattern in `src/core/fs-atomic.guard.test.ts`:
```ts
const rel = relative(process.cwd(), file).replace(/\\/g, '/')
```

L4XB:
Fixes #746

## Summary

`machineName.guard.test.ts` looks up each scanned file in `ALLOWED`, whose keys are written with forward slashes, but `relative(process.cwd(), file)` returns backslashes on Windows. The exemptions therefore never matched there, the guard scanned `OnboardingFlow.tsx` and `PtyPressureBanner.tsx`, and it reported their documented macOS copy as offenders.

The relative path is now split on `path.sep` and re-joined with `/` before the lookup, so the same keys work on every platform. The reported offender paths use the normalized form too, which keeps the failure output identical across platforms.

## Verification

`npx vitest run src/renderer/lib/machineName.guard.test.ts` passes on macOS (the normalization is a no-op there). I could not run it on Windows; the change only touches how the lookup key is built, and `split(sep).join('/')` produces exactly the `ALLOWED` keys from the report's `src\renderer\…` paths.

### DIFF
--- a/src/renderer/lib/machineName.guard.test.ts
+++ b/src/renderer/lib/machineName.guard.test.ts
@@ -1,5 +1,5 @@
 import { readdirSync, readFileSync, statSync } from 'fs'
-import { join, relative } from 'path'
+import { join, relative, sep } from 'path'
 import { describe, expect, it } from 'vitest'
 
 /**
@@ -58,7 +58,8 @@ describe('no user-visible copy calls this machine a Mac', () => {
     const offenders: string[] = []
     for (const root of ROOTS) {
       for (const file of sourceFiles(join(process.cwd(), root))) {
-        const rel = relative(process.cwd(), file)
+        // ALLOWED is keyed with forward slashes; on Windows `relative` returns backslashes.
+        const rel = relative(process.cwd(), file).split(sep).join('/')
         if (ALLOWED.has(rel)) continue
         readFileSync(file, 'utf8')
           .split('\n')


eneskirca:
**Verdict: CLOSE — superseded, with thanks.**

This is a correct diagnosis and a correct fix. It is also the same bug (#746), in the same file, as **#751**, which merged a few hours ago. `machineName.guard.test.ts` on `main` now reads:

```ts
function normalizedSourcePath(value: string): string {
  return value.replace(/\\/g, '/')
}
...
const rel = normalizedSourcePath(relative(REPO_ROOT, file))
```

Same rule, expressed with the helper shape its two sibling guards (`fs-atomic.guard.test.ts`, `localStore.guard.test.ts`) already use — which is why `replace(/\\/g, '/')` was chosen over `split(sep).join('/')`: the guards then all normalize through one recognisable idiom. #751 also carried two things beyond the reported one-liner: the scan root moved from `process.cwd()` to a `__dirname`-derived `REPO_ROOT`, and a new assertion compares the exemption keys against the keys the scan *actually produced* — the old "exemption list is honest" test read each exempt file by path, so it stayed green on Windows while the lookup was missing every entry. That is exactly how this hid.

The PR is also `CONFLICTING` against `main` now, for that reason.

Nothing here was wasted: the sweep this prompted checked every other guard test for the same unnormalized-lookup pattern (none had it). Thank you — closing as fixed, not as wrong.

L4XB:
Thanks for the detailed close, and for saying which shape the fix took. The helper idiom shared with `fs-atomic.guard.test.ts` and `localStore.guard.test.ts` is clearly the better fit, and the two extras in #751 (the `__dirname`-derived root and comparing exemption keys against the keys the scan actually produced) are the part that keeps this from coming back. Closing here.

Issue #582, opened 2026-08-31
Node title is re-sent to the hosted agent session as /rename repeatedly (bursts per message), not once


jimbobbibong-max:
## Summary

nodeterm keeps the hosted Claude Code session's name in sync with the node title by typing Claude Code's own `/rename <node title>` slash command into the pane. That sync appears to run again on every message injection / wakeup rather than once when the title changes, so a long-lived session accumulates dozens of identical `/rename` commands.

The renames themselves are idempotent, so nothing breaks. The cost is noise: each one lands in the transcript as a `local_command` plus its `Session renamed to: …` stdout plus a `<system-reminder>` telling the model the user named this session, and each consumes context in a session that is otherwise doing work.

nodeterm 0.3.4, macOS 26.2 (25C56), Apple Silicon. Claude Code agent node.

## Steps to reproduce

1. Open a Claude Code agent node on the canvas.
2. Rename the node (title used here: `Trial 1.08`).
3. Keep working in that session for a few hours — send messages, let it go idle and wake again.
4. Watch the pane, or the session's JSONL transcript, for `/rename` entries.

## Observed

Over one ~6 hour session, **33** `/rename` injections. Measured from the session transcript (counting entries of type `system`, `subtype: local_command`, `<command-name>/rename</command-name>`):

- 1 injection with the node's default title (`Claude Code`) at session start — expected.
- 32 further injections, all with the same argument (`Trial 1.08`), after the node had already been renamed and the session name already matched.
- Span 5h52m. Gap distribution between consecutive injections: 13 gaps under 200s (min 140s), 9 gaps of 200–700s, 10 gaps of 700s+. The short gaps cluster into pairs — the same title is sent twice roughly 2.5 minutes apart — which reads from inside the session as bursts.

Timestamps of a representative run (UTC), argument `Trial 1.08` in every case:

```
2026-08-31T18:32:16.206Z
2026-08-31T18:34:47.515Z
2026-08-31T18:41:58.426Z
2026-08-31T18:44:28.289Z
2026-08-31T19:03:45.223Z
```

Each one produces three transcript entries:

```
<command-name>/rename</command-name>
<command-message>rename</command-message>
<command-args>Trial 1.08</command-args>
<local-command-stdout>Session renamed to: Trial 1.08</local-command-stdout>
<system-reminder>The user named this session "…". This may indicate the session's focus or intent.</system-reminder>
```

[...]

PR #750, opened 2026-09-11
fix(nodes): gate header and AI renames on sessionNameUnchanged

### LINKED ISSUE #714: The node header's rename path does not use sessionNameUnchanged, so "Name with AI" can send a /rename that changes nothing
## What I am reporting

A static path, not an observed failure: reading `main` at `f456e888`, I cannot find anything that
stops the node header from pushing a `/rename` whose name equals the one the node already carries.
I have not reproduced it at runtime.

## The path

`renderer/lib/sessionRename.ts` exports `pushSessionRename(io, nodeId, name, current)`, which returns
early on `sessionNameUnchanged(name, current)` — compared after `oneLine` normalization.

`TerminalNode.tsx` does not use it. It imports only `renameCommand` (line 148) and defines its own
`pushSessionRename` (line 4654):

```ts
const pushSessionRename = (name: string) => {
  if (canRenameNode && name) void api.pty.sendText(id, renameCommand(name))
}
```

From there:

- **✦ Name with AI** (lines 4673-4678) calls `applyManualTitle(r.message)` whenever the generate call
  succeeds, with no comparison against the node's current title. The generated name comes from the
  same session and cwd, so getting the existing name back is an ordinary outcome — pressing it twice
  should be enough.
- The manual edit path (`commitTitleEdit`, line 4670) does compare, but against
  `titleEditStartRef.current` with a plain `trim()`. `sessionNameUnchanged`'s docblock singles out
  that shape: two titles differing only by a control character compose the same command, so a raw
  comparison reports a new name.

## On #582

I am not claiming the header was inside #582's scope — that is yours to say. What I can say is that
the guard #582 introduced is not reached from here, and the effect it describes (a `/rename` that
asserts an intent nobody expressed, costing a `local_command`, its stdout and a `<system-reminder>`
in the session's context) would apply equally if this path fires.

One nuance on the shared helper's docblock: it says the required `current` parameter is what enforces
the gate, since the components are not unit-rendered. A required parameter constrains callers of that
helper — it cannot prevent a component from writing its own two-line version, which is what happened
here. The type is a weaker guarantee than the comment claims.

## If this is worth fixing

The open question is *which* title to compare against, and the two call sites do not want the same
answer: for a manual edit, the title when the edit began is the right reference; for an AI-generated
name, the node's current title is. Whatever is chosen should be stated, because that choice is the
whole behaviour.

## Environment

Local build of `main` at `f456e888`, macOS. Read from source. Related: **#582**, **#569 §2**.

Abhirup0:
Follow-up to #582. Closes #714.

## Problem

When #582 introduced `sessionNameUnchanged` to prevent redundant `/rename` injections, it covered the canvas-control `rename` verb and Canvas's `renameSession` funnel. However, `TerminalNode.tsx` retained a local `pushSessionRename` helper that did not use `sessionNameUnchanged`:

1. **AI renaming (`nameWithAi`)**: Clicking "✦ Name with AI" had no check comparing the generated suggestion against the node's current title. When a model returned the existing session name, or when clicked repeatedly, it sent `/rename <same-name>` down the PTY. In Claude sessions, every redundant `/rename` spends context window tokens on a transcript entry, stdout notice, and `<system-reminder>`.
2. **Manual rename box (`commitTitleEdit`)**: The header's input box checked `value.trim() !== titleEditStartRef.current.trim()`, bypassing control-character normalization that `renameCommand` / `sessionNameUnchanged` perform (e.g. whitespace runs or `\r\n`).

## Changes

- **TerminalNode `pushSessionRename`**: Updated signature to `pushSessionRename(name: string, current: string)`. Now gates on `!sessionNameUnchanged(name, current)` before invoking `api.pty.sendText(id, renameCommand(name))`.
- **Manual title edits**: `commitTitleEdit` now compares against `titleEditStartRef.current ?? ''` with `!sessionNameUnchanged(value, prev)` and passes `prev` to `applyManualTitle`.
- **AI name generator**: `nameWithAi` resolves the node's active title (`titleRef.current ?? (data.title as string) ?? ''`) and passes it into `applyManualTitle`, refusing identical names.
- **Tests**:
  - Added unit test cases for empty/whitespace-only strings in `sessionRename.test.ts`.
  - Added source-level invariant assertions in `title-gate.test.ts` verifying that `TerminalNode` rename paths gate on `sessionNameUnchanged`.

## Verification

- `npm run typecheck`: clean.
- `npm test -- src/renderer/lib/sessionRename.test.ts src/renderer/nodes/title-gate.test.ts`: 40/40 tests passing.
- `npm run build`: built Electron main, preload, renderer bundle, and codex-relay with 0 errors.

### DIFF
--- a/src/renderer/lib/sessionRename.test.ts
+++ b/src/renderer/lib/sessionRename.test.ts
@@ -127,6 +127,8 @@ describe('sessionNameUnchanged', () => {
     expect(sessionNameUnchanged('Trial 1.08', 'Trial 1.08')).toBe(true)
     expect(sessionNameUnchanged('Trial 1.08', 'Trial 1.09')).toBe(false)
     expect(sessionNameUnchanged('Trial 1.08', '')).toBe(false)
+    expect(sessionNameUnchanged('', '')).toBe(true)
+    expect(sessionNameUnchanged('   ', '')).toBe(true)
     // Normalized exactly as `renameCommand` normalizes: trimmed, control runs collapsed.
     expect(sessionNameUnchanged('  Trial 1.08  ', 'Trial 1.08')).toBe(true)
     expect(sessionNameUnchanged('Trial\r\n1.08', 'Trial 1.08')).toBe(true)
--- a/src/renderer/nodes/TerminalNode.tsx
+++ b/src/renderer/nodes/TerminalNode.tsx
@@ -154,7 +154,7 @@ import { isZoomModifierHeld } from '../lib/zoomModifier'
 import { isHidden } from '../lib/ui-visibility'
 import { readsClaudeTranscript } from '../lib/transcriptGates'
 import { liveProjectJumpTarget } from '../lib/projectJump'
-import { renameCommand } from '../lib/sessionRename'
+import { pushSessionRename } from '../lib/sessionRename'
 import { useSettings } from '../state/settings'
 import { useCodexIdentity, codexSharedIdentity, codexFallbackText } from '../state/codexIdentity'
 import { useAgentStatus, agentStatusForApi, inferInterruptAfterSettle } from '../state/agentStatus'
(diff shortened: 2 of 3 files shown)


eneskirca:
**Verdict: CHANGES-NEEDED — the goal is right, but this adds a *second* rename pusher instead of using the one that already exists.**

Verified locally (merged onto `main`): typecheck clean on both projects, `sessionRename.test.ts` + `title-gate.test.ts` 40/40, and the full suite green (11171 passed). The diagnosis is also correct — `TerminalNode`'s header ✦ and rename box are genuinely the last two `/rename` sites that were not gated on `sessionNameUnchanged`. And to answer the obvious worry up front: this does **not** fight the `#582`/`#604` guard in `Canvas.renameSession` — that funnel is untouched, and the `titleAuto` contract is preserved (you still call `applyManualTitle` unconditionally on a successful ✦, so ownership flips even when the push is skipped). That is the right call.

The problem is the shape.

### `pushSessionRename` already exists, in `src/renderer/lib/sessionRename.ts`

```ts
export async function pushSessionRename(
  io: RenamePushIo, nodeId: string, name: string, current: string
): Promise<boolean> {
  if (sessionNameUnchanged(name, current)) return false
  for (let i = 0; i < RENAME_PUSH_ATTEMPTS; i++) {
    if (i > 0) await sleep(RENAME_PUSH_RETRY_MS)
    const pane = await io.paneCommand(nodeId).catch(() => null)
    if (pane && !isShellCommand(pane)) { void io.sendText(nodeId, renameCommand(name)); return true }
  }
  return false
}
```

Same name, same purpose, already does the unchanged gate you want — **plus** the pane probe that `TerminalNode`'s local copy is missing. Its own docstring says why that second gate exists, and it is not cosmetic:

> `sendText` appends Enter, so writing while a SHELL owns the pane splices the line into whatever that shell is currently reading. For a freshly opened agent node that is its own launch command being typed, and the observed damage was exactly that: `claude '<long prompt>'` cut in half by an interleaved `/rename …`, leaving the shell sitting at `quote>` with the agent never started and the prompt lost.

So today the canvas node header can still destroy a launch line, and after this PR it still can. Meanwhile every *other* ✦ surface — the sessions sidebar (`aiNameSession`) and the kanban card modal (`onRename`) — already routes through `renameSession` → the shared helper, and gets both gates. Two views of the same session, two different behaviours.

There is a second, quieter case the local copy gets wrong: a **hibernated** node (Eco exited the CLI, the pane is back to a bare shell). The local push types `/rename Foo` into that shell, which runs it as a command. The shared helper writes nothing, deliberately — "an unsynced session name is cosmetic while a destroyed launch is not."

### The ask

Delete the local `pushSessionRename` and call the shared one. It is roughly the same diff size:

```ts
const applyManualTitle = (raw: string, current: string) => {
  const name = raw.trim()
  updateNodeData(id, { title: name, titleAuto: false })
  if (canRenameNode && name) void pushSessionRename(api.pty, id, name, current)
}
```

That gets you the fix you came for, closes the splice hole in the same change, removes a duplicated rule (`sessionRename.ts`'s header literally says "this file already exists because of one such drift"), and makes the new source-pin test in `title-gate.test.ts` unnecessary — the `current` parameter is required on purpose so the **compiler** is the reviewer that sees every site.

### Two smaller notes, in case you keep the current shape

1. The `title-gate.test.ts` pin does `lines.find(l => l.includes('api.pty.sendText') && l.includes('renameCommand'))`, i.e. it depends on both tokens staying on one physical line. The new `pushSessionRename` body is ~110 chars; the first time anyone wraps it the pin stops finding its site and passes vacuously (`toBeDefined()` would fail, so it fails loudly rather than silently — but it fails for a formatting reason, not a behaviour one). Matching the joined source rather than a single line is sturdier.
2. `commitTitleEdit` now compares through `sessionNameUnchanged`, which collapses control runs as well as trimming. That is a behaviour change from `trim() !== trim()` — right for deciding whether to *push*, slightly odd for deciding whether to take **ownership** of the title (`titleAuto: false`). Routing the push through the shared helper separates the two again: commit on "the text changed", push on "the composed line would differ".

Everything else in the PR — the `sessionNameUnchanged('', '')` cases, the `titleRef.current` read for ✦, the comments — is good and should survive the rewrite.

Abhirup0:
Updated to use the shared helper as suggested:

- Removed the local `pushSessionRename` in `TerminalNode.tsx` and routed `applyManualTitle` through the shared `pushSessionRename(api.pty, id, name, current)` from `src/renderer/lib/sessionRename.ts`. This brings in the pane-probe guard against shell splicing and bare shells.
- Restored `commitTitleEdit` to compare via `trim() !== trim()` so UI title ownership (`titleAuto: false`) remains distinct from session command gating.
- Kept the `nameWithAi` active title forwarding via `titleRef.current` and the `sessionNameUnchanged('', '')` unit tests.
- Updated `title-gate.test.ts` to verify `pushSessionRename` sits behind `canRenameNode`, and dropped the redundant local source-pin test.
- Rebased onto latest `main`. `npm run typecheck`, the 6 guard suites, `title-gate.test.ts`, and `sessionRename.test.ts` all pass clean.

PR #755, opened 2026-09-11
fix(delivery): use Escape to clear prompt line on Windows shells (#710)

### LINKED ISSUE #710: Windows: `\x15` (Ctrl+U) is inserted literally by PowerShell, so retried command deliveries concatenate into one invalid command
## Environment
- nodeterm 0.3.4 (nodeterm-Setup-0.3.4.exe), Windows 11 Home 10.0.26200
- Shell: Windows PowerShell 5.1 (PSReadLine)
- Agent: Claude Code v2.1.263

## What happens
The delivery path writes `KILL_LINE` (`\x15`) to clear the prompt line before re-attempting.
PowerShell does not treat the raw `\x15` byte as readline's kill-line, so it lands in the
input buffer as text. Retries then concatenate instead of replacing:

```
PS C:\dev\teste-nodeterm> claude^Uclaude^Uclaude
claudeclaudeclaude : The term 'claudeclaudeclaude' is not recognized as the name of a
cmdlet, function, script file, or operable program.
```

The repetition count matches `DELIVERY_ATTEMPTS = 3` exactly, and it grows by one on each
further relaunch attempt (observed `claudeclaude` before an app restart, `claudeclaudeclaude`
after).

## Code
- `src/session-host/session.ts:19` — `const KILL_LINE = '\x15'`
- `src/session-host/session.ts:579,583` — `write(KILL_LINE)` before re-attempting delivery
- `src/session-host/session.ts:18` — `const DELIVERY_ATTEMPTS = 3`
- `src/renderer/terminal/command-delivery.ts:15,127` — same constant, same pattern

## Impact beyond the visible error
An armed node opened with `--after` never launches, and the play button does not recover it:
the arm is consumed but the launch string is corrupted. The dependency logic itself is
correct — the node stayed `QUEUED` while its three stations worked, and the "waits for" edges
were accurate.

## Suggested fix
Pick the line-clearing sequence per shell rather than always sending `\x15`. On Windows
PowerShell, `\x1b[2K\r` or a bare `\r` before the new command avoids the literal insertion.
Cheap guard: after writing `KILL_LINE`, read the echo back, and if the byte appears in the
buffer, fall back to the escape sequence.

## Related, not duplicate
- #575 (Windows ships no session-persistence backend) explains why sessions are gone after a
  restart. This issue is about what the user *sees* in the node afterwards: not an empty
  prompt, but a corrupted command. The two compound.
- #706 covers the delivery path on macOS (1024-byte `MAX_CANON` truncation). Same subsystem,
  different platform failure.
- CONTRIBUTING.md already names this class ("a no-op on the machine you wrote it on and a
  silent defect on Windows", issue #558).

## Possibly separate
The canvas-control shim (`canvas-control/nodeterm.sh`) fails on Windows when its arguments
contain non-ASCII characters or backslash-heavy paths, exiting 1 **silently**. An
orchestrating agent found this on its own and worked around it by restricting orchestration
text to ASCII and forward slashes. Happy to file separately if preferred.

Abhirup0:
Closes #710.

## Problem

`KILL_LINE` was hardcoded to `'\x15'` (`Ctrl+U`) in both `src/session-host/session.ts` and `src/renderer/terminal/command-delivery.ts`.

On POSIX shells (`bash`, `zsh`), `\x15` is `unix-line-discard`. On Windows shells (PowerShell with PSReadLine and `cmd.exe`), `\x15` is not bound to line-kill; it is inserted into the prompt buffer as literal text (`^U`).

When delivery verification timed out waiting for echo confirmation on Windows, retries repeatedly typed into the un-cleared buffer. Retries concatenated into corrupted inputs (e.g. `claude^Uclaude^Uclaude` running as `claudeclaudeclaude : The term 'claudeclaudeclaude' is not recognized`).

## Solution

On Windows shells, the native sequence that clears an unsubmitted command line is `\x1b` (`Escape`):
- In PowerShell with PSReadLine (default on Windows), `Escape` is bound to `RevertLine` (discards edits on the current prompt).
- In `cmd.exe`, `Escape` clears the command line.

### Changes

1. **Session Host** (`src/session-host/session.ts`):
   - Exported `WINDOWS_KILL_LINE = '\x1b'` alongside `KILL_LINE = '\x15'`.
   - Added `shellKillLineSequence(dialect?, shellExecutableName?, platform?)` to select `\x1b` for PowerShell, pwsh, cmd, and Windows hosts, while keeping `\x15` for POSIX shells (`bash`, `zsh`, `posix`).
   - Added `killLineSequence` getter to `HostSession` and used it in `deliverVerifiedCommand`.
2. **Renderer Command Delivery** (`src/renderer/terminal/command-delivery.ts`):
   - Exported `WINDOWS_KILL_LINE = '\x1b'`.
   - Added optional `DeliverCommandOptions { killLine?: string }` to `deliverCommand(io, cmd, onSettled?, options?)`.
   - Used `options?.killLine ?? KILL_LINE` in retry and cleanup writes.
3. **Agent Restart / Wake** (`src/renderer/terminal/agent-restart.ts` & `src/renderer/nodes/TerminalNode.tsx`):
   - Added `getTerminalKillLine()` in `TerminalNode.tsx` to pass `WINDOWS_KILL_LINE` for local Windows sessions (`(corePlatformRef.current === 'win32' || (isWindowsPlatform() && session.source === 'local')) && !remoteSession`), keeping `KILL_LINE` for remote SSH or POSIX sessions.
   - Forwarded `killLine` through `performRestartResume` and `performResumePhase`.
4. **Tests**:
   - `src/session-host/session.test.ts`: added unit tests for `shellKillLineSequence` across dialects, executables, and platforms; verified `\x1b` retry output on PowerShell sessions vs `\x15` on POSIX.
   - `src/renderer/terminal/command-delivery.test.ts`: tested custom `killLine` on retry, retry exhaustion, and `line-too-long` refusal.
   - `src/renderer/terminal/agent-restart.test.ts`: verified forwarding `killLine` to `deliverCommand` on verification retry.
   - `src/core/session-host-launch-ledger.test.ts`: updated `FakePty.write` to ignore `\x1b` alongside `\x15` and `\r`.

## Verification

- **Live reproduction on Windows (`powershell.exe` with `node-pty`)**:
  - `\x15`: PowerShell echoed `first_cmd^Usecond_cmd` and failed with `The term 'first_cmdsecond_cmd' is not recognized`.
  - `\x1b`: PowerShell emitted VT erase sequence (`\u001b[9X`), cleanly discarded `first_cmd`, and executed `second_cmd` alone.
- `npm run typecheck`: clean across both `tsconfig.node.json` and `tsconfig.web.json`.
- `npm run host:build`: built session host bundle clean.
- `npm run build`: production bundle built clean.
- `npm test -- src/session-host/ src/renderer/terminal/command-delivery.test.ts src/renderer/terminal/agent-restart.test.ts src/core/session-host-launch-ledger.test.ts`: 16 files, 209 passed.
- Windows CI test suite (`ci.yml:quality-windows`): 28 files, 497 passed.

### DIFF
--- a/src/core/session-host-launch-ledger.test.ts
+++ b/src/core/session-host-launch-ledger.test.ts
@@ -28,7 +28,7 @@ class FakePty {
   write(data: string): void {
     this.writes.push(data)
     this.onWrite?.(data)
-    if (this.echoWrites && data !== '\r' && data !== '\x15') this.emitData(data)
+    if (this.echoWrites && data !== '\r' && data !== '\x15' && data !== '\x1b') this.emitData(data)
   }
 
   emitData(data: string): void {
--- a/src/renderer/nodes/TerminalNode.tsx
+++ b/src/renderer/nodes/TerminalNode.tsx
@@ -109,10 +109,9 @@ import {
 import {
   cleanEcho,
   deliverCommand,
-  KILL_LINE,
-  WINDOWS_KILL_LINE,
   type DeliveryIo
 } from '../terminal/command-delivery'
+import { terminalKillLine } from '../terminal/terminal-kill-line'
 import {
   RESUME_MISS_WINDOW_MS,
   detectsResumeMiss,
@@ -1565,10 +1564,14 @@ export function TerminalNode({
     !remoteSession &&
     (data.cwd as string | undefined) !== parentWtPath
   const getTerminalKillLine = (): string => {
-    const isWindows =
-      (corePlatformRef.current === 'win32' || (isWindowsPlatform() && session.source === 'local')) &&
-      !remoteSession
-    return isWindows ? WINDOWS_KILL_LINE : KILL_LINE
+    return terminalKillLine({
+      source: session.source,
+      browserRuntime: isBrowserRuntime(),
+      viewerWindows: isWindowsPlatform(),
+      corePlatform: corePlatformRef.current,
+      remoteSession,
+      shell: data.shell || useSettings.getState().settings.defaultShell || undefined
+    })
   }
   const status = useAgentStatus((s) => s.byId[id])
   /**
--- a/src/renderer/nodes/TerminalNode.tsx
+++ b/src/renderer/nodes/TerminalNode.tsx
@@ -106,7 +106,13 @@ import {
   validCellSize,
   type Vec2
 } from '../lib/glyphGridNode'
-import { cleanEcho, deliverCommand, KILL_LINE, type DeliveryIo } from '../terminal/command-delivery'
+import {
+  cleanEcho,
+  deliverCommand,
+  KILL_LINE,
+  WINDOWS_KILL_LINE,
+  type DeliveryIo
+} from '../terminal/command-delivery'
 import {
   RESUME_MISS_WINDOW_MS,
   detectsResumeMiss,
@@ -1558,6 +1564,12 @@ export function TerminalNode({
     !parentWtStale &&
     !remoteSession &&
     (data.cwd as string | undefined) !== parentWtPath
+  const getTerminalKillLine = (): string => {
+    const isWindows =
+      (corePlatformRef.current === 'win32' || (isWindowsPlatform() && session.source === 'local')) &&
+      !remoteSession
+    return isWindows ? WINDOWS_KILL_LINE : KILL_LINE
+  }
   const status = useAgentStatus((s) => s.byId[id])
   /**
    * Which Claude account this node is ACTUALLY on. `data.accountId` is what nodeterm launched
@@ -3325,7 +3337,8 @@ export function TerminalNode({
                   if (outcome === 'line-too-long') {
                     setCo(termKey, { launchTooLongBytes: lineBytes(cmd) })
                   }
-                }
+                },
+                { killLine: getTerminalKillLine() }
               )
             )
           })
@@ -3719,6 +3732,7 @@ export function TerminalNode({
           // An unusable session id leaves this undefined and performRestartResume refuses the
           // restart on its own `resumeCommand` gate — nothing is written either way.
           command,
+          killLine: getTerminalKillLine(),
           // Session-scoped (`api`, not the global preload), like readScrollback above: a relay
           // tab's pane lives on the host, and only its own api can see it.
           paneCommand: () => api.pty.paneCommand(id),
@@ -3860,12 +3874,14 @@ export function TerminalNode({
         // reason; the wake half owes the same. Deliberately HERE and not inside
         // `performResumePhase`: that function's output is pinned byte-for-byte by Task 8's tests,
         // and the restart path (which just cleared the line itself) must not clear it twice.
-        restartIo.write(KILL_LINE)
+        const killLine = getTerminalKillLine()
+        restartIo.write(killLine)
         return performResumePhase({
           agentId,
           sessionId: agentSessionId,
           io: restartIo,
           command,
+          killLine,
           isLive: restartTarget,
           onDelivery: (cancel) => {
             if (life.dead) cancel()
--- a/src/renderer/terminal/agent-restart.test.ts
+++ b/src/renderer/terminal/agent-restart.test.ts
@@ -683,6 +683,29 @@ describe('performResumePhase', () => {
     expect(written.join('')).toContain('claude --resume sid-1 --permission-mode plan')
   })
 
+  it('forwards custom killLine to deliverCommand on verification retry', async () => {
+    const written: string[] = []
+    const io = {
+      write(d: string) {
+        written.push(d)
+      },
+      onData() {
+        return () => {}
+      }
+    }
+    const p = performResumePhase({
+      agentId: 'claude',
+      sessionId: 'sid-1',
+      io,
+      killLine: '\x1b'
+    })
+    await vi.advanceTimersByTimeAsync(2000)
+    expect(written).toContain('\x1b')
+    expect(written).not.toContain('\x15')
+    await vi.advanceTimersByTimeAsync(10000)
+    await p
+  })
+
   it('keeps the bare command as the gate even when the caller overrides it', async () => {
     const { written, io } = fakeIo()
     expect(
--- a/src/renderer/terminal/agent-restart.ts
+++ b/src/renderer/terminal/agent-restart.ts
@@ -273,6 +273,7 @@ export async function performResumePhase(d: {
   command?: string
   /** Backstop for the resume delivery; see RESTART_DELIVERY_TIMEOUT_MS. */
   deliveryTimeoutMs?: number
+  killLine?: string
   /**
    * Handed `deliverCommand`'s cancel the moment a delivery starts — and only then. The delivery
    * outlives this promise (it runs on its own echo-verify timers), so its lifetime belongs to
@@ -319,7 +320,12 @@ export async function performResumePhase(d: {
       // Two statements on purpose: `d.onDelivery?.(deliverCommand(…))` short-circuits the ARGUMENT
       // too when no callback was passed — nothing would be delivered and this promise would never
       // settle.
-      const cancelDelivery = deliverCommand(d.io, cmd, settle)
+      const cancelDelivery = deliverCommand(
+        d.io,
+        cmd,
+        settle,
+        d.killLine !== undefined ? { killLine: d.killLine } : undefined
+      )
       started = true
       d.onDelivery?.(cancelDelivery)
     } catch (e) {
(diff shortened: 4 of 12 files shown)


eneskirca:
**Verdict: CHANGES-NEEDED — the mechanism is right and the *sites* you chose are right; the renderer's decision of "is this Windows?" is not.**

Verified locally (merged onto `main`): typecheck clean on both projects, and the full suite green — **11180 passed / 40 skipped**, including your four new suites (`session.test.ts`, `command-delivery.test.ts`, `agent-restart.test.ts`, `session-host-launch-ledger.test.ts`: 136 passed). The live PowerShell repro in the description is the right kind of evidence and it is what makes this reviewable at all — thank you for measuring instead of asserting.

Two things I want to say before the asks, because they are the parts a reviewer would otherwise worry about and you got them right:

**You did not swap the one kill-line that would have been dangerous.** `performExitPhase` (`agent-restart.ts:209`) writes `KILL_LINE` into a pane owned by an **agent TUI**, not a shell, and it is still `\x15` after this PR. That matters: a lone `\x1b` into a live agent is the *interrupt* gesture, not a line-clear — this repo already treats it as exactly that, in `TerminalNode.tsx`:

```ts
// Lone Esc / Ctrl-C while the agent works: Claude Code fires NO hook on a user interrupt…
if (showStatus && (input === '\x1b' || input === '\x03')) inferInterruptAfterSettle(id)
```

Every site you *did* change targets a shell: `deliverCommand`'s retry and its `line-too-long` clear (no `\r` is ever written before verification, so the agent cannot have started), and the hibernation wake at `TerminalNode.tsx:3874`, which is gated immediately above on `isShellCommand(pane) || pane === hibernatedPane`. That split is correct and load-bearing — **please state it in a comment**, on `WINDOWS_KILL_LINE` and/or beside the surviving `\x15` in `performExitPhase`. The next person to "finish the job" will otherwise change that one too and start cancelling turns on Windows, and no test would catch it.

---

### 1. Blocking: `getTerminalKillLine()` lets the **viewer's** OS decide for a remote core

```ts
const isWindows =
  (corePlatformRef.current === 'win32' || (isWindowsPlatform() && session.source === 'local')) &&
  !remoteSession
```

`isWindowsPlatform()` is `navigator.platform` — the browser. Server Edition tabs report `session.source === 'local'` (see the comment on `FileLinkDialectFacts.browserRuntime`), so **a Windows browser pointed at a Linux Server Edition takes this branch** and writes `\x1b` into a `bash` prompt. In readline, ESC is the *meta prefix*: `ESC` then `c` is `M-c` (capitalize-word), so the line is not cleared and the retry mangles differently than it does today.

The repo already has the correct rule, as a pure tested function, for the same question — `src/renderer/terminal/file-link-dialect.ts`:

```ts
const platform =
  facts.corePlatform ??
  (facts.source === 'local' && !facts.browserRuntime ? (facts.viewerWindows ? 'win32' : 'posix') : null)
```

Note the `!browserRuntime` term — "the viewer is authoritative only for the local desktop core, because they are the same process." Its docstring states the rule this PR needs verbatim: *"A browser's OS is not evidence about a Server Edition or relay host."* Please resolve platform the same way (fail closed to `\x15` when the core has not reported yet — that is the current, harmless behaviour).

### 2. Blocking: the two halves of this PR disagree about the same pane

`session-host/session.ts` gets a properly shell-aware decision:

```ts
shellKillLineSequence(dialect?, shellExecutableName?, platform?)
```

…which correctly answers `\x15` for `dialect === 'posix'` and for `bash.exe` / `zsh`. The renderer answers from the **platform only**. On Windows, `'posix'` is a real dialect (`SessionHostShellDialect = 'posix' | 'pwsh' | 'windows-powershell' | 'cmd'`) — git-bash/MSYS is an ordinary `settings.shell` there — so for one and the same session the host would say Ctrl-U and the renderer would say Escape. That is the "a duplicated rule drifts" case CLAUDE.md calls out (rule 10 under *Adding a new agent*), and it is the more common Windows failure of the two, because git-bash users are exactly the ones whose `\x15` works today.

`shellKillLineSequence` is already the right function; it just needs to live somewhere both sides can import (`src/shared/`), with the renderer passing the node's shell (`data.shell` / the resolved default) alongside the core platform.

### 3. Blocking-ish: the decision has no test; the plumbing has all of them

Every new test exercises `deliverCommand`/`performResumePhase`/`HostSession` *given* a kill-line. Nothing tests the choice, which is where both bugs above live — and it is an inline closure inside a ~4,800-line component, so nothing can. Same fix as #1: a pure function + a table test, mirroring `file-links.windows.test.ts`. Deleting the whole renderer gate today leaves the suite green.

### 4. Please confirm one timing detail of your Windows repro

`deliverCommand` writes the kill-line and the retry **in the same synchronous tick**:

```ts
if (!write(killLine)) return
tryOnce()            // → write(cmd)
```

So ConPTY very likely receives `ESC` and the first byte of the command in one input burst, and PSReadLine's key decoder has to choose between "lone Escape" and "`Alt+<char>`" on a timeout it may never get to run. This repo has already been bitten by exactly that class — see the opencode note in `performExitPhase`: *"opencode's TUI does not submit when text and CR arrive in the same input burst… splitting CR by 100ms exits in ~500ms."* If your repro wrote ESC and the command as two separate, spaced writes, it did not reproduce `deliverCommand`'s shape. Worth re-running with them back-to-back in one tick; if it turns out to need a split, that belongs in `deliverCommand`, not at the call sites.

---

Small notes, none blocking:

- `cleanEcho`'s `ESC_SEQ` does not strip a **trailing bare** `\x1b` (`\x1b[@-_]` needs a following char). Harmless as long as PSReadLine answers with a CSI erase (`\x1b[9X`, which *is* stripped) — worth one line in the test fixture asserting that, since the whole retry chain hinges on `echoedIntact`.
- Exporting `KILL_LINE` from `session-host/session.ts` purely so the test can import it is fine, but it is now a second public spelling of a constant `command-delivery.ts` also exports. One more reason to centralise in `src/shared/`.

Take out points 1–3 and this is a clean, well-evidenced fix.

Abhirup0:
Thanks for the thorough review and clear direction, @eneskirca.

Updated in `749cec96`:

### 1. Platform resolution & Server Edition (`terminal-kill-line.ts`)
- Replaced the browser-platform inference with a pure tested function mirroring `file-link-dialect.ts`. The viewer OS is authoritative only when `source === 'local' && !browserRuntime`.
- A Windows browser connected to a Linux Server Edition or relay session now fails closed to `\x15` (Ctrl-U), never writing Escape into `bash`.
- Added `terminal-kill-line.test.ts` with table tests covering local desktop, Server Edition, relay host vs guest, remote SSH sessions, and unobserved core platform fallback.

### 2. Centralized shell decision in `@shared/shell-kill-line.ts`
- Moved `shellKillLineSequence` to `@shared/shell-kill-line.ts`, imported by both `session-host` and the renderer.
- Accepts `(dialect?, shellExecutableName?, platform?)`. Strips directory prefixes from shell paths (handling both bare names like `pwsh` and full paths like `C:\Program Files\Git\bin\bash.exe`).
- The renderer now passes the node's shell (`data.shell` or settings `defaultShell`), so git-bash / MSYS on Windows correctly receives `\x15` across both the host and renderer.
- Added comprehensive unit tests in `src/shared/shell-kill-line.test.ts`.

### 3. Load-bearing split: Shells vs Agent TUIs
- Preserved `\x15` in `performExitPhase` (`agent-restart.ts:209`) and added an explicit comment documenting that this writes to an agent TUI where lone `\x1b` is the interrupt gesture that aborts turns, unlike shells where it is line-clear.

### 4. Windows ConPTY timing measurement
- Re-tested live on a Windows 11 host with ConPTY and PowerShell / PSReadLine:
  - Synchronous back-to-back writes (`write('\x1b')` followed immediately by `write(cmd)` in the same tick) are received by ConPTY as separate input events; PSReadLine processes `\x1b` as Escape and cleanly clears the line before typing the command.
  - A single contiguous buffer (`write('\x1becho ...')`) is where the parser treats `\x1b e` as `Alt+e`. Since `deliverCommand` and `session-host` issue two distinct `write()` calls, they arrive without key-decoder conflicts and no artificial sleep is needed.

### 5. `cleanEcho` CSI erase test
- Added a test in `command-delivery.test.ts` verifying that PSReadLine's CSI erase sequence (`\x1b[18X`) is stripped by `cleanEcho`.

All local checks pass:
- `npm run typecheck` clean on both node and web
- `npm run host:build` and `npm run server:build` clean
- 149 tests pass across all touched suites (`shell-kill-line.test.ts`, `terminal-kill-line.test.ts`, `command-delivery.test.ts`, `agent-restart.test.ts`, `session.test.ts`, and `session-host-launch-ledger.test.ts`).

PR #756, opened 2026-09-11
fix(codex): restore native model on subscription restart

dkattan:
“Restart on subscription” clears the gateway environment, but Codex restores the gateway model saved with the conversation. A conversation saved with `vllm/reasoning` then fails with HTTP 400 when resumed through a ChatGPT account.

Pass an explicit OpenAI provider override so Codex resumes the same conversation using its configured or built-in default model. Capture subscription mode before spawning the shell and pass it to both the normal resume and the missing-conversation fallback. Use a plain Codex client for this restart because a shared daemon can ignore overrides for an already-loaded thread. Custom launch programs and permission settings are preserved; no model name is hard-coded and no user configuration or authentication files are changed.

This draft depends on #380 and #715. Its base is upstream `main`, so the comparison includes their four prerequisite commits. The new fix is the final commit and changes four files. It does not depend on the other gateway PRs.

Validation:

- The four focused launch, restart, and PTY suites pass: 171 tests on this PR branch and 197 tests in the combined GitButler workspace.
- `npm run typecheck` passes on both versions; `git diff --check` passes.
- Reproduced with installed Codex 0.154.0 using an isolated conversation and local test endpoint. The original resume restored `vllm/reasoning`; the fixed command resumed the same conversation with the configured `gpt-6-astra` model. No external inference request or real user conversation was used.

Desktop, Server Edition, and SSH nodes share this command builder. Windows and a real remote SSH restart were not exercised. @eneskirca: if the mobile companion has an independent subscription restart action, it needs the same provider/model reset. A custom wrapper that forces a remote daemon must apply the overrides there; a user configuration that explicitly selects a gateway-only model still needs that configuration corrected.

### DIFF
--- a/docs/codex-shared-identity.md
+++ b/docs/codex-shared-identity.md
@@ -149,6 +149,28 @@ and `codex-launcher-sh.test.ts` pins it with a start handler that sleeps.
 
 ## 3. The fallback
 
+### Subscription restarts
+
+"Restart on subscription" strips gateway environment variables and clears the node's selected
+model. Codex also saves the model with the conversation: a bare `codex resume <id>` restores that
+saved model, including a gateway-only name such as `vllm/reasoning`.
+
+Subscription resumes therefore use `codex resume <id> -c 'model_provider="openai"'`. Codex takes
+the model from its current configuration or built-in default, keeping the same conversation and
+account. This invocation uses a plain client because a shared daemon can ignore resume overrides
+for an already-loaded thread. Ordinary gateway launches continue to use the shared launcher.
+Custom launch programs still receive the provider override; a wrapper that explicitly forces its
+own remote server remains responsible for applying those settings there.
+
+Measured with Codex 0.154.0 against an isolated conversation and a local test endpoint. The
+[resume override selector](https://github.com/openai/codex/blob/rust-v0.154.0/codex-rs/tui/src/app/config_persistence.rs)
+selects current settings when a provider override is present. No model name is hard-coded and no
+user configuration or authentication file is rewritten. Desktop, Server Edition, and SSH nodes
+share this command builder; the mobile companion needs the same behavior if it offers an
+independent subscription restart action.
+
+### Launcher failures
+
 **Every failure path ends in `exec codex "$@"`, arguments intact.** Upstream exited 69 "identity
 unavailable", which turns a missing app-server, an older `codex`, a stale tmux session or a
 locked-down data dir into a **dead node**. The repo's own precedent is `gatePermissionMode`: an
--- a/src/core/pty-manager.ts
+++ b/src/core/pty-manager.ts
@@ -2768,14 +2768,20 @@ export class PtyManager {
     // historical Claude fallback does not apply here: gateway access is an explicit agent
     // capability, not a terminal default.
     //
-    // "Restart on subscription" / `vanillaLaunchDefault`: strip the gateway + inherited provider env
+    // "Restart on subscription" / launch mode: strip the gateway + inherited provider env
     // so the agent runs against its OWN default provider (Claude's subscription, Copilot's GitHub
-    // routing). `vanillaEnvStripPattern` resolves through the base harness; null ⇒ the agent has no
-    // strip set ⇒ no-op (gemini/grok/opencode are left alone). `buildPtyEnv` runs only in `spawnNew`
-    // (a fresh session), never on a warm reattach, so toggling the setting never strips an
-    // already-running session — only the next fresh launch.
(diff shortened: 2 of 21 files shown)


eneskirca:
## Verdict: MERGE-READY — with one thing the merger must know first

**Measured** on this machine: merges clean onto `main` (422767af), both typechecks exit 0, and on the merged tree `src/shared/agents/**`, `settings-store.test.ts`, `pty-single-user.test.ts` and `state/settings.test.ts` pass **433/433** across 23 files.

### This PR carries #380 and #715 with it

`git log origin/main..pr756` is five commits: three are **#380** (`vanilla-env 'Restart on subscription'` and its two follow-ups), one is **#715** (`feat(agents): choose the default gateway model`), and only `fdb8d0a4 fix(codex): reset model on subscription` is new here — 122 insertions across 4 files.

So merging this merges #380 and #715 too. All three are independently fine (I reviewed #715 today; #380 was separately assessed as still valid, unsolved in `main`, clean-merging and typechecking), but it should be a deliberate decision, not a side effect. If you merge #715 first this reduces to its own 122-line delta.

### The new commit

Tight and correctly placed: the model reset lives in `shared/agents/launch.ts` behind the base-harness mapping rather than in a frontend, `TerminalNode.tsx` only consumes it, and `launch-subscription.test.ts` is a real behaviour test rather than a restatement of the source. `docs/codex-shared-identity.md` is updated in the same commit — the right habit, and the one CLAUDE.md asks for explicitly.

### Ordering note for the series

`#715` → `#722` → `#723` → `#724` → `#725` → `#726` is one cumulative stack; this PR branches off `#715` instead. `#715`, `#722` and this one all merge clean today; `#723` and above do not. Natural order:

`#380/#715` → `#717` → `#719` → `#722` → `#756`, then rebase `#723`–`#726`.

eneskirca:
## Review — needs the author

Reviewed as the **incremental** diff (`refs/pr/726..refs/pr/756`, commit `c9898515`): +118 lines, 4 files. The smallest increment in the stack, and the easiest to reason about.

### What I ran

Isolated worktree at `c98985150` off `origin/main` (`e43d3a54`):

- Full `vitest run` — **880 files, 12,028 passed, 52 skipped, 0 failed.**

### What this actually is

Two of the three source changes are **bug fixes for defects introduced earlier in this same stack**:

```diff
-          clearEnv: data.clearEnv === true,
+          clearEnv: resumeOnSubscription,
```

plus adding the missing `clearEnv: resumeOnSubscription` at two further `assembleResumeCommand` call sites, and a stray indentation repair (`models,`).

That's worth naming plainly: the "restart on subscription" path landed in #725/#726 reading the wrong field at three of four call sites, and nothing in the 12,028-test suite caught it. It's good that you found it — but it is evidence that the middle of this stack is not settled, and it argues for the stack being broken up and landed in smaller pieces rather than reviewed as one unit.

### Concern: `model_provider="openai"` is hardcoded

```ts
const command = codexSubscription
  ? `${withModel} -c 'model_provider="openai"'`
  : withModel
```

The comment says:

> Selecting its built-in provider reloads the configured/default model too; there is no hard-coded model or edit to the user's config/auth files.

"No hard-coded **model**" is true. But `model_provider="openai"` is a hard-coded **provider**, forced onto the user's session. For a user whose `~/.codex/config.toml` sets a different default provider — an Azure route, a corporate proxy, a local model server — "Restart on subscription" now silently redirects their session to OpenAI's public endpoint. That's a bigger behaviour change than the comment admits, and on some deployments it's an egress the user did not choose.

The intent is right (strip the gateway, fall back to codex's own default). The mechanism overshoots: it doesn't restore the user's configured provider, it *imposes* one. Options: read back whatever provider codex is configured with rather than asserting one; or emit no `-c` override at all and let the stripped env produce codex's own default, which is what the env-strip is already meant to accomplish.

At minimum the comment should stop saying there is no hard-coding, since there is — just of a provider rather than a model.

### Smaller note

`sharedIdentity && !codexSubscription` silently disables the managed launcher for this one path. The reasoning (a shared app-server can retain a loaded thread and ignore resume overrides) is stated in the comment and sounds right, but it means a "restart on subscription" leaves the node outside shared identity for that launch. Worth confirming the node re-joins shared identity on its *next* ordinary restart rather than staying detached.

### Recommendation

Hold with the rest of the stack. The `clearEnv` fixes here are genuinely wanted and should not be lost — if the stack gets broken up, these three lines belong with whichever PR introduced `resumeOnSubscription`, not as a separate PR at the tip.

*Reviewed against `origin/main` @ `e43d3a54`.*

Issue #87, opened 2026-08-08
Terminal Node selection and cursor focus


luxllumen:
Selecting a terminal node doesn't reliably move keyboard focus to the text input. Often I have to click several times in different spots before the field takes focus and the caret appears. Selecting the node idealy would immediately focus the input so I can start typing.

---
```
nodeterm: v0.2.39
surface: desktop
os: macOS
electron: 42.7.1
```

eneskirca:
solved thank you for you reporting

PR #760, opened 2026-09-13
fix(windows): preserve idle agent messaging sessions

FabricioCasali:
## Summary

Agent-to-agent messaging on Windows, for both terminal backends, plus the idle-session boundary that kept a resumed agent unmessageable.

- **Backends.** Direct ConPTY terminals and the persistent session host each attest the pane owner from OS facts (console membership, one unambiguous process chain, process birth times) before a message is written. The host does it through an additive, versioned extension (`messageOwnerV1`, `messagePasteReadyV1`, `messageEnvelopeV1`), so an older live host refuses instead of falling back to name-only input.
- **Idle boundary.** A verified `idle_prompt` right after a verified `SessionStart` now commits a verified `done`. Before, a resumed CLI idling at its prompt stayed unverified until a turn it could only get from a message it could not receive.
- **Memory levers.** An idle agent CLI on a non-persistent pty is live work: releasing it forced a resume that stranded messaging.
- **Scope sync.** Desktop dispatch publishes pending canvas edits before main authorizes the pair, without answering an unresolved file conflict.

### Fixed after review and validation (second and third commits)

- **Released sessions answered `targetGone`.** Park expiry and the offscreen release drop the attached client while the backend keeps the session running. `targetGone` is terminal and never queued, so an orchestrator was told a live agent in another project was gone. `targetLive` now comes from `PtyManager.sessionExists`, and the owner, paste and envelope probes route through `sessionHostOwns`, which falls back to the release record instead of the POSIX tmux branch.
- **npm-installed CLIs read as `node`.** The console probe stopped at the first non-shell process, so every interpreter-launched agent (Codex via npm, npx custom agents) was `not-agent`. The probe now keeps only the interpreter's first positional argument (`CommandLineToArgvW`, inside PowerShell) and resolves it through the package's `bin` map, the table npm generates the `.cmd` shim from. The interpreter is the leaf, so a CLI's own children do not make the pane ambiguous.
- **The Enter was swallowed by the paste (found by validating the build above).** With the identity fixed, a `send` reached a live Codex pane, but Codex rendered the envelope in its composer and never submitted it: the session host and the direct PTY wrote the paste and its `
` back to back. The Server Edition already had the right sequence for tmux (paste, watch the footer render, submit separately); it moves to `core/settled-submit.ts` and all three backends use it. A pane that never shows the envelope gets no Enter, and the host sends none into a replaced generation.
- **`/exit` kept a pane protected forever.** SessionEnd was recorded only as `state: undefined`, which is also what an idle agent looks like. It is now a transient `sessionEnded` flag, with a live veto over the park snapshot.

## How it was checked

- `npm run typecheck`: passed.
- Focused suites (messaging, agents, session host, renderer state/terminal, canvas, nodes, lib): every failure is pre-existing and reproduces on the base commit. They are Windows environment failures (unix sockets, POSIX `0600`, symlinks), one 5 s timeout in a Git Bash hook test that passes with a longer timeout, and the machine-name guard.
- **Live Codex pane, Windows 11.** Before the fix the installed app refused `send` with `targetNotAgentPane (observed: node)`. The new probe read `codex` (`agent`) from the same console. A fake npm package in a separate console confirmed the prompt argument never appears in the probe output.
- **Installed build, end to end (Windows 11, Codex 0.154 under the session host).** With the identity fix alone, `send` reached the pane but reported `stalled` and the envelope sat unsent in the composer. With the settled submit, the same `send` returned `delivered (signal: newTurn)` and Codex answered on its own, with nobody touching the keyboard.

Not checked live: messaging a session after park expiry or the offscreen release (covered by unit tests that route a released session-host generation through its release record, never tmux), the `/exit` memory-lever change, and macOS/Linux runs of the shared lifecycle and settle changes (CI covers them). Gemini was not exercised.

## Risks and follow-up

- Codex and opencode emit no SessionEnd, so their panes stay protected after a deliberate `/quit`. That is the safe direction: memory held, no work lost.
- Installing a new build does not upgrade a running session host. It picks up the host-side probe only after the old host retires with its sessions.
- The Server Edition's settled submit now runs through the shared module. Its sequence is unchanged; the only new behaviour is that the footer match ignores Braille cells.
- Messaging a released POSIX tmux session is now allowed too (it was refused by the same gate). The owner probe and every later gate are unchanged.


FabricioCasali:
Rebased onto current `main` (`f2449a8f`). Two things the rebase needed:

1. **Conflict with #771 in `src/session-host/session.ts`.** #771 added `TerminalEmulator.bracketedPasteRequested()` as a method; this PR had added a getter with the same name, so the merged emulator carried both. Kept #771's method. `HostSession.messagePasteReady` and `NativeWindowsPane.pasteAware` now call it; `HostSession.bracketedPasteRequested()` from #771 is unchanged.
2. **One write plan for both Windows backends** (new commit `50d92d87`). `NativeWindowsPane.sendText` had its own copy of the rule #771 put in `sendKeysWrites`, and it differed on one case: an unframed payload with Enter went out as two writes instead of one. It now calls `sendKeysWrites`, so the direct PTY and the session host cannot drift. Added a test for the unframed Enter.

#755's kill-line change in the same file merged without conflict.

Checked on Windows: `npm run typecheck` green; `native-windows-pane` and `send-keys-delivery` green (22/22). Across the session-host, agents, terminal and canvas-control suites, the failures are the Windows-only ones that fail identically on `main` (0600 modes, unix sockets, the exec-bit assertion in `canvas-control.test.ts`, and the Ctrl+V default fixed by #776). Not re-verified on a device after the rebase.

eneskirca:
`quality`, `quality-windows` and dependency review are all green here; the one red check is **CodeQL alert #87 (high, `js/insecure-temporary-file`)**, and it is a real finding rather than noise:

```
scripts/sim-agent-messaging-windows.ts:404
fs.writeFileSync(path.join(os.tmpdir(), 'nodeterm-sim-messaging-last.log'), log.join('\n'))
```

A fixed name in the shared temp directory is pre-creatable by any other account on the machine (symlink or a file they keep open), so the run either clobbers or is clobbered. Everything else in this script already does the right thing — the scenario root is `fs.mkdtempSync(...)` and every other artifact is joined onto that `dir`. This one line is the only one that escapes it, so the fix is to write the log inside `dir` too (or `mkdtemp` a second root for it).

We fix CodeQL findings at the source rather than dismissing them, including in `scripts/` — the alert is what flips the check back to green, and a dismissal leaves the pattern for the next person to copy.

That is the only thing standing between this PR and a review of the substance. Once it is green I will go through the ownership attestation and the versioned host extension properly.

FabricioCasali:
Fixed at the source in `1bcec106`, plus one related fix in `01dda532`.

**CodeQL #87.** The log now gets its own `fs.mkdtempSync` root instead of a fixed name in the shared temp dir. It does not go inside the scenario `dir`, because that root is removed a few lines later in the same `finally` and the log would go with it. The script prints the log path.

**Same class, not flagged by CodeQL.** `smoke-session-host-messaging.ts` required a pre-existing `%TEMP%\opencode` directory, which only existed in my local sandbox. It now uses `mkdtemp` under `os.tmpdir()`.

**The two Windows smoke tests were stale after the settled submit** (`01dda532`). Their fake readers never printed what they received, so the settle saw no change, sent no Enter, and the smoke timed out. The readers now echo the pasted text like a composer. The assertion on the received bytes (`\x1b[200~…\x1b[201~\r`) is unchanged.

How it was checked, on Windows 11:
- `npm run typecheck` green.
- `smoke-windows-agent-messaging` bundled and run under `ELECTRON_RUN_AS_NODE`: passed. The pre-fix version of the same script fails with `native reader timed out`.
- `smoke-session-host-messaging` after `npm run host:build` (confirmed the rebuilt `host.cjs` contains the settled submit): passed, probe 772 ms.

Not run: `sim-agent-messaging-windows.ts` after the log change (the change is only where the log is written).

FabricioCasali:
**Not tested on a device: the memory-lever protection for idle agents on a non-persistent pty.**

This is the `agentProcessInPane` / `sessionEnded` change: an idle agent CLI on a direct Windows pty must survive the offscreen release and park expiry, and must be released again once `/exit` ends its session.

I tried to check it on the installed build (`0.3.6-windows-messaging.2` at `9d47507f`, with Persistent sessions off and Release offscreen terminals at 1 minute). The measurement needs the node offscreen for over a minute while it is timed and observed. Panning the canvas takes the controlling terminal off screen too, so I found no practical way to run it by hand. I stopped, and I am not reporting a result.

It is covered only by the unit tests in `live-work.test.ts` and `park-budget.test.ts`. The `/exit` half was never attempted.

FabricioCasali:
Rebased onto current `main` (`e43d3a54`, v0.3.7). Two conflicts:

1. **`package.json`**: took `main`'s `dist` / `dist:linux` / `release` lines (they gained `npm run host:build`). This PR's `extraResources` for the Windows host is unchanged, and `session-host-launcher.ts` on `main` already looks in `resources/session-host` first.
2. **`src/core/context-link.ts`**: this PR carried its own `opencode-export.ts`, which ran the npm shim through PowerShell. #655 has since put the same read on `directExecutableInvocation` (`opencodeExportAt`). Kept `main`'s path and deleted the module and its test, so the PR is two files smaller. `cbc41290` fixes the two doc sentences that still described the PowerShell route.

The removed module had an output bound, a timeout and a session-id check that `opencodeExportAt` lacks. Those are in #820, separate from this PR.

`src/session-host/host.ts` merged cleanly over #808.

Checked on Windows 11: `npm run typecheck` green. The session-host, `native-windows-pane`, agent-messaging, pane-owner, `live-work`, `park-budget` and `message-scope-sync` suites: 512 pass, 25 fail, and all 25 are `spawn /bin/sh ENOENT` in the three context-link shim suites, which this PR no longer touches. The smoke scripts were not re-run after this rebase, and the device check from my previous comment is still open.

PR #761, opened 2026-09-13
perf(renderer): hold decorative animations still while nobody is looking

eneskirca:
nodeterm measures **Energy Impact 150.4** in Activity Monitor on an otherwise idle Mac — third heaviest on that machine, ahead of Raycast (59) and twenty times Chrome (6.9). This is where it goes, and it is not a timer.

**A running CSS animation obliges the compositor to produce a frame every vsync — 120/s on a ProMotion display — for as long as it runs, and each of those frames re-rasters and re-composites the window.** That cost is paid once for the *window*, not once per animation, and it does not care whether anybody is looking: a visible-but-unfocused window is not `document.hidden` by any definition Chromium uses, so a canvas of agent nodes keeps the frame loop open at full display rate while its owner works in another app.

## Measured

Server Edition, 40 terminal nodes on one canvas, headless Chrome, 25 s idle windows, total CPU across every Chrome process read from `/proc/<pid>/stat`:

| state | CPU |
|---|---|
| idle, nothing animating | **1.5 %** |
| **one** visible node with the `working` glow | **33 %** |
| five | 66 % |
| twenty | 101 % |
| twenty, but panned off screen | 2.2 % |
| twenty, `animation-play-state: paused` | **1.6 %** |

Four things there contradict a reasonable guess, and each shaped the fix:

- **The step is at the FIRST animation, not the twentieth.** What costs is that frames are produced at all. So a gate covering *most* of the app's animations buys nothing — one that keeps running holds the loop open, while everything the gate *does* cover still looks correctly frozen. That is why this applies to every `infinite` animation in `styles.css` and is enforced by scan rather than by hand.
- **Offscreen nodes are already free** (2.2 %) — Chromium skips raster and compositing outside the viewport. There was nothing to fix there, and a viewport-gated animation would have been work with no return.
- **`paused` is as cheap as `none`**, so the gate freezes each animation where it stands instead of snapping it to a resting frame. Nothing *moves* at the moment focus is lost, and refocusing resumes rather than restarts.
- **The main thread is idle throughout.** With one node pulsing, `Performance.getMetrics` reported 0.15 s of `TaskDuration`, 1 layout and 51 style recalcs over 25 s, while the renderer process burned 14.8 % and the GPU process 17.9 %. This is compositor and raster work, invisible to every JS-level profiler.

For contrast, the thing that *looked* like the suspect: over the same idle window the renderer fired **56 timer callbacks in 30 s** (~1.9/s — 40 per-node liveness polls at 1/30 Hz, the 2 s focus mirror, the 30 s host-RAM read) and **zero** `requestAnimationFrame` callbacks. At those frequencies a JS wakeup is nothing beside one frame of compositing, which is why this PR adds no timer gating.

## What changed

`renderer/lib/windowActivity.ts` sets `data-nt-window="idle"` on the document element when the window loses focus or the page hides; `:root[data-nt-window='idle']` flips `--nt-anim-state` to `paused`, and all 53 gated animations inherit it.

The three per-node glows take a **static-lit** rule instead of the shared pause. `nt-unread-glow` rests at `opacity: 0`, so pausing it is a coin flip on whether the glow that says *"this agent finished while you were away"* is still on screen when you come back to look for it — which is exactly the moment it exists for.

`hud.css` is deliberately excluded: the notch HUD's window is never focused, so the shared gate would freeze it permanently rather than while nobody is looking.

The fail-safe direction is `active`. Everything here is decoration, so being wrong costs a needlessly running animation, never a hidden one.

## How it was checked

- Unit tests for the state machine, including the visible-but-unfocused case the whole thing exists for, the ⌘H blur-and-hide-in-one-tick race, and the fail-open path.
- A guard test that scans `styles.css` and fails on any `infinite` animation that does not join the gate — mutation-checked by adding an ungated animation, which reds it. Given the first-animation cliff, that guard *is* the feature.
- End to end against the built Server Edition under headless Chrome: with the page focused the attribute is absent and 20 animations run; activating another tab sets `data-nt-window="idle"`, `getAnimations()` drops to 0 and the glow renders at its static opacity; refocusing restores both. Same scenario measured **101.5 % → 1.36 %**.
- Computed `animation-play-state` checked on 12 synthesized real selectors (status dots, both mascots, all three minimap dots, sidebar, kanban badge, subagent dot, spinner, presence chip): all 12 read `running` focused and `paused` idle. This was worth doing — an earlier draft of a related rule lost on specificity and had no effect at all while looking correct in the diff.

## What was NOT verified

**Any macOS number.** Headless Chrome rasters in software (SwiftShader) here, so the absolute percentages are inflated and none of them predicts Energy Impact. What the A/B establishes is the mechanism and its direction; the magnitude on a given machine has to be measured there.

Also not fixed: the **focused** case. A visible canvas with agents working still animates, by design — that is what the badges are for. Two things I measured and rejected for it: layer promotion (`transform: translateZ(0)` on the glow pseudo-element) made it *worse*, 33 % → 42.6 %; and there is no viewport gain to take, since offscreen is already free.


eneskirca:
One behaviour I could not test here, flagged so a reviewer does not have to find it: **focusing a `<webview>` guest** (a browser or web node).

Focus then moves from the top-level document into an out-of-process frame. Per spec `document.hasFocus()` counts descendant browsing contexts, so it should stay `true` and nothing should pause — but I have no Electron `<webview>` on this box (the Server Edition renders none), so that is reasoning, not a measurement.

If it *does* read as unfocused, the visible effect is benign: while you work in an embedded browser the other nodes' glows go static-lit rather than pulsing. They stay readable — `animation: none` with an explicit lit opacity, not a disappearance. It would only ever be a surprise, never a loss of state.

Worth one click on a real build before merging.

PR #784, opened 2026-09-15
fix(grok): fall back to session_summary while grok has not named the session yet

xaviguardia:
Supersedes #585. That PR's head was this fork's `main`, so a rebase there would have mixed unrelated history. This is the same remaining change, cut down against current `main`.

## What is left

`TITLE_KEYS` on main is `['title', 'generated_title']`. There is no `session_summary` leg, so a session whose generated title has not been written yet still reads as nameless.

Measured across 49 local `summary.json` files (grok 1.0.13):

- 20 of 49 carry `session_summary` and no `generated_title` (young sessions)
- where both exist they are identical in 28 of 29
- longest value in the corpus is 49 characters — a title, not a sentence

So: `TITLE_KEYS = ['title', 'generated_title', 'session_summary']`.

`session_summary` is accepted only under an 80-character cap. A longer value is **refused, not truncated**: this result is adopted as the node title with no other length bound (header, sidebar, kanban card, window title, `project.json`). Null is the node keeping its own name; a truncated sentence would be a wrong name.

## What this does not do

**`'title'` stays first.** Main's comment already argues that a key absent from every file costs one failed lookup and would win the day grok splits a manual title from a generated one. This PR does not re-litigate that.

The captured fixture and the "no `title` key appears" doc note already landed with #627. They are not in this diff.

## Testing

- `npx vitest run src/core/grok-session.test.ts src/core/agent-session-name.test.ts`: 35/35
- new cases: fallback when only `session_summary` is present; `generated_title` still wins; 81 chars → `null` (not a prefix); 80 chars accepted; an over-long summary does not hide a generated title


xaviguardia:
@eneskirca This is the rebase+shrink of #585 you asked for, against current `main`. Diff is three files: the fallback, the cap, and the doc sentence.

`'title'` is kept first, per main's comment. `session_summary` is the addition. Happy to take another pass if the cap bound (80) is the wrong number in that 60–80 range.

PR #789, opened 2026-09-16
feat(tabbar): Chrome-style tab strip on a 36px bar, height behind one token

eneskirca:
The project tab bar now looks and behaves like Google Chrome's strip, and it is 8px shorter (44 → 36). Inactive tabs are flat with a 1px divider between them that drops on both sides of a hovered or active tab; hover is a soft pill inset from the bar's edges; the active tab takes the canvas colour with rounded top corners, a square bottom and two concave flares that run it into the page. Tabs share one width and shrink evenly as more open, the title fades out instead of ending in an ellipsis, and **the name has priority as tabs shrink**: a measured density level sheds the SSH chip, then makes the inactive caret hover-only, before the 60px name floor makes the strip scroll. The active tab keeps its board toggle and options caret at every width.

| | |
|---|---|
| Hover + active, dark |  |
| Hover + active, light |  |
| 8 tabs, 1340px macOS window (SSH chips, a badge) |  |
| 12 tabs, same window: every name at the 60px floor, strip scrolls 46px |  |
| 12 tabs, an inactive tab hovered: its caret comes back |  |
| The regression this PR's first commit shipped, same 12 tabs |  |

**How it is built**

- The active tab's flares are one `::after`, a flare wider than the tab on each side, painted with a radial gradient in each corner in `--canvas-bg`. That is also the kanban overlay's colour, so the tab merges under both views. The strip gets `--tab-flare` of side padding so the first and last tab's flare is not clipped by the scroller.
- No hairlines any more. In the light theme `--panel` and `--canvas-bg` are one value apart, so a canvas-coloured tab would vanish; instead the strip steps back to `--surface-deep` (new `--tabbar-bg`, themed in both blocks), which is how Chrome's own light strip works: grey behind a paler active tab.
- Dividers are each tab's own `background-image` at its left edge, so a hover or active fill replaces the divider in one declaration; `.tab:hover + .tab` and `.tab.active + .tab` drop the neighbour's. The strip carries the one before the `+`, dropped via `:has()` when the last tab is hovered or active.
- Even shrinking: every tab is `width: var(--tab-w)` (168px) with `flex: 0 1 auto`, the name is `flex: 1 1 0; width: 0; min-width: 24px` under a mask fade. Two things there are measured, not chosen: `width: 0` on the name is what keeps a tab's automatic minimum at its fixed parts instead of the whole label, and the tab's width is a `width` rather than a `flex-basis` because the strip is content-sized (so the + sits right after the last tab) and Chromium sizes a flex container from its items' max-content contributions, which a basis does not enter. The first draft used a basis and four tabs in a 1280px window sat at their 83px floor.
- **The bar height is one number in two places that cannot read each other.** `--tabbar-h` in styles.css is now read by everything that positions against the bar, including the kanban overlay's `top` and the usage popover's cap, which were literal 44s. `TABBAR_HEIGHT_PX` in the new `@shared/window-chrome-metrics` is the same number for main, which derives the macOS traffic-light `y` from it (`trafficLightY`, 12 for a 36px bar; the old literal 15 was right for 44px and only for 44px). `styles.tabbar.test.ts` pins the token to the constant and every dependant to the token, and refuses a literal copy of the height on any bar-related rule.
- **Density** (second commit, after a regression report from the running app). At 8 tabs in a 1340px window the ACTIVE tab's name was 24px — 0 characters — because it carries the most furniture (padding 14 + dot 9 + gap 6 + SSH chip 36 + toggle+caret 47 = 118px) and flex shrink is weighted by basis, so it hit its floor first and the name took the loss; SSH tabs had 41px, and at 12 tabs every name sat on the 24px floor. Now: the name floor is 60px (~6–7 chars) and is the one place the strip decides to scroll; the active tab's basis is wider by its toggle so its name shrinks in step; and `lib/tabDensity.ts` derives a per-strip level from measured strip width ÷ tab count (TabBar stamps `data-density`) — `compact` hides the SSH chip, `tight` makes the inactive caret hover-only (kept while its menu is open). Thresholds are the furniture sums (149px / 113px per tab), not chosen numbers. Measured after, same window: 8 tabs fit with no scroll, active name 70px (7 chars), plain 75px (9), SSH 75px, badge tab 60px (6), caret on every tab; 12 tabs every name 60px (6–9 chars), strip scrolls 46px (was 89px at 0 chars), hover shows the caret (89 → 119px tab, name stays 60).
- One deliberate visible change beyond the restyle: the active tab's label is neutral text now. The CSS comment already said "neutral, never the project colour", but an inline `style={{color: p.color}}` on the tab root was overriding it. The glyph still carries the project colour; the inline colour stays only for the omni swimlane highlight, whose underline is `currentColor`.

**Kept byte-for-byte in function** (the last five TabBar commits): caret on every tab, board toggle on the active tab between the name and caret, unread badge, drag-reorder with the insertion line at both ends (re-centred for the inset tabs), click-outside menu close, the `nodeterm` brand + drag region, and the `+` outside the scroller. The non-mac branch is unchanged (`padding-left: 86px` is still gated on `data-window-chrome='inset'`).

**Checked**
- `npm run typecheck` clean; full vitest suite 853 files / 11667 tests green, including the new `styles.tabbar.test.ts` and the updated `window-chrome.test.ts`.
- Real browser: Server Edition built from this branch, headless Chrome 153 at 2x, the macOS 86px traffic-light reservation emulated so the width math matches the desktop. Seeded with 4, 8 and 12 projects (SSH-chipped and plain, one unread badge). Bar 36px; 4 tabs at 168px each; the 8- and 12-tab numbers are in the Density bullet and the screenshots above. The full before/after per-tab breakdown (padding, dot, chip, badge, actions, name, readable characters) is in the second commit's message.

**Not verified**
- Desktop macOS: traffic-light centring is derived from the constant, not seen on a device. Worth one look on a Mac, together with the flares on a Retina panel.
- Windows/Linux desktop title bar: only the CSS branch was inspected, not run.
- The density observer's measurement runs only in a real layout, so `TabBar.test.tsx` pins the attribute and the menu-open marker while the thresholds and the browser run above carry the rest. A window resized live (not reloaded) re-measures through the ResizeObserver but was not exercised on a device.

**Surfaces**: Desktop and Server Edition share the renderer, so both get this. Mobile: N/A, the companion has no tab bar.


eneskirca:
**Regression fixed in be73b31e** — 8 tabs in a ~1340px window showed 3-letter names.

Measured on the first commit's build, 1340px with the macOS reservation emulated, 8 projects mirroring the report:

| | tab | fixed furniture | name | readable |
|---|---|---|---|---|
| active (dot, SSH chip, toggle + caret) | 142px | 118px | **24px** | 0 chars |
| inactive, plain | 137px | 53px | 77px | 9 chars |
| inactive, SSH chip | 137px | 89px | 41px | 2–5 chars |
| 12 tabs, every tab | 83px | — | 24px | 0 chars |

The cause was not one number. Flex shrink is weighted by basis, so with one basis the tab carrying the most furniture hit its floor first and its name absorbed the loss — the active tab, the one you most need to read. And the 24px name floor let every tab be squeezed to nothing before the strip scrolled.

Fix, each part from a row above: name floor 60px (the one place the strip decides to scroll); active basis wider by its toggle so its name shrinks in step; a measured density (`lib/tabDensity.ts`, strip width ÷ tab count, stamped as `data-density`) that sheds the SSH chip at `compact` and makes the inactive caret hover-only at `tight` — kept while its menu is open; the active tab keeps both buttons everywhere. Thresholds are furniture sums (149 / 113px), not tuned.

Same window, after:

| | 8 tabs | 12 tabs |
|---|---|---|
| level | compact | tight |
| strip | fits, no scroll | scrolls 46px (was 89px) |
| active name | 70px, 7 chars | 60px, 7 chars |
| plain inactive | 75px, 9 chars | 60px, 6–9 chars |
| SSH inactive | 75px (chip hidden) | 60px |
| badge tab | 60px, 6 chars | 60px, 7 chars |
| caret on inactive | every tab | on hover (89 → 119px, name stays 60) |

Screenshots for both counts (and the before shot) are in the PR body. Full suite green. Not device-checked: live window resize re-measures through a ResizeObserver, exercised only by reload in the browser run.

eneskirca:
Note: this PR was merged before be73b31e was pushed to its branch, so the fix above is NOT in main. It is re-opened as its own PR from a branch off main — see the follow-up PR linked in this repo's PR list (fix/tabstrip-name-priority).

eneskirca:
Correction to the note above: the follow-up is #790 (branch fix/tabstrip-name-priority, cherry-pick of be73b31e onto main).

PR #798, opened 2026-09-16
feat(agents): Antigravity CLI (agy) as a builtin agent, with a hook that cannot deny tools

FabricioCasali:
Adds Google's Antigravity CLI (`agy`, measured on 1.2.3) as a builtin agent. An Antigravity node
gets:
- the RUNNING / NEEDS YOU badge (NEEDS YOU comes from `agy`'s `ask_question` tool);
- unread and notifications;
- `--after` dependencies, including the errored-turn hold from #521.

The delicate part is the hook, not the badge. `agy` runs hooks **synchronously** from the **global**
`~/.gemini/config/hooks.json` and reads their stdout as a **decision**. We subscribe `PreToolUse`, so
the hook stands in front of every tool call of every `agy` on the machine, inside nodeterm and in the
user's own terminals. Most of this PR exists so that the hook can never deny a tool.

**Who gets the hook.** It is installed **only on machines where `agy` is installed**. The installer
runs at every launch, in the desktop app and in the Server Edition:
- it looks for `agy` as a file only (PATH, then the vendor's install directories) and never runs it;
- the lookup happens twice per launch. At boot the process may only have the PATH it inherited (a
  GUI app on macOS/Linux starts with a minimal one), so a miss there does nothing at all. After the
  login-shell PATH probe settles, or fails, the lookup is repeated, and only then does a miss count
  (if the probe fails, that second lookup still sees only the inherited PATH);
- where `agy` is still missing after that second lookup, the installer writes nothing and withdraws
  a bundle of ours if one is there;
- an `agy` installed later gets the hook the next time nodeterm opens;
- there is **no setting to turn it off** yet.

### What `agy` 1.2.3 does with `PreToolUse` stdout (measured on Windows 11, in a temporary HOME)

| hook stdout | exit | result |
|---|---|---|
| nothing | 0 | tool runs |
| `{}` | 0 | **denied** |
| `{"decision":""}` | 0 | **denied** |
| `{"decision":"ask"}` | 0 | runs under the user's normal policy (loses to `--dangerously-skip-permissions`) |
| any non-JSON byte | 0 | **denied** |
| nothing | 1 | **denied** |

In 1.2.3 a silent `PreToolUse` executes; `{}`, `{"decision":""}`, non-JSON output and a non-zero exit
deny.

**What we answer.** The answers live in one table (`antigravity-decision.ts`):
- `PreToolUse` → `{"decision":"ask"}`;
- `Stop` → `{"decision":""}`;
- the lifecycle events → `{}`;
- an unknown or empty event → **nothing**, because `{}` there would deny the tool if the event name
  were ever lost.

Never `allow`, never `force_ask`, never `"continue"`.

**How the script keeps that answer clean.** It answers first. Then it sends stdout and stderr to
`/dev/null`, because a stray byte is a deny and `PreInvocation` stdout can be injected into the
conversation. The other agents' generated scripts are byte-identical, and their hashes are pinned in
a test.

### Other design points

- **The event name is not in `agy`'s payloads.** The command exports it, the script POSTs it as
  `nodeterm_hook_event`, and the hook server merges it after `JSON.parse`, the same way it already
  merges `nodeterm_pending_id`.
- **State mapping** (`normalizeAntigravity`, pure):
  - `Stop` with `fullyIdle:false` stays RUNNING;
  - `terminationReason` is UPPER_SNAKE, and only `ERROR` marks the turn errored;
  - `newTurn` rides the `PreInvocation` whose `invocationNum` is `0`. Without it, one errored turn
    would hold `--after` dependents for the rest of the app run.
- **Launch.** `agy` has no positional prompt, so `AgentConfig` gains an optional `promptFlag` and the
  node launches `agy --prompt-interactive '<prompt>'`. Copilot's command is unchanged.
- **Branding.** No brand mark yet: on the desktop the node uses the generic terminal icon, and the
  colour (`#00a3a3`) is provisional.

### Windows command form

`agy` passes the command to `cmd /c` as one argument with its inner quotes escaped as `\"`, which
cmd.exe cannot read. The quoted form the codex hook uses (`cmd.exe /d /c call "<wrapper> " Event`)
therefore exited 1 under the real `agy`, which **denies every tool**. The command has no quotes:

```
if exist ..\..\.nodeterm\agent-hooks\antigravity-hook.cmd (call ..\..\.nodeterm\agent-hooks\antigravity-hook.cmd Stop) & exit 0
```

- **Relative path.** `agy` runs hooks with the hooks.json directory as cwd (documented, and
  measured). So the command carries no profile name, which matters for `C:\Users\John Doe`.
- **If a future `agy` changes that cwd,** the guard finds nothing and the hook falls silent: tools
  keep running, only the badge goes dark.
- **Missing wrapper.** Without the guard, a missing wrapper is also exit 1 (measured). With it, it is
  silence.
- **One `.cmd` for all events.** The event name comes as its first argument, which survives `agy`'s
  dispatch (measured).
- **Tests reproduce `agy`'s dispatch.** They run Windows dispatch through Node's default argument
  escaping, which reproduces `agy`'s failure byte for byte. A test that uses
  `windowsVerbatimArguments` passes while the real `agy` fails.

**AutoRun.** `agy`'s `cmd /c` has no `/d`, so a registry `AutoRun` command runs first, and anything
it prints would deny tools. On Windows, the installer reads the three `Command Processor\AutoRun`
values:
- **How:** `reg query`, by absolute path, with `/reg:64` for HKLM. A missing key is decided by
  listing the parent key, because `reg` localizes its errors.
- **When any value is set, or the registry cannot be read:** the installer installs nothing,
  **withdraws** a bundle an earlier launch wrote, and logs what the withdrawal actually did.

**The installer's footprint.** It owns only the `nodeterm-status` bundle and never touches
`~/.gemini/settings.json` or `GEMINI.md`.
- Foreign keys are kept, including a literal `__proto__`.
- An unparseable file is left alone.
- A file with nothing of ours is never rewritten.

### Numbers

**Cost per event** on Windows 11, measured over 20 runs dispatched the way `agy` dispatches:
- the answer reaches stdout in about **100 ms**;
- `agy` waits for the process to exit, which takes about **520 ms** inside a nodeterm node and about
  **155 ms** in a plain terminal.

A tool step fires three of the four subscribed events. That cost is accepted for now (see
follow-ups).

**Tests.**
- Six new or extended test files for this agent, 141 tests.
- The script runs under a real `sh` with a fake `curl`, covering every table row, no
  `NODETERM_NODE_ID`, empty stdin, 200 KB stdin, and bytes planted after the answer.
- The wrapper runs under a real `cmd.exe` in a home with a space in its path.
- The errored-turn sequence runs against the real `agentStatus` store and the real `--after` gate.
- The `agy` lookup is tested against a fake PATH and a fake install directory, and the two passes
  with a detector that misses at boot and hits after the probe (and the reverse cases).
- Existing inventory tests were updated for the new builtin (capabilities, keybindings).
- `pane-owner-predicate.ts` gained the `agy` binary name, which its existing coverage test requires
  for every builtin.

### Verified

- `npm run typecheck` is clean.
- The six antigravity test files pass, as do `src/shared` (1252 tests), all after rebasing on
  current `main`. The add-menu and canvas-control suites (97 tests) pass on the final commits.
- Mutation checks:
  - removing the `/dev/null` redirect, the `if exist` guard, the `newTurn`, the AutoRun refusal, the
    withdrawal-result reporting, the `__proto__`-safe copy, the `agy` lookup, the boot-pass
    "do nothing on a miss" rule or the second pass each turns at least one test red;
  - adding a `{}` catch-all to the table does too.
- **Real `agy` 1.2.3 in a temporary HOME** (Windows 11):
  - **With the hooks.json, script and wrapper generated by this code:** events arrived with the
    right names, a file-read prompt completed, and `run_command` ran without a prompt under
    `--dangerously-skip-permissions`.
  - **With the same command form assembled by hand, before the code generated it:** it worked with a
    space in the profile path, with the wrapper missing (the tool ran, silently) and with the script
    missing (the wrapper answered).
  - **The codex-style quoted command:** it denied every tool.
  - **Hangs:** in the later bypass runs, 5 of 6 hung until `agy`'s print timeout. Three control runs
    **without any hook** hung the same way, which points at the account quota (see below) rather
    than the hook.
- **Real `agy` with the real global hooks.json installed by a dev build, run in a plain terminal
  outside nodeterm:**
  - two tool calls ran, and `agy`'s own log shows no hook denial;
  - outside nodeterm the hook POSTs nothing, so its firing was not observed directly;
  - the turn then stopped on the account's quota error (HTTP 429).
- **Full suite.** The last full run was on this Windows machine, before the final review rounds and
  before rebasing on current `main`:
  - 396 failures: the 395 that clean `main` (the old base) also fails here, mostly tests that time
    out under `sh`/tmux, plus one timeout in `managed-script.test.ts` that passes when run in
    isolation;
  - it was not re-run afterwards, because the suite rewrites real hook files on the developer's
    machine (follow-up below).

### Not verified

- **The badge and NEEDS YOU in the app window.** The test account's quota ran out before the in-app
  test could be done. This must pass before merge.
- A `Stop` with `terminationReason: 'ERROR'` from a real `agy`: never produced.
- POSIX (macOS/Linux) with a real `agy`. The script ran only under Git Bash, and the `agy` lookup on
  those platforms is only unit-tested.
- `--mode accept-edits` with an edit tool.
- A real registry `AutoRun` affecting `agy`: that would require writing the registry.
- What `agy` does when a hook exceeds the 5 s `timeout`.
- `invocationNum` restarting at 0 inside one turn in scenarios other than the ones measured.

The full device checklist is §8 of `docs/antigravity-agent.md`.

**Known limits, documented there:**
- `agy`'s own permission prompt fires no hook, so the node shows RUNNING while it waits.
- ESC fires no `Stop`, so a cancelled turn stays RUNNING.

### Surfaces

- **Desktop:** full; Windows is the target of this change.
- **Server Edition:** by construction. The normalizer is in `src/shared`, and the field merge and
  installer are in `src/core`. Neither raw hook listener needed a branch.
- **Mobile:** N/A for launching. The status mirror is agent-agnostic, so state should reach the phone
  as is (not tested). @eneskirca: the phone's hand-copied agent colours and brand marks do not know
  Antigravity yet.

### Follow-ups

- **An off switch.** Should the hook sit behind a setting, beyond "only where `agy` is installed"?
- **Hook cost.** Move everything after the stdin read to the background for this agent; measured
  about 100–150 ms instead of about 520 ms. Open question: what a slow `PreToolUse` does against the
  5 s timeout.
- **`npm test` rewrites the real managed hook files** of the other agents on the developer's machine.
  The installer-registry test's `os.homedir` spy does not reach the named imports. This predates this
  PR; this agent's installer resolves through the default import and is covered.
- **Global writers follow the last instance to run.** The global installers (hook scripts,
  canvas-control skill and instruction blocks) write the userData path of whichever instance ran
  last, so a second `NT_MULTI` dev instance repoints them.
- **Not done yet:**
  - a permission-mode mapping for `agy` (`--mode`, `--dangerously-skip-permissions`);
  - resume via `agy --conversation`;
  - a transcript reader (`transcript_full.jsonl`);
  - a brand icon;
  - hardening the NEEDS YOU correlation.


FabricioCasali:
Superseded by #799.

PR #804, opened 2026-09-17
feat(canvas): work on two projects at once by tearing a tab off into its own window

Scotty108:
I want to look at two canvases at once. Today every project lives behind one tab in one window, so working on a second project means switching away from the first — on a two-monitor desk that is the one thing the spatial canvas cannot do. This adds **pop-out project windows**: drag a project tab off the tab strip (or pick **Open in new window** from its caret menu) and that project opens in its own OS window, with its own terminals, agents, board and camera. Do it again for another project. Close the window (its ×, **Back to main window**, or the ghost tab's **Bring back to this window**) and the project returns to the main strip exactly as it was, sessions untouched.

### What had to be true for this to be safe

A pop-out is a second full renderer on the same page. Every renderer saves the WHOLE workspace from its own copy (`toWorkspace()` on every autosave), so the naive version — just open another `BrowserWindow` — loses data within minutes: main edits A and saves; the pop-out, showing B, autosaves and carries the copy of A it loaded ten minutes ago; `saveNow` skips only an *unchanged* candidate, and a stale copy is by definition changed, so A goes back to the old version with no bar and no error. So the whole design is one rule, enforced where the writes land:

- **A project is shown by exactly one window at a time, and a window may only WRITE what it shows.** `WorkspaceStore.save(ws, scope)` takes a `SaveScope` (`core/workspace-scope.ts`) that the Electron shell derives from the *sending window* (`main/popout-windows.ts`), and `scopeIndex` keeps the store's previous entry verbatim for everything outside it: main cannot delete a detached project, nothing can introduce one, a pop-out's save keeps the previous tab order and active project. The Server Edition sets no resolver — unscoped, byte-identical.
- **Save before popping out.** The new window is booted from the project as last saved, a one-project slice served from the store's memory — never a second `load()`, which is the boot path (sidelines, migrations, `revs`/`lastWritten` re-seeding) and would race a live main window's autosaves.
- **A pop-out cannot switch** (`useProjects.setActive` refuses — the one funnel every switch path uses), never joins presence, and never ghosts anything. The main window ghosts the tab (click = bring that window forward) and keeps its serialized copy fresh from the pop-out's saves (`window:popout-project-saved` → `replaceProject`), so the sidebar, the board and the store-answered control verbs stay honest.
- **Per-node traffic goes to every app window** (`sendToAppWindows`: agent status, unread clears, external changes, pressure) and **round trips go to the window showing the node** (`windowForNode`: control requests, browser popups, the browser resolve, a notification click), so exactly one window answers. Pop-outs count as attached clients, or the reaper would read their sessions as detached.
- **Closing runs a bounded flush handshake** (2.5 s): main asks the renderer to commit + save, then really closes. The ack means "you may close me", never "the save landed"; on quit it is skipped, the same exposure the main window has.

Full write-up with the reasoning and the device checklist: `docs/popout-windows.md`. The condensed rules are in `CLAUDE.md` ("Pop-out project windows") and a house rule in `CONTRIBUTING.md`.

### How it was checked

- **Unit**: the scope merge (`workspace-scope.test.ts`), the scoped save through the real store on a temp dir (`workspace-store.scope.test.ts` — the stale-copy scenario above, both directions, plus the memory boot slice and its disk fallback), the window registry and sender scope (`popout-windows.test.ts`), the tear-off / eligibility / next-active decisions (`lib/popout.test.ts`), the hash (`popout-window.test.ts`), the windows store (`windows.test.ts`), and the tab strip (`TabBar.test.tsx`: ghost tab, the menu rows, the pop-out strip, and a drag released below the strip tearing off while a drop on another tab reorders). `window-raise.guard.test.ts` counts the four new click-driven raises with their reasons.
- **Driven end to end on macOS** with Playwright over a throwaway `NT_MULTI` instance (own userData + `TMUX_TMPDIR`, three seeded projects): caret-menu pop-out, a real drag tear-off of the active tab, a rename inside the pop-out reaching the main window's ghost within one autosave, a rename in main persisting beside it, the return putting the tab back with the pop-out's canvas. On disk afterwards each project file carries exactly its owner's edits (`alpha` rev 2 untouched, `beta` and `gamma` rev 3 with their renames), index order unchanged, zero renderer errors or warnings across three windows.
- `npm run typecheck` clean. Full `vitest run`: 46 failures on this machine are pre-existing and identical on an untouched `main` checkout (pty shadow/background-write/idle-reap, session-host hello-probe's socket path length, `remote-atomic-write`, `cardModalSize`'s `localStorage.clear`) — none in files this PR touches.

### Not verified / known gaps

- Windows and Linux: the tear-off's `dragend` coordinates when released *outside* the window (Chromium reports client coordinates on macOS; if not elsewhere, that release is a cancelled drag and the menu row still works), the Linux pop-out icon, and closing the main window with pop-outs open. Listed in the doc's device checklist.
- Both windows share `localStorage` (same origin), so the per-viewer stores — view mode, explorer expansion, the persisted unread/session map — are last-writer-wins across windows.
- The keyboard-shortcut mirrors (`uiTerminalFocus` / `uiShortcutRecording`) stay sender-guarded to the main window, so under `terminal-first` a pop-out's intercepts follow the main window's terminal focus. Relaxing the guard was considered and refused: the bits are global and change-deduped, and a closing pop-out would strand the main window's state.
- Pop-out geometry is not remembered and the set of pop-outs does not survive a restart.
- Not available in the Server Edition or a relay tab (`windows.popout` rejects, the row is hidden, the drag is disarmed). Mobile: N/A.

Two small refactors ride along in `createWindow` so the pop-out cannot be a less locked-down window than the main one: the web preferences, the Linux icon, the navigation lock-down and the trackpad ledger are now shared helpers, byte-identical for the main window.

Model: Claude Fable 5.1 (claude-fable-5-1), Claude Code CLI.

https://claude.ai/code/session_01HRJRTXUTXNVv3RUQ5nzcbk


eneskirca (commented):
Thanks for this. It is a careful piece of work, and it is the feature we wanted: a user asked to *"drag multiple tabs out into their own windows for a multi monitor setup"*. The central design choice is right. The save scope is enforced in main, derived from the sending window, and `scopeIndex` keeps the previous entry verbatim for everything outside the sender's scope. That is the correct place for the rule, and the stale-copy test shows it holds.

**Verdict: we'd take it once the fixes below land.** The first two lose data or kill sessions, so they block the merge. The rest are smaller.

## What I ran

- Merged onto current `main` (the branch is 51 commits behind). `git merge-tree` reports **no conflicts**. `npm run typecheck` is clean.
- **Full `vitest run` on the merged tree: 884 files / 12,116 tests pass, 0 failures** (Linux). The 46 failures you saw on your machine don't reproduce here.
- I built the merged tree and **drove it in Electron 42.11.3 under xvfb on Linux** with Playwright. The instance ran as a throwaway `NT_MULTI` with its own `HOME`, `TMUX_TMPDIR` and userData, seeded with three projects, one of which held a terminal. Pop-out from the caret menu works on Linux: a second window with `#popout=beta`, a single tab, *Back to main window*, the terminal attached (one tmux client, attached), and the main window ghosts the tab and activates Alpha. The drag tear-off released outside the window is still unverified. Playwright can't produce a native `dragend` outside the window.
- I also ran a store-level probe through the real `workspace:load` / `workspace:save` handlers, with the same scaffolding as `workspace-store.scope.test.ts`. Blocking item 2 below comes from it.

## Blocking

### 1. The main window can still edit, close and **end the sessions of** a popped-out project

Only the tab caret menu was gated (`detached.has(menuProject.id)` → two rows). Three other surfaces in the main window still act on a project the main window doesn't own:

- **The sessions sidebar project menu** (`onProjectContextMenu`, Canvas.tsx:13642). Right-click on a detached project's header offers *Rename*, *Set folder…*, *Project settings…*, colors and *Close project*.
- **The sidebar session rows / session-memory kill**, which go through `closeStoredNodes` (Canvas.tsx:12236).
- **The Omni Kanban lanes.**

The save scope then drops the project.json half of each edit silently. Any side effect outside project.json still happens.

Measured on the built app:

1. Pop Beta out (it holds one terminal).
2. In the **main** window's sidebar, right-click *Beta* → *Rename* → "Beta renamed-in-main". The main window's sidebar and ghost tab show the new name. `workspace.json` still says `Beta`. The next pop-out save (`replaceProject`) reverts it with no signal.
3. In the main window's sidebar, right-click *Beta* → *Close project* → tick *End its session too* → **Close & end 1 session**. Result:
   - `tmux ls` on the sandbox socket: **no server** (the pop-out's session is gone).
   - The pop-out is still open on Beta, and its terminal reads **"Closed by another user — this session was ended."**
   - Beta has disappeared from the main window's strip and sidebar.
   - On disk Beta is **not** closed, because the scope dropped the `closed` flag. So the user killed live work in a window they weren't looking at, and the project comes back open when the pop-out closes.

This also contradicts the doc's claim that the sidebar and Omni Kanban "actions route to the window". Only row clicks do (`travelToNode` / `focusNodeById`).

**Fix direction:** for a detached project, the sidebar project menu should offer the same two rows as the ghost tab (*Show window* / *Bring back to this window*). Destructive or editing actions on its nodes should either route to the owning window or be disabled with a reason. Ideally add a single renderer guard (`ownsProjectHere`) in the store mutators the main window uses for non-active projects (`renameProject`, `closeProject`, `setProjectFolder`, `removeNode`, `moveNodeToGroup`, kanban writes, …). That would make a future surface fail closed instead of writing into a copy that gets thrown away.

### 2. A pop-out reload boots from a stale `lastSaved` and its boot save erases store-side writes

`loadFor` (workspace-store.ts:260) answers a pop-out from `lastSaved`. That map is updated **only** by `saveNow`. The store also writes project files outside `saveNow`:

- `appendRemoteNode`: a phone registering a session, relay `projects.registerNode`
- the watcher's `readLocalRefByPath`: git pull, a teammate's commit
- `refreshSshProject`: SSH reconcile / mobile append on the host
- the kanban / `removeProjectNode` writers

The pop-out adopts those through `workspace:external-change` without marking itself dirty, so nothing re-saves them. The hash is designed to survive a reload (⌘R, crash auto-reload), and a reload re-enters `loadFor`. `Canvas`'s boot then saves immediately (`api.workspace.save(toWorkspace())` right after `hydrate`).

Probe against the real store and IPC handlers:

1. `save(MAIN, [a, b])`, then detach `b`.
2. `load(POPOUT)` + `save(POPOUT, …)`.
3. `appendRemoteNode('b', {id: 'term-mabc123-xyz789'})`. The file now contains the phone's node.
4. `load(POPOUT)` (the reload) returns `['b-term']`, without the phone's node.
5. `save(POPOUT, thatSlice)`: **the file on disk is back to `['b-term']`**, and the phone's registered session node is deleted.

A git pull on the popped-out project followed by a reload of the pop-out rolls the pull back the same way.

The "main saves before it opens a pop-out, so the slice is exactly the content on disk" argument holds for the *first* boot only. **Fix direction:** answer a pop-out's load from the store's current knowledge of that project rather than a renderer snapshot. For a local ref that's `readLocalRef(id)`, the SSH entry's `cache`, or the inline entry. Alternatively, update or invalidate `lastSaved` on every store-side write. Please add the probe above as a test.

## Should fix

3. **Every agent interrupt fires in both windows.** `agentStatus` now goes to every app window (`sendToAppWindows`). Canvas's `alert()` (around Canvas.tsx:12900) has no ownership gate, so for each done / needs-you edge:
   - **Chimes:** both renderers play the chime, each with its own cooldown ref.
   - **Main window unfocused:** `document.hasFocus()` is false there, so the main renderer calls `notify`. `app:notify` checks only `getMainWindow().isFocused()`, so the OS notification is shown even while the user is sitting in the pop-out watching that exact node finish.
   - **Unread state:** the main window also marks the node unread while the pop-out acks it, so the ghost tab and sidebar dot disagree with what the user saw.

   The state and session-id bookkeeping should stay in both windows, since the sidebar needs it. The interrupts (unread, chime, notify) should run only where `ownsProjectHere(owner(nodeId))`.
4. **App-menu accelerators always target the main window.** `buildAppMenu(win)` binds its `send` to the main window. With a pop-out focused, ⌘⇧B toggles the board in the main window behind it, and ⌘, opens Settings there. The same goes for the Snap-to-Grid / view items. Menu clicks should go to `BrowserWindow.getFocusedWindow()` when that window is an app window.
5. **`popOutProject` ignores a failed save.** `writeDisk()` swallows a refused save (it records `saveDelivery` and returns), and `popOutProject` then asks main to open the window anyway. `lastSavedProject` still holds the *previous* save, so the pop-out boots without the edits that just failed, and its autosave makes that permanent over the main window's copy. `writeDisk` should report success, and the pop-out should be refused (tab un-ghosted) when the save didn't land.
6. **Team sync from a pop-out is silent until a peer speaks first.** The publisher's `shouldPublish` is `hasPeersRef.current`. A pop-out never joins presence, so that stays false until an inbound mutation arrives. The pop-out's edits are not cast to teammates in the meantime. Inbound mutations do reach it (`broadcast` now includes pop-outs). Either seed `hasPeersRef` from main or state it as a v1 gap in the doc.

## Nits / device checklist

- **WebGL:** the budget is per renderer (`setWebglBudget` at boot), so two windows can hold 2×24 (2×16 on macOS) contexts. The macOS number was chosen to cap compositor pressure. Please add a two-window zoom-out soak on a Mac to the checklist.
- **Traffic lights:** the pop-out's are laid out from `tabBarHeight` at creation, and only the main window is re-centred when the setting changes.
- **Quit comment:** the `close` handler says "the main window's quit flush covers the store". There is no renderer flush on quit for either window (the quit flush is `remoteWorkspaceIO.flush()` + `ptyManager.killAll()`). The exposure is the same ≤800 ms as the main window, as the PR body says, but the code comment claims more than exists.
- **Parked terminals:** the main window's parked terminals of the detached project stay co-subscribed for `TERM_PARK_MS`. Parked views report a `null` size, so they don't constrain the pop-out's pty size. I checked `applySize` / `resize` and this is fine as is.

## Scope vs. the original request

This gives **one project per window, any number of windows**. Each can be dragged to another monitor. It doesn't cover several tabs in one pop-out, dragging a tab from one pop-out into another, remembering a pop-out's position/size, or reopening pop-outs on relaunch (everything returns to the main strip). All of that is stated in the doc, and it's a reasonable v1. Geometry memory per project is the follow-up multi-monitor users will notice first.

Happy to re-review once 1 and 2 are in. For item 2, please include the store probe as a test.

Scotty108:
Thanks for the careful review. Every item is addressed on the branch, which is now merged with current `main` (no conflicts). Each fix is its own commit.

## Blocking

**1. The main window can no longer edit, close or end the sessions of a popped-out project.** (`3e1e3692`)
- **Store guard.** Every per-project mutator in `useProjects` refuses a project this window doesn't own (`OWNERSHIP_GUARDED` lists what each returns when refused). The ones that must keep working are in `OWNERSHIP_EXEMPT`, each with a reason. For example, `replaceProject` is the mirror refresh. `projects.ownership.test.ts` fails on any store method in neither table, so a new mutator can't skip the rule by being forgotten.
- **Side effects outside the store** (ending a session, a typed `/rename`, and the Omni handlers that call `useProjects.setState` directly) now go through `refuseForeignProject` first. That shows an info strip, "Beta is open in its own window. Make that change there.", with a **Show window** button. This covers `closeStoredNodes`, `closeSession` (the sidebar ×, the session-memory kill), `endProjectSessions`, `performCloseProject`, `renameSession`, the sidebar drag, reorder and "+", and every `nodeterm:global-*` / `create-node` handler.
- **Menus and lanes.** A detached project's sidebar project menu now shows only **Show window** / **Bring back to this window**. Its session rows show **Go to (in its window)** / **Bring project back**. Its Omni Kanban lane is a one-line header with **Show window** instead of cards.
- The doc no longer claims these actions route to the window. Only navigation does.

**2. A pop-out reload now boots from the store's current copy.** (`96b7eb83`)
- `loadFor` → `currentProject` → `buildEntry`. `buildEntry` is the per-entry assembly `loadV3` already ran, pulled out so both paths share it: the folder's project.json, an ssh entry's cache, or an inline data file. It is queued on `saveChain`, without sidelining.
- Your probe is now a test in `workspace-store.scope.test.ts`: save(MAIN) → detach → load+save(POPOUT) → `appendRemoteNode('b', {id:'term-mabc123-xyz789'})` → load(POPOUT) includes it → save(POPOUT) keeps it on disk. I checked that it fails against the previous `workspace-store.ts`.

## Should fix

3. **Alerts fire in one window.** (`ece70270`) Unread, chime and notification now fire only where the node's project is owned (`alertsHere`). State and session-id bookkeeping still run in every window. `app:notify` checks the focus of the window that asked, not the main window's.
4. **Menu commands go to the focused app window.** (`9bcab420`) `menuTarget` is resolved at click time. A focused window that isn't an app window, such as DevTools or the notch HUD, falls back to the main window.
5. **A failed save refuses the pop-out.** (`f878575c`) `writeDisk` now resolves whether the save landed. On `false` the pop-out is refused, the tab is un-ghosted, the canvas it was showing comes back, and the user is told.
6. **Team sync is documented as a v1 gap** in §6 of the doc. Seeding `hasPeersRef` needs a new main→pop-out channel carrying the main window's peer count. Casting while alone isn't free (a ~20 Hz stream during a drag), so I left it rather than guess at the wiring. Inbound mutations do reach the pop-out, and it starts casting after the first one.

## Nits (`898dc596`)
- The traffic lights of every open pop-out are re-centred on a tab-bar height change.
- The quit comment in the pop-out `close` handler is corrected. No renderer is asked to flush on quit, so the exposure is one autosave debounce, the same as the main window.
- The device checklist gains a two-window WebGL zoom-out soak on macOS, and §6 now states the per-renderer budget (2×16 on macOS against a shared GPU process cap of 32). The checklist also has an item to re-run your blocking scenarios on a device.
- Also (`b1513abb`): a pop-out no longer shows the onboarding tour or the one-shot launch card.

## How it was checked
- `npm run typecheck`: clean.
- Full `vitest run` on this Mac: 12,638 pass, 47 fail, all in 8 files. Every failure is in a file that already failed on untouched `main` here: pty-background-write / pty-idle-reap / pty-shadow, session-host-client, hello-probe, remote-atomic-write (unix-socket path length under `/var/folders`), plus cardModalSize and contextWindow (`localStorage` in the test env). I re-checked the last two against a clean `origin/main` worktree just now. Your Linux run is the better signal for those. Every suite this branch touches passes (49 files / 785 tests).
- **Your blocking scenarios re-run in the built app** (Electron, `NT_MULTI` sandbox with its own `HOME`, `TMUX_TMPDIR` and userData, the repo's bundled tmux 3.7b, Playwright). Beta held one terminal and was popped out from the caret menu. All 11 checks passed:
  - the sidebar menu for Beta in the main window is exactly "Show window / Bring back to this window";
  - the row × is refused with the strip, and `tmux ls` still shows `nt-term-beta1 (attached)`;
  - the Omni lane for Beta is the one-line "Open in its own window · Show window" header;
  - after an external write to `beta/.nodeterm/project.json` (the git-pull shape), a `webContents.reload()` of the pop-out shows the new node, and after its boot save the file still has it (rev 4, both nodes);
  - with the pop-out focused, the app menu's Settings item opened Settings in the pop-out and not in the main window;
  - zero renderer console errors.
- Not verified: the drag tear-off released outside the window on Linux/Windows (still item 1 on the checklist), and the two-window WebGL soak on real hardware.

Model: Claude Opus 5.5 (1M context), in Claude Code.

Scotty108:
A follow-up with four more commits. Three close gaps that my first reply left documented, and one corrects a claim I got wrong.

- **Team sync from a pop-out (should-fix 6) is fixed, no longer a documented gap.** (`3d6773ce`) The pop-out now listens to presence passively. It seeds the peer table from the hub through a new `window:presence-peers` channel (answered only for pop-out senders) and applies the join/leave diffs that `broadcast` already delivered to it. It still never says hello, so `myId` stays null: no cursors, no facepile, no name prompt. The table includes the main window's own entry, so "more than one peer" means "a teammate is here", the same rule the main window uses. Tested in `presence-session.test.ts`.
- **Tear-off released outside the window, off macOS.** (`f45908c5`) On Linux, Chromium can report `dragend` outside the window as `(0, 0)`. That lands inside the strip and reads as a cancelled drag. The strip now tracks the drag at the document level: a `dragleave` with no `relatedTarget` marks "left the window", and the last `dragover` position is kept. `isTabTearOff` checks both before the coordinates. It is unit-tested and re-checked on macOS with real drags: dropping on another tab reorders and opens no window, and releasing on the canvas tears the tab off. Still to confirm on Linux/Windows: that the document `dragleave` fires there. If it doesn't, the old behaviour stands.
- **Keyboard stand-down bits are now per window.** (`7eae3ca4`) This was a v1 gap in the doc. The "recorder armed" and "terminal focused" bits were one global pair guarded to the main window. Each app window now reports and reads its own (`window-key-state.ts`). Each window's bits are cleared at the same three sites as before. The app menu's stand-down follows the focused app window and re-syncs on `browser-window-focus`. In the built app, main's log showed the pop-out's own terminal focus disabling the menu while the pop-out was focused. Switching focus between windows couldn't be driven reliably under automation, because macOS often reports no focused window to a Playwright-driven app. That case is item 6 on the device checklist.
- **Two-window WebGL soak, run on a Mac, and a correction.** (`1e58c901`) I ran it on a MacBook Pro (M5 Max), macOS 26.6.2, Electron 42.10.1. Two windows sat side by side with 24 live terminals each, through 61 cycles of fit / zoom in / zoom out past fit / pan. Each window held its 16-context budget the whole time (32 live contexts total), with **0** `webglcontextlost` events. Every terminal was painted in the final screenshots, with no lost-context placeholder and nothing black. My earlier reply said the two windows share "a GPU process cap of 32". That was wrong: Chromium counts that cap per renderer process, and each window is its own renderer. The doc now says so. Not covered: Intel or low-memory GPUs, and terminals streaming output during the zoom.

The full blocking-scenario run in the sandbox now has 13 checks, all passing. That includes both real-drag checks above.

Model: Claude Opus 5.5 (1M context), in Claude Code.

Scotty108:
@eneskirca thanks again for the thorough review. Everything in it has been addressed and pushed to the branch, which is now merged with current `main`:

- **Blocking 1:** the main window can no longer rename, close, end the sessions of, or edit a popped-out project from the sidebar, the session-memory panel or the Omni Kanban. The projects store refuses those edits, and a test fails if a new store method isn't classified.
- **Blocking 2:** a pop-out reload now boots from the store's current copy of the project. Your `appendRemoteNode` probe is now a passing test, and it fails on the old code.
- **Should-fix 3–6:** alerts fire only in the owning window, menu shortcuts go to the focused window, a failed save refuses the pop-out, and team sync from a pop-out now works without waiting for a peer to speak first.
- **Nits:** traffic lights re-centre, the quit comment is fixed, and the two-window WebGL soak ran on a Mac with 0 lost contexts.

Two extras: tear-off detection no longer relies on `dragend` coordinates off macOS, and the keyboard stand-down bits are now per window.

The two comments above list each fix with its commit and how it was checked, including a 13-check run of your blocking scenarios in the built app. **Ready for a re-review whenever you have time.**

PR #820, opened 2026-09-18
fix(context-link): bound the opencode export and refuse an option-like session id

FabricioCasali:
`opencodeExportAt` (Context Link's local read of an opencode session, added in #655) ran `execFile` with its default 1 MiB `maxBuffer` and no timeout. An export is the whole conversation on stdout, so a long session overflowed the buffer, `execFile` errored, and the linked agent was told there was no transcript. A CLI that never exited held that read open with nothing to end it.

**What changed**

- The export allows 32 MiB of output and 60 s.
- `fetchOpencodeExport` refuses a session id outside `[A-Za-z0-9][A-Za-z0-9._-]{0,255}`, on the local leg and the SSH leg. The id comes from a hook payload and ends up as an argv entry. `directExecutableInvocation` already keeps it from being read as shell syntax; an id starting with `-` would still reach `opencode export` as an option.

The check sits in `fetchOpencodeExport`, not in `opencodeExportAt`, so `context-link.windows.test.ts` keeps proving that the cmd-shim path passes a metacharacter-bearing argument through untouched.

This came out of rebasing #760, which had carried its own PowerShell export with these same bounds. That module is dropped there in favour of #655's path, and the bounds move here.

**How it was checked** (Windows 11)

- `npm run typecheck` green.
- New `context-link.opencode-export.test.ts`: a fake `opencode` (cmd shim on Windows, sh script elsewhere) writes 3 MiB and the full output comes back. With the `maxBuffer` line removed the same test fails (`expected undefined to be 3145728`).
- `context-link.windows.test.ts` still passes.

**Not checked**

- The POSIX branch of the fake shim did not run here; CI's `quality` job is its first run.
- The timeout has no test. On Windows the kill reaches the hidden `cmd.exe`, and I did not measure whether the `node` under it exits with it.
- Not run against a real opencode session over 1 MiB.


FabricioCasali:
Fixed the red `quality` check in `aceccbe2`. The failing test was `context-link.handler.test.ts › shell-quotes the session id it sends to a remote opencode export`. It used an injection id (`x'; rm -rf ~ #`), and this PR now refuses that id before the remote branch, so nothing was sent and the assertion saw `undefined`. The code was right and the test was outdated.

The test now checks quoting with a valid id (`opencode export 'ses_6f2a9c1d'`). Both an injection id and `--help` assert that nothing is sent to the host. `context-link.handler.test.ts` and `context-link.opencode-export.test.ts` pass locally (18 tests).

